PR #4, opened 2018-11-02
Add continuous integration for druid

dsp:


raphlinus:
Looks like that worked. Not 100% sure if there's anything more I need to configure so new PR's will have their CI automatically reported.

PR #6, opened 2018-11-27
Move to platform independence

snoe:
Here's a PR that I hope helps move druid in a direction where it has a place to put code for other platforms. Users should only need to call `piet`, `druid`, or `druid-shell` functions and shouldn't need to interface directly with `winapi`, `directdraw` or `directwrite`

It's possible this PR is too early or unwanted at this time, I'm ok with that, I thought this would help at least crystallize discussions.

With this structure we'd have a place and direction to add cairo and other OS shell support. Here's a WIP example: https://github.com/snoe/druid/tree/support-macos

The piet for 2d graphics crate I think/hope follows the discussion here:
https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/2D.20abstraction.20layer

It's likely the public api for piet needs to be cleaned up and tightened instead of just exposing direct2d, but I wonder if that can come later as other platforms may force different apis.

I'm also a rust noob so if there are more idiomatic ways to do anything here I'd appreciate the feedback.


snoe:
>Apologies for the delay in reviewing.

No worries, my time can be pretty constrained as well.

> have you tried making the corresponding patch to xi-editor?

Oh I thought I had but i guess I was confused between xi-editor and xi-win, you can see the changes for xi-win here: https://github.com/xi-editor/xi-win/compare/master...snoe:deplatform

> For "author" in the metadata, I'm ok with that being you.

That feels a bit disingenuous to me since I just moved your code around, I just added myself to the AUTHORS file.

> I'm slightly concerned about the order that changes happen


A big danger I see from exporting:* from directdraw is we don't know and can't control  what pieces others are using. Doing a [search on github](https://github.com/search?q=filename%3ACargo.toml+druid) it seems that xi-editor and your synthesizer-io are the only users visible there.

An option that could mitigate some of this danger of exposing the directdraw apis, is I could lock down the pub exports to the set of vars used by known clients of piet. 

> Another possibility to explore is to get a minimum functioning piet library
> A last possibility is to have a druid branch

Either of these could be good options only I'd worry about them not tracking the users/usages of druid and getting out of sync - since there are only a few maybe this concern isn't as important.

> I think it will be necessary fairly soon to have a new repo for piet. That could be done now (I'm happy to create the repo in the xi-editor org), or as a separate step.

xi-editor org would definitely be the right place for a piet repo. I worry about publishing a standalone crate before the API is better resolved.

A thought I had was that druid could re-export piet and druid_shell so that it's users didn't need to care where the crate was located, but that may not be idiomatic rust.

snoe:
https://github.com/linebender/piet aptly replaces a large part of this pr. closing.

raphlinus:
@snoe I wanted to thank you for getting this rolling and apologize for not working with this PR more directly. Building out piet is confirming what I suspected, that there are significant changes needed to make things efficient and Rust-native compared with just replicating the Direct2D interfaces. My hope is to get that in reasonable shape soon, then we can direct attention to making druid-shell platform independent and reworking all the Direct2D code to piet. This PR will I think be a useful roadmap for that work, though we won't be able to use much of it directly.

snoe:
@raphlinus Getting the ball rolling is all I wanted to accomplish, so thank you for taking it on. I'm super happy with your direction.

PR #11, opened 2019-01-25
Cross Platforming Drawing

### LINKED ISSUE #8: Platform Independent Drawing
So after looking over the code, I have an idea of where to start with platform independence. I'd like to start from the inside and work our way out, first by changing the signature for the `Widget` trait and then simply following the types. Before progress on this can be made, there are some questions to answer:

* **What should the signature of `paint` be?** Currently it is ` fn paint(&mut self, paint_ctx: &mut PaintCtx, geom: &Geometry)`. `paint_ctx` is simply a wrapper around the rendering context (both direct2d and directwrite) plus the information of whether the widget is hovered (a.k.a hot) or pressed (a.k.a active). Swapping the inner render context should be a straight 1 to 1 swap with piet `RenderContext`, but is this the right place for the hot and active bools to live?
* **Should `LayoutCtx` include a `RenderContext`?** The `LayoutCtx` currently takes a `directwrite::Factory` so that it can do things like layout text to get its size. If we pass in a `RenderContext`  we'll be able to get this info. We can also pass in a `TextLayoutBuilder` or `TextLayout` which will also get us that information. However, we then need to parameterize `LayoutCtx` on one of these traits which will force us to bubble up a generic to `LayoutCtx` and `HandlerCtx` which is unfortunate.

rylev:
Fixes #8

### DIFF
--- a/src/widget/button.rs
+++ b/src/widget/button.rs
@@ -120,11 +120,13 @@ impl Widget for Button {
                 _ => 0x404048,
             };
             let brush = paint_ctx.render_ctx.solid_brush(bg_color).unwrap();
+            let (x, y) = geom.pos;
+            let (width, height) = geom.size;
             let rect = Rect::new(
-                geom.pos.0.into(),
-                geom.pos.1.into(),
-                geom.size.0.into(),
-                geom.size.1.into(),
+                x as f64,
+                y as f64,
+                x as f64 + width as f64,
+                y as f64 + height as f64,
             );
             paint_ctx
                 .render_ctx
--- a/src/widget/button.rs
+++ b/src/widget/button.rs
@@ -16,11 +16,9 @@
 
 use std::any::Any;
 
-use direct2d::brush::SolidColorBrush;
-use direct2d::RenderTarget;
-use directwrite::{self, TextFormat, TextLayout};
-
-use druid_win_shell::util::default_text_options;
+use kurbo::Rect;
+use piet::{FillRule, FontBuilder, RenderContext, Text, TextLayoutBuilder};
+use piet_common::Piet;
 
 use widget::Widget;
 use {BoxConstraints, Geometry, LayoutResult};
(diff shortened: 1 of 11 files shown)


rylev:
@raphlinus Besides the fact that `druid` is now relying on `piet` and `piet-common` through github instead of crates.io this is ready to go. `druid` is now only dependent on `piet`, `piet-common`, `kurbo` and `druid-win-shell` which means the only Windows specific dependency is `druid-win-shell`. I think now we can go through the use of `druid-win-shell` in `druid` and itemize what is Windows specific. Hopefully that will make it clear what the exact work items are.

src/widget/button.rs
```diff
@@ -47,34 +45,39 @@ impl Label {
         ctx.add(self, &[])
     }
 
-    fn get_layout(&self, dwrite_factory: &directwrite::Factory) -> TextLayout {
+    fn get_layout(
+        &self,
+        rt: &mut Piet,
+        font_size: f32,
+    ) -> <<Piet as RenderContext>::Text as Text>::TextLayout {
```
rylev: This is not super pleasant to have to write out. Since TextLayout is also the name of a trait, I'm not super sure what the best way to expose this from `piet_common` would be.

src/widget/button.rs
```diff
@@ -47,34 +45,39 @@ impl Label {
...
+            .new_font_by_name("Segoe UI", font_size)
+            .unwrap()
             .build()
             .unwrap();
-        let layout = TextLayout::create(&dwrite_factory)
-            .with_text(&self.label)
-            .with_font(&format)
-            .with_width(1e6)
-            .with_height(1e6)
+        // TODO: Layout height and width: 1e6
```
raphlinus: This TODO can be removed, though we're going to have to revisit this. The use of 1e6 was basically a hack to get a single line. But down the road we'll probably want a `Label` to be able to do line-breaking to fit the layout width as well. This is dependent on a high-level text interface, which is a glimmer in my eye at the moment.

src/widget/button.rs
```diff
@@ -47,34 +45,39 @@ impl Label {
...
-        let text_layout = self.get_layout(paint_ctx.dwrite_factory());
-        let rt = paint_ctx.render_target();
-        let fg = SolidColorBrush::create(rt)
-            .with_color(0xf0f0ea)
-            .build()
+        let font_size = 15.0;
+        let text_layout = self.get_layout(paint_ctx.render_ctx, font_size);
+        let brush = paint_ctx.render_ctx.solid_brush(0xf0f0eaff).unwrap();
+
+        // TODO: bring back default_text_options from win-shell
```
raphlinus: I think this is the wrong place for this TODO. There's already [one](https://github.com/linebender/piet/blob/7a13480b19936b020726499f81f6388545e2b807/piet-direct2d/src/lib.rs#L274) in piet-direct2d. (It should be noted that this change will regress color emoji, but it's not going to be that difficult to fix, it's basically a factor of having Windows version detection at the right place).

raphlinus:
Ok, from what I can tell it's good now, thanks!

Issue #13, opened 2019-01-29
Cross Platform `WinHandler`


rylev:
With #11 drawing will be cross platform. The question is then "what next?". I propose we next try to make `WinHandler` cross platform. `WinHandler` is the trait that defines how to respond to window events like painting, size changes and keyboard/mouse events. 

Here's a list of items that need to be addressed:
* `WinHandler::connect` takes a `WindowHandle` which has a weak reference to a `WindowState` which is highly Windows specific as it has a reference to `HWND` and `WndProc`. `connect` stores the `WindowHandle` on the state's `LayoutContext`. The handle is used for the following: invalidating widgets, closing the window, file dialog, converting physical pixels to px units, getting the current dpi. Perhaps this can all be abstracted away in a trait that is made concrete in a similar way that we do with `piet-common`
* The various keyboard handlers take modifier keys which can be different on different platforms. We could solve this the same way web APIs do: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/getModifierState
* Realted to the first issue: `WinHandler::mouse`, `WinHandler::mouse_move` and `WinHandler::size` all use the windows specific `WindowHandle` that is stored on the `UiState`'s `LayoutContext`
* `WinHandler::destroy` calls the windows specific `win_main::request_quit()`

Thoughts on how to abstract `WindowHandle` and how to handle modifier keys?

raphlinus:
Closing this, as we've been cross platform for a while, and I think all other relevant details are captured in other issues.

PR #17, opened 2019-02-13
Rename druid-win-shell to druid-shell

raphlinus:
As part of platform independence, rename druid-win-shell to remove the
reference to a specific platform.

Also fixes druid-shell examples, which were slightly broken.

Normalizes line endings to LF (not CRLF).

Part of #13


raphlinus:
One thing that came up, the druid-shell examples didn't compile, so we should make sure CI gets those. I think a good way to do that is set up a workspace; this should also help with RLS.

PR #20, opened 2019-02-16
Create windows on mac

raphlinus:
Implement a subset of the window creation API. At this point, it
compiles and runs the examples in druid-shell, but with much
functionality stubbed out, and no attempt to run the higher level.

Uploading this as a checkpoint.

### DIFF
--- a/druid-shell/examples/perftest.rs
+++ b/druid-shell/examples/perftest.rs
@@ -26,9 +26,12 @@ use time::get_time;
 use kurbo::{Line, Rect};
 use piet::{FillRule, FontBuilder, RenderContext, Text, TextLayoutBuilder};
 
+#[cfg(target_os = "windows")]
+use druid_shell::platform::PresentStrategy;
+
+use druid_shell::platform::WindowBuilder;
 use druid_shell::win_main;
 use druid_shell::window::{WinHandler, WindowHandle};
-use druid_shell::platform::{PresentStrategy, WindowBuilder};
 
 struct PerfTest(RefCell<PerfState>);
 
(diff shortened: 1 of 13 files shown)


druid-shell/examples/perftest.rs
```diff
@@ -26,9 +26,12 @@ use time::get_time;
 use kurbo::{Line, Rect};
 use piet::{FillRule, FontBuilder, RenderContext, Text, TextLayoutBuilder};
 
+#[cfg(target_os = "windows")]
+use druid_shell::platform::PresentStrategy;
```
rylev: Q: What is PresentStrategy?

druid-shell/src/mac/mod.rs
```diff
@@ -13,8 +13,277 @@
...
+
+use util::assert_main_thread;
 
 #[derive(Clone, Default)]
-pub struct WindowHandle;
+pub struct WindowHandle {
+    /// This is an NSView, as our concept of "window" is more the top-level container holding
+    /// a view. Also, this is better for hosted applications such as VST.
+    ///
+    /// TODO: remove option (issue has been filed against objc, or we could manually impl default with nil)
```
rylev: Can you link the issue number?

druid-shell/examples/perftest.rs
raphlinus: See https://github.com/xi-editor/xi-win/issues/17 . It's part of my quest to make window resizing smooth on Windows. It's not something that's meaningful on other platforms, and most actual users will use the default. It's exposed for experimentation.

raphlinus:
FWIW, I have an implementation of IdleHandler and anim (the latter is crude, based on continual invalidation, rather than CVDisplayLink, but is ok to get this into a working state) in my local repo. I can add to this PR or file a new one.

rylev:
This looks good. Might be worth getting this in before adding the other stuff

PR #21, opened 2019-02-17
Add idle handler and animation

raphlinus:
The animation performance is poor, but it works. Idle handler is not tested.

### DIFF
--- a/druid-shell/src/mac/mod.rs
+++ b/druid-shell/src/mac/mod.rs
@@ -57,7 +57,7 @@ pub struct WindowHandle {
     /// TODO: remove option (issue has been filed against objc, or we could manually impl default with nil)
     /// https://github.com/SSheldon/rust-objc/issues/77
     nsview: Option<WeakPtr>,
-    queue: Weak<Mutex<Vec<Box<IdleCallback>>>>,
+    idle_queue: Weak<Mutex<Vec<Box<IdleCallback>>>>,
 }
 
 /// Builder abstraction for creating new windows.
@@ -70,7 +70,7 @@ pub struct WindowBuilder {
 #[derive(Clone)]
 pub struct IdleHandle {
     nsview: WeakPtr,
-    queue: Weak<Mutex<Vec<Box<IdleCallback>>>>,
+    idle_queue: Weak<Mutex<Vec<Box<IdleCallback>>>>,
 }
 
 // TODO: move this out of platform-dependent section.
@@ -128,7 +128,7 @@ impl WindowBuilder {
             window.setTitle_(make_nsstring(&self.title));
             window.makeKeyAndOrderFront_(nil);
 
-            let (view, queue) = make_view(self.handler.unwrap());
+            let (view, idle_queue) = make_view(self.handler.unwrap());
             let content_view = window.contentView();
             let frame = NSView::frame(content_view);
             view.initWithFrame_(frame);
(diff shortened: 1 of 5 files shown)


druid-shell/src/mac/mod.rs
```diff
@@ -53,6 +57,7 @@ pub struct WindowHandle {
     /// TODO: remove option (issue has been filed against objc, or we could manually impl default with nil)
     /// https://github.com/SSheldon/rust-objc/issues/77
     nsview: Option<WeakPtr>,
+    queue: Weak<Mutex<Vec<Box<IdleCallback>>>>,
```
rylev: `queue` is a bit of a weak name, but `idle_callback_queue` might be too long...

druid-shell/src/mac/mod.rs
```diff
@@ -62,9 +67,26 @@ pub struct WindowBuilder {
...
+
+impl<F: FnOnce(&Any) + Send> IdleCallback for F {
+    fn call(self: Box<F>, a: &Any) {
+        (*self)(a)
+    }
+}
 /// This is the state associated with our custom NSView.
 struct ViewState {
     handler: Box<dyn WinHandler>,
+    idle_queue: Arc<Mutex<Vec<Box<IdleCallback>>>>,
```
rylev: This should probably match the name above ^^

druid-shell/src/mac/mod.rs
```diff
@@ -235,14 +271,39 @@ extern "C" fn draw_rect(this: &mut Object, _: Sel, dirtyRect: NSRect) {
...
+        }
 
         let superclass = msg_send![this, superclass];
         let () = msg_send![super(this, superclass), drawRect: dirtyRect];
     }
 }
 
+extern "C" fn run_idle(this: &mut Object, _: Sel) {
+    unsafe {
+        let view_state: *mut c_void = *this.get_ivar("viewState");
```
rylev: Do we need to do any null checking or pointer validation?

druid-shell/src/mac/mod.rs
```diff
@@ -235,14 +271,39 @@ extern "C" fn draw_rect(this: &mut Object, _: Sel, dirtyRect: NSRect) {
...
+                withObject: nil waitUntilDone: NO);
+        }
 
         let superclass = msg_send![this, superclass];
         let () = msg_send![super(this, superclass), drawRect: dirtyRect];
     }
 }
 
+extern "C" fn run_idle(this: &mut Object, _: Sel) {
+    unsafe {
```
rylev: Looks like you might be wrapping more code in unsafe than is strictly necessary. I think it's best to only wrap the absolute minimum in unsafe as possible.

druid-shell/src/mac/mod.rs
raphlinus: Actually I might want to reconsider "idle", as that suggests a lower priority than other events on the UI thread, and I don't think that's accurate. The name is taken from gtk. But I think "idle_queue" is probably the best compromise for now.

druid-shell/src/mac/mod.rs
raphlinus: No. I consider `viewState` ivar to basically be a Rust object (a Box) owned by the NSView. There is a much more subtle case later in `add_idle`, where the nsview pointer loaded might be nil, but in that case the default behavior of msg_send is valid. I should add a comment there.

druid-shell/src/mac/mod.rs
raphlinus: This is a style point I don't have strong feelings about, but have tightened up the scope of unsafe in cases where it doesn't force gymnastics for visibility.

PR #24, opened 2019-02-24
Mac events

jjl:
I have built open your work to add the infrastructure for making menus and mouse move events work.

There's probably some room for improvement, but it works.


jjl:
Looks like it fails because of the import of Menu in the calc app, which it appears isn't public on windows - how is one supposed to use the menu if it is not public?

Cargo.toml
```diff
@@ -19,6 +19,9 @@ piet-common = "0.0.2"
 piet = "0.0.2"
 kurbo = "0.2.1"
 
+[[bin]]
```
raphlinus: I don't think this is necessary, you run the example with `cargo run --example calc`. It is very much an example, not really a useful binary in and of itself.

druid-shell/src/mac/menu.rs
```diff
@@ -13,17 +13,105 @@
...
+    }
+}
+
+impl Default for Menu {
+    fn default() -> Menu {
+        let mut menu = Menu::new();
+        let mut submenu = Menu::new();
+        submenu.add_item(1, "Quit", "q");
+        submenu.add_separator();
+        submenu.add_item(1, "Also Quit", "q");
```
raphlinus: I understand the desire to test but probably just the one quit is enough :)

druid-shell/src/mac/mod.rs
```diff
@@ -145,13 +163,47 @@ impl WindowBuilder {
...
+struct WindowClass(*const Class);
+unsafe impl Sync for WindowClass {}
+
+lazy_static! {
+    static ref WINDOW_CLASS: WindowClass = unsafe {
+        let mut decl =
+            ClassDecl::new("DruidWindow", class!(NSWindow))
+            .expect("Window class defined");
+
+        decl.add_ivar::<BOOL>("acceptsMouseMove");
```
raphlinus: I'm not sure I fully understand the motivation for this variable. I've been studying other code bases and it seems like the best practice for enabling mouse movement is to set up an NSTrackingRegion. Examples in [minifb](https://github.com/emoon/rust_minifb/blob/master/src/native/macosx/OSXWindowFrameView.m#L88) and [rutabaga](https://github.com/wrl/rutabaga/blob/master/src/platform/cocoa/window.m#L108) linked. I haven't studied it hard enough to know the true best way.

Cargo.toml
jjl: Heh, I wasn't quite sure, so I just added this so I could test it

druid-shell/src/mac/menu.rs
jjl: It's only there to show the separator works (the separator will not render if you remove the second item)

jjl: I'm not actually sure what a sensible default is here. My initial thought was None, but actually we probably are better to just defaulting to Quit. That said, I have no real idea how it all threads through the rest of the library yet.

druid-shell/src/mac/mod.rs
jjl: I have no idea what the "best" way is, this is more the only way I found to do when I needed to last time. I'll take a look at those examples if you think that way is better.

The major dislike I have of the way I've done it here is the need for the extra subclass of NSWindow

druid-shell/src/mac/menu.rs
raphlinus: Yeah, defaulting to quit is pretty reasonable, I think. It's different than Windows, where it's common to have a window that doesn't have any menus at all.

jjl:
I've addressed the two simple ones here, but I'd like to know that an NSTrackingRegion-based solution would be the 'best' one before I sink time into it. I suspect you've got a better network to answer that question than I do.

druid-shell/src/mac/menu.rs
```diff
@@ -13,17 +13,106 @@
...
+
+fn make_proxy(id: u32) -> id {
+    unsafe {
+        let mut target: id = msg_send![MENU_ITEM_PROXY_CLASS.0, alloc];
+        (*target).set_ivar("menu_id", id);
+        target.autorelease()
+    }
+}
+fn make_menu_item(id: u32, text: &str, key: &str) -> id {
+    unsafe {
```
rylev: I've talked with @raphlinus before about this and I think we have different opinions. I prefer to wrap the absolute minimum in unsafe so that debugging is easier. With the code as it is you'd have to inspect `make_proxy` only to realize it's not actually unsafe. In this case that's not a huge deal since it's right above, but if this function were in another crate, you'd have to do some digging that could be avoided if `unsafe` were used a bit less liberally.

jjl: I'm not sure I fully understand. The reason I've done it this way is because basically if you're going to touch cocoa integration, you're fundamentally in unsafe territory because of the need to interop with the objectionable c runtime. I don't see a good way of avoiding unsafe in the general case with what we're doing, so i haven't worried too hard about it.

rylev: I'm refering to the scope of the `unsafe` block. You could move the use of the `unsafe` keyword down one line and the code would still compile. I'm arguing that this is the right thing to do because it properly highlights which code is actually unsafe in that function.

raphlinus:
Also, since I've merged #23 , I think this needs to be updated otherwise it will try to merge into the branch for that. Shouldn't be any merge conflicts, just a procedural issue.

raphlinus:
What I'm going to suggest is to strip out the stub mouse tracking and get the menu stuff merged, then do a follow-on PR for mouse. This is in part because there are some other PR's in flight, so it would be good not to have any open for a long period.

Also keep in mind one of the newer requirements is that it work as a [VST plugin](https://github.com/rust-dsp/rust-vst/wiki/Rust-Audio-GUI-Development-Progress). This limits to some extent what can be done with the NSWindow, as the plugin is given an NSView handle. This is one of the main reasons I'm looking at the [rutabaga](https://github.com/wrl/rutabaga/tree/master/src/platform/cocoa) codebase, as it contains patterns known to work.

jjl:
I just tried to cherry-pick this onto master only to discover that master doesn't actually build on a mac anymore.

```
error[E0432]: unresolved import `druid_shell::application`
  --> src/lib.rs:33:18
   |
33 | use druid_shell::application::Application;
   |                  ^^^^^^^^^^^ could not find `application` in `druid_shell`

error[E0432]: unresolved import `druid_shell::windows`
  --> src/widget/key_listener.rs:17:18
   |
17 | use druid_shell::windows::M_ALT;
   |                  ^^^^^^^ could not find `windows` in `druid_shell`

error[E0599]: no method named `invalidate` found for type `druid_shell::window::WindowHandle` in the current scope
   --> src/lib.rs:749:29
    |
749 |                 self.handle.invalidate();
    |                             ^^^^^^^^^^

error[E0599]: no method named `pixels_to_px_xy` found for type `druid_shell::window::WindowHandle` in the current scope
   --> src/lib.rs:944:46
    |
944 |         let (x, y) = state.layout_ctx.handle.pixels_to_px_xy(x, y);
    |                                              ^^^^^^^^^^^^^^^

error[E0599]: no method named `pixels_to_px_xy` found for type `druid_shell::window::WindowHandle` in the current scope
   --> src/lib.rs:951:46
    |
951 |         let (x, y) = state.layout_ctx.handle.pixels_to_px_xy(event.x, event.y);
    |                                              ^^^^^^^^^^^^^^^

error[E0277]: the trait bound `f64: piet::RoundFrom<f32>` is not satisfied
  --> src/widget/button.rs:56:14
   |
56 |             .new_font_by_name("Segoe UI", font_size)
   |              ^^^^^^^^^^^^^^^^ the trait `piet::RoundFrom<f32>` is not implemented for `f64`
   |
   = note: required because of the requirements on the impl of `piet::RoundInto<f64>` for `f32`

error: aborting due to 6 previous errors
```

raphlinus:
I'm not seeing that problem, I was able to rebase on top of master without difficulty. The "calc" example runs, but the "hello" example in druid-shell has small errors with the menu calls.

jjl:
I cherry-picked just the one commit directly relevant to the menus, on the basis that there is another PR in flight which addresses the 2018 edition, which was another consequence of this branch.

jjl:
Okay, gave up on the cherry pick and just did as you suggested, rebased and then removed it manually. Calc example still runs.

jjl:
This should now merge cleanly onto master.

raphlinus:
The "hello" example in druid-shell is still not compiling, the signature of `.add_item()` is mismatched. I'm going to do another pass over the code now.

raphlinus:
Generally this looks good, but I see a few small issue in addition to the menu confusion above. I think it's fastest for me to fix those and push a branch, because I want to do some work on top of it. Lemme try that, then we'll figure out the best way to get that merged.

raphlinus:
Ok, I uploaded https://github.com/xi-editor/druid/tree/fix-menu-sig , which is a pretty small change to just fix compilation. The *easiest* thing is for me to just PR that and merge it, but I think that doesn't give you proper credit in the commit stats, so if you care about that, feel free to pull it into your branch, then I'll merge this PR. Thanks!

Then I'm working on some of the other issues in a local branch based on top of that.

jjl:
I don't much care whether I'm credited in the stats, I just think it's a cool thing, so do whatever you will.

In any case, it's likely to be far from my last contribution.

raphlinus:
No problem, closing this one in favor of #27, then I should have another one up soon. Thanks for your contributions, looking forward to more!

PR #28, opened 2019-03-01
More refinement of cross-platform menus

raphlinus:
As of this commit, menus aren't fully wired, but the API provided to clients is a lot closer to what it should be.

### DIFF
--- a/druid-shell/src/mac/menu.rs
+++ b/druid-shell/src/mac/menu.rs
@@ -13,12 +13,12 @@
 // limitations under the License.
 
 //! macOS implementation of menus.
-use lazy_static::lazy_static;
-use objc::runtime::{Class, Object, Sel};
-use objc::declare::ClassDecl;
+use cocoa::appkit::{NSApplication, NSMenu, NSMenuItem};
 use cocoa::base::{id, nil, BOOL, NO, YES};
 use cocoa::foundation::{NSAutoreleasePool, NSPoint, NSRect, NSSize, NSString};
-use cocoa::appkit::{NSMenu,NSMenuItem, NSApplication};
+use lazy_static::lazy_static;
+use objc::declare::ClassDecl;
+use objc::runtime::{Class, Object, Sel};
 
 struct MenuItemProxyClass(*const Class);
 unsafe impl Sync for MenuItemProxyClass {}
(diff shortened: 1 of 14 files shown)


chris-morgan:
A uniformly higher-performance `strip_access_key` (10–20% faster on pathological microbenchmarks, typically a lot more due to avoiding allocation altogether where no access key is present, and precise allocation in other cases):

```rust
use std::borrow::Cow;

fn strip_access_key_hunkwise(mut raw_menu_text: &str) -> Cow<str> {
    if let Some(mut i) = raw_menu_text.find('&') {
        // This is both the most likely size and the size it will not exceed.
        let mut result = String::with_capacity(raw_menu_text.len() - 1);
        loop {
            let offset = match raw_menu_text.as_bytes().get(i + 1) {
                Some(&b'&') => i + 1,
                _ => i,
            };
            result.push_str(&raw_menu_text[..offset]);
            raw_menu_text = &raw_menu_text[offset + 1..];
            if let Some(j) = raw_menu_text.find('&') {
                i = j;
            } else {
                result.push_str(raw_menu_text);
                break result.into();
            }
        }
    } else {
        raw_menu_text.into()
    }
}
```

I haven’t thought about it much yet, but remember how in Windows you specify the keyboard shortcut to appear right-aligned on the right-hand side with the ␉ separator, e.g. `"E&xit\tAlt-F4"`. I have no idea how other platforms represent that. The formatting of the shortcut gets you into heavily platform- and even locale-dependent behaviour, too. Fun times.

raphlinus:
Huh, usually I'm the one writing bytewise algorithms that carefully attempt to avoid allocation :) I'll put this into the code though, no use leaving performance on the table.

As I mentioned on Zulip, my current plan is to format the `MenuKey` into a string using Windows conventions. Actually wiring up the accelerators is a different issue.

It might be interesting to take a look at code which does Windows menus now in [xi-win](https://github.com/xi-editor/xi-win/blob/master/src/menus.rs). That will have to change of course.

druid-shell/src/mac/menu.rs
```diff
@@ -44,38 +43,79 @@ lazy_static! {
...
+fn make_key_equivalent(key: impl Into<MenuKey>) -> String {
+    // TODO: handle modifiers and other fun things
+    let key = key.into();
+    match key.key {
+        KeySpec::Char(c) => c.to_string(),
+        KeySpec::None => "".to_string(),
+    }
+}
+
+/// Strip the access keys from the menu strong.
```
jjl: Typo here, strong -> string.

PR #30, opened 2019-03-06
Proof of concept druid running on web

tedsta:
Hey Raph long time no chat!

I've been wanting to do some hobby web tools and since somebody already did the heavy lifting getting piet working on wasm I figured I'd give this a shot.

Is web something you are interested in targeting with Druid?

image

### DIFF
--- a/druid-shell/src/web/mod.rs
+++ b/druid-shell/src/web/mod.rs
@@ -32,7 +32,7 @@ use std::sync::{Arc, Mutex};
 
 use wasm_bindgen::prelude::*;
 use wasm_bindgen::JsCast;
-use web_sys::{window, HtmlCanvasElement};
+use web_sys::{HtmlCanvasElement};
 
 use piet::RenderContext;
 
@@ -87,8 +87,11 @@ pub struct WindowHandle(Weak<WindowState>);
 /// message may be sent to a stray window).
 #[derive(Clone)]
 pub struct IdleHandle {
+    window_state: WindowHandle,
     queue: Arc<Mutex<Vec<Box<IdleCallback>>>>,
 }
+
+//impl Send for IdleHandle { }
 
 trait IdleCallback: Send {
     fn call(self: Box<Self>, a: &Any);
@@ -121,7 +124,7 @@ impl Default for PresentStrategy {
 impl WindowState {
     // Renders but does not present.
     fn render(&self) {
-        let window = web_sys::window().expect("web_sys window");
+        let window = window();
         let ref mut canvas_ctx = *self.canvas_ctx.borrow_mut();
         canvas_ctx.clear_rect(0.0, 0.0, self.get_width() as f64, self.get_height() as f64);
         let mut piet_ctx = piet_common::Piet::new(
@@ -174,7 +177,6 @@ fn setup_web_callbacks(window_state: &Rc<WindowState>) {
             };
             state.handler.mouse(&event);
 
-            state.process_idle_queue();
             state.render();
         }) as Box<dyn FnMut(_)>);
         window_state.canvas
@@ -189,7 +191,6 @@ fn setup_web_callbacks(window_state: &Rc<WindowState>) {
 	    let mods = 0;
 	    state.handler.mouse_move(x, y, mods);
 
-            state.process_idle_queue();
             state.render();
         }) as Box<dyn FnMut(_)>);
         window_state.canvas
@@ -215,7 +216,6 @@ fn setup_web_callbacks(window_state: &Rc<WindowState>) {
             };
             state.handler.mouse(&event);
 
-            state.process_idle_queue();
             state.render();
         }) as Box<dyn FnMut(_)>);
         window_state.canvas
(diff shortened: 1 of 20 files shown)


raphlinus:
Hi @tedsta! Indeed, it's been a while.

Yes, targeting web is *definitely* something I'm interested in. I wanted to respond quickly before doing a full review. There is some foundational work in druid remaining that will affect ports, so it's a project management question whether we should limit the number of ports while that's churning, or bring all ports along. There's similarly a [Linux port](https://github.com/xi-editor/druid/compare/master...bobtwinkles:gtk) (using a gtk dependency) where we need to make some similar decisions.

I hope to review this in more detail soon; this week is challenging for finding chunks of time.

tedsta:
Take your time - I should emphasize the proof-of-concept-ness of this PR a bit more. Also note this is the first time I've even looked at wasm and `web_sys` stuff.

I switched over to the calculator example and while it renders ok it doesn't react to button presses properly, so I'll be looking at that next.

I think it could be helpful to have the web port in the churn so that assumptions aren't locked in that aren't valid on web. But I understand if it's too much - I could go play with the linux port instead.

Current state (questions are just things I'm asking myself):
- Everything except rendering and mouse events is stubbed out
- Relies on the DOM already having a canvas element named "canvas". Should we just create a canvas in the body? Assume the body is empty at start?
- There is no main loop, just set up the callbacks. The rust entry point actually exits. This makes the "idle queue" a bit problematic... my code assumes that idle callbacks will only be queued after some kind of UI event, but I that isn't valid when external things like network messages can update the GUI.

I need to do some learning here - I don't know when callbacks I register with `web_sys` are called. I assumed they are called on the main thread, and so if I hog the thread with a main loop, they would never be called? (haven't actually tested) Or maybe the callbacks will preempt my main loop?

- `std::time` doesn't work on wasm32-unknown-unknown, so I made a little `Timer` object with an implementation for wasm32 and another for everyone else.

- How do you exit the web app? Close the tab? Do web APIs give a way to do that?

druid-shell/src/web/mod.rs
```diff
@@ -166,7 +167,7 @@ fn setup_web_callbacks(window_state: &Rc<WindowState>) {
 
             let event = MouseEvent {
                 x: event.offset_x() as i32,
-                y: event.offset_x() as i32,
```
tedsta: .... figured out why half of the calc example buttons iddn't work... :roll_eyes:

msand:
Hello! This seems like a interesting project. I'm thinking you probably want to use request_animation_frame for the main loop: https://github.com/rustwasm/wasm-bindgen/blob/fcee4656923a1128d0a2faf781242ee2d7835db4/examples/request-animation-frame/src/lib.rs#L10-L14

https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.Window.html#method.request_animation_frame
This should allow the js event loop to function performantly and not hog the main ui thread.

tedsta:
Thanks for the tip @msand, idle callbacks now use `request_animation_frame(_)`

msand:
@tedsta You're welcome 👍Thanks for working on this! I just realized, if it's actually idle you want rather than high fps render loop, then you might want to change to requestIdleCallback instead: https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.Window.html#method.request_idle_callback

jjl:
This rebases cleanly against master and works (more or less).

I've pushed some changes to my [web-hidpi](https://github.com/jjl/druid/tree/web-hidpi) branch which is a continuation of your web branch:
- The DPI calculation caused 1/3 of the content in each dimension to be clipped, this is rectified
- Mouse positions are now translated into physical positions instead of css positions
- Added window resizing

jjl:
I have been thinking about next steps.

At the minute the event loop is handled through the window. I would like to move this to the run loop but it's not immediately clear what the correct way of doing this is, particularly with regard to supporting other platforms - where the cross platform api stops and the per-platform begins is not entirely clear to me at present.I am thinking to copy what is done in windows for now, but will probably wait to see what happens for multiple window support on other platforms before looking for an alternative solution.

Speaking of windows... We should decide what constitutes a window on the web backend. My current thoughts are that yes, it should be a canvas but no, it should not necessarily be full screen. Perhaps full screen can be a mode opted into in the same way one might for other platforms.

This leaves open a few questions:

* Should we map windows to canvas elements such that when multiple windows are supported on other platforms, it maps multiple canvas elements rendered in one document? (my view: yes)
* Should these be optionally individually resizable? (i think so, because we'd want that on other platforms)
* Should creating and closing a window map to creating and destroying a canvas? (seems logical)
* What are we to do about ownership of windows and the run loop? (???)

jjl:
Seems I may have spoken too soon, my master was out of date and the muggle merge has broken it. Perhaps going through it will answer some of my other questions

jjl:
I seem to have everything except Text working in my local copy - a shame we need text everywhere for the calculator.

It's a bit awkward trying to infer what the web structure should be from the windows version, but I've gotten most of it down, I think. I'm currently just working out if i can get a mutable canvas context ref to the WinCtxOwner without requesting the context afresh.

xStrom:
Now that #759 has landed, is there still a reason to keep this PR open?

raphlinus:
Nope, this was the basis of that, so safe to close. Thanks for the heads-up!

PR #32, opened 2019-03-12
Remove mac warnings

rylev:
Branched off of #31


rylev:
@raphlinus totally understand. I think given the state of the project and the rest of your work in the surrounding ecosystem, we should lean more on actual issues versus warnings as todos and TODO code comments.

PR #36, opened 2019-04-23
Support mouse scrolling

zaynetro:
Tested on mac. I am not certain whether this is the best design. Creating this MR to get some feedback and to let others play with scroll events.

At the moment there is one issue: scroll events start to arrive to the widget only after mouse is moved. It is most certainly due to the usage of `self.layout_ctx.hot`. I will take a look at it later.

Added a mouse example to play around with mouse events.

image


zaynetro:
Thanks for the quick response!

I have unified scroll and hscroll methods in Widget trait. I am not familiar with `hasPreciseScrollingDeltas` at all so I just followed the same approach as in makepad. The docs for `scrollingDeltaY` say:

> When hasPreciseScrollingDeltas is false, multiply the value returned by this method by the line or row height.

This leaves me with a question if we should handle it in druid shell or let the user decide.

Few other things:
* Some events are defined in `druid-shell/src/window.rs` and some in `druid/src/widget/mod.rs`. Should there be a single place for the events? Or is there a separation between some of those?
* I am definitely lost with units :) Should they be `i32`, `f32` or `f64`?

raphlinus:
This looks pretty good and I'm willing to merge it. I was thinking that x and y scroll would be unified at the druid-shell level also, but this can be followup work - quite a bit of that is needed to get key events right, and I'm interested supporting multitouch, using an interface like [W3C pointer events](https://www.w3.org/TR/pointerevents/) while we're at it.

Thanks!

zaynetro:
I think it is a good idea to begin with a minimal working version and continue based on user needs. Feel free to merge whenever you have time! I will not be pushing any new changes in this MR.

 If I have time I might experiment with scrollable view as I would like to write a basic GUI app where scroll is a must have.

raphlinus:
Thanks!

PR #41, opened 2019-05-25
Expose BoxConstraints members, add some Debug impls

cmyr:
Without these exposed it's impossible to implement a widget in an external crate.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -132,7 +132,7 @@ enum AnimState {
     AnimFrameRequested,
 }
 
-#[derive(Clone, Copy)]
+#[derive(Clone, Copy, Debug)]
 pub struct BoxConstraints {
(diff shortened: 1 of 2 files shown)


raphlinus:
I know it's possible to implement widgets in external crates, the [synthesizer does it](https://github.com/raphlinus/synthesizer-io/blob/master/synthesizer-io-win/src/ui/patcher.rs#L144), but that's because it uses public methods like `constrain`. I'm slightly curious what you need, but can easily imagine that implementing *container* widgets externally would be hard; the synthesizer is only leaf widgets.

cmyr:
Hmm okay I'll double check if I actually need this before merging.

cmyr:
I've added `BoxConstraints::max()/min()` instead of exposing the members directly.

PR #42, opened 2019-06-04
Fix order of makeKeyAndOrderFront call

raphlinus:
The action of making the window key only really works after the app is
activated. This patch moves both actions into the Window::show() method.

Also snaps deps to latest, which resolves deprecation warnings.

### DIFF
--- a/druid-shell/src/mac/mod.rs
+++ b/druid-shell/src/mac/mod.rs
@@ -136,7 +136,6 @@ impl WindowBuilder {
             window.autorelease();
             window.cascadeTopLeftFromPoint_(NSPoint::new(20.0, 20.0));
             window.setTitle_(make_nsstring(&self.title));
-            window.makeKeyAndOrderFront_(nil);
             // TODO: this should probably be a tracking area instead
             window.setAcceptsMouseMovedEvents_(YES);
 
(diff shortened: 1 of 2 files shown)


raphlinus:
A tiny polish papercut is that when you launch the app from the command line it first shows the window, generally under the terminal, then has a second transition to bring the window to front. This probably won't be an issue when we bundle into an app. I'm sure there's a right way to do it, but I couldn't figure it out quickly.

raphlinus:
Continuing to fiddle with this, I'm now doubting `makeKeyAndOrderFront()`, as doing a separate `orderFrontRegardless()` and `makeKeyWindow()` seems not to have the extra transition. But I have no theory why this should be, or what's the right way.

druid-shell/src/mac/mod.rs
```diff
@@ -136,7 +136,6 @@ impl WindowBuilder {
             window.autorelease();
             window.cascadeTopLeftFromPoint_(NSPoint::new(20.0, 20.0));
             window.setTitle_(make_nsstring(&self.title));
-            window.makeKeyAndOrderFront_(nil);
```
cmyr: an alternative solution is to move this line down to the end of this function, after we've done all of our view setup.

I'm not totally sure which one is 'better', but both fix the issue of not having mouse when we launch.

PR #48, opened 2019-06-11
Setup rustfmt

cmyr:
- Add rustfmt.toml
- run rustfmt
- add rustfmt check to CI


cmyr:
this is going to smash #44 but I'll sort that out, would like to rip this bandaid off now.

PR #49, opened 2019-06-11
First stab at an environment

cmyr:
Not sure about this, but was fun to put together. The basic idea is that there's an `Environment` struct that gets passed up into the layout & draw calls.

Currently this is only used for theming. How _this_ works is a bit experimental; it's based around the idea of typed keys and values, so we can store various copyable types in a collection but use keys that know their generic type.

We define a bunch of default keys/values for things like background color, text color, text size etc. Widgets can define their own default values for new things they need; these are registered into the global environment the first time a new kind of widget is added to the graph.

When you're creating your `UiState`, you can then go and specify overrides for any known keys.

### DIFF
--- a/examples/calc.rs
+++ b/examples/calc.rs
@@ -20,6 +20,7 @@ extern crate druid_shell;
 use druid_shell::platform::WindowBuilder;
 use druid_shell::win_main;
 
+use druid::environment::text;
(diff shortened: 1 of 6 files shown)


cmyr:
closing this for now, some ideas in here that might be worth revisiting but that will be new work.

PR #54, opened 2019-06-18
Set level to 4 on window creation (#43)

### LINKED ISSUE #43: Make mac Window creation smoother

futurepaul:
This fixes #43 as far as I can tell. I just followed @ratmice's ideas and this seemed the correct place to set the level.

On my Mac: without this level it flashes behind the terminal before going in front, with level 4 set it starts up in front of the terminal as expected.

### DIFF
--- a/druid-shell/src/mac/mod.rs
+++ b/druid-shell/src/mac/mod.rs
@@ -133,6 +133,7 @@ impl WindowBuilder {
                 NO,
             );
 
+            window.setLevel_(4);
             window.autorelease();
             window.cascadeTopLeftFromPoint_(NSPoint::new(20.0, 20.0));
             window.setTitle_(make_nsstring(&self.title));
--- a/druid-shell/src/mac/mod.rs
+++ b/druid-shell/src/mac/mod.rs
@@ -408,7 +408,8 @@ impl WindowHandle {
             current_app.activateWithOptions_(NSApplicationActivateIgnoringOtherApps);
             if let Some(ref nsview) = self.nsview {
                 let window: id = msg_send![*nsview.load(), window];
-                window.makeKeyAndOrderFront_(nil)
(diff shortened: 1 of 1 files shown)


futurepaul:
Wait!

I'm clearly not paying close enough attention. The window starts in front, but also it remains in front forever. This is not correct, sorry!

futurepaul:
Okay, sorry for all the noise. This works now. (I believe!)

cmyr:
So this is an improvement, but I would second the concern about `setLevel_(0)`. I briefly fired up LLDB and didn't see any unexpected `order:` calls, so I'm not sure about that hunch. I don't have anything more helpful to suggest, though. :(

futurepaul:
I'm pretty stumped on how to do it with that extra setlevel! Fwiw, the [hello_world example here](https://github.com/servo/core-foundation-rs/blob/2aac8fb85b5b114673280e273c04219c0c360e54/cocoa/examples/hello_world.rs) shows the same behind-then-in-front behavior, so I don't think we're doing anything strange that's causing this in the first place.

ratmice:
one more wild guess and i'm stumped,
I would try moving the makeKeyAndOrderFront call, from before app.run()
into a function called from the app.setDelegate_(delegate) method applicationDidFinishLaunching,
or WillFinish...

There is some delegate examples here (but those are window delegates):
https://github.com/servo/core-foundation-rs/blob/2aac8fb85b5b114673280e273c04219c0c360e54/cocoa/examples/fullscreen.rs

using the delegate function applicationDidFinishLaunching or ...WillFinish...
IIRC the main runloop isn't actually running/doing anything with events until app.run().

https://developer.apple.com/documentation/appkit/nsapplicationdelegate/1428385-applicationdidfinishlaunching?language=objc

cmyr:
This is resolved in #309.

PR #59, opened 2019-06-23
Receive right mouse events on mac

cmyr:
At some point we will have to look into making this support contextual menus, but for now I think it makes sense to plumb these through as just mouse events.


cmyr:
I realized a better way to handle the mouse-up stuff. This is still a bit messy, but much better.

PR #60, opened 2019-06-25
Druid v0.3.0

cmyr:
- reexport druid_shell as 'shell'
- Update for removal of Coord & Point from piet
- make expose updated kurbo
- update examples

Before publishing, I'd like to discuss which bits of druid_shell should be exported directly. My gut feeling is that it's more or less 'all of it'; consumers of druid shouldn't need to reach into `shell` for anything? What am I missing?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -173,14 +173,14 @@ dependencies = [
 
 [[package]]
 name = "druid"
-version = "0.2.0"
+version = "0.3.0"
 dependencies = [
- "druid-shell 0.2.0",
+ "druid-shell 0.3.0",
 ]
 
 [[package]]
 name = "druid-shell"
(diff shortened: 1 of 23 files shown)


raphlinus:
It's a valid question how much of the shell to expose. If we do, for example, VST's, then I think it *will* be necessary to do deal with platform pointers to NSView or HWND or whatever. But I agree for the mainstream case it can be simplified.

cmyr:
I've taken a minimal approach: making `druid_shell` available as `druid::shell`, and exposing `WindowBuilder` from the root of `shell`, so that to get going you just need `use druid::shell::{runloop, WindowBuilder}`. (I'm exporting `win_main` as `runloop`).

PR #69, opened 2019-07-10
A scroll container

raphlinus:
Adds a scrollable container. This implementation is quite basic as it
does not draw scroll bars or allow them to be dragged by mouse clicking,
but the fundamental scroll mechanics (with wheel) should be in place.

Scrolling is discussed in #5 among other things.

### DIFF
--- a/src/widget/scroll.rs
+++ b/src/widget/scroll.rs
@@ -1,0 +1,114 @@
+// Copyright 2019 The xi-editor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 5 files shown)


raphlinus:
This works on macOS but is broken on Windows due to https://github.com/linebender/piet/pull/64.

src/widget/scroll.rs
```diff
@@ -0,0 +1,114 @@
...
+use crate::{
+    Action, BaseState, BoxConstraints, Data, Env, Event, EventCtx, LayoutCtx, PaintCtx, Point,
+    Rect, Size, UpdateCtx, Vec2, Widget, WidgetPod,
+};
+
+use crate::piet::{FillRule, RenderContext};
+
+use crate::kurbo::Affine;
+
+pub struct Scroll<T: Data> {
```
cmyr: I'd like to get in the habit of having some documentation for all public types.

raphlinus:
I'll file an issue with the remaining refinements.

PR #70, opened 2019-07-10
Add "handled" state to event

raphlinus:
Allow an event to be handled, blocking further propagation.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -145,6 +145,7 @@ pub struct EventCtx<'a> {
     window: &'a WindowHandle,
     base_state: &'a mut BaseState,
     had_active: bool,
+    is_handled: bool,
 }
 
 pub struct UpdateCtx {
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -526,6 +530,17 @@ impl<'a> EventCtx<'a> {
     pub fn window(&self) -> &WindowHandle {
         &self.window
     }
+
+    /// Set the event as "handled", which stops its propagation to other
+    /// widgets.
+    pub fn set_handled(&mut self) {
```
cmyr: There's an asymmetry here between `set_handled` and `set_active`. I prefer this to not take an argument, but maybe `set_active` doesn't need to either?

raphlinus: No, `set_active` can definitely take `false`, for example on mouse-up in a button. I don't think it makes any sense to unset handled. Open to ideas of improving naming/docs tho.

PR #72, opened 2019-07-10
Expose setting the cursor

cmyr:
This adds support for the set of cursors shared between macOS
and Windows. It does not include support for custom cursors.


cmyr:
At least on Cocoa there's the convention of treating the cursor as a stack, with `push` and `pop` operations for this use case. We could consider implementing something similar.

PR #77, opened 2019-07-12
 Rely less on interior mutability in druid-shell

### LINKED ISSUE #76: Panics due to wndproc reentrancy
You can provoke a panic in the master branch by running the [sample](https://github.com/xi-editor/druid/blob/88366718606109114645da3d1d0d56124503beea/examples/sample.rs) example, selecting File Open from the menu, then doing a dynamic monitor change (for example, unplugging an external monitor). This is because Windows requires the wndproc to be reentrant, and our implementation uses a (panicking) `RefCell::borrow_mut()` for safety.

There's a [Reddit thread](https://www.reddit.com/r/rust/comments/cbwlv3/reentrancy_mut_and_undefined_behavior/) which goes into some more detail, including pointers to how winit handles this. There's also some [discussion on the Zulip](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/interior.20mutability.20in.20druid-shell.20interface).

The upshot is that we want to refactor the way we handle interior mutability. In particular, druid-shell should handle it, so the `WinHandler` can be called mutably and not need its own interior mutability. There's also followup work to make the file dialog not hold long-lived borrows, but that's significantly lower priority.

This came up because we need a good way to plumb additional resources, particularly a factory for creating text layouts, to user methods.

raphlinus:
This makes methods on `WinHandler` mostly `&mut self`, and does `try_borrow_mut` for this state in the wndproc handler.

It's not complete, as there are more methods that should move from `WindowHandle` to `WinCtx`, including dpi calculation, but it does fix the panics in the attached bug. These further refinements should be possible without significant reorganization. And at some point I should be able to add text measurement in layout and event contexts.

Fixes #76

### DIFF
--- a/druid-shell/src/mac/mod.rs
+++ b/druid-shell/src/mac/mod.rs
@@ -46,7 +46,7 @@ use piet_common::{Piet, RenderContext};
 use crate::keyboard::{KeyEvent, KeyModifiers};
 use crate::platform::dialog::{FileDialogOptions, FileDialogType};
 use crate::util::make_nsstring;
-use crate::window::{Cursor, MouseButton, MouseEvent, WinHandler};
+use crate::window::{Cursor, MouseButton, MouseEvent, WinCtx, WinHandler};
 use crate::Error;
 
 use util::assert_main_thread;
@@ -88,8 +88,13 @@ impl<F: FnOnce(&dyn Any) + Send> IdleCallback for F {
 }
 /// This is the state associated with our custom NSView.
 struct ViewState {
+    nsview: WeakPtr,
     handler: Box<dyn WinHandler>,
     idle_queue: Arc<Mutex<Vec<Box<dyn IdleCallback>>>>,
+}
+
+struct WinCtxImpl<'a> {
+    nsview: &'a WeakPtr,
 }
 
 impl WindowBuilder {
@@ -149,18 +154,21 @@ impl WindowBuilder {
                 _ => (),
             }
             content_view.addSubview_(view);
-            let handle = WindowHandle {
-                nsview: Some(WeakPtr::new(view)),
-                idle_queue,
-            };
             let view_state: *mut c_void = *(*view).get_ivar("viewState");
             let view_state = &mut *(view_state as *mut ViewState);
+            let handle = WindowHandle {
+                nsview: Some(view_state.nsview.clone()),
+                idle_queue,
+            };
             (*view_state).handler.connect(&crate::window::WindowHandle {
                 inner: handle.clone(),
             });
+            let mut ctx = WinCtxImpl {
+                nsview: handle.nsview.as_ref().unwrap(),
+            };
             (*view_state)
                 .handler
-                .size(frame.size.width as u32, frame.size.height as u32);
+                .size(frame.size.width as u32, frame.size.height as u32, &mut ctx);
 
             Ok(handle)
         }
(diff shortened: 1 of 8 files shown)


druid-shell/src/window.rs
```diff
@@ -35,6 +35,14 @@ impl Deref for WindowHandle {
     }
 }
 
+/// A context supplied to most `WinHandler` methods.
+pub trait WinCtx {
```
cmyr: What is the relationship between this and `WindowHandle`?

src/lib.rs
```diff
@@ -24,21 +24,20 @@ mod lens;
 mod value;
 
 use std::any::Any;
-use std::cell::RefCell;
-use std::collections::BTreeMap;
-use std::ffi::OsString;
-use std::mem;
-use std::ops::{Deref, DerefMut};
-use std::time::Instant;
```
cmyr: thanks for the cleanup.

druid-shell/src/windows/mod.rs
```diff
@@ -240,18 +253,31 @@ impl MyWndProc {
...
+        eprintln!(
+            "dropped message 0x{:x}, hwnd={:?}, wparam=0x{:x}, lparam=0x{:x}",
+            msg, hwnd, wparam, lparam
+        );
+    }
+}
+
 impl WndProc for MyWndProc {
-    fn connect(&self, handle: &WindowHandle, state: WndState) {
+    fn connect(&self, handle: &WindowHandle, mut state: WndState) {
```
cmyr: So just to understand, doing anything before calling `connect` will be a crash?

druid-shell/src/windows/mod.rs
```diff
@@ -268,300 +294,378 @@ impl WndProc for MyWndProc {
...
+                } else {
+                    self.log_dropped_msg(hwnd, msg, wparam, lparam);
                 }
                 Some(0)
             },
             WM_COMMAND => {
-                self.handler.command(LOWORD(wparam as u32) as u32);
+                if let Ok(mut s) = self.state.try_borrow_mut() {
+                    let s = s.as_mut().unwrap();
+                    let mut ctx = WinCtxImpl {
```
cmyr: the fact that this gets repeated so often makes me think it might be worth having a `WinCtxImpl::new` fn, save us a couple lines in each match arm.

druid-shell/src/mac/mod.rs
```diff
@@ -88,10 +88,15 @@ impl<F: FnOnce(&dyn Any) + Send> IdleCallback for F {
...
 /// This is the state associated with our custom NSView.
 struct ViewState {
+    nsview: WeakPtr,
     handler: Box<dyn WinHandler>,
     idle_queue: Arc<Mutex<Vec<Box<dyn IdleCallback>>>>,
 }
 
+struct WinCtxImpl<'a> {
+    nsview: &'a WeakPtr,
+}
```
cmyr: Future thought: we stash our host view, but I wonder if we won't also want to stash a reference to the window itself? we'll want to have access to this at some point.

druid-shell/src/window.rs
raphlinus: There's a bunch of overlap, but the difference is ownership - `WindowHandle` has shared ownership and the `WinCtx` is mutable access to state singly owned by the platform window handler. I'd like to basically migrate as much as possible to the latter, possibly even get rid of the `connect` method.

druid-shell/src/windows/mod.rs
raphlinus: Anything before `connect` is considered part of the construction of the struct. This is basically just a way to get a window handle into MyWndProc - for reasons that are not very interesting it needs to be created before that window handle is available, though I wouldn't be surprised if there was a way to change the order up a bit. It'll be a crash if it tries to do anything requiring the handle. I think that's not exposed to client code though.

druid-shell/src/windows/mod.rs
raphlinus: That'd be nice, but I tried a couple variants and wasn't able to make it work. That function would need to hold a reference to `&mut self`, and calling the handler method also requires `&mut self`.

There could well be a way to reduce the duplication, I just don't see it.

druid-shell/src/mac/mod.rs
raphlinus: Why not just call NSView's [window](https://developer.apple.com/documentation/appkit/nsview/1483301-window?language=objc) method? I think that's what most native code would do.

PR #78, opened 2019-07-15
Update Slider and ProgressBar widgets for muggle

futurepaul:
Both of these are "stateless" from a use perspective, but slider needs to keep track of its width so it can calculate a new value during a mouse event. I believe the plan is to wrap these in a similar way to DynLabel so they can accept a closure.


src/widget/progress_bar.rs
```diff
@@ -14,67 +14,69 @@
...
         bc: &BoxConstraints,
-        _children: &[Id],
-        _size: Option<Size>,
-        _ctx: &mut LayoutCtx,
-    ) -> LayoutResult {
-        LayoutResult::Size(bc.constrain((bc.max.width, BOX_HEIGHT)))
+        _data: &f64,
+        _env: &Env
+    ) -> Size {
+        bc.constrain((bc.max.width, bc.max.height))
```
cmyr: I think you can just write `bc.max()` here.

src/widget/progress_bar.rs
```diff
@@ -14,67 +14,69 @@
...
-}
+
+    fn update(
+        &mut self,
+        ctx: &mut UpdateCtx,
+        _old_data: Option<&f64>,
+        _data: &f64,
+        _env: &Env,
+    ) {
+        ctx.invalidate();
```
cmyr: I think this is unnecessary. `update` is currently poorly described and I'm not totally clear what its role is supposed to be, but this should work without it.

src/widget/slider.rs
```diff
@@ -14,93 +14,104 @@
...
-            value: initial_value,
-        }
-    }
 
-    pub fn ui(self, ctx: &mut Ui) -> Id {
-        ctx.add(self, &[])
-    }
-}
+impl Widget<f64> for Slider {
+    fn paint(&mut self, paint_ctx: &mut PaintCtx, base_state: &BaseState, data: &f64, _env: &Env) {
```
cmyr: I think for control widgets we should always have some visual indication of 'hover' and 'active' state. This doesn't need to be fancy now, but might be fun to paint the knob differently in these two cases?

src/widget/slider.rs
```diff
@@ -14,93 +14,104 @@
...
-                .max(0.0)
-                .min(1.0);
+        None
+    }
 
-            ctx.send_event(self.value);
-            ctx.invalidate();
-        }
+    fn update(&mut self, ctx: &mut UpdateCtx, _old_data: Option<&f64>, _data: &f64, _env: &Env) {
+        ctx.invalidate();
```
cmyr: unnecessary, as per last comment

raphlinus:
The `update` methods should call `invalidate`. Apologies this isn't clear from the current code.

futurepaul:
I believe I covered all your comments, other than the padding idea.

Adding hover and active made me realize I was doing some things slightly wrong with how the slider was behaving (it now will go to your cursor on click, even before you move your mouse, which seems like standard behavior).

But! It also exposes how roughly the slider is implemented. I'm not differentiating between the knob and the track — the whole widget is "hot" if the cursor is inside its layout box. Any ideas how to improve that situation? Should perhaps the knob be its own widget?

futurepaul:
> The `update` methods should call `invalidate`. Apologies this isn't clear from the current code.

Just to clarify: you're saying I should add back the ctx.invalidate() to those update methods?

cmyr:
I think we're still figuring out how composability should work in situations like this. I would be tempted to just do manually hit testing on the knob to determine how we paint it, but I haven't thought about this hard.

futurepaul:
Okay, added a manual hit test for the knob. It adds a little bit more state but doesn't feel too bad to me.

cmyr:
Another chance to add some more state: currently, when you mouse down on the knob it immediately moves the knob to be centered on the click location. I think that if the user clicks the knob, it shouldn't move until the mouse does. 😇

futurepaul:
Thanks for being picky about behavior, this def feels better now. If you click the knob it doesn't jump, and if you click and drag the knob it maintains the offset from knob center (which is how it works on mac).

cmyr:
Here's a funny one: if I click and drag the knob, and with the mouse down move the cursor over the buttons, they highlight. I don't think this is your fault, but something to keep in mind.

@futurepaul are you happy with this as-is, or did you want to keep fiddling with styling?

futurepaul:
Up to y'all: I was working on styling stuff in a different branch, but I can add styles to ProgressBar and Slider for this PR if you'd like!

raphlinus:
I'm in favor of getting stuff into the tree, so would be happy to see this merged and refine styling in a followup.

PR #79, opened 2019-07-17
Send key_up/key_down when lone modifiers change (macOS)

cmyr:
This introduces the convention that we send key_up and key_down when
modifier keys are pressed on their own. This implements this behaviour on
macOS, but not on windows, because I'm not sure what's going on there.

if we don't end up keeping muggle (which seems very unlikely) we'll have to backport this to master.


raphlinus:
I believe Windows sends these as-is (Windows tends to be lower level) but at some point we'll want to go through these carefully and verify that we're getting pretty much the same events on both platforms.

PR #81, opened 2019-07-19
KeyEvent ergonomics improvements

cmyr:
For discussion.

I was finding myself writing a lot of boilerplate in runebender for handling various combinations of modifiers. This is an attempt to simplify that somewhat, by adding an alternative form of `KeyModifiers`, `ModState`, that is an exhaustive enum of modifier combinations. This means we can replace code like,

```rust
match event {
    e if e.text() == Some("a") && e.modifiers.meta && !e.modifiers.shift => ...
}
```

with
```rust
match event {
    e if e == ("a", ModState::Meta) => ...
}
```

or, for a mouse event,
```rust
match event {
    e if e.mods == ModState::AltShift => ...
}
```

And for added convenience, I've gotten a bit fancy, and you can also do,

```rust
if event == (KeyCode::Escape, ModState::ANY) {
...
}
```

Which might be a bit much, I'm not sure, but it's in a separate commit so easy to drop if we don't like.

I'd like to try actually using this API a bit before merging, but wanted to open a PR for comment.

### DIFF
--- a/druid-shell/src/keyboard.rs
+++ b/druid-shell/src/keyboard.rs
@@ -33,6 +33,8 @@ use std::fmt;
 /// assert!(event != ("a", ModState::Alt));
 /// // can also match against the raw keycode + modstate.
 /// assert!(event == (KeyCode::KeyA, ModState::AltCtrl));
+/// // match any modstate:
+/// assert!(event == (KeyCode::KeyA, ModState::ANY));
 /// ```
 #[derive(Debug, Clone, Copy)]
 pub struct KeyEvent {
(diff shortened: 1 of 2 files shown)


druid-shell/src/keyboard.rs
```diff
@@ -679,6 +740,159 @@ impl fmt::Debug for KeyModifiers {
...
+        KeyModifiers {
+            alt,
+            ctrl,
+            meta,
+            shift,
+        }
+    }
+}
+
+impl std::cmp::PartialEq<(&'static str, ModState)> for KeyEvent {
```
raphlinus: Is this really eq? I think I would be slightly more comfortable with it just being a function, like `key.matches("a", ModState::Shift)` or whatnot.

cmyr:
Yea, `KeyMatcher` might make sense, this started out fairly small in scope and then kind of snowballed.

raphlinus:
I know how that goes :)

PR #83, opened 2019-07-22
Add text layout capability to widget contexts

raphlinus:
With this patch, the event, update, and layout contexts all have a `text` method which allows the creation of text layout objects. There's also a bit of documentation and other cleanup.

This patch is part of a general migration of functionality into methods on mutable contexts which are passed down. This pattern is flexible and robust than the previous pattern of stashing a shared reference to a "window handle" and using interior mutability.

Thanks to @SimonSapin for helping with some particularly tricky borrow checker issues.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -139,7 +139,9 @@ pub struct PaintCtx<'a, 'b: 'a> {
     pub render_ctx: &'a mut Piet<'b>,
 }
 
-pub struct LayoutCtx {}
+pub struct LayoutCtx<'a, 'b: 'a> {
+    text: &'a mut druid_shell::window::Text<'b>,
+}
 
 pub struct EventCtx<'a> {
     window: &'a WindowHandle,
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -139,16 +139,33 @@ pub struct PaintCtx<'a, 'b: 'a> {
...
+/// A mutable context provided to event handling methods of widgets.
+///
+/// Widgets should call [`invalidate`] whenever an event causes a change
+/// in the widget's appearance, to schedule a repaint.
+///
+/// [`invalidate`]: #method.invalidate
+pub struct EventCtx<'a, 'b> {
+    win_ctx: &'a mut dyn WinCtx<'b>,
+    // TODO: migrate most usage of `WindowHandle` to `WinCtx` instead.
     window: &'a WindowHandle,
```
cmyr: I might just do this as part of this patch, if it isn't too much work.

raphlinus: Do an invalidate? I ***really*** don't want to, it feels like widgets should have the opportunity to avoid doing extra work.

cmyr: no, I mean get rid of `WIndowHandle`.

raphlinus: Ah, heh. This means that an event can't change the cursor, until I add that functionality. Separate PR maybe?

cmyr: sure, sounds reasonable.

cmyr: (I was maybe confused, it looked like there was a `WindowHandle` inside the `WinCtx`.)

PR #86, opened 2019-07-26
WIP TextBox update for muggle

futurepaul:
Problems (that I'm aware of):
1. Need latest Piet to position the cursor correctly
2. Only works with String (need help implementing some sort of adapter to convert between types in and out) — that's why I built a separate demo for it
3. Styling is pretty pixel-fiddly, would appreciate any ideas to make it based on fewer constants.

### DIFF
--- a/src/widget/textbox.rs
+++ b/src/widget/textbox.rs
@@ -14,114 +14,114 @@
 
 //! A textbox widget.
 
-use crate::widget::Widget;
 use crate::{
-    BoxConstraints, HandlerCtx, Id, KeyCode, KeyEvent, LayoutCtx, LayoutResult, MouseEvent,
-    PaintCtx, Ui,
+    Action, BaseState, BoxConstraints, Env, Event, EventCtx, KeyCode, LayoutCtx, PaintCtx,
+    UpdateCtx, Widget,
 };
 
-use crate::kurbo::{Line, Rect, Size, Vec2};
+use crate::kurbo::{Line, Point, RoundedRect, Size, Vec2};
 use crate::piet::{
     Color, FillRule, FontBuilder, Piet, RenderContext, Text, TextLayout, TextLayoutBuilder,
 };
 
-const ACTIVE_BORDER_COLOR: Color = Color::rgb24(0xff_00_00);
-const INACTIVE_BORDER_COLOR: Color = Color::rgb24(0x55_55_55);
+const BACKGROUND_GREY_LIGHT: Color = Color::rgba32(0x3a_3a_3a_ff);
+const BORDER_GREY: Color = Color::rgba32(0x5a_5a_5a_ff);
+const PRIMARY_LIGHT: Color = Color::rgba32(0x5c_c4_ff_ff);
+
 const TEXT_COLOR: Color = Color::rgb24(0xf0_f0_ea);
 const CURSOR_COLOR: Color = Color::WHITE;
 
 const BOX_HEIGHT: f64 = 24.;
-const BORDER_WIDTH: f64 = 2.;
+const FONT_SIZE: f64 = 14.0;
+const BORDER_WIDTH: f64 = 1.;
+const PADDING_TOP: f64 = 5.;
+const PADDING_LEFT: f64 = 4.;
 
+#[derive(Debug, Clone)]
 pub struct TextBox {
-    text: String,
     width: f64,
-    font: Option<<<Piet<'static> as RenderContext>::Text as Text>::Font>,
 }
 
 impl TextBox {
-    pub fn new(default_text: Option<String>, width: f64) -> TextBox {
-        TextBox {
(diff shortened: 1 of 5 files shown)


raphlinus:
/AzurePipelines run

I *think* the above is necessary because it's only set to run PR's to the master branch from external contribs.

src/widget/textbox.rs
```diff
@@ -133,34 +133,58 @@ impl Widget for TextBox {
...
+                    ctx.set_active(false);
+                }
+                ctx.invalidate();
             }
-            event if event.key_code.is_printable() => {
-                self.text.push_str(event.text().unwrap_or(""))
+            Event::KeyDown(key_event) => {
+                match key_event {
+                    event if event.key_code == KeyCode::Backspace => {
+                        let mut text = data.clone();
```
cmyr: I think we can just do `data.pop()`? I forget exactly how this works, but I think that the state that is passed into `event` can be mutated directly, although we might want this to be an `Arc<String>`. @raphlinus can clarify?

(It likely goes without saying that the actual logic for what backspace does is complicated, and is definitely not just popping the last unicode scalar, but this is okay for now)

src/widget/textbox.rs
```diff
@@ -133,34 +133,58 @@ impl Widget for TextBox {
...
+                        text.pop();
+                        *data = text.to_string();
+                    }
```
cmyr: we should also be handling at least the arrow keys, here.

src/widget/textbox.rs
raphlinus: Yes, we should be able to do a `data.pop()` here. I have the `clone` dance in places where data has interior mutability (in some cases, `make_mut` might be better, but we might want to experiment, as refcount == 1 might be rare in practice).

Whether this should be `String` or `Arc<String>` is a pretty complex question. If it's used for relatively small fields I don't think it matters much and would lean toward the former for ergonomics. A general heuristic is that any time you get a chunk of data big enough you don't want to compare it when processing `update`, wrap it in `Arc`.

Regarding backspace, there is a big discussion to be had about this. I have on my local machine an embryonic "xi-text-core" crate that has the backspace logic factored out, and I know @cmyr has also been exploring this. I'd like to do something like that, but there are some really big decisions to be made; for example, this code currently depends on xi-rope, and that seems possibly too heavyweight for this. I definitely don't want to block basic text widget functionality on figuring out those big questions.

raphlinus:
The cursor moving on space is fixed in https://github.com/linebender/piet/pull/69.

Cursor blinking can be done by the existing animation code, but with performance lossage; basically it has to be animating every frame. A more power-efficient strategy is to wire up timers starting at druid-shell and then up through the event system. Alternatively, we can implement the timer API in druid now and use animation to back it, with an open issue to replace it with timers later.

Agree with keeping the cursor visible. That I don't think is too hard, even with the very primitive text measurement facilities.

Even doing arrow keys will require better text measurement than what piet provides now. That's work that should be done on piet soon.

raphlinus:
@cmyr This seems like a good checkpoint to me. As we discussed on Zulip, there are tons more things that can be done, but anything blocking this going in now?

cmyr:
let me fire it up quickly but I expect I'm happy.

src/widget/textbox.rs
```diff
@@ -133,34 +139,54 @@ impl Widget for TextBox {
...
+        event: &Event,
+        ctx: &mut EventCtx,
+        data: &mut String,
+        _env: &Env,
+    ) -> Option<Action> {
         match event {
-            event if event.key_code == KeyCode::Backspace => {
-                self.text.pop();
+            Event::MouseDown(_) => {
+                if ctx.is_hot() {
```
raphlinus: I believe with the new focus logic, what you want to do here is just `ctx.request_focus()`.

src/widget/textbox.rs
```diff
@@ -14,114 +14,120 @@
...
-            ACTIVE_BORDER_COLOR
+impl Widget<String> for TextBox {
+    fn paint(
+        &mut self,
+        paint_ctx: &mut PaintCtx,
+        base_state: &BaseState,
+        data: &String,
+        _env: &Env,
+    ) {
+        let is_active = base_state.is_active();
```
raphlinus: With the new focus stuff, I think this should be:

```rust
    let has_focus = base_state.has_focus();
```

(Now I'm slightly worried that I'm exposing inner details on the event flow by calling this "has_focus" instead of "is_focussed". Would love feedback on that)

futurepaul: I don't know about inner details, but: I guess is_focussed is more consistent with is_hot / is_active. If you're relying on auto complete, if you type is_ you get all the relevant suggestions?

futurepaul:
Will do!

raphlinus:
Looks good, and the rebase is fine, but it's failing CI because of the rustfmt check.

PR #87, opened 2019-07-27
Improved set_cursor handling

raphlinus:
Move druid-shell handling from WinHandle to WinCtx.

Add a `set_cursor` method to `EventCtx`. Previously you'd get a window
handle and call `set_cursor` on that, but it's been removed.

The `EventCtx::set_cursor` method doesn't immediately delegate it
to the WinCtx, but instead defers it to the end of event propagation.

Don't set a default cursor in the window class on Windows (this
doesn't work). Instead expect it to be set in the mouse move handler,
and set the default to arrow when dispatching the mouse move event.

### DIFF
--- a/druid-shell/src/window.rs
+++ b/druid-shell/src/window.rs
@@ -48,6 +48,9 @@ pub trait WinCtx<'a> {
 
     /// Get a reference to an object that can do text layout.
     fn text_factory(&mut self) -> &mut Text<'a>;
+
+    /// Set the cursor icon.
+    fn set_cursor(&mut self, cursor: &Cursor);
 }
 
 /// App behavior, supplied by the app.
@@ -165,6 +168,7 @@ pub enum MouseButton {
 //both Windows and macOS. We may want to provide polyfills for various additional cursors,
 //and we will also want to add some mechanism for adding custom cursors.
 /// Mouse cursors.
+#[derive(Clone)]
 pub enum Cursor {
     /// The default arrow cursor.
     Arrow,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -151,6 +151,7 @@ pub struct LayoutCtx<'a, 'b: 'a> {
 /// [`invalidate`]: #method.invalidate
 pub struct EventCtx<'a, 'b> {
     win_ctx: &'a mut dyn WinCtx<'b>,
+    cursor: &'a mut Option<Cursor>,
     // TODO: migrate most usage of `WindowHandle` to `WinCtx` instead.
     window: &'a WindowHandle,
     base_state: &'a mut BaseState,
@@ -254,6 +255,7 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
         let had_active = self.state.has_active;
         let mut child_ctx = EventCtx {
             win_ctx: ctx.win_ctx,
+            cursor: ctx.cursor,
             window: &ctx.window,
             base_state: &mut self.state,
             had_active,
@@ -367,8 +369,13 @@ impl<T: Data> UiState<T> {
     fn do_event(&mut self, event: Event, win_ctx: &mut dyn WinCtx) -> bool {
         // should there be a root base state persisting in the ui state instead?
         let mut base_state = Default::default();
+        let mut cursor = match event {
+            Event::MouseMoved(..) => Some(Cursor::Arrow),
+            _ => None,
+        };
         let mut ctx = EventCtx {
             win_ctx,
+            cursor: &mut cursor,
             window: &self.handle,
             base_state: &mut base_state,
             had_active: self.root.state.has_active,
@@ -378,6 +385,9 @@ impl<T: Data> UiState<T> {
         let _action = self.root.event(&event, &mut ctx, &mut self.data, &env);
         let needs_inval = ctx.base_state.needs_inval;
         let is_handled = ctx.is_handled();
+        if let Some(cursor) = cursor {
+            win_ctx.set_cursor(&cursor);
+        }
 
         let mut update_ctx = UpdateCtx {
             win_ctx,
@@ -539,6 +549,17 @@ impl<'a, 'b> EventCtx<'a, 'b> {
     /// Get an object which can create text layouts.
     pub fn text(&mut self) -> &mut Text<'b> {
         self.win_ctx.text_factory()
+    }
+
+    /// Set the cursor icon.
+    ///
+    /// Call this when handling a mouse move event, to set the cursor for the
+    /// widget. A container widget can safely call this method, then recurse
+    /// to its children, as a sequence of calls within an event propagation
+    /// only has the effect of the last one (ie no need to worry about
+    /// flashing).
+    pub fn set_cursor(&mut self, cursor: &Cursor) {
+        *self.cursor = Some(cursor.clone());
     }
 
     /// Set the "active" state of the widget.
--- a/druid-shell/src/mac/mod.rs
+++ b/druid-shell/src/mac/mod.rs
@@ -551,23 +551,6 @@ impl WindowHandle {
         }
     }
 
-    /// Set the current mouse cursor.
-    pub fn set_cursor(&self, cursor: &Cursor) {
-        unsafe {
-            let nscursor = class!(NSCursor);
-            let cursor: id = match cursor {
-                Cursor::Arrow => msg_send![nscursor, arrowCursor],
-                Cursor::IBeam => msg_send![nscursor, IBeamCursor],
-                Cursor::Crosshair => msg_send![nscursor, crosshairCursor],
-                Cursor::OpenHand => msg_send![nscursor, openHandCursor],
-                Cursor::NotAllowed => msg_send![nscursor, operationNotAllowedCursor],
-                Cursor::ResizeLeftRight => msg_send![nscursor, resizeLeftRightCursor],
-                Cursor::ResizeUpDown => msg_send![nscursor, ResizeUpDownCursor],
-            };
-            msg_send![cursor, set];
-        }
-    }
-
     /// Get a handle that can be used to schedule an idle task.
     pub fn get_idle_handle(&self) -> Option<IdleHandle> {
         // TODO: maybe try harder to return None if window has been dropped.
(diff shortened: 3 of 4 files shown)


druid-shell/src/window.rs
```diff
@@ -165,6 +168,7 @@ pub enum MouseButton {
 //both Windows and macOS. We may want to provide polyfills for various additional cursors,
 //and we will also want to add some mechanism for adding custom cursors.
 /// Mouse cursors.
+#[derive(Clone)]
```
cmyr: `Copy`?

src/lib.rs
```diff
@@ -367,8 +369,13 @@ impl<T: Data> UiState<T> {
     fn do_event(&mut self, event: Event, win_ctx: &mut dyn WinCtx) -> bool {
         // should there be a root base state persisting in the ui state instead?
         let mut base_state = Default::default();
+        let mut cursor = match event {
+            Event::MouseMoved(..) => Some(Cursor::Arrow),
```
cmyr: This seems weird? If the mouse moved but we're dragging an item, for instance, we don't want arrow.

druid-shell/src/window.rs
raphlinus: I was reluctant in case it expands to hold custom cursors.

src/lib.rs
raphlinus: Maybe I should add a comment to clarify, but the intent here is that arrow is the default. Basically the contract is that a widget that sets the cursor should *always* set it in its `MouseMoved` handler.

raphlinus:
I want to get this right on mac, so let's pull out a couple threads.

First, I believe that this mechanism is quite fine for, eg, setting an ibeam in a textbox. The implementation is for that widget to always set the cursor in its handling of MouseMoved.

I'd like to dig into the philosophy of this just a bit, as it might not be immediately clear. (Also, this should be written down in the docs somewhere). You can solve this problem using either an "immediate mode" or a "DSL" approach. In "immediate mode," events are piped to user code, which gets a chance to decide what to do with them, then the result of that is (immediately) presented to the user. In "DSL", user code does not expect to be invoked every event, so creates a specification that can be evaluated by the lower levels of the system. There's a fair amount of DSL in mac - cursor rectangles, CoreAnimation, responsive scrolling (where trackpad events directly move the scroll window without needing to get into the app event loop), and I can think of others. Basically the "druid way" is immediate, and a basis for that is the expectation that the Rust code is always fast enough to respond to events. My feeling (and strongly expressed by Rik Arends as well) is that immediate is overall a system simplification, because as soon as the expected behavior falls out of the range of what can be readily expressed in the DSL, things get complicated quickly.

There might be a second issue here (my mac is in the shop because the butterfly keyboard got a crumb in it, so I can't test), which is whether [`NSCursor::set`](https://developer.apple.com/documentation/appkit/nscursor/1526148-set) leaks globally into other app windows. If so, we should probably change it so there's a single rect that covers the entire window (or, to be more precise, the area of the NSView that serves as the root of the druid widget hiearchy, when thinking about VST guest use cases). I think that change can be made at the shell level without affecting event flow or API, if needed.

I'll point out that the Windows counterpart *is* global, but the ecosystem expects this: the system delivers [WM_SETCURSOR](https://docs.microsoft.com/en-us/windows/win32/menurc/wm-setcursor) to windows when the mouse hovers over them, not just when they're front. Apps are expected to set the cursor in response, and if they don't, you get stale cursor icons. (As just a bit more color to this, if an app sets a default cursor on window creation, then that cursor automatically gets applied when the mouse goes over the window. I'm not 100% sure whether this is done in the default windowproc or whether it sets metadata which is processed by the window manager, but it doesn't matter too much; apps that dynamically set the cursor icon basically don't set this so it's all done in immediate style)

src/lib.rs
cmyr: Ah okay I misunderstood, but am still a little confused.

What is the distinction of `cursor` being `Some` or `None`, here? Afaict in either case we allow the widget to call `set_cursor`, and it also seems like this means that we will always be calling `set_cursor` on each `MouseMoved` event, which seems unnecessary?

cmyr:
> I want to get this right on mac, so let's pull out a couple threads.
> 
> First, I believe that this mechanism is quite fine for, eg, setting an ibeam in a textbox. The implementation is for that widget to always set the cursor in its handling of MouseMoved.

I guess this depends on how `MouseMoved` gets handled, and whether it can be received by multiple widgets. My understanding from your explanation is that `MouseMoved` can also serve as a sort of hit-test, that is if your widget gets `MouseMoved` then the mouse is guaranteed to be in the layout bounds of your widget?


> I'd like to dig into the philosophy of this just a bit, as it might not be immediately clear. (Also, this should be written down in the docs somewhere). You can solve this problem using either an "immediate mode" or a "DSL" approach. In "immediate mode," events are piped to user code, which gets a chance to decide what to do with them, then the result of that is (immediately) presented to the user. In "DSL", user code does not expect to be invoked every event, so creates a specification that can be evaluated by the lower levels of the system. There's a fair amount of DSL in mac - cursor rectangles, CoreAnimation, responsive scrolling (where trackpad events directly move the scroll window without needing to get into the app event loop), and I can think of others. Basically the "druid way" is immediate, and a basis for that is the expectation that the Rust code is always fast enough to respond to events. My feeling (and strongly expressed by Rik Arends as well) is that immediate is overall a system simplification, because as soon as the expected behavior falls out of the range of what can be readily expressed in the DSL, things get complicated quickly.

This is great, as long as it works.

> There might be a second issue here (my mac is in the shop because the butterfly keyboard got a crumb in it, so I can't test), which is whether [`NSCursor::set`](https://developer.apple.com/documentation/appkit/nscursor/1526148-set) leaks globally into other app windows. If so, we should probably change it so there's a single rect that covers the entire window (or, to be more precise, the area of the NSView that serves as the root of the druid widget hiearchy, when thinking about VST guest use cases). I think that change can be made at the shell level without affecting event flow or API, if needed.

This **is** currently an issue. I'm not sure about how `NSCursor.set` interacts with cursor rects, but if we are just using `NSCursor.set` we're going to clobber the global cursor; I've seen this in my prototyping.

> I'll point out that the Windows counterpart _is_ global, but the ecosystem expects this: the system delivers [WM_SETCURSOR](https://docs.microsoft.com/en-us/windows/win32/menurc/wm-setcursor) to windows when the mouse hovers over them, not just when they're front. Apps are expected to set the cursor in response, and if they don't, you get stale cursor icons. (As just a bit more color to this, if an app sets a default cursor on window creation, then that cursor automatically gets applied when the mouse goes over the window. I'm not 100% sure whether this is done in the default windowproc or whether it sets metadata which is processed by the window manager, but it doesn't matter too much; apps that dynamically set the cursor icon basically don't set this so it's all done in immediate style)

I think this is a good model to follow (with a default, per-window cursor) but I suspect we will need to use `NSTrackingArea` to do it; I may well be wrong, this hunch is just from reading docs.

src/lib.rs
raphlinus: On Windows I'm pretty sure that is indeed what you want - it's hard to be sure if something else has set the cursor, so always setting it seems safe. Also I think the call is pretty cheap.

On mac if we end up setting tracking rectangles, then it's pretty clear that setting it to the same value accomplishes nothing, so I'd be fine with checking equality with the past value on the Rust side before calling into Cocoa.

raphlinus:
> I guess this depends on how MouseMoved gets handled, and whether it can be received by multiple widgets. My understanding from your explanation is that MouseMoved can also serve as a sort of hit-test, that is if your widget gets MouseMoved then the mouse is guaranteed to be in the layout bounds of your widget?

The logic is slightly more complicated than that. You also get MouseMoved if you're active, which I believe is desired behavior. (ie if you click inside a textbox and drag outside, you'll still get Ibar).

> using NSCursor.set [will] clobber the global cursor

Ok, let's fix that. Followup CL to this one, as I believe that was existing behavior? (also, I don't have a mac handy now but will by Monday). I believe the correct strategy is to set the `NSTrackingArea` for the view, although this may not give ideal results when dragging outside the view. I might want to experiment with that some more.

raphlinus:
So I'm looking at [Apple docs](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/EventOverview/TrackingAreaObjects/TrackingAreaObjects.html) and I'm *very* confused. It seems to me that you set up these tracking areas, and it drills down to Cocoa calling your view's `cursorUpdate` method, of which their suggested impl (see Listing 6-4) is to just call `NSCursor::set`. So I'm wondering if this is really what's causing the clobbering of the global cursor, or whether it might be something else such as a grab that's not being let go.

src/lib.rs
cmyr: I think I'd feel better if we knew the last cursor we'd set, and only called the system api if necessary? At the least it might be saving us a context switch? This isn't very substantiated but it feels unnecessary.

It also makes me think that it might be worth doing this in something like `ActiveChanged`, instead of `MouseMoved`? still just spitballing.

cmyr: I think I'd feel better if we knew the last cursor we'd set, and only called the system api if necessary? At the least it might be saving us a context switch? This isn't very substantiated but it feels unnecessary.

It also makes me think that it might be worth doing this in something like `ActiveChanged`, instead of `MouseMoved`? still just spitballing.

raphlinus: I think it'd be `HotChanged`, as that gets triggered on hover, but active state only changes on mouse press. I can make that change, but it will require a tad bit more event flow to get right, particularly (on Windows) I need to make sure that all hot state is reset when the mouse leaves the window. This turns out to be slightly tricky, I think I need to call [`TrackMouseEvent` ](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-trackmouseevent) to ensure that I get a WM_MOUSELEAVE. If I don't do this, then I'll skip setting the cursor when (eg) the mouse moves from a nonclient region such as a window sizing border back into the main window; the app will think the cursor is set, but in reality it was set in the hover over the nonclient region. Just doing it on mouse moved seems to be the windows way and avoids this extra complication. I can make this change if you feel strongly (an argument can be made that it gives more precise hot highlighting anyway).

Incidentally, https://devblogs.microsoft.com/oldnewthing/20031013-00/?p=42193 makes some interesting reading, explaining why there's WM_MOUSELEAVE but no WM_MOUSEENTER. The example is problematic, as it doesn't mention anything about TrackMouseEvent and I'm not seeing the WM_MOUSELEAVE events on my system.

I'm still confused about mac.

PR #90, opened 2019-07-27
How to build instructions added

Dmitry-Borodin:


Dmitry-Borodin:
Looks like something with nightly rust causing problems on CI
error: component 'rustfmt' for target 'x86_64-apple-darwin' is unavailable for download for channel 'nightly'

CONTRIBUTING.md
```diff
@@ -3,6 +3,14 @@
 We'd love to accept your patches and contributions to this project. There are
 just a few small guidelines you need to follow.
 
+## How to build
+Fulfill [gtk-rs dependencies](http://gtk-rs.org/docs/requirements.html) for cairo crate
+
+```
+cargo build --target=x86_64-pc-windows-msvc
+```
+Only windows and macos platforms are supported at the moment
```
Dmitry-Borodin: readme.md starting with it's windows-only. Is that outdated or to point that macos support is weaker?

CONTRIBUTING.md
```diff
@@ -3,6 +3,14 @@
 We'd love to accept your patches and contributions to this project. There are
 just a few small guidelines you need to follow.
 
+## How to build
+Fulfill [gtk-rs dependencies](http://gtk-rs.org/docs/requirements.html) for cairo crate
```
Dmitry-Borodin: readme.md states it's just cargo run, while without this dependency I'm getting compilation errors.
Remove the statement from readme.md?

CONTRIBUTING.md
cmyr: We currently support macOS and windows approximately equally.

CONTRIBUTING.md
cmyr: If you can't build on windows without specifying the target, could you open an issue? I'm not sure if that's a known thing or not.

In any case, I'm a bit confused: windows should be using d2d, so you shouldn't need the cairo dependency I don't think?

And related, the instructions added here are, at the very least, windows specific, so if we want to add this info it would be worth having separate `### windows` and `### macos` sections I think!

cmyr: The mention of `cargo run` in the readme is more referring to what is involved in getting a client application to run.

Dmitry-Borodin: I'm using GNU/Linux (Ubuntu 18.4 dependencies, Mint 19) and cannot compile because my platform have no implementation. And for me the only way to find it out was to read the code and see, that it is not a dependency problem, there is not implementation at all and I have to specify supported platform explicitly.

cmyr: Ah, yes okay that makes sense. :)

I think the solution to this problem is to clearly state in the readme what platforms are supported. We could also just deny compilation based on the current platform, e.g. something like,

```rust
#[cfg(not(any(target_os = "macos", target_os = "windows")))]
compile_error!("druid does not current support your platform.");
```

which would at least give a clear and immediate diagnostic? I'm not sure this is necessary though, it feels a bit cute.

cmyr:
All this said, @Dmitry-Borodin if you're interested in working on linux support for druid we would be very happy to help, linux support is important but there just isn't anyone right now who is willing to own that platform and we don't want to have a half-supported platform blocking other work.

CONTRIBUTING.md
Dmitry-Borodin: I think this small check make sense
In general, supporting GNU/Linux is something I'm planning to work on, but I would like to play around with it before declaring to maintain a platform. I'm new in Rust yet.

Regarding this docs, I will note that gtk-rs needed only for MacOs targeted builds, links at the bottom and general readme update regarding two platforms, right?

Dmitry-Borodin: updated readme

README.md
```diff
@@ -5,7 +5,7 @@
 Druid is a new Rust-native UI toolkit, still in early stages. Its main
 goal is performance, also aiming for small binary size and compile time,
 fast startup, and very easy build configuration (just `cargo run`). It
```
cmyr: I would replace this sentence with, 'It current supports Windows and macOS, with GNU/Linux support planned.'

CONTRIBUTING.md
```diff
@@ -3,6 +3,13 @@
 We'd love to accept your patches and contributions to this project. There are
 just a few small guidelines you need to follow.
 
+## How to build
```
cmyr: I would say,

## Building

Currently, druid only builds on windows and macOS.

### macOS

On macOS, druid requires cairo; see [gtk-rs dependencies] for installation instructions.
You may also need to set your `PKG_CONFIG_PATH`; assuming you have installed `gtk` through homebrew,
you can builld with,

```sh
$> PKG_CONFIG_PATH="/usr/local/opt/libffi/lib/pkgconfig" cargo build
```

raphlinus: I'm a little confused about the build instructions. My understanding is that we're asking people to install cairo on the mac build (ie `brew install cairo`), not all of gtk. Just cairo is a considerably lighter weight dependency.

README.md
```diff
@@ -5,7 +5,7 @@
...
-is currently Windows-only, but with hope for porting.
+is currently Windows and MacOS only, porting to GNU/Linux in plans.
```
raphlinus: This is currently spelled "macOS".

CONTRIBUTING.md
```diff
@@ -3,6 +3,13 @@
...
+> Only windows and macos platforms are supported at the moment
```
raphlinus: "Windows and macOS". Let's make sure the capitalization is consistent, there are a couple other instances.

CONTRIBUTING.md
cmyr: Updated.

Dmitry-Borodin: I think we are mixing target platform and building platform.
Brew is only available in MacOS. But any platfrom, trying to build macos targeted binaries has to have hits dependencies. 
So we cannot just ask here to use homebrew.

raphlinus: Are you talking about cross-compiling? I don't think we're supporting that, I can think of a number of ways in which that would be tricky, more so going forward (we may well want to depend on xcode-specific tools to, for example, packresources into an app bundle).

Dmitry-Borodin: Well, at least right now I can build for both platforms. But let's not support it officially until tested properly. I will update PR soon.

README.md
Dmitry-Borodin: fixed everywhere

CONTRIBUTING.md
Dmitry-Borodin: done

CONTRIBUTING.md
Dmitry-Borodin: Readme updated

README.md
Dmitry-Borodin: fixed

README.md
```diff
@@ -4,8 +4,8 @@
 
 Druid is a new Rust-native UI toolkit, still in early stages. Its main
 goal is performance, also aiming for small binary size and compile time,
-fast startup, and very easy build configuration (just `cargo run`). It
-is currently Windows-only, but with hope for porting.
+fast startup, and very easy build configuration (just `cargo run`). 
+It current supports Windows and macOS, with GNU/Linux support planned.
```
cmyr:
```suggestion
It currently supports Windows and macOS, with GNU/Linux support planned.
```

Dmitry-Borodin: Thank you
I thought it's weird but decided to not change since I'm not native English speaker.

cmyr:
Okay I'm happy to merge this shortly, although linux support may be landing sooner than expected! Anyway, thanks for the PR. :)

PR #96, opened 2019-08-03
Initial try to build environment infrastructure

raphlinus:
As of this checkpoint, some of the bones of environments are in place, especially for retrieving theme values. There's more to be done, including:

* Persisting the initial theme so that it doesn't get recreated every time.

* Storing an environment reference in `WidgetPod` so that we can propagate `update` if either the data or environment (or both) have changed.

* Nicer methods for dealing with environment values.

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -28,25 +28,111 @@ use crate::Data;
 // maybe we should use Rc instead of Arc.
 use std::sync::Arc;
 
+#[derive(Clone)]
 pub struct Env(Arc<EnvImpl>);
 
 pub struct EnvImpl {
     theme: EnvValue,
+    // This is always a map, maybe should be a map here.
+    values: EnvValue,
 }
 
 #[derive(Clone)]
 pub enum EnvValue {
     Float(f64),
+    Bool(bool),
+    Color(Color),
     String(String),
     List(Arc<Vec<EnvValue>>),
     Map(Arc<HashMap<String, EnvValue>>),
     Null,
 }
 
+impl Env {
+    pub fn from_theme(theme: EnvValue) -> Env {
+        Env(Arc::new(EnvImpl {
+            theme,
+            values: EnvValue::empty_map(),
+        }))
+    }
+
+    /// Get a theme value.
+    ///
+    /// The value will be EnvValue::Null if the key is not present.
+    // TODO: make key more flexible
+    pub fn theme_value(&self, key: &str) -> &EnvValue {
+        &self.0.theme[key]
+    }
+
+    /// Get a theme color.
+    ///
+    /// Prints a debug warning and returns a garish color if the key
+    /// isn't present or is of the wrong type.
+    pub fn theme_color(&self, key: &str) -> Color {
+        if let Some(c) = self.theme_value(key).as_color() {
+            c
+        } else {
+            println!("missing theme color for {}", key);
+            Color::rgb24(0xff_00_ff)
+        }
+    }
+
+    /// Get a theme float value.
+    ///
+    /// Prints a debug warning and returns 0.0 if the key isn't present
+    /// or is of the wrong type.
+    ///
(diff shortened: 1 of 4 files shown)


src/env.rs
```diff
@@ -0,0 +1,278 @@
...
+/// A widget can retrieve theme parameters (colors, dimensions, etc.). In
+/// addition, it can pass custom data down to all descendants. An important
+/// example of the latter is setting a value for enabled/disabled status
+/// so that an entire subtree can be disabled ("grayed out") with one
+/// setting.
+#[derive(Clone)]
+pub struct Env(Arc<EnvImpl>);
+
+struct EnvImpl {
+    theme: EnvValue,
```
cmyr: Something to think about with theme is how dynamic it is. That's to say: for system widgets, we will know all of the recognized theme keys at compile time, so `theme` could just be a struct. This would let us do things like `theme.text_color` and `theme.selected_text_bg_color`, without having to mess around with keys; and we could also guarantee that all values were present in the theme.

There are other options. We could have something like,

```rust
pub struct Theme(EnvMap);

impl Theme {
    fn text_color(&self) -> Color {
            self.0.theme_value(TEXT_COLOR_KEY).as_color()
    }
}
```

which lets us keep using `EnvValue` under the covers but gives a simpler API for this specific (theme) case. (I'm not sure I like this much.)

src/env.rs
```diff
@@ -0,0 +1,278 @@
...
+/// This is an enum with a variety of common types.
+#[derive(Clone)]
+pub enum EnvValue {
+    Float(f64),
+    Bool(bool),
+    Color(Color),
+    String(String),
+    List(Arc<Vec<EnvValue>>),
+    Map(Arc<HashMap<String, EnvValue>>),
+    Null,
```
cmyr: This sets of my spidey sense. What does this give us that we can't do with `Option`? Having indexing work is a very questionable gain if the thing we get back may be invalid; if something can be missing, forcing the user to handle an `Option` feels better, because that's just a common Rust pattern.

src/env.rs
```diff
@@ -0,0 +1,278 @@
...
+    // TODO: make key more flexible
+    pub fn theme_value(&self, key: &str) -> &EnvValue {
+        &self.0.theme[key]
+    }
+
+    /// Get a theme color.
+    ///
+    /// Prints a debug warning and returns a garish color if the key
+    /// isn't present or is of the wrong type.
+    pub fn theme_color(&self, key: &str) -> Color {
```
cmyr: I don't love these two methods either, but it's kind of a finicky problem.

I think getting this right demands we think more about what a theme is. There's a very reasonable case to be made that a theme is either a fixed set of keys/values (as in: it is an invariant that a given set of keys exists, and each key has a known value type), *or* a theme is that *plus* some custom keys and values provided by the user.

In either case, (for themes, maybe not for `env` more generally) keys and the type of their values should always be known at compile time.

In my environment branch, I was playing around with various ways of encoding this in the type system. The best approach I've found is to encode the type of the value in the key itself, like:

```rust
pub struct Key<T> {
    key: &'static str,
    value_type: PhantomData<T>,
}

pub const TEXT: Key<Color> = Key::new("io.xi-editor.text_color");

let color: Color = theme.get(&TEXT); // Color or panic
```

For all built-in keys, we can guarantee that they are of the correct type. This also lets us do things like disallow overwriting a key with a value of a different type.

Another option is to have a fn `get`, that is generic over it's return type:

```rust
fn get<T: FromEnvValue>(key: &Key) -> T;
```

where `FromEnvValue` is a trait implemented by the raw `EnvValue` types.

raphlinus:
This was discussed in depth on Zulip. I'm going to take another pass, incorporating a more statically typed approach and opaque keys.

raphlinus:
I'm going to close this in favor of #107; there are some intriguing ideas in here but it's not the direction we want to go.

PR #107, opened 2019-08-05
Another try at environments

raphlinus:
This PR has some elements of my last try at environments, and some of
Colin's ideas (and code) from the "environment" branch.

The button is wired up to use the theme.

One thing that's *not* wired up is the propagation of `update` when
the env changes.


src/env.rs
```diff
@@ -0,0 +1,233 @@
...
+    // Also &V.
+    pub fn try_get<V: TryFrom<Value, Error = String>>(&self, key: Key<V>) -> Option<V> {
+        self.0
+            .map
+            .get(key.key)
+            .map(|value| value.to_owned().into_inner_unchecked())
+    }
+
+    /// Adds a key/value, acting like a builder.
+    pub fn adding<K: Into<String>, V: Into<Value>>(mut self, key: K, value: V) -> Env {
```
cmyr: I want to highlight this as something that is a bit unidiomatic; it's basically bundling a builder pattern into the type itself. I like this because you don't need to import a separate builder type and you also don't need to call `build`, but it might seem a bit weird to some folks.

src/theme.rs
```diff
@@ -0,0 +1,33 @@
...
+use crate::{Env, Key};
+
+pub const BACKGROUND_COLOR: Key<Color> = Key::new("background_color");
+pub const HOVER_COLOR: Key<Color> = Key::new("hover_color");
+pub const PRESSED_COLOR: Key<Color> = Key::new("pressed_color");
+pub const LABEL_COLOR: Key<Color> = Key::new("label_color");
+
+/// An initial theme.
+pub fn init() -> Env {
+    Env::default()
```
cmyr: I suspect we'll want to put theme keys in their own map; this also motivates having `Key` be a 'key path' type. I think this is totally reasonable for now though.

raphlinus:
I rebased this and added the logic to propagate. Things to perhaps criticize:

* I haven't exercised this very much except as a way to provide theme values. In particular I haven't exercised the logic to override env values.

* Now that we've done the piet work, it would make sense to store gradients as env values. Not sure if we want to get this in and do that as followup work, or get it in here.

For the former, I think an `EnvWrapper` widget might be convenient. That basically takes an `FnMut(&mut Env, data: &T)`. Again, that could go in this PR, or could be a followup. It's also subject to bikeshed on the name and exact signature.

src/env.rs
```diff
@@ -0,0 +1,296 @@
...
+// we could do some serious deriving here: the set of types that can be stored
+// could be defined per-app
+// Also consider Box<Any> (though this would also impact debug).
+/// A dynamic type representing all values that can be stored in an environment.
+#[derive(Clone)]
+pub enum Value {
+    Point(Point),
+    Size(Size),
+    Rect(Rect),
+    Color(Color),
```
cmyr: How _would_ we want to include gradient? An option would be to use the `PaintBrush` type, here, or we could have distinct members for `LinearGradient` and `RadialGradient`.

src/env.rs
```diff
@@ -0,0 +1,296 @@
...
+/// Note that for "expensive" types this is the reference. For example,
+/// for strings, this trait is implemented on `&'a str`. The trait is
+/// parametrized on a lifetime so that it can be used for references in
+/// this way.
+pub trait ValueType<'a>: Sized {
+    /// The corresponding owned type.
+    type Owned;
+
+    /// Attempt to convert the generic `Value` into this type.
+    fn try_from_value(v: &'a Value) -> Result<Self, String>;
```
cmyr: We'll need to figure out an error type for this at some point; for now I don't mind this being a `TODO`.

src/env.rs
```diff
@@ -0,0 +1,296 @@
...
+    /// Gets a value from the environment, expecting it to be present.
+    ///
+    /// Note that the return value is a reference for "expensive" types such
+    /// as strings, but an ordinary value for "cheap" types such as numbers
+    /// and colors.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the key is not found, or if it is present with the wrong type.
+    pub fn get<'a, V: ValueType<'a>>(&'a self, key: Key<V>) -> V {
```
cmyr: It might be cleaner to implement this in terms of `try_get`, below?

I'm also now thinking about the names a bit more; `get` in other Rust API would be expected to return an `Option`, and then we would give _this_ a different name.

There _is_ an argument for keeping it as-is; this is intended only to be used with known keys, and the illusion of `Env` is that it just always works. But we could also just remove `try_get`, make this return `Option<V>`, and document that unwrapping is reasonable for known keys.

src/env.rs
```diff
@@ -0,0 +1,296 @@
...
+    /// Panics if the value for the key is found, but has the wrong type.
+    pub fn try_get<'a, V: ValueType<'a>>(&'a self, key: Key<V>) -> Option<V> {
+        self.0
+            .map
+            .get(key.key)
+            .map(|value| value.to_inner_unchecked())
+    }
+
+    /// Adds a key/value, acting like a builder.
+    pub fn adding<'a, V: ValueType<'a>>(mut self, key: Key<V>, value: impl Into<V::Owned>) -> Env
```
cmyr: I'm not sure I love mixing generic syntax with argument position `impl`? It might be clearer here just as,

```rust
pub fn adding<'a, V, T>(mut self, key: Key<V>, value: T) -> Env
where
    V: ValueType<'a>,
    V::Owned: Into<Value>,
    T: Into<V::Owned>,
{
// ...
}
```

But I don't feel strongly about this.

This also makes me wonder if we couldn't have the `Into<Value>` bound be a bound on the associated type in the trait? This would clean up these function signatures somewhat.

src/env.rs
```diff
@@ -0,0 +1,296 @@
...
+            Value::Rect(r) => write!(f, "Rect {:?}", r),
+            Value::Color(c) => write!(f, "Color {:?}", c),
+            Value::Float(x) => write!(f, "Float {}", x),
+            Value::String(s) => write!(f, "String {:?}", s),
+        }
+    }
+}
+
+impl<T> Key<T> {
+    pub const fn new(key: &'static str) -> Self {
```
cmyr: For docs:

```rust
/// Create a new strongly typed `Key` with the given string value.
/// The type of the key will be inferred.
/// 
/// # Examples
///
/// ```
/// use druid::env::Key;
/// use druid::piet::Color;
///
/// let float_key: Key<Float> = Key::new("a.very.good.float");
/// let color_key: Key<Color> = Key::new("a.very.nice.color");
/// ```
```

src/env.rs
```diff
@@ -0,0 +1,296 @@
...
+        Key {
+            key,
+            value_type: PhantomData,
+        }
+    }
+}
+
+impl Value {
+    pub fn to_inner_unchecked<'a, V: ValueType<'a>>(&'a self) -> V {
```
cmyr: Also deserves a docstring.

src/env.rs
```diff
@@ -0,0 +1,296 @@
...
+        Arc::ptr_eq(&self.0, &other.0) || self.0.deref().same(other.0.deref())
+    }
+}
+
+impl Data for EnvImpl {
+    fn same(&self, other: &EnvImpl) -> bool {
+        if self.map.len() != other.map.len() {
+            return false;
+        }
+        for (k, v1) in self.map.iter() {
```
cmyr: 🏌 
```rust
        self.map
            .iter()
            .all(|(k, v1)| other.map.get(k).map(|v2| v1.same(v2)).unwrap_or(false))
```

src/theme.rs
```diff
@@ -0,0 +1,50 @@
...
+
+pub const BACKGROUND_COLOR: Key<Color> = Key::new("background_color");
+pub const HOVER_COLOR: Key<Color> = Key::new("hover_color");
+pub const PRESSED_COLOR: Key<Color> = Key::new("pressed_color");
+pub const LABEL_COLOR: Key<Color> = Key::new("label_color");
+pub const FONT_NAME: Key<&str> = Key::new("font_name");
+
+/// An initial theme.
+pub fn init() -> Env {
+    let mut env = Env::default()
```
cmyr: Are theme keys all just top-level keys in the environment, or does the environment have a `theme` object?

Related, but do we want to introduce the idea of keypaths here, so that `theme.background` will resolve to the key 'background' in the object with the key "theme"?

I suppose not for now, since env doesn't contain any collection types, but might be useful at some point.

src/widget/button.rs
```diff
@@ -72,10 +73,11 @@ impl Label {
 }
 
 impl<T: Data> Widget<T> for Label {
-    fn paint(&mut self, paint_ctx: &mut PaintCtx, _base_state: &BaseState, _data: &T, _env: &Env) {
+    fn paint(&mut self, paint_ctx: &mut PaintCtx, _base_state: &BaseState, _data: &T, env: &Env) {
+        let font_name = env.get(theme::FONT_NAME);
         let font_size = 15.0;
```
cmyr: this should go in theme too, while we're at it?

src/env.rs
raphlinus: When we discussed this on Zulip, consensus was that it should individual variants. But that was before PaintBrush actually landed. I personally like the idea of including PaintBrush, as then we can switch between gradient and solid themes, and the drawing code doesn't have to care.

src/env.rs
raphlinus: I'll typedef it with a TODO.

src/env.rs
raphlinus: The specific reason for not implementing it in terms of `try_get` is that `try_get` consumes its key by move, making it unavailable for an error message.

I'm going to argue for keeping this as-is for reasons of conciseness. Yes, this is expected to always succeed. I don't think adding any more verbosity, either through a longer name or an explicit unwrap (ugh), actually helps anything. To a large extent, the distinctions are self-documenting through types.

src/env.rs
cmyr: I think that works. One thing I'm realizing is I think `PaintBrush` doesn't derive `PartialEq` (checked, and it actually doesn't derive anything 🤦‍♂)

src/env.rs
raphlinus: Yeah, I think `Into<Value>` on `ValueType::Owned` is a win. Here's the minor edit:

```rust
    pub fn adding<'a, V: ValueType<'a>>(mut self, key: Key<V>, value: impl Into<V::Owned>) -> Env
```

And here's what it would be with no impl but on one line:

```rust
pub fn adding<'a, V: ValueType<'a>, T: Into<V::Owned>>(mut self, key: Key<V>, value: T) -> Env
```

And with 'where':

```rust
    pub fn adding<'a, V, T>(mut self, key: Key<V>, value: T) -> Env
    where
        V: ValueType<'a>,
        T: Into<V::Owned>,
```

Preference?

cmyr: The first or the last, first is cleaner but depends how we feel about mixing `impl` args with normal generics. Your call.

src/env.rs
raphlinus: +1 (I had to clean it up very slightly to compile)

src/env.rs
raphlinus: Hmm, I *personally* find the `for` version easier to read, but suspect people will have different feelings about this.

src/env.rs
raphlinus: I'm gonna go with the first, because `impl Into` feels really idiomatic to me. The third is also appealing because it reads a little less like a "line noise" type.

src/theme.rs
raphlinus: Earlier versions had this. Right now I want to keep things simple.

src/env.rs
cmyr: either works for me.

raphlinus: Heh, I can get rid of all explicit control flow:

```rust
    fn same(&self, other: &EnvImpl) -> bool {
        self.map.len() == other.map.len()
            && self
                .map
                .iter()
                .all(|(k, v1)| other.map.get(k).map(|v2| v1.same(v2)).unwrap_or(false))
    }
```

This isn't bad, actually, I think I'll go with it.

src/env.rs
raphlinus: Hmm. I'm running into a problem here, which is that `PaintBrush` isn't `Clone`. That said, I'm thinking the owned type maybe should be `Arc<PaintBrush>`, which of course is `Clone` whether the inner type is or not. I'm going to have to play with types a bit to see how this works, so what I think I'll do it push a commit with all other feedback addressed.

PR #108, opened 2019-08-05
Impl deref for paintctx (#97)

### LINKED ISSUE #97: Deref PaintCtx into Piet
As a quality-of-life improvement, we'd like `PaintCtx` to deref into a `Piet` render context, so you don't have to write stuff like `paint_ctx.render_ctx.draw_text()` all the time.

Basically, the work is to add a `Deref` and `DerefMut`, and then to clean up all the uses in the `paint` methods.

futurepaul:
Fixes #97. 

I sort of banged on this until it compiled, so I'd appreciate a close look at the actual impl:

```rust
impl<'a, 'b: 'a> Deref for PaintCtx<'a, 'b> {
    type Target = &'a mut Piet<'b>;

    fn deref(&self) -> &Self::Target {
        &self.render_ctx
    }
}

impl<'a, 'b: 'a> DerefMut for PaintCtx<'a, 'b> {
    fn deref_mut(&mut self) -> &mut Self::Target {
        &mut self.render_ctx
    }
}
```

I ran all the examples in druid and druid-shell and everything works. I also tweaked the docs where appropriate as far as I could find.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -302,10 +302,10 @@ pub struct PaintCtx<'a, 'b: 'a> {
 }
 
 impl<'a, 'b: 'a> Deref for PaintCtx<'a, 'b> {
-    type Target = &'a mut Piet<'b>;
+    type Target = Piet<'b>;
 
     fn deref(&self) -> &Self::Target {
-        &self.render_ctx
+        self.render_ctx
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -291,7 +291,7 @@ pub struct Env {
 
 /// A context passed to paint methods of widgets.
 ///
-/// Widgets paint their appearance by calling methods on the 
+/// Widgets paint their appearance by calling methods on the
 /// `render_ctx`, which PaintCtx derefs to for convenience.
 /// This struct is expected to grow, for example to include the
 /// "damage region" indicating that only a subset of the entire
@@ -314,7 +314,6 @@ impl<'a, 'b: 'a> DerefMut for PaintCtx<'a, 'b> {
         &mut self.render_ctx
     }
 }
-
 
 /// A context provided to layout handling methods of widgets.
 ///
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -24,6 +24,7 @@ mod lens;
 mod value;
 
 use std::any::Any;
+use std::ops::Deref;
 use std::ops::DerefMut;
 use std::time::Instant;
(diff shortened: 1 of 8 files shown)


src/lib.rs
```diff
@@ -24,6 +24,7 @@ mod lens;
 mod value;
 
 use std::any::Any;
+use std::ops::Deref;
```
raphlinus: Generally we would write these on the same line, especially as they're related: `std::ops::{Deref, DerefMut}`

futurepaul:
Yeah, sorry for all the noise, I'll get that set up.

src/lib.rs
```diff
@@ -290,14 +292,29 @@ pub struct Env {
...
+/// This struct is expected to grow, for example to include the
+/// "damage region" indicating that only a subset of the entire
+/// widget hierarchy needs repainting.
 pub struct PaintCtx<'a, 'b: 'a> {
     /// The render context for actually painting.
     pub render_ctx: &'a mut Piet<'b>,
 }
 
+impl<'a, 'b: 'a> Deref for PaintCtx<'a, 'b> {
+    type Target = &'a mut Piet<'b>;
```
raphlinus: Actually I had a closer look, and it should just be `type Target = Piet<'b>`. Then the body needs no additional deref, it's just `&self.render_ctx`.

This is very subtle because it does work as written, but not as efficiently. What you get with the current state of the code is a reference to the field within the `PaintCtx` object. You can still call methods on it because magic deref will follow both references, but it's a double indirection (although probably will get optimized).

cmyr: Hmm, this is subtle, I wouldn't have thought of it. Although having the target of `Deref` (not `DerefMut`) be `&mut` maybe looks a little strange.

futurepaul: Thank you for this explanation! Yeah, I should've taken more pause at the &mut in the regular Deref.

src/lib.rs
```diff
@@ -290,14 +292,29 @@ pub struct Env {
...
+    type Target = Piet<'b>;
+
+    fn deref(&self) -> &Self::Target {
+        self.render_ctx
+    }
+}
+
+impl<'a, 'b: 'a> DerefMut for PaintCtx<'a, 'b> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.render_ctx
```
raphlinus: The `&mut` is also not needed here.

PR #109, opened 2019-08-06
More ergonomic brushes

raphlinus:
This patch lets the user specify a color or a higher-level gradient
to fill and stroke methods.

Closes https://github.com/linebender/piet/issues/72

### DIFF
--- a/src/paint.rs
+++ b/src/paint.rs
@@ -1,0 +1,235 @@
+// Copyright 2019 The xi-editor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Paint context and helper methods.
+
+use std::borrow::{Borrow, Cow};
+use std::ops::{Deref, DerefMut};
+
+use crate::kurbo::{Point, Rect, Shape};
+
+use crate::piet::{
+    Color, FillRule, GradientStop, LinearGradient, Piet, RenderContext, StrokeStyle,
+};
+
+/// A context passed to paint methods of widgets.
+///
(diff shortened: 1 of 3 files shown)


raphlinus:
This is a draft PR mostly because it depends on https://github.com/linebender/piet/issues/74 landing. Patch locally to experiment.

I'll lift out the discussion questions here:

* Should this brush trait (with the resolution of `UnitPoint` to actual `Point`) be in druid at all, or does it make more sense to do in piet? The tradeoffs are tricky.

* Should `UnitPoint` move to kurbo? It's pretty much pure geometry.

* Should the ergonomic gradient stop conversions move to piet as simply impl's of `Into<Vec<GradientStop>>`?

* Should the `Brush` type take a closure to resolve bbox instead of shape?

Many of these questions involve "where should this go?" We'll need to do new releases of at least piet, maybe kurbo as well.

src/paint.rs
```diff
@@ -0,0 +1,235 @@
...
+}
+
+impl<'a, 'b: 'a> DerefMut for PaintCtx<'a, 'b> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        self.render_ctx
+    }
+}
+
+// The block below expands the range of brush arguments to allow anything
+// implementing the brush trait. This is a signifcant ergonomics improvement.
```
cmyr:
```suggestion
// implementing the brush trait. This is a significant ergonomics improvement.
```

src/paint.rs
```diff
@@ -0,0 +1,235 @@
...
+
+impl<'a> GradientStops for (Color, Color, Color, Color) {
+    fn to_vec(self) -> Vec<GradientStop> {
+        let stops: &[Color] = &[self.0, self.1, self.2, self.3];
+        GradientStops::to_vec(stops)
+    }
+}
+
+impl Gradient {
+    pub fn new(start: UnitPoint, end: UnitPoint, stops: impl GradientStops) -> Gradient {
```
cmyr: this is unclear enough to deserve a doc comment, ideally with a couple `# Examples`.

raphlinus:
If you think this is tricky, wait'll you see the version that integrates it into piet :). That said, the end-user confusion should be less.

The main reason I'm thinking closure is for the text bounding box. I'm not too concerned about being able to draw text with gradients, but being able to use Color sure would be nice imho.

cmyr:
@raphlinus is this obsolete?

raphlinus:
Yes, this got done in the piet 0.0.6 release.

PR #110, opened 2019-08-07
Add timers

### LINKED ISSUE #105: Add timers
Applications should be able to schedule timers on the runloop that will call them back with a given `Action` or `Event`.

raphlinus:
Adds a timer event to druid_shell, and also event flow for timers.

WIP: Windows only

Fixes #105

### DIFF
--- a/druid-shell/src/mac/mod.rs
+++ b/druid-shell/src/mac/mod.rs
@@ -688,7 +688,7 @@ impl<'a> WinCtx<'a> for WinCtxImpl<'a> {
     }
 
     fn request_timer(&mut self, deadline: std::time::Instant) -> TimerToken {
-        let ti = time_interval_to_deadline(deadline);
+        let ti = time_interval_from_deadline(deadline);
         let token = next_timer_id();
         unsafe {
             let nstimer = class!(NSTimer);
@@ -705,8 +705,8 @@ impl<'a> WinCtx<'a> for WinCtxImpl<'a> {
 /// Convert an `Instant` into an NSTimeInterval, i.e. a fractional number
 /// of seconds from now.
 ///
-/// This may loose some precision for multi-month durations.
-fn time_interval_to_deadline(deadline: std::time::Instant) -> f64 {
+/// This may lose some precision for multi-month durations.
+fn time_interval_from_deadline(deadline: std::time::Instant) -> f64 {
     let now = Instant::now();
     if now >= deadline {
         0.0
--- a/druid-shell/src/mac/mod.rs
+++ b/druid-shell/src/mac/mod.rs
@@ -37,6 +37,7 @@ use std::ffi::c_void;
 use std::ffi::OsString;
 use std::mem;
 use std::sync::{Arc, Mutex, Weak};
+use std::time::Instant;
 
 use cairo::{Context, QuartzSurface};
 
@@ -46,7 +47,7 @@ use piet_common::{Piet, RenderContext};
 use crate::keyboard::{KeyCode, KeyEvent, KeyModifiers};
 use crate::platform::dialog::{FileDialogOptions, FileDialogType};
 use crate::util::make_nsstring;
-use crate::window::{Cursor, MouseButton, MouseEvent, Text, WinCtx, WinHandler};
+use crate::window::{Cursor, MouseButton, MouseEvent, Text, TimerToken, WinCtx, WinHandler};
 use crate::Error;
 
 use util::assert_main_thread;
@@ -256,6 +257,10 @@ lazy_static! {
         );
         decl.add_method(sel!(runIdle), run_idle as extern "C" fn(&mut Object, Sel));
         decl.add_method(sel!(redraw), redraw as extern "C" fn(&mut Object, Sel));
+        decl.add_method(
+            sel!(handleTimer:),
+            handle_timer as extern "C" fn(&mut Object, Sel, id),
+        );
         ViewClass(decl.register())
     };
 }
@@ -523,6 +528,25 @@ extern "C" fn redraw(this: &mut Object, _: Sel) {
     }
 }
 
+extern "C" fn handle_timer(this: &mut Object, _: Sel, timer: id) {
+    let view_state = unsafe {
+        let view_state: *mut c_void = *this.get_ivar("viewState");
+        &mut *(view_state as *mut ViewState)
+    };
+    let mut ctx = WinCtxImpl {
+        nsview: &(*view_state).nsview,
+        text: Text::new(),
+    };
+    let token = unsafe {
+        let user_info: id = msg_send![timer, userInfo];
+        msg_send![user_info, unsignedIntValue]
+    };
+
+    (*view_state)
+        .handler
+        .timer(TimerToken::new(token), &mut ctx);
+}
+
 impl WindowHandle {
     pub fn show(&self) {
         unsafe {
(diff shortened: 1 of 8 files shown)


src/event.rs
```diff
@@ -92,6 +92,17 @@ pub enum Event {
...
+    ///
+    /// Request a timer event through [`EventCtx::request_timer()`]. That will
+    /// cause a timer event later.
+    ///
+    /// Note that timer events from other widgets may be delivered as well. Use
+    /// the id returned from the `request_timer()` call to filter events more
+    /// precisely.
+    ///
+    /// [`EventCtx::request_timer()`]: struct.EventCtx.html#method.request_timer
+    Timer(usize),
```
cmyr: I would prefer a `TimerToken` type.

src/lib.rs
```diff
@@ -1006,6 +1016,21 @@ impl<'a, 'b> EventCtx<'a, 'b> {
...
+    pub fn request_anim_frame(&mut self) {
+        self.base_state.request_anim = true;
+    }
+
+    /// Request a timer event.
+    ///
+    /// The return value is an id, which can be used to associate the
+    /// request with the event. The id is guaranteed not to be 0, and can
+    /// also be assumed to be a small integer.
+    pub fn request_timer(&mut self, deadline: Instant) -> usize {
```
cmyr: I might call this `add_timer`. 'request' is tricky; can requests be refused? Are we given a timer if our request is accepted? etc.

There's also the question of what the argument should be. There's a reasonable case that this could just be a duration, which is taken relative to the time when the call is made. This would also allow for things like repeating timers, and prevent things like scheduling timers for times in the past.

If we make our rumoured changes to `Action`, we can also associate an `Action` with a timer which would be a nicer API.

We might also want to consider giving timers a name, which would let callers use const strs to identify timers, instead of needing to stash the returned token.

src/event.rs
raphlinus: I can make it an alias or a newtype. (preference?)

I would prefer the concrete type to match Windows though, otherwise we have to do translation.

cmyr: I'd prefer a newtype; and I don't think we should necessarily be making decisions in `druid` based on the implementation details of a single platform.

cmyr: to clarify: we should pick the best reasonable abstraction that can be implemented well across multiple backends, and we should be okay leaning on Rust is Fast if we need to map between a general and a platform representation of some token.

src/lib.rs
raphlinus: I'm fine with changing the name to `add_timer`. The existing name is designed to be parallel to `request_anim_frame`, which basically comes from web world.

My understanding is that instant is the gold standard for timers. Fuchsia iterated on this a bit. One motivation is that if you want to blink something at a fixed rate, you repeatedly add the period to the base instant. This is especially important if the underlying mechanism is sloppy. (As an aside, I'm personally interested in precise time for, eg, music production, but this is an extremely difficult problem, so it needs to go on the back burner)

I think names are *highly* problematic, as there's no mechanism to prevent collisions. Remember, widgets don't have identity, though this is perhaps a question we could come back to.

I'm open to making `Action` include timers, but confess to being very fuzzy what that would actually look like. Maybe more clarity will come when we wire up other things such as context menus.

So, in sum, while this API is perhaps not the absolute most friendly (the user needs to stash the id and filter on it), I'll defend it as being a pretty solid way to get things done, requiring a smidge of extra work on the part of the widget author that I don't see any easy way to remove without adding significant extra complexity in the system.

cmyr: Not sure I understand, about `Instant`: if you want to blink something at a fixed rate with `Duration`, you just set a repeating timer, and the framework reschedules whenever it delivers; and on Cocoa, at least, we can just schedule a repeating timer directly. (Cocoa provides API for using both duration (the default) and also firing at a specific datetime).

named timers would require some convention like reverse dns, for namespacing. I'm happy not to do this, but I'm not sure it's a huge problem in practice; any timer names used by the framework would be collision-resilient, and it would be up to the application developer to pick their own names appropriately.

raphlinus: Having an explicit repeating timer is one way to do this, and very common in timer API's. I believe it complicates the API for not too much benefit - you then need to be able to cancel the timer. But this might just be my preference for a small number of simple primitives.

I think for our use cases, duration and instant aren't that much different. I think for advanced uses, instant is more robust.

src/event.rs
raphlinus: One advantage of using a "small integer" is that it can be used as an index of a Vec if you need to maintain a map. But I'm willing to write that off as premature optimization.

It feels to me like the newtype should derive Ord, Hash, Clone, Debug, PartialEq, and Eq, so it can be used as map keys as needed. Sound reasonable?

src/lib.rs
cmyr: I would suggest that the prevalence of API for repeated timers suggests they may be useful.

I also think that in 99% of cases, this API will be called with some duration relative to `Instant::now()`, that is, the `Instant` bit is redundant.

That said, it's really a wash unless we want repeating timers; the only advantage to `Duration` is requiring one less import and one less line of boilerplate; no big deal, but I do think these things are worth weighing.

druid-shell/src/windows/mod.rs
```diff
@@ -1191,6 +1232,22 @@ impl<'a> WinCtx<'a> for WinCtxImpl<'a> {
             SetCursor(cursor);
         }
     }
+
+    /// Request a timer event.
+    ///
+    /// The return value is an identifier.
+    fn request_timer(&mut self, deadline: std::time::Instant) -> usize {
```
cmyr: okay taking a closer look now, this feels weird; we're basically using `0` as a null return value, but are we expecting the user to handle this?

'no' is okay, but i think it should be better documented. Otherwise this should probably be returning an `Option`, which also feels bad. This is probably okay, but it feels a bit iffy; we're designing a general API around a platform convention.

raphlinus: I don't think it'll fail, or if it does it's proabably ok to panic. (there might be a limit but it's extremely unlikely to hit it in practice). I'm basically just avoiding 0 in the call to SetTimer because it has a special purpose.

druid-shell/src/mac/mod.rs
```diff
@@ -662,6 +686,36 @@ impl<'a> WinCtx<'a> for WinCtxImpl<'a> {
...
+            msg_send![nstimer, scheduledTimerWithTimeInterval: ti target: view selector: selector userInfo: user_info repeats: NO];
+        }
+        TimerToken::new(token)
+    }
+}
+
+/// Convert an `Instant` into an NSTimeInterval, i.e. a fractional number
+/// of seconds from now.
+///
+/// This may loose some precision for multi-month durations.
```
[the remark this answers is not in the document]
cmyr: ah, I wondered why my editor was complaining about some unsaved change; i'd made this edit but I guess not saved before the commit. 🤦‍♀

raphlinus:
I'm good with this change, but it feels slightly awkward figuring out who approves it.

Oh, and I have perhaps a better motivation for Instant: a soft cursor blink animation will be made up of `request_anim_frame` for the smooth fade, mixed with timers for the periods of time when nothing changes. If you use duration, you have to take the time spent by the animated frames off, and it's not crystal-clear how to account for those. But with instant, very easy to reason about.

PR #113, opened 2019-08-09
Update for piet 0.0.6 (druid 0.4.0)

cmyr:
This fixes breakage and adopts new idioms where possible.

Looking at this, we probably don't need a version bump since we haven't published since 0.1.1 😨

### DIFF
--- a/examples/custom_widget.rs
+++ b/examples/custom_widget.rs
@@ -17,8 +17,7 @@
 use druid::kurbo::{Affine, BezPath, Point, Rect, Size};
 
 use druid::piet::{
-    Color, FillRule, FontBuilder, ImageFormat, InterpolationMode, RenderContext, Text,
-    TextLayoutBuilder,
+    Color, FontBuilder, ImageFormat, InterpolationMode, RenderContext, Text, TextLayoutBuilder,
 };
 
 use druid::shell::{runloop, WindowBuilder};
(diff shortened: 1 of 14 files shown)


examples/custom_widget.rs
```diff
@@ -53,17 +52,17 @@ impl Widget<String> for CustomWidget {
             (80.0, 90.0),
             (base_state.size().width, base_state.size().height),
         );
-        // Create a solid brush
-        let brush = paint_ctx.solid_brush(Color::rgb24(0x00_80_00));
+        // Create a color
+        let brush = Color::rgb8(0x00, 0x80, 0x00);
```
raphlinus: Very minor point, but I wonder if we should call this a "brush".

src/widget/button.rs
```diff
@@ -21,13 +21,13 @@ use crate::{
...
 
-use crate::piet::{Color, FillRule, FontBuilder, Text, TextLayoutBuilder};
+use crate::piet::{Color, FontBuilder, Text, TextLayoutBuilder};
 use crate::{Piet, Point, RenderContext};
 
-const BUTTON_BG_COLOR: Color = Color::rgba32(0x40_40_48_ff);
-const BUTTON_HOVER_COLOR: Color = Color::rgba32(0x50_50_58_ff);
-const BUTTON_PRESSED_COLOR: Color = Color::rgba32(0x60_60_68_ff);
-const LABEL_TEXT_COLOR: Color = Color::rgba32(0xf0_f0_ea_ff);
+const BUTTON_BG_COLOR: Color = Color::rgba8(0x40, 0x40, 0x48, 0xff);
```
raphlinus: I think if the alpha is 0xff, it should be omitted. OTOH this is going to go away soon :)

examples/custom_widget.rs
cmyr: you mean the variable name? or the comment? I don't think it's appropriate in the comment, since 'brush' isn't really a well-defined term. The variable name I don't mind.

raphlinus: The variable name. It used to be a brush type, now it's a color, which is "IntoBrush" to be sure, but is not actually a brush.

Issue #111, opened 2019-08-08
use text measurement to determine label width


cmyr:
this is currently hard-coded.

I think `Label` and `DynLabel` should also get moved into their own file.

andlr:
[This](https://github.com/xi-editor/druid/compare/master...andlr:label-width) gets the work done, but I am not sure if it is okay to use hard-coded `font_size` in both `paint` and `layout` methods.

cmyr:
@andir feel free to open a PR! keeping font_size hard-coded for now is okay, we can figure that out later.

LiHRaM:
Is this closed then, or are we using this to track the work on `font_size`?

PR #116, opened 2019-08-11
Use text measurement to determine label width

andlr:
#111

### DIFF
--- a/src/widget/button.rs
+++ b/src/widget/button.rs
@@ -21,8 +21,10 @@ use crate::{
     UpdateCtx, Widget,
 };
 
-use crate::piet::{Color, FontBuilder, Text, TextLayoutBuilder};
-use crate::{Piet, Point, RenderContext};
+use crate::piet::{
+    Color, FontBuilder, PietText, PietTextLayout, Text, TextLayout, TextLayoutBuilder,
+};
+use crate::{Point, RenderContext};
 
 const BUTTON_BG_COLOR: Color = Color::rgb8(0x40, 0x40, 0x48);
 const BUTTON_HOVER_COLOR: Color = Color::rgb8(0x50, 0x50, 0x58);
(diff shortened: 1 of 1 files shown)


src/widget/button.rs
```diff
@@ -55,16 +55,18 @@ impl Label {
         Label { text: text.into() }
     }
 
-    fn get_layout(&self, rt: &mut Piet, font_size: f64) -> <Piet as RenderContext>::TextLayout {
+    fn get_layout(
+        &self,
+        rt: &mut <Piet as RenderContext>::Text,
```
cmyr: when #113 lands, we'll be able to use `PietText` here (and `PietTextLayout`) below.

src/widget/button.rs
```diff
@@ -55,16 +55,18 @@ impl Label {
...
+        font_size: f64,
+    ) -> <Piet as RenderContext>::TextLayout {
         // TODO: caching of both the format and the layout
```
cmyr: If you're in here (and you feel like it) it might be a good time to resolve this TODO; we should only to construct the format once per font/size combo, and we should only need to construct the layout when the string changes.

src/widget/button.rs
cmyr: (that's merged, so if you rebase this now you can use the type aliases)

src/widget/button.rs
andlr: Well, honestly, I am not sure how exactly this should be implemented. 
>once per font/size combo

Do you mean using some hash map (per label or globally) to store formats or just constructing it only if font/size changes?

src/widget/button.rs
andlr: Okay :slightly_smiling_face: Maybe it's also a good time to move Label and DynLabel into their own file.

src/widget/button.rs
cmyr: What I mean is: we can store the font as an `Option<PietFont>`, in the label, and then only make it if it's missing; and we can store the layout as an `Option<(String, PietTextLayout)>`, and only make it when it's missing, or the string has changed.

cmyr: What I mean is: we can store the font as an `Option<PietFont>`, in the label, and then only make it if it's missing; and we can store the layout as an `Option<(String, PietTextLayout)>`, and only make it when it's missing, or the string has changed.

src/widget/button.rs
cmyr: How about we get this change merged, and then you can open a separate PR that adds the new file?

andlr: Sure :slightly_smiling_face:

src/widget/button.rs
cmyr: @andlr are you interested in making this change now, or would you like me to merge this, and then you can do a follow-up (or not) later if you like?

andlr: Yes, I think it is better to merge this now and add caching later with a separate PR

cmyr:
Awesome, thank you @andlr!

PR #117, opened 2019-08-13
Blinking cursor + text editing improvements

futurepaul:
If this is too messy I can strip out everything but the blinking, but thought I'd run it by you like this first. This version needs `unicode-segmentation` as a dependency, so that's also worth discussing.

Also, the get_layout function is copied from Label originally, so if we end up with [caching in Label](https://github.com/xi-editor/druid/pull/116) it would be good to pull that in here as well.

### DIFF
--- a/druid-shell/src/lib.rs
+++ b/druid-shell/src/lib.rs
@@ -14,6 +14,9 @@
 
 //! Platform abstraction for druid toolkit.
 
+pub use clipboard;
+pub use unicode_segmentation;
+
 pub use piet_common as piet;
 pub use piet_common::kurbo;
 
--- a/druid-shell/src/lib.rs
+++ b/druid-shell/src/lib.rs
@@ -17,8 +17,6 @@
 pub use piet_common as piet;
 pub use piet_common::kurbo;
 
-pub use unicode_segmentation;
-
 #[cfg(target_os = "windows")]
 #[macro_use]
 extern crate winapi;
--- a/druid-shell/src/lib.rs
+++ b/druid-shell/src/lib.rs
@@ -14,11 +14,10 @@
 
 //! Platform abstraction for druid toolkit.
 
-pub use clipboard;
-pub use unicode_segmentation;
-
 pub use piet_common as piet;
 pub use piet_common::kurbo;
+
+pub use unicode_segmentation;
 
 #[cfg(target_os = "windows")]
 #[macro_use]
--- a/src/widget/textbox.rs
+++ b/src/widget/textbox.rs
@@ -14,18 +14,25 @@
 
 //! A textbox widget.
 
+use std::cmp::{max, min};
+use std::time::{Duration, Instant};
+
 use crate::{
     Action, BaseState, BoxConstraints, Cursor, Env, Event, EventCtx, KeyCode, LayoutCtx, PaintCtx,
-    UpdateCtx, Widget,
+    TimerToken, UpdateCtx, Widget,
 };
 
 use crate::kurbo::{Affine, Line, Point, RoundedRect, Size, Vec2};
 use crate::piet::{Color, FontBuilder, Piet, RenderContext, Text, TextLayout, TextLayoutBuilder};
 
+use druid_shell::clipboard::{ClipboardContext, ClipboardProvider};
+use druid_shell::unicode_segmentation::GraphemeCursor;
+
 const BACKGROUND_GREY_LIGHT: Color = Color::rgba8(0x3a, 0x3a, 0x3a, 0xff);
 const BORDER_GREY: Color = Color::rgba8(0x5a, 0x5a, 0x5a, 0xff);
 const PRIMARY_LIGHT: Color = Color::rgba8(0x5c, 0xc4, 0xff, 0xff);
 
+const SELECTION_COLOR: Color = Color::rgb8(0xf3, 0x00, 0x21);
 const TEXT_COLOR: Color = Color::rgb8(0xf0, 0xf0, 0xEA);
 const CURSOR_COLOR: Color = Color::WHITE;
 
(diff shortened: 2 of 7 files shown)


druid-shell/src/lib.rs
```diff
@@ -17,6 +17,8 @@
 pub use piet_common as piet;
 pub use piet_common::kurbo;
 
+pub use unicode_segmentation;
```
raphlinus: Hmm. I'm not sure I want druid-shell to become a "batteries included" aggregator for libraries we want. It feels like this dependency is better in druid than druid-shell. We're doing this for kurbo and piet because those are necessary dependencies from druid-shell.

That said, the idea that druid depends on unicode-segmentation seems quite reasonable to me.

src/widget/textbox.rs
```diff
@@ -35,30 +44,131 @@ const BORDER_WIDTH: f64 = 1.;
...
             .unwrap()
             .build()
             .unwrap();
         text.new_text_layout(&font, data).unwrap().build().unwrap()
     }
+
+    fn insert(&mut self, src: &mut String, new: &str) {
+        // TODO: handle incomplete graphemes
+        let (left, right) = (self.selection.min(), self.selection.max());
+        src.replace_range(left..right, new);
```
raphlinus: TIL this method! It's great, such an improvement over doing truncation and push of slices by hand.

src/widget/textbox.rs
```diff
@@ -35,30 +44,131 @@ const BORDER_WIDTH: f64 = 1.;
...
+    }
+
+    fn cursor(&self) -> usize {
+        self.selection.end
+    }
+
+    fn backspace(&mut self, src: &mut String) {
+        if self.selection.is_caret() {
+            let cursor = self.cursor();
+            let new_cursor = prev_grapheme(&src, cursor);
```
raphlinus: I'd put in a TODO that grapheme isn't really the right unit. FYI, I consider https://github.com/xi-editor/xi-editor/blob/master/rust/core-lib/src/backspace.rs to be pretty much the authoritative source of what this logic should be.

src/widget/textbox.rs
```diff
@@ -90,30 +199,63 @@ impl Widget<String> for TextBox {
...
+                let text_pos = Point::new(0.0 + PADDING_LEFT, text_height + PADDING_TOP);
+
+                let cursor_x = self.substring_measurement_hack(rc, data, 0, self.cursor());
+
                 // If overflowing, shift the text
-                if text_layout.width() + (PADDING_LEFT * 2.) > self.width {
-                    let offset = text_layout.width() - self.width + (PADDING_LEFT * 2.) + 1.;
-                    rc.transform(Affine::translate(Vec2::new(-offset, 0.)));
+                let padding = PADDING_LEFT * 2.;
+                if cursor_x > self.width + self.hscroll_offset - padding {
```
raphlinus: In *general* I like paint to be a more or less pure function that just paints the view; this logic feels like it should be split out and invoked from anything that causes a change (event, update). But that's more of a feeling; paint is in fact allowed to do stuff like this.

src/widget/textbox.rs
```diff
@@ -90,30 +199,63 @@ impl Widget<String> for TextBox {
...
+                    self.hscroll_offset = cursor_x - self.width + padding;
+                } else if cursor_x < self.hscroll_offset {
+                    // If cursor goes past left side, match the offset
+                    //    <-
+                    // **[I****]****
+                    //   ^
+                    self.hscroll_offset = cursor_x
                 }
+
+                rc.transform(Affine::translate(Vec2::new(-self.hscroll_offset, 0.)));
```
raphlinus: Minor style point, the `Vec2::new` can be elided because the translate constructor takes its argument by `Into<Vec2>`.

src/widget/textbox.rs
```diff
@@ -144,18 +289,102 @@ impl Widget<String> for TextBox {
...
+                    ctx.invalidate();
+                    let deadline = Instant::now() + Duration::from_millis(500);
+                    self.cursor_timer = ctx.request_timer(deadline);
+                }
+            }
             Event::KeyDown(key_event) => {
                 match key_event {
+                    // Copy (Ctrl+C || Cmd+C)
+                    event
+                        if (event.mods.meta || event.mods.ctrl)
```
raphlinus: Note that `druid_shell::keycodes::command_modifier()` tells you which modifier is the command modifier for the platform. It's not available in easy-to-use form though. Also see #81 for related ideas.

This question of keybindings etc needs attention (including the fact that a lot of these are likely best expressed as shortcuts available on an "Edit" menu), but I'm happy getting something useful in early.

src/widget/textbox.rs
futurepaul: That's a good rule of thumb. I moved it to a method and am calling from event now. In doing so, I realized that there were definitely times I was doing this calculation that I didn't need to be doing it!

src/widget/textbox.rs
futurepaul: I couldn't figure out how to compare a `KeyModifier` with a `Modifier`. Would it be okay to do like this?

```rust
const fn command_modifier() -> KeyModifiers {
    #[cfg(target_os = "macos")]
    {
        KeyModifiers {
            shift: false,
            alt: false,
            ctrl: false,
            meta: true,
        }
}
    #[cfg(not(target_os = "macos"))]
    {
        KeyModifiers {
            shift: false,
            alt: false,
            ctrl: false,
            meta: true,
        }
    }
}
```

cmyr: mm, this is interesting.

I think that the correct way to handle this is to have some sort of `CommandList` struct where you register the commands you're looking for, which can have per-platform overrides, and then you pass events to that struct to determine if a command has occurred, but I'll need to think about that a bit more.

For the time being, I think adding `fn contains_sys_command_modifier(self) -> bool` to `KeyModifier` is... okay. 🤢

src/widget/textbox.rs
```diff
@@ -35,30 +44,153 @@ const BORDER_WIDTH: f64 = 1.;
...
+            .new_font_by_name("Segoe UI", font_size)
             .unwrap()
             .build()
             .unwrap();
         text.new_text_layout(&font, data).unwrap().build().unwrap()
     }
+
+    fn insert(&mut self, src: &mut String, new: &str) {
+        // TODO: handle incomplete graphemes
+        let (left, right) = (self.selection.min(), self.selection.max());
```
cmyr: Non-blocking: you could have a fn `Selection::range() -> Range<usize>`, if you felt like it.

src/widget/textbox.rs
```diff
@@ -35,30 +44,153 @@ const BORDER_WIDTH: f64 = 1.;
...
+            let new_cursor = prev_grapheme(&src, cursor);
+            src.replace_range(new_cursor..cursor, "");
+            self.cursor_to(new_cursor);
+        } else {
+            src.replace_range(self.selection.min()..self.selection.max(), "");
+            self.cursor_to(self.selection.min());
+        }
+    }
+
+    fn copy_text(&self, input: String) {
```
cmyr: maybe throw in a `TODO` note here, explaining you're waiting for copy/past to get implemented?

src/widget/textbox.rs
futurepaul: I like it.

futurepaul:
Alright, I believe I've addressed everything. I also added a few more doc strings. Since I'm not quite following your ambitions for event matching ergonomics, I think `fn contains_sys_command_modifier(self) -> bool` looks great! :)

cmyr:
Feel like I had a bit of a brain fart, since `contains_sys_command_modifier` will also return true for multi-modifier combos, which we might not want. This might be okay for now, but has clear limitations longer-term. I'm going to sketch out some command handling stuff today, and see if we can't improve that situation.

futurepaul:
Okay, rebased to #121. I think it went well. Because of the `matches` method instead of PartialEq I figured it would be best to use HotKeys throughout, even where it isn't strictly required, so that nobody gets confused with `==`.

Also I caught an edgecase with hscroll that I was missing! As simple as hscroll logic looks, it's really been a brain challenge for me to do it correctly.

cmyr:
okay, cool.

This makes me wonder if I should actually have a fn `matches` on `KeyEvent`, which takes an `impl Into<HotKey>`; this would let us just have code like,

```rust

match event {
    k_e if k_e.matches((SysMod::Cmd, "c")) => (),
    k_e if k_e.matches((None, KeyCode::LeftArrow)) => (),
}
```

Although this might require a bunch of this stuff to go into druid-shell. That might make sense anyway, though...

src/widget/textbox.rs
```diff
@@ -14,18 +14,28 @@
...
 use crate::kurbo::{Affine, Line, Point, RoundedRect, Size, Vec2};
-use crate::piet::{Color, FontBuilder, Piet, RenderContext, Text, TextLayout, TextLayoutBuilder};
+use crate::piet::{
+    Color, FontBuilder, PietText, PietTextLayout, RenderContext, Text, TextLayout,
+    TextLayoutBuilder,
+};
+
+use crate::unicode_segmentation::GraphemeCursor;
 
 const BACKGROUND_GREY_LIGHT: Color = Color::rgba8(0x3a, 0x3a, 0x3a, 0xff);
```
cmyr: one last thought: since env has landed, we can start moving these sorts of things into that, if we like? But I'm happy for this to be followup.

futurepaul: I was planning on a styled widgets PR next, so I think I'd prefer to merge now.

cmyr: sounds good!

PR #118, opened 2019-08-13
Lens derive and lens_wrap!()

futurepaul:
I'd love input on everything, of course, but I'm especially hoping for some suggestions on the lens_wrap!() macro. For instance, right now it can parse this:

```rust
    let label = DynLabel::new(|data: &String, _env| data.clone());
    let display = lens_wrap!(label, CalcState::value);
```

but not this:

```rust
    let display = lens_wrap!(DynLabel::new(|data: &String, _env| data.clone()), CalcState::value);
```

Also, the `CalcState::value` syntax itself might need some bikeshedding. Not exactly sure how discoverable it is outside of copy / pasting from existing examples.

Additionally, the folder pollution bugs me, especially for something this small. Maybe there's somewhere better to hide all this other than the root level?


lens-macro-impl/src/lib.rs
```diff
@@ -0,0 +1,180 @@
...
+fn is_camel_case(name: &str) -> bool {
+    let name = name.trim_matches('_');
+    if name.is_empty() {
+        return true;
+    }
+
+    // start with a non-lowercase letter rather than non-uppercase
+    // ones (some scripts don't have a concept of upper/lowercase)
+    !name.chars().next().unwrap().is_lowercase()
+        && !name.contains("__")
```
cmyr: didn't we strip this above?

lens-macro-impl/src/lib.rs
```diff
@@ -0,0 +1,180 @@
...
+        && !name.chars().collect::<Vec<_>>().windows(2).any(|pair| {
```
cmyr: A fancy zero-allocation way to do this would be with `peek`, but this is totally fine as is.

lens-macro-impl/src/lib.rs
```diff
@@ -0,0 +1,180 @@
...
+        named
+    } else {
+        unimplemented!();
+    };
+
+    let twizzled_name = if is_camel_case(&name.to_string()) {
+        let temp_name = to_snake_case(&mut name.to_string());
+        Ident::new(&temp_name, Span::call_site())
+    } else {
+        unimplemented!();
```
cmyr: is there some other reasonable path we can take here? like just prepending some prefix and using the name as-is?

cmyr: Or if we want to say that you need to use camelCase, we can make this a `compile_error` with some diagnostic. If we want helpful diagnostics, we can also return a `syn::Error` here, which should be associated with the span of the type.

lens-macro/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "lens-macro"
```
cmyr: We might consider calling this `lens-derive`, which is a common naming convention in other crates, but this does contain a non-derive macro as well so this seems okay too.

lens-macro-impl/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "lens-macro-impl"
```
cmyr: I wonder if we can't put this crate _inside_ the main `lens-macro` crate's root? That would clean things up somewhat.

lens-macro/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+use proc_macro_hack::proc_macro_hack;
```
cmyr: I would still include at least a module-level docstring here, and likely also a copyright notice, because I'm anal-retentive like that.

futurepaul:
After a long slumber I'm finally revisiting this! Borrowing from the structure of the Data derive macro was helpful, and I'm sure there are ways this could be cleaned up further by someone with a stronger understanding.

Sorry for the discontinuity between the commits... I did a bit too aggressive of a rebase and the files needed to be in new locations anyways.

I'm dropping the LensWrap macro for now for simplicity's sake (and because @cmyr raises some good points about ergonomics!)

lens-macro-impl/src/lib.rs
futurepaul: you would think! but I copy / pasted all of these [case-related functions from rustc](https://github.com/rust-lang/rust/blob/590ae0ec4d0c782f7cf97cff7474dc4012c1b615/src/librustc_lint/nonstandard_style.rs#L48)... it feels like if I agree with you here I'll end up needing to put in a pull request :)

lens-macro-impl/src/lib.rs
futurepaul: another rustc PR!

lens-macro/src/lib.rs
futurepaul: now that this is gone do I still need a dosctring somewhere? not sure where to put it

lens-macro/Cargo.toml
futurepaul: I went with druid-derive-lens to match the data one

lens-macro-impl/src/lib.rs
futurepaul: would need some way of telling the user that they need to prefix in this scenario, yes?

futurepaul:
Cool! I'll open an issue for further Lens work.

Afaik the only failure states here are at compile time. Am I misunderstanding?

cmyr:
Oops yea I wasn't clear: allowing lenses to get items from collections is fallible, because you might have an invalid index.

PR #121, opened 2019-08-14
Add better hotkey support

cmyr:
This is another attempt at improving the ergonomics of handling
keyboard events; as a bonus, this will also be useful for doing
menus.

closes #81

This spun off of some other work I was doing on commands, but I think it can stand alone. I'd be curious to get @futurepaul's opinion, in terms of whether this would improve the code in #117; in particular it offers a workaround for the "system command key". For example, with this, you can do:

```rust
let event: KeyEvent = ...;
match event {
    e if e == HotKey::new(SysMods::Cmd, "c") => self.copy(),
    e if e == HotKey::new(SysMods::Cmd, "v") => self.paste(),
    _ => (),
}
```
and it will work with 'Command' on macOS and 'Ctrl' elsewhere.

### DIFF
--- a/src/hotkey.rs
+++ b/src/hotkey.rs
@@ -18,8 +18,8 @@ use crate::{KeyCode, KeyEvent, KeyModifiers};
 
 /// A description of a keyboard shortcut.
 ///
-/// This type implements `PartialEq<KeyEvent>`, and it is only intended to
-/// be used to describe shortcuts, and recognize them when they arrive.
+/// This type is only intended to be used to describe shortcuts,
+/// and recognize them when they arrive.
 ///
 /// # Examples
 ///
@@ -31,21 +31,21 @@ use crate::{KeyCode, KeyEvent, KeyModifiers};
 /// let hotkey = HotKey::new(SysMods::Cmd, "a");
 ///
 /// #[cfg(target_os = "macos")]
-/// assert!(hotkey == KeyEvent::for_test(RawMods::Meta, "a", KeyCode::KeyA));
+/// assert!(hotkey.matches(KeyEvent::for_test(RawMods::Meta, "a", KeyCode::KeyA)));
 ///
 /// #[cfg(target_os = "windows")]
-/// assert!(hotkey == KeyEvent::for_test(RawMods::Ctrl, "a", KeyCode::KeyA));
+/// assert!(hotkey.matches(KeyEvent::for_test(RawMods::Ctrl, "a", KeyCode::KeyA)));
 /// ```
 ///
-/// `None` matches only the key withuot modifiers:
+/// `None` matches only the key without modifiers:
 ///
 /// ```
 /// use druid::{HotKey, KeyEvent, RawMods, SysMods, KeyCode};
 ///
 /// let hotkey = HotKey::new(None, KeyCode::ArrowLeft);
 ///
-/// assert!(hotkey == KeyEvent::for_test(RawMods::None, "", KeyCode::ArrowLeft));
-/// assert!(hotkey != KeyEvent::for_test(RawMods::Ctrl, "", KeyCode::ArrowLeft));
+/// assert!(hotkey.matches(KeyEvent::for_test(RawMods::None, "", KeyCode::ArrowLeft)));
+/// assert!(!hotkey.matches(KeyEvent::for_test(RawMods::Ctrl, "", KeyCode::ArrowLeft)));
 /// ```
 ///
 /// [`SysMods`]: enum.SysMods.html
@@ -113,7 +113,21 @@ impl HotKey {
         }
         self
     }
+
+    /// Returns `true` if this [`KeyEvent`] matches this `HotKey`.
+    ///
+    /// [`KeyEvent`]: struct.KeyEvent.html
+    pub fn matches(&self, event: impl Borrow<KeyEvent>) -> bool {
+        let event = event.borrow();
+        self.mods == event.mods
+            && match self.key {
+                KeyCompare::Code(code) => code == event.key_code,
+                KeyCompare::Text(text) => Some(text) == event.text(),
+            }
+    }
 }
+
+use std::borrow::Borrow;
 
 /// A platform-agnostic representation of keyboard modifiers, for command handling.
 ///
@@ -154,22 +168,6 @@ pub enum RawMods {
     AltMetaShift,
     CtrlMetaShift,
     AltCtrlMetaShift,
-}
-
-impl std::cmp::PartialEq<KeyEvent> for HotKey {
-    fn eq(&self, other: &KeyEvent) -> bool {
-        self.mods == other.mods
-            && match self.key {
-                KeyCompare::Code(code) => code == other.key_code,
-                KeyCompare::Text(text) => Some(text) == other.text(),
-            }
-    }
-}
-
-impl std::cmp::PartialEq<HotKey> for KeyEvent {
-    fn eq(&self, other: &HotKey) -> bool {
-        other == self
-    }
 }
 
 impl std::cmp::PartialEq<KeyModifiers> for RawMods {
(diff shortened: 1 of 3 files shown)


futurepaul:
I love this SysMods -> RawMods mapping. Seems to cover all my current uses super well.

One point of concern from a learnability perspective: with this setup, the syntax for Shift+ArrowLeft is very different than ArrowLeft alone. Would `HotKey::new(None, KeyCode::ArrowLeft)` be a crazy thing to have? (Some applications like video editors do have a lot of menu-level "hotkeys" bound to simple keycodes, so if this becomes a menu-related thing it makes sense in that regard). Not a big deal but just want to minimize the number of concepts people need if possible.

This also made me think of the mousewheel + command / shift + drag / command + click situations. Not sure if this impacts those at all?

cmyr:
There are both `SysMods::None` and `RawMods::None`, so you can do `HotKey::new(RawMods::None, KeyCode::ArrowLeft)`. I could maybe play some type-tetris to get just `None` to work, will take a look at that.

Thanks for bringing up mods + mouse action, hadn't thought of that, will make sure everything makes sense.

A natural extension of this is something like this (I had this written already, just didn't include it in this PR):

```rust
/// A mapping of hotkeys to commands.
///
/// This is a way of encapsulating relatively static key event logic.
pub struct CmdList<T> {
    cmds: Vec<Entry<T>>,
}

/// A hotkey/command pair.
#[derive(Debug)]
struct Entry<T> {
    cmd: HotKey,
    event: T,
}

impl<T> CmdList<T> {
    pub fn new() -> Self {
        CmdList { cmds: Vec::new() }
    }

    /// Add a new hotkey, mapped to an event.
    ///
    /// The first two arguments to this function describe the hotkey, and
    /// the last argument is the event to return if that hotkey is matched.
    ///
    /// See [`HotKey::new`] for an explanation of these arguments.
    ///
    /// # Platform considerations
    ///
    /// Hotkeys are often platform specific. For this reason, you will often
    /// want to set per-platform hotkeys, which you can do using `cfg` blocks.
    /// [`HotKey::new`]: struct.HotKey.html#method.new
    pub fn add(&mut self, mods: impl Into<RawMods>, key: impl Into<KeyCompare>, event: T) {
        let cmd = HotKey::new(mods, key);
        let entry = Entry { cmd, event };
        self.cmds.push(entry);
    }

    pub fn cmd_for_event(&self, event: &KeyEvent) -> Option<&T> {
        self.cmds
            .iter()
            .find(|entry| &entry.cmd == event)
            .map(|entry| &entry.event)
    }
}

// example:

use druid::{CmdList, RawMods, SysMods};
enum MyCommands {
    Clear,
    Copy,
    Paste,
    Quit,
}

fn make_cmd_list() -> CmdList<MyCommands> {
    let mut cmds = CmdList::new();
    // Commands that use 'Command' on macOS and 'Ctrl' elsewhere can be represented
    // with `SysMods::Cmd`.`
    cmds.add(SysMods::Cmd, "c", MyCommands::Copy);
    cmds.add(SysMods::Cmd, "v", MyCommands::Paste);
    // If a mapping only makes sense on one platform, put it behind a cfg flag.
    #[cfg(target_os = "macos")]
    cmds.add(SysMods::Cmd, "q", MyCommands::Quit);
    #[cfg(target_os = "windows")]
    cmds.add(SysMods::Alt, KeyCode::F4, MyCommands::Quit);
    // When adding commands for a specific platform, you have more control
    // if you use `RawMods`; for instance, if you want Command + Ctrl on macOS.
    #[cfg(target_os = "macos")]
    cmds.add(RawMods::CtrlMeta, "k", MyCommands::Clear);
    
    cmds
}

// in some type that has a copy of the above command list
fn handle_key_event(&self, event: &KeyEvent) {
    match self.cmds.cmd_for_event(event) {
        Some(MyCommands::Clear) => self.clear(),
        Some(MyCommands::Copy) => self.copy(),
        // ...
        _ => (),
    }
}
```

The idea here is that you can express complicated command logic once, and then reuse that mapping later; This also means that we can trivially include like a `TextCmdList` that will handle the Cocoa text system shortcuts.

I'm not really sure yet if this is much of a win, though, so I wanted to wait until I could play around a bit more.

cmyr:
Okay, made some tweaks: In particular, for other event types you can now do, say,

```rust
let scroll = WheelEvent;
if scroll.mods == SysMods::Cmd {
    self.zoom();
}
```

This will also work for mouse events, although I'd like to make some other additional improvements there as well.

src/hotkey.rs
```diff
@@ -0,0 +1,269 @@
...
+/// let hotkey = HotKey::new(SysMods::Cmd, "a");
+///
+/// #[cfg(target_os = "macos")]
+/// assert!(hotkey == KeyEvent::for_test(RawMods::Meta, "a", KeyCode::KeyA));
+///
+/// #[cfg(target_os = "windows")]
+/// assert!(hotkey == KeyEvent::for_test(RawMods::Ctrl, "a", KeyCode::KeyA));
+/// ```
+///
+/// `None` matches only the key withuot modifiers:
```
raphlinus: "without"

src/hotkey.rs
```diff
@@ -0,0 +1,269 @@
...
+    /// or an instance of either [`SysMods`], or [`RawMods`]. [`SysMods`] unify the
+    /// 'Command' key on macOS with the 'Ctrl' key on other platforms.
+    ///
+    /// The second argument describes the non-modifier key. This can be either
+    /// a `&'static str` or a [`KeyCode`]. If it is a `&str`, it will be compared
+    /// against the [`unmodified text`] for a given key event; if it is a [`KeyCode`]
+    /// it will be compared against the event's key code.
+    ///
+    /// In general, [`KeyCode`] should be preferred for non-printing keys, like
+    /// the arrows or backspace.
```
raphlinus: This docstring leaves me not certain whether Shift-A is `(RawMods::Shift, "a")` or `(RawMods::Shift, "a")`. I see a TODO below; that seems reasonable and we can get back to this (especially win vs mac).

src/hotkey.rs
```diff
@@ -0,0 +1,269 @@
...
+    Cmd,
+    /// Command + Alt on macOS, Ctrl + Alt on windows/linux
+    AltCmd,
+    /// Command + Shift on macOS, Ctrl + Shift on windows/linux
+    CmdShift,
+    /// Command + Alt + Shift on macOS, Ctrl + Alt + Shift on windows/linux
+    AltCmdShift,
+}
+
+//TODO: should something like this just _replace_ keymodifiers?
```
raphlinus: This is a pretty deep question. There are potentially more than four modifier (this isn't even exotic if you distinguish left and right shift).

src/hotkey.rs
```diff
@@ -0,0 +1,269 @@
...
+    CtrlMeta,
+    MetaShift,
+    AltCtrlMeta,
+    AltCtrlShift,
+    AltMetaShift,
+    CtrlMetaShift,
+    AltCtrlMetaShift,
+}
+
+impl std::cmp::PartialEq<KeyEvent> for HotKey {
```
raphlinus: I am a bit uncomfortable with `PartialEq` here because it doesn't strictly follow the transitive rule; you can get a violation if you can for example make two KeyEvents with different codes but the same text. I'm willing to suspend disbelief because I do see the ergonomics advantage, but would prefer acknowledging that it's a bit of abuse of notation.

src/hotkey.rs
cmyr: yea I'm personally confused about this situation right now. On mac, shift is applied, so 'unmod text' is "A" for shift+a, and "a" for ctrl+a. We could case normalize, but I haven't thought through all the edge cases.

src/hotkey.rs
cmyr: I can take it or leave it for `KeyEvent`, `hotkey.matches(&event)` isn't a huge headache.

cmyr:
Okay I'm going to merge this, may iterate a bit but I think it will clean up #117 quite a bit, and it's important for menus (my next/current project)

PR #122, opened 2019-08-19
Add alignment widget

raphlinus:


src/widget/align.rs
```diff
@@ -0,0 +1,85 @@
...
+    fn layout(
+        &mut self,
+        layout_ctx: &mut LayoutCtx,
+        bc: &BoxConstraints,
+        data: &T,
+        env: &Env,
+    ) -> Size {
+        let size = self.child.layout(layout_ctx, &bc.loosen(), data, env);
+        let my_size = bc.constrain(size);
+        let extra_size = (my_size.to_vec2() - size.to_vec2()).to_size();
```
futurepaul: When this is negative it gets pretty weird. This fixed it for me:

```rust
let extra_width = extra_size.width.max(0.0);
let extra_height = extra_size.height.max(0.0);
let origin = self.align.resolve(Rect::from_origin_size(
    Point::ORIGIN,
    Size::new(extra_width, extra_height),
));
```

raphlinus: Yeah, we need to be a lot more systematic about the way we handle constraint violations / negative sizes. I'm happy to put this in as a patch but do want a careful review at some point, including figuring out logging of warnings. I've wanted to study Flutter to see how it handles that also, but haven't had the time.

futurepaul:
I'm fine to merge this. My local version handles infinity:

```rust
if bc.max().width == std::f64::INFINITY {
    my_size.width = bc.constrain(size).width;
}
if bc.max().height == std::f64::INFINITY {
    my_size.height = bc.constrain(size).height;
}
```

But that could be a whole additional conversation about what (if any) methods to add to BoxConstraints, and I'm ready to have that convo in my own branch which builds off of Align.

Also, it seems like we might want methods for each of the nine basic UnitPoints, but I figure it would be good have that motivated by actual use first — maybe we'll want to wrap this in a Container and use it rarely by itself, who knows.

raphlinus:
I tried to handle infinity too, just in the general case - in the common case, I think extra_width and extra_height will compute to zero. I'm curious if you have a specific example where the logic in this version of the PR can be improved.

futurepaul:
I'm sorry, I left off a line:

```rust
let mut my_size = bc.max();
if bc.max().width == std::f64::INFINITY {
    my_size.width = bc.constrain(size).width;
}
if bc.max().height == std::f64::INFINITY {
    my_size.height = bc.constrain(size).height;
}
```

The point is for Align's size to be bc.max() (not bc.constrain(size)), except cases of infinity.

If you put this into widget_demo you can see the problem:
```rust
let mut col = Column::new();
let bar = ProgressBar::default();

let button_1 = ActionWrapper::new(Button::new("increment "), move |data: &mut f64, _env| {
    *data += 0.1
});

col.add_child(Padding::uniform(5.0, bar), 1.0);
col.add_child(Padding::uniform(5.0, Align::centered(button_1)), 1.0);

let state = UiState::new(Align::centered(col), 0.7f64);
```

With my code, the bar still fills the whole width and the button is centered. In your version, without bc.max(), Align sizes itself to be the size of the button, and therefore ends up positioned on the left. If you remove the (admittedly silly) Align::centered wrapper from the col:
```
let state = UiState::new(col, 0.7f64);
```
then both versions work correctly.

raphlinus:
Ah, got it. I missed some logic in the Flutter version I was working from. I'll revise it.

futurepaul:
👍 
This looks good and thanks for the is_bounded methods those are going to be useful!

PR #124, opened 2019-08-20
Localization

cmyr:
This adds basic localization, based on Fluent.

tldr;

We have added a `LocalizedString` type. To convert this to an actual string, we pass in the `Env` and the  `Data`; this will lazily compute the concrete value for this string, for the current locale.

As a demo, `LocalizedString` can now be used in `Label`.

This is fairly rough, but is also not cutting corners; the strategy employed here is a reasonable long-term approach to localization.

There are a few things that could be refined:

- **lenses** a localized string can accept arguments, such as when you want to format a number. Currently you provide arguments indirectly, by providing a closure that can generate the required value from the current data and environment. I think this is an okay initial solution, but this is another place where lenses could really shine; in particular it would let us invalidate strings more efficiently.

- **`Env`**: I've jammed a bunch of new stuff in here, via a new `L10nManager` struct. I'm not sure if this should live in `Env` or somewhere else.

- **resource bundles**: The main limitation of this right now is that there is no system for packaging a druid application along with other necessary resources, so this will only work correctly for examples that are run from the `druid` root directory. We use `include_str!` to include the `en_US` localization with the app, so that we have a reasonable fallback.

### DIFF
--- a/druid-shell/src/windows/util.rs
+++ b/druid-shell/src/windows/util.rs
@@ -280,3 +280,13 @@ fn attach_console() {
         }
     }
 }
+
+/// Returns the current locale string.
+///
+/// This should a [Unicode language identifier].
+///
+/// [Unicode language identifier]: https://unicode.org/reports/tr35/#Unicode_language_identifier
+pub fn get_locale() -> String {
+    //TODO ahem
+    "en_US".into()
+}
--- a/druid-shell/src/windows/util.rs
+++ b/druid-shell/src/windows/util.rs
@@ -288,5 +288,5 @@ fn attach_console() {
 /// [Unicode language identifier]: https://unicode.org/reports/tr35/#Unicode_language_identifier
 pub fn get_locale() -> String {
     //TODO ahem
-    "en_US".into()
+    "en-US".into()
 }
--- a/examples/hello.rs
+++ b/examples/hello.rs
@@ -13,8 +13,8 @@
 // limitations under the License.
 
 use druid::shell::{runloop, WindowBuilder};
-use druid::widget::{ActionWrapper, Align, Button, Column, DynLabel, Padding};
-use druid::{UiMain, UiState};
+use druid::widget::{ActionWrapper, Align, Button, Column, Label, Padding};
+use druid::{LocalizedString, UiMain, UiState};
 
 fn main() {
     druid::shell::init();
@@ -22,7 +22,9 @@ fn main() {
     let mut run_loop = runloop::RunLoop::new();
     let mut builder = WindowBuilder::new();
     let mut col = Column::new();
-    let label = DynLabel::new(|data: &u32, _env| format!("value: {}", data));
+    let text =
+        LocalizedString::new("hello-counter").with_arg("count", |_env, data: &u32| (*data).into());
+    let label = Label::new(text);
     let button = Button::new("increment");
     col.add_child(Align::centered(Padding::uniform(5.0, label)), 1.0);
     col.add_child(Padding::uniform(5.0, button), 1.0);
--- a/resources/i18n/en-US/builtin.ftl
+++ b/resources/i18n/en-US/builtin.ftl
@@ -1,3 +1,3 @@
-app-name = Druid
+-app-name = Druid
 
 hello-counter = Current value is { $count }
--- a/resources/i18n/en-US/builtin.ftl
+++ b/resources/i18n/en-US/builtin.ftl
@@ -1,0 +1,9 @@
+app-name = Druid
+macos-menu-about-app = About { $app_name }
+macos-menu-preferences = Preferables...
+macos-menu-hide-app = Hide { $app_name }
+macos-menu-hide-others = Hide Others
+macos-menu-show-all = Show All
+macos-menu-services = Services
+macos-menu-application-menu = { $app_name }
+macos-menu-quit = Quit { $app_name }
--- a/resources/i18n/en-US/builtin.ftl
+++ b/resources/i18n/en-US/builtin.ftl
@@ -1,9 +1,3 @@
 app-name = Druid
-macos-menu-about-app = About { $app_name }
-macos-menu-preferences = Preferables...
-macos-menu-hide-app = Hide { $app_name }
-macos-menu-hide-others = Hide Others
-macos-menu-show-all = Show All
-macos-menu-services = Services
-macos-menu-application-menu = { $app_name }
-macos-menu-quit = Quit { $app_name }
+
+hello-counter = Current value is { $count }
--- a/src/env.rs
+++ b/src/env.rs
@@ -37,14 +37,12 @@ use crate::Data;
 /// so that an entire subtree can be disabled ("grayed out") with one
 /// setting.
 #[derive(Clone)]
-pub struct Env {
-    items: Arc<EnvImpl>,
-    l10n: Arc<L10nManager>,
-}
+pub struct Env(Arc<EnvImpl>);
 
-#[derive(Clone, Default)]
+#[derive(Clone)]
 struct EnvImpl {
     map: HashMap<String, Value>,
+    l10n: Arc<L10nManager>,
 }
 
 /// A typed key.
@@ -108,7 +106,7 @@ impl Env {
     ///
     /// Panics if the key is not found, or if it is present with the wrong type.
     pub fn get<'a, V: ValueType<'a>>(&'a self, key: Key<V>) -> V {
-        if let Some(value) = self.items.map.get(key.key) {
+        if let Some(value) = self.0.map.get(key.key) {
             value.to_inner_unchecked()
         } else {
             panic!("key for {} not found", key.key)
@@ -121,7 +119,7 @@ impl Env {
     ///
     /// Panics if the value for the key is found, but has the wrong type.
     pub fn try_get<'a, V: ValueType<'a>>(&'a self, key: Key<V>) -> Option<V> {
-        self.items
+        self.0
             .map
             .get(key.key)
             .map(|value| value.to_inner_unchecked())
@@ -129,7 +127,7 @@ impl Env {
 
     /// Adds a key/value, acting like a builder.
     pub fn adding<'a, V: ValueType<'a>>(mut self, key: Key<V>, value: impl Into<V::Owned>) -> Env {
-        let env = Arc::make_mut(&mut self.items);
+        let env = Arc::make_mut(&mut self.0);
         env.map.insert(key.into(), value.into().into());
         self
     }
@@ -141,7 +139,7 @@ impl Env {
     /// Panics if the environment already has a value for the key, but it is
     /// of a different type.
     pub fn set<'a, V: ValueType<'a>>(mut self, key: Key<V>, value: impl Into<V::Owned>) {
-        let env = Arc::make_mut(&mut self.items);
+        let env = Arc::make_mut(&mut self.0);
         let value = value.into().into();
         let key = key.into();
         // TODO: use of Entry might be more efficient
@@ -161,7 +159,7 @@ impl Env {
     ///
     /// [`L10nManager`]: struct.L10nManager.html
     pub fn localization_manager(&self) -> &L10nManager {
-        &self.l10n
+        &self.0.l10n
     }
 }
 
@@ -249,7 +247,7 @@ impl Data for Value {
 
 impl Data for Env {
     fn same(&self, other: &Env) -> bool {
-        Arc::ptr_eq(&self.items, &other.items) || self.items.deref().same(other.items.deref())
+        Arc::ptr_eq(&self.0, &other.0) || self.0.deref().same(other.0.deref())
     }
 }
 
@@ -266,10 +264,11 @@ impl Data for EnvImpl {
 impl std::default::Default for Env {
     fn default() -> Self {
         let l10n = L10nManager::new(vec!["builtin.ftl".into()], "./resources/i18n/");
-        Env {
+        let inner = EnvImpl {
             l10n: Arc::new(l10n),
-            items: Arc::default(),
-        }
+            map: HashMap::new(),
+        };
+        Env(Arc::new(inner))
     }
 }
 
--- a/src/env.rs
+++ b/src/env.rs
@@ -21,7 +21,7 @@ use std::ops::Deref;
 use std::sync::Arc;
 
 use fluent::{FluentArgs, FluentBundle, FluentResource};
-use unic_langid::langid;
+use unic_langid::{langid, LanguageIdentifier};
 
 use crate::kurbo::{Point, Rect, Size};
 use crate::piet::{Color, LinearGradient};
@@ -286,7 +286,11 @@ impl std::default::Default for Env {
         let resource =
             FluentResource::try_new(en_strings.to_owned()).expect("Could not parse en-US/builtin.ftl");
         let langid_en = langid!("en");
-        let mut bundle = FluentBundle::new(&[langid_en]);
+        let locale: LanguageIdentifier = crate::shell::get_locale()
+            .parse()
+            .unwrap_or_else(|_| langid_en.clone());
+        eprintln!("using locale {:?}", locale);
+        let mut bundle = FluentBundle::new(&[locale, langid_en]);
         bundle
             .add_resource(resource)
             .expect("Failed to add FTL resources to the bundle.");
--- a/src/env.rs
+++ b/src/env.rs
@@ -20,17 +20,11 @@ use std::marker::PhantomData;
 use std::ops::Deref;
 use std::sync::Arc;
 
-use fluent::{FluentArgs, FluentBundle, FluentResource};
-use unic_langid::{langid, LanguageIdentifier};
-
 use crate::kurbo::{Point, Rect, Size};
 use crate::piet::{Color, LinearGradient};
 
+use crate::localization::L10nManager;
 use crate::Data;
-
-type I18nBundle = Arc<FluentBundle<FluentResource>>;
-
-pub const APP_NAME: Key<&str> = Key::new("druid.app-name");
 
 /// An environment passed down through all widget traversals.
 ///
@@ -45,7 +39,7 @@ pub const APP_NAME: Key<&str> = Key::new("druid.app-name");
 #[derive(Clone)]
 pub struct Env {
     items: Arc<EnvImpl>,
-    builtin_i18n: I18nBundle,
+    l10n: Arc<L10nManager>,
 }
 
 #[derive(Clone, Default)]
@@ -162,23 +156,12 @@ impl Env {
         env.map.insert(key, value);
     }
 
-    /// Fetch a localized string from the current bundle by key.
-    pub fn localize<'args>(
-        &'args self,
-        key: &str,
-        args: impl Into<Option<&'args FluentArgs<'args>>>,
-    ) -> Option<String> {
-        let args = args.into();
-        let value = match self.builtin_i18n.get_message(key).and_then(|msg| msg.value) {
-            Some(v) => v,
-            None => return None,
-        };
-        let mut errs = Vec::new();
-        let result = self.builtin_i18n.format_pattern(value, args, &mut errs);
-        for err in errs {
-            eprintln!("localization error {:?}", err);
-        }
-        Some(result.to_string())
+    /// Returns a reference to the [`L10nManager`], which handles localization
+    /// resources.
+    ///
+    /// [`L10nManager`]: struct.L10nManager.html
+    pub fn localization_manager(&self) -> &L10nManager {
+        &self.l10n
     }
 }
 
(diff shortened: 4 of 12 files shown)


raphlinus:
Before getting to the code, the bloat regression:

|  | Before | After |
| --- | --- | --- |
| Windows debug build time | 25s | 40s |
| Windows release build time | 29s | 85s |
| Windows release exe size | 372k | 640k | 
| macOS debug build time | 16s | 49s |
| macOS release build time | 21s | 126s |
| macOS release exe size | 600k | 912k |

Methodology is building the caclulator example starting with an empty target subdir, but with the cargo caches warmed.

The blog post on Rust bloat is coming. I'm not going to block this PR on that though.

druid-shell/src/windows/util.rs
```diff
@@ -280,3 +280,13 @@ fn attach_console() {
...
     }
 }
+
+/// Returns the current locale string.
+///
+/// This should a [Unicode language identifier].
+///
+/// [Unicode language identifier]: https://unicode.org/reports/tr35/#Unicode_language_identifier
+pub fn get_locale() -> String {
+    //TODO ahem
```
raphlinus: I believe the correct API here is [GetUserPreferredUILanguages function](https://docs.microsoft.com/en-us/windows/win32/api/winnls/nf-winnls-getuserpreferreduilanguages), but am not 100% sure. This function returns a list of languages, but in my testing I was only able to get one. For what we have now, I think just grabbing the first language is good.

druid-shell/src/windows/util.rs
```diff
@@ -280,3 +280,13 @@ fn attach_console() {
...
+    "en_US".into()
```
raphlinus: Modern locale identifier syntax is of the form "en-US", though the underscore is still accepted. I think we should use those.

src/env.rs
```diff
@@ -35,8 +36,11 @@ use crate::Data;
 /// example of the latter is setting a value for enabled/disabled status
 /// so that an entire subtree can be disabled ("grayed out") with one
 /// setting.
-#[derive(Clone, Default)]
-pub struct Env(Arc<EnvImpl>);
+#[derive(Clone)]
+pub struct Env {
+    items: Arc<EnvImpl>,
+    l10n: Arc<L10nManager>,
```
raphlinus: I think I'd rather this goes into EnvImpl rather than Env because it's an additional pointer. But probably not that big a deal in the grand scheme of things.

src/localization.rs
```diff
@@ -0,0 +1,361 @@
...
+/// A closure that generates a localization value.
+type ArgClosure<T> = Arc<dyn Fn(&Env, &T) -> FluentValue<'static> + 'static>;
+
+/// Wraps a closure that generates an argument for localization.
+#[derive(Clone)]
+struct ArgSource<T>(ArgClosure<T>);
+
+/// A string that can be localized based on the current locale.
+///
+/// At it's simplest, a `LocalizedString` is a key that can be resolved
```
raphlinus: "its"

src/localization.rs
```diff
@@ -0,0 +1,361 @@
...
+        self.resolved
+            .as_ref()
+            .map(|s| s.as_str())
+            .or(self.placeholder)
+            .unwrap_or(self.key)
+    }
+}
+
+impl<T: Data> LocalizedString<T> {
+    /// Add a named argument and a cooresponding [`ArgClosure`]. This closure
```
raphlinus: "corresponding"

src/widget/button.rs
```diff
@@ -96,18 +101,20 @@ impl<T: Data> Widget<T> for Label {
         None
     }
 
-    fn update(&mut self, _ctx: &mut UpdateCtx, _old_data: Option<&T>, _data: &T, _env: &Env) {}
+    fn update(&mut self, _ctx: &mut UpdateCtx, _old_data: Option<&T>, data: &T, env: &Env) {
+        self.text.display_text(data, env);
```
raphlinus: What I'd *really* like to see here is that your resolver returns a boolean indicating whether the resolved string has changed, and if so, invalidate. As it is, I think you're underinvalidating here, as the text could change and it won't ask for a repaint.

I think we need to implement at least a basic version of finer-grained invalidation before too long, otherwise there's no real motivation to get invalidation right.

src/widget/button.rs
```diff
@@ -22,17 +22,25 @@ use crate::{
...
 use crate::theme;
 use crate::{Point, RenderContext};
 
+/// The text for the label; either a localized or a specific string.
+pub enum LabelText<T> {
+    Localized(LocalizedString<T>),
+    Literal(String),
+}
+
 /// A label with static text.
```
raphlinus: This description feels a little stale, it's not necessarily static. And I think the distinction between this an `DynLabel` is getting fuzzier, not sure we still need both.

src/widget/button.rs
```diff
@@ -22,17 +22,25 @@ use crate::{
...
-pub struct Label {
-    text: String,
+pub struct Label<T> {
+    text: LabelText<T>,
 }
 
 /// A button with a static label.
```
raphlinus: Same concern about "static".

src/localization.rs
```diff
@@ -0,0 +1,361 @@
...
+            .get_or_insert(HashMap::new())
+            .insert(key, ArgSource(Arc::new(f)));
+        self
+    }
+
+    /// Lazily compute the localized value for this string based on the provided
+    /// environment and data.
+    pub fn resolve<'a>(&'a mut self, env: &Env, data: &T) -> &'a str {
+        //TODO: this recomputes the string if either the language has changed,
+        //or *anytime* we have arguments. Ideally we would be using a lens
```
raphlinus: I don't think we have to do anything with lenses here, this logic can piggy back on just the changes to `T`. The way to do that is to have `paint` call a method that resolves the string iff the resolution hasn't been done yet, otherwise just fetches the resolution, then have `update` recompute the string. As I said below, even better is if the "recompute" method lets us decide whether to propagate an invalidation. Just redoing the string every time `T` changes seems cheap enough to me.

resources/i18n/en-US/builtin.ftl
```diff
@@ -0,0 +1,3 @@
+app-name = Druid
```
zbraniecki: I recommend using Fluent Terms for that - https://projectfluent.org/fluent/guide/terms.html

src/localization.rs
```diff
@@ -0,0 +1,361 @@
...
+/// A string that can be localized based on the current locale.
+///
+/// At it's simplest, a `LocalizedString` is a key that can be resolved
+/// against a map of localized strings for a given locale.
+#[derive(Debug, Clone)]
+pub struct LocalizedString<T> {
+    key: &'static str,
+    placeholder: Option<&'static str>,
+    args: Option<HashMap<&'static str, ArgSource<T>>>,
+    resolved: Option<String>,
```
zbraniecki: Please, note that Fluent supports compound messages which are value + attributes. This is particularly useful for UI toolkits which may use multiple related messages to localize a single UI widget.

See for example main Firefox menubar - https://searchfox.org/mozilla-central/source/browser/locales/en-US/browser/menubar.ftl or Preferences file - https://searchfox.org/mozilla-central/source/browser/locales/en-US/browser/preferences/preferences.ftl

And DOM - https://searchfox.org/mozilla-central/source/browser/base/content/browser-menubar.inc and https://searchfox.org/mozilla-central/source/browser/components/preferences/in-content/preferences.xul

src/localization.rs
```diff
@@ -0,0 +1,361 @@
...
+    /// Loads a new localization resource from disk, as needed.
+    fn get_resource(&mut self, res_id: &str, locale: &str) -> Arc<FluentResource> {
+        let path = self
+            .path_scheme
+            .replace("{locale}", locale)
+            .replace("{res_id}", res_id);
+        if let Some(res) = self.resources.get(&path) {
+            res.clone()
+        } else {
+            let string = fs::read_to_string(&path).unwrap_or_else(|_| FALLBACK_STRINGS.to_string());
```
zbraniecki: The way you wrote the fallback here will only use the fallback in case of a missing file. Much more common scenario where we lazily iterate to fallback is when a string is missing or there are runtime errors during resolution of a string.

That's why we use `Localization` API with cachediterator and return an iterator out of the `L10nManager`.

examples/hello.rs
```diff
@@ -13,16 +13,19 @@
...
     druid::shell::init();
 
     let mut run_loop = runloop::RunLoop::new();
     let mut builder = WindowBuilder::new();
     let mut col = Column::new();
-    let label = DynLabel::new(|data: &u32, _env| format!("value: {}", data));
+
+    let labeltext =
+        LocalizedString::new("hello-counter").with_arg("count", |_env, data: &u32| (*data).into());
+    let label = Label::new(labeltext);
```
zbraniecki: What I would recommend here is to design the fluent-druid binding that allows for L10nId/L10nArgs to be passed to the label, rather than eagerly resolved into a string that is passed to the label.

The way it is done now resembles the pre-fluent paradigm in Firefox and it eventually led to developers resolving strings very deep in their code and then carrying around data with resolved strings up to the UI.
This is counter-productive (if the widget doesn't get displayed, why localize it?), and it locks a lot of language selection logic deep in the imperative code without any means to "retranslate" on fly.

What I'd suggest here is sth like:

```rust
let label = Label::new();
label.set_l10n_attributes("hello-counter", fluent_args!["count" => counter);

...

fn update_counter() {
  counter += 1;
  label.set_l10n_attributes("hello-counter", fluent_args!["count" => counter);
}
```

This will bring this API much closer to how Fluent paradigm operates and in the future will allow you to bring a lot of features like runtime pseudolocales, l10n cycle indepenent of release cycle and runtime locale switching.

It also allows your layout/drawing part handle the actual localization in batches, right before the next paint, rather than synchronously string-by-string.

src/env.rs
cmyr: This is tricky, because `L10nManager` can't impl `Clone`, which isn't impl'd for `FluentBundle`; this is the main reason it's behind the `Arc`.

raphlinus: Understood. To be clear, I'm not talking about getting rid of the `Arc`. Because of the extremely large amount of sharing that's expected, I expect multiple levels of `Arc` and tried to design it so that it's a single pointer. If it's two pointers, it's still ok, but if it becomes a lot of state, it could be a problem.

src/widget/button.rs
cmyr: Okay, i've made this change, seems clean.

src/widget/button.rs
cmyr: Yea, it's interesting. `DynLabel` is definitely more flexible, but how useful that flexibility is in practice is hard to say.

src/localization.rs
cmyr: My thinking here is that with something like lenses, you could indicate per-arg where in the data tree it comes from, and then we could stash that and do better invalidation automatically by comparing whether those items have changed. I'm not sure how much of a saving this this, but would definitely be a handful fewer allocations per localized string, per update.

src/localization.rs
cmyr: Right: attributes were something that I didn't really understand, and so had initially ignored.

Is the relationship between `value` and `attributes` an 'either' relationship? Can a message have no value and no attributes? Can a message have both?

In terms of the design of this API, I could imagine us translating `attributes` into something like a `LocalizationGroup` that contains all the strings for a given resource; I'll look into this when I do the work on window menus, since they want both a title and a tooltip string.

zbraniecki: >  Can a message have no value and no attributes? Can a message have both?

Yes, Yes.

> In terms of the design of this API, I could imagine us translating attributes into something like a LocalizationGroup that contains all the strings for a given resource; 

Why not just use `FluentMessage` returned by `get_message`? https://github.com/projectfluent/fluent-rs/blob/master/fluent-bundle/src/bundle.rs#L28

src/localization.rs
cmyr: Okay, thanks for pointing that out, I think I was confused. My `L10nManager` is intended to be at about the level of `fluent_fallback::Localization`; it keeps a stack of bundles for the current locale, localizes messages with that stack, and rebuilds that stack when the locale changes. I'll flesh this out a bit more and see how well I'm understanding things.

examples/hello.rs
cmyr: I *think* what this is doing is close to what you describe? It probably isn't clear from this example how druid works, but this is just describing how to get the arguments required from the application state as needed.

Basically the label knows the fluent 'message' is it supposed to display, and for each arg required by that message, the label has a closure that can extract the value from the application state.

When the application state changes, we will automatically update the localization with the new data, and then invalidate and paint only if it has changed. Does that sound close to your suggestion?

zbraniecki: > When the application state changes, we will automatically update the localization with the new data, and then invalidate and paint only if it has changed.

Ahhh! That makes sense!

> Does that sound close to your suggestion?

Yes!

So, to stay closer to Fluent terminology, I'd name your `LocalizationString` a `LocalizationMessage` (for it may have more than one string in compound case), and the rest makes sense now. You pass it to `Label` and then before painting the `Label` resolve its localization. That looks good!

The one interesting piece you seem to be aiming for is something we played with as "responsive localization" where a localization is hooked to potential changes in the data state and react with retranslation.

Our main focus at the time was mobile UI l10n and we wanted to react to functions, so we had sth like:

```ftl
preferences = { SCREEN_WIDTH() ->
    [narrow] Security
   *[medium] Privacy & Security
    [wide] Privacy & Security Settings
}
```

here's a video: https://www.youtube.com/watch?v=V5K6YN7MMOA

We ended up finding this model very complicated on the Fluent bindings side and we couldn't justify the complexity at the time.
I still am intrigued by it, so if you think you can get it working, I'd say go for it!

I would still recommend having a "non-automagic" approach available. In other words, it should be possible to do:

```
label.set_l10n(l10n_id, l10n_args);
```

and have the label retranslate.

src/localization.rs
cmyr: Thank you very much for your review and attention here, @zbraniecki, it's very appreciated.

Tried finding this in the docs, without luck: under what circumstance would a message not have a value? What are the implications of this?

And for attributes, after I've called `get_message` I call `format_pattern` for each attribute, passing in the same `FluentArgs` each time?

I don't have a very clear vision of this `LocalizationGroup` type in my head, I'm just imagining ways to let application developers indicate that some widget references known attributes of some message.

zbraniecki: > Tried finding this in the docs, without luck: under what circumstance would a message not have a value? What are the implications of this?

That depends on bindings. For DOM scenario, we use `value` to indicate the content of the element, and `attributes for attributes.

So:

```html
<p data-l10n-id="key1"/>
<menuitem data-l10n-id="key2"/>
```

```ftl
key1 = Hello World
key2 =
    .label = My Menu Item
    .accesskey = M
    .tooltip = Open { key2.label }
```

will result in:

```html
<p data-l10n-id="key1">Hello World</p>
<menuitem data-l10n-id="key2"
                    label="My Menu Item"
                    accesskey="M"
                    tooltip="Open My Menu Item"/>
```

`key1` has only a value, `key2` has only attributes. There are cases where the message will have both, but it has to have at least one or the other for the parser to accept it.

> And for attributes, after I've called get_message I call format_pattern for each attribute, passing in the same FluentArgs each time?

Yep! That's how you'd format a full message :)
In my POC of a patch to swap JS for Rust in Gecko - https://phabricator.services.mozilla.com/D36286

there's (sth similar to, but with ugly C FFI bindings):

```rust
pub fn fluent_bundle_compound(bundle: &FluentBundleRc, id: &nsAString, args: &FluentArgs) -> LocalizationMessage {
    if let Some(message) = bundle.get_message(id.to_string().as_str()) {
        let mut errors = vec![];
        let value = message.value.map(|v| bundle.format_pattern(v, args, &mut errors);
        let mut attributes = vec![];
        for (name, value) in message.attributes {
            let val = bundle.format_pattern(value, args.as_ref(), &mut errors);
            attributes.push((name, val));
        }
        return LocalizationMessage { value, attributes };
    }
}
```

examples/hello.rs
cmyr: > So, to stay closer to Fluent terminology, I'd name your `LocalizationString` a `LocalizationMessage` (for it may have more than one string in compound case), and the rest makes sense now. 

Yea, this is much closer to a fluent 'message' than anything. I'm not sure how much of fluent's naming we want to surface through our API, but at the very least this relationship should be clarified in our docs.

This also relates to the `LocalizationGroup` I was mentioning earlier; my thinking was that it might make sense to separate the compound from the non-compound case, because it could let us provide simpler APIs to our users.

> 
> The one interesting piece you seem to be aiming for is something we played with as "responsive localization" where a localization is hooked to potential changes in the data state and react with retranslation.
> 
> Our main focus at the time was mobile UI l10n and we wanted to react to functions, so we had sth like:
> 
> ...
>
> here's a video: https://www.youtube.com/watch?v=V5K6YN7MMOA

Ah, very cool. We may definitely want to explore something like this, given our cross-platform aspirations. For the time being, the use case is just things like the following, where a button increments a counter and that automatically updates a label:



but there's obviously room for more sophistication here down the road.


> I would still recommend having a "non-automagic" approach available. In other words, it should be possible to do:
> 
> ```
> label.set_l10n(l10n_id, l10n_args);
> ```
> 
> and have the label retranslate.

We do expose an API that let's you pass a msg + args directly to the 'current bundle stack' and just get a string back; this will need some tweaking in light of attributes, but is at least a fallback option, but we're going to strongly encourage using abstractions like this `LocalizedString`; we'll see how far that gets us, I guess, and if it just isn't flexible enough we can revisit.

src/localization.rs
cmyr: okay that's very helpful, thanks!

My take away from this is that we need to also define the expected structure of our user's `ftl` files; we could say for the time being that all messages must have a value, and that attributes are not supported, and then we could revisit this down the road.

examples/hello.rs
zbraniecki: > We do expose an API that let's you pass a msg + args directly to the 'current bundle stack' and just get a string back;

That's not what I suggested. This is an anti-pattern in the Fluent world.

I suggest an ability to update the id/args combo for a given UI widget.

A litmus test for when it works is:

```rust
let msg = LocalizedMessage::new("hello-counter");
let label = Label::new(msg);

fn update_label(label: &Label) {
  let l10n_id = if (User::is_logged_in()) {
     "welcome-back"
  } else {
     "welcome-new-user"
  };
  let msg = LocalizedMessage::new(l10n_id);
  label.set_l10n(msg);
}
```

This makes a clear distinction between what the engineer does (maintain the id/args status of all of the UI) from what the bindings do (they can iterate over the whole UI and retranslate at any point, or localize only right before layout etc.)

src/localization.rs
zbraniecki: > My take away from this is that we need to also define the expected structure of our user's ftl files;

I don't understand why would you do that, but your call :)

examples/hello.rs
cmyr: Our label type is parameterized over its state, so for the time being I'd like to see if we can enforce the idea that you don't set args directly, you just describe how to get them. I think I understand your point, though, which is that depending on state you might want to use completely different messages. One possible solution that fits this model is that you could set the message conditionally based on some attribute in the state, but have this be declarative.

In any case thanks for the clarification, I definitely have some more thinking to do here!

src/localization.rs
cmyr: It's very possible that my thoughts here will change as my understanding improves. :)

druid-shell/src/windows/util.rs
cmyr: This would be annoying for me to do without access to a windows machine, so I'm going to leave this stubbed out for now.

examples/hello.rs
zbraniecki: > One possible solution that fits this model is that you could set the message conditionally based on some attribute in the state, but have this be declarative.

Hmm, that may work, but it feels very complicated. As I said, our experience with trying to do stuff reactively on the state, led us to sticking to simple one-way assignments, and postponing the reactive magic for later.

Based on that experience I think it would be reasonable to provide the same basic functionality for setting/updating the l10n annotation of UI widgets independently of the state-reacting flow.

In `@fluent/dom` we provide `setAttributes` method to do so - https://github.com/projectfluent/fluent.js/blob/master/fluent-dom/src/dom_localization.js#L87 and you can see how it's used all around Firefox UI - https://searchfox.org/mozilla-central/search?q=setAttributes&path=.js

I think you can build the state-reactive model on top of the simple model that we provide, but not the other way around.

cmyr: I think I see what you're saying, and I can totally imagine having an additional API that just takes args directly; this would be helpful in circumstances where for some reason there is some state that isn't in the explicit state object, which is possible.

zbraniecki:
@raphlinus I opened https://github.com/projectfluent/fluent-rs/issues/139 to investigate reducing the size of `fluent-bundle` crate. I completely agree with your blog post and consider `fluent-bundle` to be such `foundational` crate.

sparecycles:
This PR introduces second versions of `proc-macro2`, `quote`, `syn`, and `unicode-xid`.  Relevant to the concerns in the [blog](https://raphlinus.github.io/rust/2019/08/21/rust-bloat.html)?

> I sometimes hear that it’s ok to depend on commonly-used crates, because their cost is amortized among the various users that share them. I’m not convinced, for a variety of reasons. For one, it’s common that you get different versions anyway.

cmyr:
Okay: I think there's more work to do here at some point but I'm happy to get this in as a checkpoint, we can always iterate. @Eijebong thanks for the find, is saving me 10s on a clean debug build. 😁

PR #125, opened 2019-08-23
Fixed Windows keyboard input handling, and some other features to Textbox

8176135:
Fixed Windoes keyboard handing so that
 - Shortcuts (e.g. Ctrl-A) actually work
 - Allows for key combinations involving Alt,
 - Equals/Plus and Backtick/Tilde works
 - Backspace doesn't delete 2 characters per press
 - Keydown for modifier keys are parsed into `KeyCode` correctly

Also on Textboxes:
 - Added the delete key functionality (The `backspace` function probably could be rewritten to be more general)
 - Disabled text input when Ctrl or Alt or Meta is held down

Some of the code is slightly hacky, but I don't know Windows API enough to properly implement them. It seems to still work fine for different keyboard layouts (tested Dvorak), not sure about other languages though. 

Any feedback or changes required is of cause welcome.


src/widget/textbox.rs
```diff
@@ -386,8 +398,15 @@ impl Widget<String> for TextBox {
...
-                    // Actual typing
-                    k_e if k_e.key_code.is_printable() => {
+                    // Delete
+                    k_e if (HotKey::new(None, KeyCode::Delete)).matches(k_e) => {
+                        self.delete(data);
+                        self.reset_cursor_blink(ctx);
+                    }
+                    // Actual typing (when Ctrl or Alt or Meta is not held down)
+                    k_e if k_e.key_code.is_printable()
+                        && !(k_e.mods.alt || k_e.mods.ctrl || k_e.mods.meta) =>
```
cmyr: On some platforms (macOS) alt produces output that we expect to see in the text field; for instance it is how you would type åççéñ†é∂ and alternate characters.

druid-shell/src/keyboard.rs
```diff
@@ -256,6 +256,72 @@ impl KeyCode {
             _ => false,
         }
     }
+
+    pub fn to_char(self) -> Option<char> {
+        use KeyCode::*;
+        match self {
+            KeyA => Some('a'),
```
cmyr: Unfortunately I'm pretty sure this mapping will only be correct on an ANSI keyboard, using american english or compatible; it will not work with alternate keyboard layouts, or alternate languages. We should probably document this better; `KeyCode` here identifies a particular physical key, but that has no guaranteed relationship with the text  produced when that key is pressed.

druid-shell/src/keyboard.rs
```diff
@@ -419,19 +485,21 @@ impl From<i32> for KeyCode {
 
         match src {
             VK_LSHIFT => KeyCode::LeftShift,
+            VK_SHIFT => KeyCode::LeftShift,
```
cmyr: I'd be happy to merge all of these changes immediately; it might be worth opening a new PR for this?

druid-shell/src/windows/mod.rs
```diff
@@ -537,16 +545,23 @@ impl WndProc for MyWndProc {
                     let s = s.as_mut().unwrap();
                     let key_code: KeyCode = (wparam as i32).into();
                     s.stashed_key_code = key_code;
-                    if key_code.is_printable() {
-                        //FIXME: this will fail to propogate key combinations such as alt+s
+
```
cmyr: The logic here is tricky.

The problem with alt+s is that we don't actually know what the character value of 's' is, here. For `WM_KEYDOWN` (no alt key pressed) the system will call [`TranslateMessage`](https://docs.microsoft.com/en-ca/windows/win32/api/winuser/nf-winuser-translatemessage) and then we'll get a `WM_CHAR` message with the actual character value; but there's no equivalent for `WM_SYSKEYDOWN`.

How do we get the actual character value? I'm not sure, but I can think of a couple options:

- see if we can call `TranslateMessage` ourselves in this case, and then just handle `WM_CHAR` as normal;
- see if there's another API where we can look up the current char value of this key in the current keyboard layout, and then use that value to construct the appropriate `KeyEvent`.
- or something else entirely?

8176135:
Alright, I'll open a new PR for the VK_SHIFT mod keys, and the backspace fix.

>  Backspace: this sounds like a bug. On windows, in the general case, when the user presses a key we get the `WM_KEYDOWN` event; and then afterwards, if the key has a text value, we also get called with `WM_CHAR`. We want to send one event per key, so what we try to do is to send non-printable events when we handle `WM_KEYDOWN`, (because we won't get called with `WM_CHAR` for these keys) but then stash the `KeyCode` and send the event in `WM_CHAR` if possible. It is guaranteed that our implementation of this is buggy, but this is what we're _trying_ to do. 😁 Anyway: the bug here is that `BackSpace` is returning `true` from `is_printable`. A PR for that would be welcome!

Well backspace is not returning true from `is_printable` at the moment, but Windows still thinks it is a character for some reason. I think just having in the following in the WM_CHAR branch will make sure all cases of this are prevented, and not introduce any side effects.
```
if !key_code.is_printable() {
   // Avoid double calls on some non-printable keys (e.g. Backspace)
   return None;
}
```

8176135:
> when you tested dvorak was that a hardware dvorak keyboard or just setting your keyboard layout? 

It was just the windows keyboard layout options.

druid-shell/src/keyboard.rs
8176135: Well. on windows it seems you are converting virtual key codes (i.e. wparam) into `KeyCode` which is hardware independent (as opposed to scancodes in lparam which are the raw inputs sent from the keyboard). 

So it should be fine? I can't really test this as I only have American English keyboards.

8176135: Never mind, going to research this some more.

cmyr: yea our idea is that `KeyCode` is a platform-independent equivalent to virtual key codes. It is hardware independent, but it is not language independent; for instance you can in settings change your keyboard to greek, without the `KeyCode`s changing.

Good reading for the windows side: https://docs.microsoft.com/en-us/windows/win32/inputdev/about-keyboard-input

cmyr:
So basically (and this should be documented) the intended meaning of `is_printable` is basically, "does this key have a text/character value that will change based on the user's selected language / input method"? Going by this, backspace and delete should both be non-printable.

If windows is calling `WM_CHAR` with backspace, I would be tempted to handle that explicitly, e.g. checking if the stashed keycode is backspace, and return.

My thinking is that by just checking `is_printable` we will mask other special cases that we might want to know about.

PR #126, opened 2019-08-23
Dispatch Cocoa menu events via the responder chain

cmyr:
This changes how we route menu events on mac. Instead of having
a custom proxy class that we instantiate for each menu, we set
the menu's 'tag' property to the command number, and set the
'action' to be an action we define on our custom view class.

This also stubs out the `command` fn in druid to receive these events.

### DIFF
--- a/druid-shell/src/mac/menu.rs
+++ b/druid-shell/src/mac/menu.rs
@@ -13,36 +13,11 @@
 // limitations under the License.
 
 //! macOS implementation of menus.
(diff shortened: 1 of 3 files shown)


cmyr:
If a menu item does not have an explicit `target`, its `action` is sent to the first responder, and then down the responder chain, from there. By setting the `action` to a method defined on our NSView subclass, which is in the responder chain, that method will be called (on the view that is owned by the active window, which is always part of the responder chain) when the menu is selected.

PR #127, opened 2019-08-23
First draft of multiple window support

raphlinus:
This isn't quite ready to merge, but I think is far enough along that it's worth reviewing. Open issues:

* There's no event flow for closing windows, so we're not cleaning up that state. The example still works though.

* The existing `UiMain` and `UiState` should be deleted, and examples moved to the new framework.

* It's *very* likely that the API can be cleaned up, especially exposing less of the internals of WindowId / WindowBuilder on first-window creation.

* The example should have both globally shared and window-local state; the latter is implemented basically as a map keyed off WindowId. This means making sure the window_id is exposed in all relevant contexts. We could also consider some kind of lens to make this more natural, so (a) you don't always have to access the map explicitly and (b) invalidations to window-local state don't propagate to other windows.

* Apparently it doesn't merge cleanly against master. I'll merge or rebase :/

### DIFF
--- a/examples/multi_win.rs
+++ b/examples/multi_win.rs
@@ -1,0 +1,94 @@
+// Copyright 2019 The xi-editor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use druid::shell::{runloop, WindowBuilder};
+use druid::widget::{ActionWrapper, Button, Column, DynLabel, Padding};
+
+use druid::{
+    Data, DruidHandler, Env, Event, EventCtxRoot, KeyCode, LayoutCtxRoot, PaintCtxRoot, RootWidget,
+    UpdateCtxRoot, Widget, WindowId, WindowSet,
+};
+
+pub struct SharedWindow<T: Data> {
+    windows: WindowSet<T>,
+}
+
+impl<T: Data> SharedWindow<T> {
+    pub fn new(root: impl Widget<T> + 'static, id: WindowId) -> SharedWindow<T> {
+        SharedWindow {
+            windows: WindowSet::new(root, id),
+        }
+    }
+}
+
+impl RootWidget<u32> for SharedWindow<u32> {
+    fn event(&mut self, event: &Event, ctx: &mut EventCtxRoot, data: &mut u32, env: &Env) {
+        match event {
+            Event::KeyDown(e) => {
+                // TODO: I know this is not the best practice to match
+                if e.mods.ctrl && e.key_code == KeyCode::KeyN {
+                    let mut builder = WindowBuilder::new();
+                    builder.set_title("Additional window");
+                    let (id, window_pod) = ctx.new_win(builder, make_ui());
+                    self.windows.add_window(id, window_pod);
+                    println!("new window {:?}", id);
+                }
+            }
+            _ => (),
+        }
+        self.windows.event(event, ctx, data, env);
+    }
+
+    /// Propagate a data update to all windows.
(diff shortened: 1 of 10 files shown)


src/window.rs
```diff
@@ -0,0 +1,268 @@
...
+        self.windows.paint(paint_ctx, data, env);
+    }
+}
+
+impl WindowId {
+    /// Allocate a new, unique window id.
+    ///
+    /// Do note that if we create 4 billion windows there may be a collision.
+    pub fn new() -> WindowId {
+        let id = WINDOW_ID_COUNTER.fetch_add(1, Ordering::Relaxed);
```
cmyr: I would just declare the `static` in this fn, feels a bit tidier?

examples/multi_win.rs
```diff
@@ -0,0 +1,94 @@
...
+
+impl<T: Data> SharedWindow<T> {
+    pub fn new(root: impl Widget<T> + 'static, id: WindowId) -> SharedWindow<T> {
+        SharedWindow {
+            windows: WindowSet::new(root, id),
+        }
+    }
+}
+
+impl RootWidget<u32> for SharedWindow<u32> {
```
cmyr: All of the code in this example feels like code that should be provided by druid.

What if `RootWidget` wasn't a trait, but was a struct? It could, by default, implement a bunch of general behaviour, keep a stack of windows, handle system commands, etc; and then if the user wanted custom event handling, for instance, we could use a delegate pattern, like:

```rust
```rust
struct RootWidget<T> {
    ...
    event_delegate: Option<Box<dyn EventDelegate<T>>>,
}

trait EventDelegate<T> {
    /// Recieves an event before it is sent up the widget tree.
    /// Can return either the original event or a new event to be passed
    /// up the tree (for instance, replacing a key press with a command),
    /// or can return `None` to swallow this event.
    fn event(&mut self, event: Event, ctx: &mut dyn WinCtx) -> Option<Event>;
}
```

I guess my question is: what logic, exactly, do we expect to want at the application level? Do we ever imagine the user wanting to override `paint`? `layout`? What would that mean?

I'm also thinking about this in terms of menus. Currently, if you ever want to modify a menu you will need to do it manually in a widget. It might be nice if, instead, there was a `MenuDelegate` trait, where we would call you on `update` and you could request to change a menu.

src/event.rs
```diff
@@ -44,6 +44,17 @@ use druid_shell::window::{MouseEvent, TimerToken};
 /// [`WidgetPod`]: struct.WidgetPod.html
 #[derive(Debug, Clone)]
 pub enum Event {
+    /// Called on the root widget when the window size changes.
```
cmyr: I could imagine wanting a bunch of events that only go to the root widget; things like application life-cycle, opening files, or doing state restoration. I could imagine something like,

```rust
pub enum RootEvent {
    LocaleChanged(Locale),
    Start,
    WillQuit,
    // ...
    WindowEvent(Event),
}
```

cmyr:
I remembered something I wanted to say:


I've been playing around with menus, and I wanted to make an example widget that did some fun custom-menu stuff, but then just had a nice declarative widget tree without a lot of logic.

This is hard! to do custom menu stuff right now I need to have a custom widget, and once I have a custom root widget it isn't easy to just use a normal widget graph for your layout; you need to manually have a widget pod or a child widget that you forward everything to, explicitly.

basically: for a good ergonomics story  I think we'll want to think a lot more about composability, and how we can make these sorts of things easier.

raphlinus:
I think I'm going to ask if I can hand this off to you, I'm not sure I have the motivation right now, and won't have time next week in any case.

Some general thoughts:

I'm fairly happy with the lower levels of this patch, the way it creates a state object which is shared by the windows. My original plan was to follow the basic widget structure, so the app-delegate object owns the windows, but I now see that this is kinda unpleasant from an API point of view and there might not be that much benefit. To my mind, the main thing is providing plumbing for the event flow, so the app delegate can handle the event itself, and possibly do some other wrapping of its children (this might be a place for undo logic to live). I also think we need to wire up a command into this event handler, but I think that's orthogonal to the window plumbing.

The macOS crashes are probably just not cleaning up state for deleted windows. We can do the cleanup, but it also might make sense to soft-unwrap weak references in case.

cmyr:
I'm basically also not around next week; with that in mind I'm happy to take this ticket, but if you have a burst of motivation/inspiration next week you should feel free to take it back. 😉

In any case I think we're on the same page, it's really a question of what the customization points are.

cmyr:
superseded by #149

Issue #119, opened 2019-08-13
TextBox uses is_printable to filter text input which is unrelieable.


cmyr:
The `is_printable` fn on `KeyCode` is not very useful, because it alone does not tell you whether a key event will have printable text.

As an example: `ctrl + e` on macOS uses the 'printable' `KeyE` keycode, but sends the unprintable ascii value 0x5. similarly with `ctrl + u`, `ctrl + k`, and others.

cbondurant:
I think this is now a non-issue, since I cannot find reference to is_printable in the codebase, and it is most likely a function that was removed in some part of the event rework.

xStrom:
Alright, let's close this for now.

PR #129, opened 2019-08-24
Fixes backspace, and VK mapping

8176135:
Pulled out just the backspace and VK mapping fixes from #125


druid-shell/src/windows/mod.rs
```diff
@@ -508,6 +508,12 @@ impl WndProc for MyWndProc {
                     let s = s.as_mut().unwrap();
                     //FIXME: this can receive lone surrogate pairs?
                     let key_code = s.stashed_key_code;
+
+                    if !key_code.is_printable() {
```
cmyr: I'm now confused about this. If `Backspace` is returning `false` for `is_printable()`, how are we receiving two events? I would assume the first event comes from `WM_KEYDOWN`, below, but we should be returning on L548, and not sending that event? Do you know what's happening here?

8176135: From what I can tell Windows just sends 2 messages (a WM_KEYDOWN then a WM_CHAR) like any other printable key for backspace (the handle messages loop goes around twice)

cmyr: right, but _we_ should only send a message during `WM_KEYDOWN` if `is_printable() == false`?

8176135: Hmm... I don't think I understand what you are asking?

`backspace` is returning false for `is_printable`, which means we are letting it through in `WM_KEYDOWN`, but windows considers `backspace` a character, so is also sending another event with `msg == WM_CHAR`. Since currently there are no checks in `WM_CHAR` of whether we consider what windows is giving us is or is not printable, we are also letting that second event through, resulting in two backspace events per press of the key. 

This will also be an issue for any other keys that windows considers characters, but we don't. 
So the `if !key_code.is_printable() {` I added just makes sure that if we let the key through in `WM_KEYDOWN` we don't also let a duplicate through in `WM_CHAR` 

> My thinking is that by just checking `is_printable` we will mask other special cases that we might want to know about.

Can you give me an example of a case we want both events to propagate?

cmyr: Okay, I think I was confused. If `Backspace` produces `WM_CHAR`, I think the correct behaviour is to not emit a `KeyEvent` while handling `WM_KEYDOWN` for that key, and emit in `WM_CHAR`.

My thinking is that in `WM_CHAR` we have strictly more information about the key than we do in `WM_KEYDOWN`, so we should use that if we can.

`is_printable` is a mess (see for instance #119) and what we really want, here, is some way of checking if a key is a 'character key'. I'm not sure if there's API for this. For now, I think the best solution would be to just special-case `Backspace` in `WM_KEYDOWN`, maybe with a comment linking to this discussion.

PR #130, opened 2019-08-25
Allow limiting scroll to horizontal/vertical direction

zaynetro:
This will guide scroll to limit itself in one direction only. Child will then be
constrained to `Scroll`'s parent max bounds in width/height.

As discussed: [in zulip](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/druid.20design.20language/near/173964242)

### DIFF
--- a/src/widget/scroll.rs
+++ b/src/widget/scroll.rs
@@ -33,6 +33,10 @@ enum ScrollDirection {
 }
 
 impl ScrollDirection {
+    /// Return the maximum size the container can be given
+    /// its scroll direction and box constraints.
+    /// In practice vertical scrolling will be width limited to
+    /// box constraints and horizontal will be height limited.
     pub fn max_size(&self, bc: &BoxConstraints) -> Size {
         match self {
             ScrollDirection::Horizontal => Size::new(INFINITY, bc.max().height),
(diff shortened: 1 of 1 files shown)


src/widget/scroll.rs
```diff
@@ -25,6 +25,23 @@ use crate::piet::RenderContext;
...
 
+#[derive(Debug, Clone)]
+enum ScrollDirection {
+    Horizontal,
+    Vertical,
+    All,
+}
+
+impl ScrollDirection {
+    pub fn max_size(&self, bc: &BoxConstraints) -> Size {
```
cmyr: - does this need to be `pub`?
- A short doc comment would be good here.

src/widget/scroll.rs
```diff
@@ -61,6 +84,20 @@ impl<T: Data> Scroll<T> {
             false
         }
     }
+
+    /// Limit scroll behavior to allow only vertical scrolling (Y-axis).
+    /// The child is laid out with width constrained layout bounds.
```
cmyr: maybe, "The child is laid out with constrained width and infinite height."?

if you prefer this you can make the same change below.

src/widget/scroll.rs
zaynetro: I don't really know what are the best practices for methods on a private struct. My intent was to emphasize that this method will not be used within the struct but from the outside.

I will add docs.

src/widget/scroll.rs
zaynetro: changed

src/widget/scroll.rs
zaynetro: Added docs

cmyr:
Thanks!

PR #132, opened 2019-08-25
Fix TextBox panic due to selection going off-bounds on data string

zaynetro:
@futurepaul let me know if this is the right fix and in the right place.

* [x] <s>Should we replace all usages of `s.len()` with `s.chars().count()`?</s> No!

### DIFF
--- a/src/widget/textbox.rs
+++ b/src/widget/textbox.rs
@@ -446,3 +446,32 @@ fn prev_grapheme(src: &str, from: usize) -> usize {
         0
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// Test that when data is mutated externally widget
+    /// can still be used to insert characters.
+    #[test]
+    fn data_can_be_changed_externally() {
+        let mut widget = TextBox::new(100.);
+        let mut data = "".to_string();
+
+        // First insert some chars
+        widget.insert(&mut data, "o");
+        widget.insert(&mut data, "n");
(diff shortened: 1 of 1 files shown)


cmyr:
What's the bug here? It might be worth including a test case.

In any case, we don't want to use `chars().count()`, which will return the number of unicode scalars in the string; we should be representing selections as utf-8 byte indexes.

It should be an invariant that the selection always describes a valid range in the string, and we should be validating that when the selection is changed.

zaynetro:
Added a test that panics without the fix.

I am not sure if I am fixing it in the right place. Alternatively, we can check if selection is out of bounds in the `update` method.

It would be nice to have some kind of integration test setup. I would like to simulate sending events to the widget. Currently, I couldn't find a way.

cmyr:
Ah cool I understand what was happening now. My personal feeling is that `update` is the right place to handle this, but i'd be curious for @raphlinus's opinion. I also think the logic should be pretty simple; if the text changes externally we should reset the selection state, moving the cursor to the end of the text. We could maybe special-case if the cursor is currently at 0, and keep it there; but if the selection state is any more complicated then we can end up with a weird selection that we probably don't want.

I agree that we will want some better ways to test at some point.

zaynetro:
I don't think we currently have a way of knowing if data was mutated externally or internally. So in `update` we won't know if data has changed outside the widget.

cmyr:
Right, that makes sense, and this seems like a good solution for now. Thanks!

PR #133, opened 2019-08-25
Rename examples/widget_demo -> slider, textbox_demo -> textbox

cmyr:
'demo' feels redundant?


zaynetro:
I tend to agree with you.

PR #134, opened 2019-08-27
Add styled widgets (and lay them out correctly)

futurepaul:
There's a lot here, but I think this is a good spot to make it official.

Things to discuss:
1. Opinions on colors and aesthetic! I've mostly based my design on [this Figma doc](https://www.figma.com/file/YudE0StDEPdemyIIVdUMzr/Druid-GUI-concept?node-id=0%3A1), so a fork of the Figma doc might be a good way to propose alternatives.
2. Better names for "how tall things are" and "taller things."
3. Thoughts on whether these are easy to theme (for instance they mostly have hard-coded gradients based on Env colors, maybe those gradients should be moved to Env?) for people who don't like the colors and aesthetic.
4. Thoughts on making light mode default instead of dark mode (or just support for OS-defined dark mode so we support both out of the box).
5. I'm using this `DynWidget` component I made up in order to make the CheckBox work in widget_demo. This doesn't seem optimal (right now hovering over the CheckBox widget sets Data to 0.0), but I wasn't sure what the correct answer is.
6. I used wrapper widgets for most of these to make it easy to be left-aligned, vertically centered without adding a lot of complication to the widget (custom hit testing was especially annoying). We talked about this in Zulip so I think we're okay with this, just making sure we're sure.
7. I added an optional axis constraint to Align so that it doesn't always take up the max constraint in both y and x.
8. I added some sanity checks to `Flex` so it works inside scroll. Hopefully these are non-controversial.
9. I added a .check("debug name") method to BoxConstraints (which helped me figure out the right thing to do with `Flex`)


cmyr:
Sorry to let this linger. It's great to see the polish. That said, it feels like there's a bunch of stuff going on here, and I wonder if we couldn't break it up a little bit. In particular, maybe the checkbox widget can get its own PR? I'm not sure if it also makes sense to separate the layout from the theming stuff; I defer to your judgement there.

Another thought I have, here, is that we may want to have different button 'styles'. I imagine this would be an enum, and then when we actually want to style the button we take the theme and the style and use that to resolve the various properties.

This can definitely be follow-up work, just brainstorming.

futurepaul:
I was worried this was doing too many different things, but I didn't quite know where to slice it. I can drop the checkbox from the PR, and that also means we don't need DynWidget. (I included the checkbox because it was an example of a widget that has a fixed height and width).

I could separate out the layout stuff, but there's a sense in which these styled widgets aren't correct without proper layout. So maybe I could do:
1. Layout improvements (adding axis to Align is especially important)
2. Styled Widgets
3. Checkbox

If it would help, layout could be further broken into:
1. Align axis
2. BoxConstraints `check` method (this isn't needed in production but it's really helpful for development)
3. Flex sanity checks

src/theme.rs
```diff
@@ -18,21 +18,49 @@ use crate::piet::Color;
...
+pub const BUTTON_DARK: Key<Color> = Key::new("button_dark");
+pub const BUTTON_LIGHT: Key<Color> = Key::new("button_light");
+pub const BORDER: Key<Color> = Key::new("border");
+pub const BORDER_LIGHT: Key<Color> = Key::new("border_light");
+pub const SELECTION_COLOR: Key<Color> = Key::new("selection_color");
+pub const CURSOR_COLOR: Key<Color> = Key::new("cursor_color");
+
 pub const FONT_NAME: Key<&str> = Key::new("font_name");
 pub const TEXT_SIZE_NORMAL: Key<f64> = Key::new("text_size_normal");
+//TODO: what's a better name for these?
```
raphlinus: Yes, we need better names. I propose `BASIC_WIDGET_HEIGHT` and `BORDERED_WIDGET_HEIGHT`. Definitely open to bikeshedding on these, but want to get a "probably good enough" proposal out there.

src/widget/flex.rs
```diff
@@ -166,20 +166,25 @@ impl<T: Data> Widget<T> for Flex<T> {
         }
```
raphlinus: Where it has `minor = 0.0f64` above, I think that should be `minor = self.direction.minor(bc.min())`. (Another nit, we should probably be calling the accessors rather than getting the fields directly, in case people want to cut'n'paste this code). I don't think this is a very serious issue, but I'd like to move toward making sure sizes actually meet the constraints by construction.

src/widget/flex.rs
```diff
@@ -200,7 +205,7 @@ impl<T: Data> Widget<T> for Flex<T> {
             child.widget.set_layout_rect(rect.with_origin(pos));
             major += self.direction.major(rect.size());
         }
-        if flex_sum > 0.0 {
+        if flex_sum > 0.0 && total_major != std::f64::INFINITY {
```
raphlinus: Hmm, this seems to be dealing with the case where at least one child is flex but the major dimension is unbounded. In Flutter, that [seems to be an error](https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/rendering/flex.dart#L661). Are you sure you want to add special-case logic to deal with it? It might be useful to, at the very least, log a warning when that case is detected.

src/widget/align.rs
```diff
@@ -21,10 +21,16 @@ use crate::{
 
 use crate::piet::UnitPoint;
 
+enum Axis {
```
raphlinus: I have a question about the axis option for Align, as Flutter doesn't have precisely that. Is Axis::Horizontal roughly equivalent to setting `heightFactor = 1.0` in Flutter?

(I asked this in Zulip and the answer seems to be yes, but wanted to leave a record in the PR)

src/theme.rs
futurepaul: I went with those names for now. Should I leave in this todo?

src/widget/textbox.rs
```diff
@@ -312,17 +325,26 @@ impl Widget<String> for TextBox {
         _layout_ctx: &mut LayoutCtx,
         bc: &BoxConstraints,
         _data: &String,
-        _env: &Env,
+        env: &Env,
     ) -> Size {
-        bc.constrain((self.width, BOX_HEIGHT))
+        // QUESTION: is this the right place to put this default width?
+        let default_width = 100.0;
```
futurepaul: The slider and the progress bar each have a default width for when they're given an infinite width to fill. Is it okay in here, or should it be part of the theme?

src/theme.rs
```diff
@@ -42,7 +70,7 @@ pub fn init() -> Env {
     {
         // Ideally this would be a reference to San Francisco, but Cairo's
         // "toy text" API doesn't seem to be able to access it easily.
-        env = env.adding(FONT_NAME, "Arial");
+        env = env.adding(FONT_NAME, "SF Pro Display Regular");
```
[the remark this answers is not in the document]
futurepaul: Got it. I forgot that I manually installed SF on my system which is why this worked for me.

src/widget/textbox.rs
raphlinus: I'm personally ok leaving this; it seems useful for prototyping but probably in a "real" layout it'll be constrained by parent. I think we can revisit though.

src/theme.rs
futurepaul: (I removed it)

futurepaul:
> Another thought I have, here, is that we may want to have different button 'styles'. I imagine this would be an enum, and then when we actually want to style the button we take the theme and the style and use that to resolve the various properties.

I'd love to talk this through in Zulip or in an issue. I don't know exactly the scope for styling beyond what has been implemented so far. I do know a good feature for a button would be to have a disabled state, along with "primary" and "warning" styles.

raphlinus:
> I'd love to talk this through in Zulip or in an issue. I don't know exactly the scope for styling beyond what has been implemented so far. I do know a good feature for a button would be to have a disabled state, along with "primary" and "warning" styles.

I'm also going to be in the market for a disabled style very soon (specifically for the "delete master" button, which should only be enabled when a master is selected). I think that be orthogonal to the other styles; what I was thinking (and this is borrowed to some extent from SwiftUI) is that the enabled state becomes part of env (probably with specific API surface for it; doesn't have to go through the typed key/value mechanism) and widgets respond to it. There are definitely questions though. I think maybe an issue is best as we're somewhat async right now with timezones.

src/lib.rs
```diff
@@ -929,6 +927,20 @@ impl BoxConstraints {
...
+    /// Check to see if these constraints are legit.
+    pub fn check(&self, name: &str) {
+        if !(0.0 <= self.min.width
+            && self.min.width <= self.max.width
+            && self.max.width <= std::f64::INFINITY)
+            || !(0.0 <= self.min.height
+                && self.min.height <= self.max.height
+                && self.max.height <= std::f64::INFINITY)
+        {
+            eprintln!("Bad BoxConstraints passed to {}:", name);
```
raphlinus: Should this be using the (now merged) logging functionality?

src/lib.rs
```diff
@@ -929,6 +927,20 @@ impl BoxConstraints {
     pub fn is_height_bounded(&self) -> bool {
         self.max.height.is_finite()
     }
+
+    /// Check to see if these constraints are legit.
+    pub fn check(&self, name: &str) {
+        if !(0.0 <= self.min.width
+            && self.min.width <= self.max.width
+            && self.max.width <= std::f64::INFINITY)
```
raphlinus: Checking less than or equal to infinity seems strange to me; the only thing that wouldn't is a NaN, I believe.

src/lib.rs
futurepaul: Ah yes! Would this be:
```
warn!("Bad BoxConstraints passed to {}:", name);
warn!("{:?}", self);
```
?

raphlinus: That looks right to me, yes.

src/lib.rs
futurepaul: This is pretty funny now that I think about it.

futurepaul:
Okay, I feel good about this, and look forward to iterating from here!

PR #135, opened 2019-08-31
Use logging crate

### LINKED ISSUE #99: Add logging
we're currently dumping a bunch of stuff to stderr, but to properly support application use we should be using (I suspect) the `log` crate.

LiHRaM:
fixes #99

Looking to start contributing again, thought this might be a nice place to start.

I've added logging using the `log` crate.

Todo:
- [x] Replace calls to println and eprintln with appropriate logging macros
- [x] Add example with logging impl


druid-shell/src/mac/mod.rs
```diff
@@ -204,7 +206,7 @@ lazy_static! {
         }
         decl.add_method(sel!(dealloc), dealloc as extern "C" fn(&Object, Sel));
         extern "C" fn dealloc(this: &Object, _sel: Sel) {
-            eprintln!("view is dealloc'ed");
+            error!("view is dealloc'ed");
```
cmyr: I think this is just `info`, deallocation is part of normal program execution.

src/localization.rs
```diff
@@ -149,7 +151,7 @@ impl ResourceManager {
     /// Return the best localization bundle for the provided `LanguageIdentifier`.
     fn get_bundle(&mut self, locale: &LanguageIdentifier, resource_ids: &[String]) -> BundleStack {
         let resolved_locales = self.resolve_locales(locale.clone());
-        eprintln!("resolved: {}", PrintLocales(resolved_locales.as_slice()));
+        info!("resolved: {}", PrintLocales(resolved_locales.as_slice()));
```
cmyr: I vote `debug!`.

src/localization.rs
```diff
@@ -213,7 +215,7 @@ impl L10nManager {
             .parse()
             .unwrap_or_else(|_| default_locale.clone());
         let locales = get_available_locales(base_dir).unwrap_or_default();
-        eprintln!(
+        info!(
```
cmyr: ditto

src/localization.rs
```diff
@@ -265,7 +267,7 @@ impl L10nManager {
             .current_bundle
             .format_pattern(key, value, args, &mut errs);
         for err in errs {
-            eprintln!("localization error {:?}", err);
+            error!("localization error {:?}", err);
```
cmyr: I vote `warn!`. This is tricky; but basically `error!` is the highest log level we have, and should probably be reserved for things that are fatal or otherwise compromise the execution of the application. This is maybe an error though, it's hard to say. I think that [this summary](https://stackoverflow.com/a/2031209/1529922) is a reasonable overview of how to think about the different log levels, this is definitely a 'feel' thing, though!

src/widget/textbox.rs
```diff
@@ -193,11 +195,11 @@ impl TextBox {
 
     // TODO: waiting on druid clipboard support for copy / paste.
     fn copy_text(&self, input: String) {
-        eprintln!("COPY: {}", input);
+        error!("COPY: {}", input);
```
cmyr: I'd go with `warn!`, and change the message to indicate this is not implemented. 👍

src/lib.rs
```diff
@@ -764,7 +766,7 @@ impl<T: Data + 'static> WinHandler for UiMain<T> {
     }
 
     fn command(&mut self, id: u32, ctx: &mut dyn WinCtx) {
-        eprintln!("got command {}", id);
+        error!("got command {}", id);
```
cmyr: maybe `debug!`?

examples/logging.rs
```diff
@@ -0,0 +1,61 @@
...
+    builder.set_handler(Box::new(UiMain::new(state)));
+    let window = builder.build().unwrap();
+    window.show();
+    run_loop.run();
+}
+
+// Typically the user would use a crate for logging
+// Here we implement the Log trait for a struct and print everything to STDOUT
+// See https://docs.rs/log/ for more info
+use log::{Level, Metadata, Record};
```
cmyr: I appreciate the thoroughness, but I wonder if this isn't out of scope for this PR?

What I would personally be inclined to do is to find the simplest, most minimal logging crate ([simple_logger]](https://github.com/borntyping/rust-simple_logger) was the best i could find in a quick search) and just use that? It's closer to what the user will ever actually do.

Something we might consider for follow-up work is to add something like this to druid itself, and then we could let the user ask for stderr logging when they start the runloop.

druid-shell/src/windows/mod.rs
```diff
@@ -513,7 +515,7 @@ impl WndProc for MyWndProc {
                     let text = match s.stashed_char {
                         Some(c) => c,
                         None => {
-                            eprintln!("failed to convert WM_CHAR to char: {:#X}", wparam);
+                            error!("failed to convert WM_CHAR to char: {:#X}", wparam);
```
cmyr: `warn!` I think.

druid-shell/src/windows/mod.rs
```diff
@@ -865,7 +867,7 @@ impl WindowBuilder {
             }
 
             let dcomp_state = create_dcomp_state(self.present_strategy, hwnd).unwrap_or_else(|e| {
-                println!("Error creating swapchain, falling back to hwnd: {:?}", e);
+                error!("Error creating swapchain, falling back to hwnd: {:?}", e);
```
cmyr: `warn!` I think?

druid-shell/src/windows/mod.rs
```diff
@@ -930,9 +932,9 @@ unsafe fn create_dcomp_state(
             &IID_IDXGIFactory2,
             &mut factory as *mut *mut IDXGIFactory2 as *mut *mut c_void,
         ))?;
-        println!("dxgi factory pointer = {:?}", factory);
+        info!("dxgi factory pointer = {:?}", factory);
```
cmyr: these can all be `debug!`.

examples/logging.rs
LiHRaM: I wrote it like that to avoid any extra dependencies. Little did I know about [this](https://doc.rust-lang.org/rust-by-example/testing/dev_dependencies.html)...

So I added it as a dev-dependency.
Anyway, sounds good. I'll leave more concrete logging implications on the druid library up to another PR.

src/localization.rs
LiHRaM: Yeah. I just ventured a guess with most of them, but didn't feel familiar enough with the internals of the library in a lot of cases. Thanks for taking the time. 😄

examples/logging.rs
cmyr: yea, dev dependency is perfect. 👍

Cargo.toml
```diff
@@ -1,27 +1,31 @@
-[package]
```
cmyr: do you have any idea why this is showing all these changes? Did the encoding change or something??

edit: looked into this, and it's a different line ending?

src/widget/null.rs
```diff
@@ -1,28 +1,30 @@
...
-
-pub struct NullWidget;
-
-impl Widget for NullWidget {
-    fn poke(&mut self, _payload: &mut dyn Any, _ctx: &mut HandlerCtx) -> bool {
-        eprintln!("Poke to null widget: probable use-after-free");
-        true
-    }
-}
+// Copyright 2018 The xi-editor Authors.
```
cmyr: ditto here, something confusing is going on. I wonder if it's just the diff is screwing up..

examples/logging.rs
```diff
@@ -0,0 +1,33 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use druid::shell::{runloop, WindowBuilder};
+use druid::widget::Column;
+use druid::{UiMain, UiState};
+use simple_logger;
+
+fn main() {
+    simple_logger::init().unwrap();
```
cmyr: I actually think we can get rid of this file, but I would like to add logging to a few examples; in particular `hello`, `calc`, and `custom_widget`. (These are the examples I mostly use when experimenting & debugging)

Cargo.toml
LiHRaM: Weird. My git settings should be enforcing LF. Sorry about that

examples/logging.rs
LiHRaM: Fair enough. ;) I've added logging to those examples.

PR #141, opened 2019-09-06
Fail build on warnings

### LINKED ISSUE #139: Make warnings fail CI
We have an informal policy not to commit code that prints warnings; it would be nice to make this an official policy, by baking it into CI.

If anyone wants to dig into this, that would be very welcome!

LiHRaM:
Run the build command for `druid` and `druid_shell`, failing on warnings.

fixes #139

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -25,6 +25,14 @@ jobs:
         parameters:
           crate_path: '.'
 
+      - template: ci/build.yml
+        parameters:
+          crate_path: 'druid-shell'
+      
+      - template: ci/build.yml
+        parameters:
+          crate_path: '.'
+
       - template: ci/test.yml
         parameters:
           toolchain: stable
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -1,0 +1,5 @@
+steps:
+  - script: |
+        cargo rustc -- -D warnings
+    workingDirectory: '$(Build.SourcesDirectory)/${{parameters.crate_path}}'
+    displayName: Run cargo rustc, deny warnings
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -2,4 +2,4 @@ steps:
   - script: |
         cargo rustc -- -D warnings
     workingDirectory: '$(Build.SourcesDirectory)/${{parameters.crate_path}}'
-    displayName: Run cargo rustc, deny warnings
+    displayName: Run cargo rustc -- -D warnings on ${{parameters.crate_path}}


.azure-pipelines.yml
```diff
@@ -25,6 +25,14 @@ jobs:
         parameters:
           crate_path: '.'
 
+      - template: ci/build.yml
+        parameters:
+          crate_path: 'druid-shell'
```
cmyr: I'm not sue this is necessary? I think that because druid builds druid-shell, the warnings will be propagated.

ci/build.yml
```diff
@@ -0,0 +1,5 @@
+steps:
+  - script: |
+        cargo rustc -- -D warnings
+    workingDirectory: '$(Build.SourcesDirectory)/${{parameters.crate_path}}'
+    displayName: Run cargo rustc -- -D warnings on ${{parameters.crate_path}}
```
cmyr: I would just call this "Check for build warnings" (this is the name that shows up in the report on the web dashboard).

.azure-pipelines.yml
LiHRaM: I suspected that as well, but it didn't seem to be the case when I was testing it locally.

```
C:\Users\hilma\git\druid
λ cargo rustc -- -D warnings
   Compiling druid-shell v0.3.0 (C:\Users\hilma\git\druid\druid-shell)
warning: unused variable: `warning`
  --> druid-shell\src\window.rs:37:13
   |
37 |         let warning = "warning";
   |             ^^^^^^^ help: consider prefixing with an underscore: `_warning`
   |
   = note: #[warn(unused_variables)] on by default

   Compiling druid v0.3.0 (C:\Users\hilma\git\druid)
    Finished dev [unoptimized + debuginfo] target(s) in 8.89s
```

ci/build.yml
LiHRaM: I was trying to be consistent with the apparent naming convention of the other scripts, which seemed to be more verbose. Should I change them as well? See e.g. rustfmt and test.

.azure-pipelines.yml
cmyr: Okay cool, thanks. It’s possible there’s another compiler flag for this but I’m satisfied as-is!

ci/build.yml
cmyr: Consistency is good, happy with this as is!

.azure-pipelines.yml
LiHRaM: Yeah, there is an environment variable for denying warnings, but this fails on warnings on any dependency, which I'm not sure is desired behavior?

cmyr: yea I could see that causing headaches.

PR #142, opened 2019-09-06
Add scroll bars

zaynetro:
scroll-bars

### DIFF
--- a/examples/scroll_colors.rs
+++ b/examples/scroll_colors.rs
@@ -1,0 +1,121 @@
+// Copyright 2019 The xi-editor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! This example allows to play with scroll bars over different color tones.
+
+use druid::shell::kurbo::{Rect, Size};
+use druid::shell::piet::{Color, RenderContext};
+use druid::shell::{runloop, WindowBuilder};
+use druid::widget::{Column, Row, Scroll};
+use druid::{
+    Action, BaseState, BoxConstraints, Data, Env, Event, EventCtx, LayoutCtx, PaintCtx, UiMain,
+    UiState, UpdateCtx, Widget,
+};
+
+struct FixedSize {
+    size: Size,
+    color: Color,
+}
+
+impl FixedSize {
+    pub fn new(size: Size, color: Color) -> Self {
+        Self { size, color }
+    }
+}
+
+impl<T: Data> Widget<T> for FixedSize {
+    fn paint(&mut self, paint_ctx: &mut PaintCtx, base_state: &BaseState, _data: &T, _env: &Env) {
+        let color = if base_state.is_hot() {
+            Color::BLACK
+        } else {
+            self.color.clone()
+        };
+        let brush = paint_ctx.render_ctx.solid_brush(color);
+        let rect = Rect::from_origin_size((0.0, 0.0), base_state.size());
+        paint_ctx.render_ctx.fill(rect, &brush);
+    }
+
+    fn layout(
+        &mut self,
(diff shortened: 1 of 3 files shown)


src/widget/scroll.rs
```diff
@@ -15,17 +15,23 @@
...
-use crate::piet::RenderContext;
+use crate::piet::{Color, RenderContext};
 
-use crate::kurbo::Affine;
+use crate::kurbo::{Affine, RoundedRect};
+
+const SCROLL_BAR_WIDTH: f64 = 8.;
+const SCROLL_BAR_PAD: f64 = 2.;
+const SCROLL_BAR_BG: Color = Color::rgb8(255, 255, 255);
+const SCROLL_BAR_BORDER: Color = Color::rgba8(0, 0, 0, 76);
```
zaynetro: It seems there is now a `theme.rs` file. Which values should live there? Only colors?

src/widget/scroll.rs
```diff
@@ -147,11 +235,35 @@ impl<T: Data> Widget<T> for Scroll<T> {
         } else {
             None
         };
+
+        match event {
+            Event::AnimFrame(_) => {
```
raphlinus: A couple things here. First, the scroll bars will fade immediately if there's some other widget requesting animation. I'd guard this by the timer id being invalid.

Second, on high refresh rate monitors the fade will be faster. I'd use the  interval to control the fade speed.

src/widget/scroll.rs
```diff
@@ -147,11 +235,35 @@ impl<T: Data> Widget<T> for Scroll<T> {
...
+                // Animate scroll bars opacity
+                let opacity = self.scroll_bars.opacity.abs();
+                let diff = 0.004 / opacity;
+                self.scroll_bars.opacity -= diff;
```
raphlinus: I think it'd be cleaner to use saturating subtraction here, to prevent opacity from going negative (though see you do have a `<=` check above, so the overall effect is the same).

src/widget/scroll.rs
```diff
@@ -104,6 +126,69 @@ impl<T: Data> Scroll<T> {
...
+    /// Draw scroll bars.
+    fn draw_bars(&self, paint_ctx: &mut PaintCtx, viewport: &Rect) {
+        if self.scroll_bars.opacity <= 0.0 {
+            return;
+        }
+
+        let brush = paint_ctx
+            .render_ctx
+            .solid_brush(SCROLL_BAR_BG.with_alpha(self.scroll_bars.opacity));
+        let border_brush = paint_ctx.render_ctx.solid_brush(SCROLL_BAR_BORDER);
```
raphlinus: I think there's a little fine-tuning that can be done on the appearance. In particular, the border does not fade smoothly during the fade animation, but disappears suddenly at the end. This is pretty subtle because its opacity is low, but could be better. I think the right way to do that is to put both fill and stroke in an opacity group, but this is not in piet yet.

src/widget/scroll.rs
zaynetro: These are good findings! I wasn't aware of any.

src/widget/scroll.rs
futurepaul: The way I've been doing it is: colors in the theme, sizes in the theme only if they're standard sizes used by other widgets.

zaynetro: Moved colors to the `theme.rs`

src/widget/scroll.rs
zaynetro: Changed.

src/widget/scroll.rs
zaynetro: Changed border color to have a dynamic opacity.

examples/scroll_colors.rs
```diff
@@ -0,0 +1,121 @@
...
+    color: Color,
+}
+
+impl FixedSize {
+    pub fn new(size: Size, color: Color) -> Self {
+        Self { size, color }
+    }
+}
+
+impl<T: Data> Widget<T> for FixedSize {
```
raphlinus: This is in an example, so less important, but I'm wondering if its use could be replaced with `SizedBox` now that's landed.

zaynetro: There is a reason below:

>       // There is a SizedBox widget but it's not currently possible
>       // to set a background on it.

I wanted to use `SizedBox` but we currently have no way to style it. Would be nice to discuss how we do basic stylings (background, border, ...)

raphlinus:
Thanks, that makes sense. Thanks for this, I'm merging this now, and also adding you to as a committer.

zaynetro:
Thanks for the collaborator invitation!

PR #145, opened 2019-09-09
Add Checkbox widget

futurepaul:
I believe this widget is straightforward in its implementation, but I'm curious if there's a better way to use it. I put it in the slider example, with the help of `Lens`. It felt pretty simple to use once I got past the setup boilerplate, but I'm sure there's another strategy I'm not thinking of.

I believe one possibility, which I did not explore, is to return a "toggle" action or something from `event` (like how the button returns "hit") and then handle it with an `ActionWrapper`.

### DIFF
--- a/src/widget/checkbox.rs
+++ b/src/widget/checkbox.rs
@@ -1,0 +1,128 @@
+// Copyright 2019 The xi-editor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


src/widget/checkbox.rs
```diff
@@ -0,0 +1,128 @@
...
+
+use crate::kurbo::{BezPath, Point, RoundedRect, Size};
+use crate::piet::{LineCap, LineJoin, LinearGradient, RenderContext, StrokeStyle, UnitPoint};
+use crate::theme;
+use crate::widget::Align;
+use crate::{
+    Action, BaseState, BoxConstraints, Env, Event, EventCtx, LayoutCtx, PaintCtx, UpdateCtx, Widget,
+};
+
+/// A checkbox that toogles a boolean
```
LiHRaM: Not really qualified to make comments about the implementation, but noticed a typo here. 😄 
```suggestion
/// A checkbox that toggles a boolean
```

futurepaul: Thank yoo for catching! Also: I love this suggestion thing.

PR #146, opened 2019-09-10
Add Del key behavior

mralve:
This just adds the default text editing behavior for the delete key, at least on windows.

### DIFF
--- a/src/widget/textbox.rs
+++ b/src/widget/textbox.rs
@@ -427,7 +427,7 @@ impl Widget<String> for TextBoxRaw {
                     k_e if (HotKey::new(None, KeyCode::Delete)).matches(k_e) => {
                         if self.selection.is_caret() {
                             // Never touch the characters before the cursor.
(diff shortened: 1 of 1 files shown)


src/widget/textbox.rs
```diff
@@ -460,6 +474,18 @@ fn next_grapheme(src: &str, from: usize) -> usize {
     }
 }
 
+/// Checks if there is a next character from the given index.
+fn next_grapheme_exsists(src: &str, from: usize) -> bool {
```
zaynetro: Typo `next_grapheme_exsists` -> `next_grapheme_exists`

mralve: Wops thanks for catching that.

futurepaul:
Thanks for this!

I haven't tested on Windows but this works on my Macbook with fn+delete and with my Bluetooth keyboard's del key.

raphlinus:
Thanks for the contribution, and to @futurepaul for the review.

PR #147, opened 2019-09-11
Add app.rs & AppLauncher

cmyr:
This simplifies the API for creating windows and starting a druid
application.

This was broken out of multi-window work. That work requires a
platform-agnostic way to describe a window; that window description
is used here as the basis for simpler initial setup.

This approach will also make it easier, in the future, for us
to implement things like state restoration.

-----
notes:

There's some stuff in here (like the title being a localized string) that doesn't make a ton of sense right now, but will end up being relevant for multi-win.

I'm not sure about names, in particular `WindowDesc`. I'm also not sure about `app.rs`. 🤷‍♀ 

@LiHRaM If you were interested, this is where a custom logging impl would go; we would add a method to `AppLauncher` that would enable logging, and if the user called that we would set up some (maybe custom?) simple logger that prints to stdout.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -1,0 +1,113 @@
+// Copyright 2019 The xi-editor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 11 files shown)


src/app.rs
```diff
@@ -0,0 +1,112 @@
...
+        //TODO: launch all windows, when multi-win lands
+        let window = self.windows.pop().expect("launch called with no window");
+        let WindowDesc {
+            root_builder,
+            title,
+            ..
+        } = window;
+
+        //TODO: use title when multi-win lands
+        let _title = title.unwrap_or(LocalizedString::new("app-name-exclaim"));
```
raphlinus: It looks like this regresses setting a title at all. I'm patient with that landing, but if it's a simple change to keep basic (non-localized) functionality in place, that'd be cool.

cmyr: ah fair, changed so it will use the non-localized `localized_str()` value (e.g. the placeholder if one is set, or the key if it isn't).

PR #149, opened 2019-09-12
multiple windows, round two

cmyr:
This is my attempt to get very basic multi-win support working.

This is simpler and less flexible than the original approach in #127. My thinking is that we should do as little as possible, and then figure out what missing features we will need for runebender.

This will require some follow-up work; in particular I want to add a `NewWindow` event that announces a new window, and passes along the window id so that other windows can send that window messages.

I'm not especially confident in this design; it's hard to really anticipate problems until I've used it more, but I hope it's a reasonable starting point.

I'm not totally sure I have the event/update logic correct.

In any case, I want to get this up now as a checkpoint. I'm going to start trying to weld menus onto this, and we'll see how that goes? If I'm having a lot of trouble I may need to reconsider some things here. 🤷‍♂


src/command.rs
```diff
@@ -0,0 +1,118 @@
...
+/// let selector = Selector::new("process_rows");
+/// let rows = vec![1, 3, 10, 12];
+/// let command = Command::new(selector, rows);
+///
+/// assert_eq!(command.get_object(), Some(&vec![1, 3, 10, 12]));
+/// ```
+#[derive(Debug, Clone)]
+pub struct Command {
+    pub selector: Selector,
+    pub object: Option<Arc<dyn Any>>,
```
raphlinus: Discussion question: is the option dynamic or static (ie some selectors just don't need arguments)? In the latter case, it might be a convention to use `()` as the argument. On the other hand, that's maybe less efficient, as it would still need to be boxed...

src/command.rs
```diff
@@ -0,0 +1,118 @@
...
+impl From<Selector> for Command {
+    fn from(selector: Selector) -> Command {
+        Command {
+            selector,
+            object: None,
+        }
+    }
+}
+
+impl PartialEq<Selector> for Command {
```
raphlinus: These violate transitivity. On the other hand, I would be perfectly fine with them if they also enforced that the argument was null.

src/lib.rs
```diff
@@ -1048,12 +859,35 @@ impl<'a, 'b> EventCtx<'a, 'b> {
         self.base_state.request_timer = true;
         self.win_ctx.request_timer(deadline)
     }
+
+    /// Submit a [`Command`] to be run after this event is handled.
+    ///
+    /// Commands are run in the order they are submitted; all commands
+    /// submitted during the handling of an event are executed before that
```
raphlinus: The language seems odd here, maybe "before the update method is called"?

src/win_handler.rs
```diff
@@ -0,0 +1,458 @@
...
+impl<'a, T: Data> SingleWindowState<'a, T> {
+    fn paint(&mut self, piet: &mut Piet, ctx: &mut dyn WinCtx) -> bool {
+        let request_anim = self.send_anim_frame(ctx);
+        self.send_layout(piet);
+        piet.clear(BACKGROUND_COLOR);
+        self.send_paint(piet);
+        request_anim
+    }
+
+    fn send_anim_frame(&mut self, ctx: &mut dyn WinCtx) -> bool {
```
raphlinus: Not wild about the "send" naming here as it seems to imply async or deferred operation, but I do see why you want a different name than "do" as below.

druid-shell/src/mac/mod.rs
```diff
@@ -581,7 +581,13 @@ impl WindowHandle {
...
         if let Some(ref nsview) = self.nsview {
             unsafe {
-                let window: id = msg_send![*nsview.load(), window];
+                let view = nsview.load();
+                let window: id = msg_send![*view, window];
+                //FIXME: It's possible that we're leaking windows here,
+                //but if we don't set this we segfault after close. Maybe we should
+                //have an appdelegate that handles the windowShouldClose notification?
+                //See https://developer.apple.com/documentation/appkit/nswindow/1419062-releasedwhenclosed?language=objc
+                msg_send![window, setReleasedWhenClosed: NO];
```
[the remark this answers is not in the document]
cmyr: this doesn't fix that segfault, which I haven't tried tracking down yet. It does fix a segfault that was happening when we called `close()` ourselves, though.

examples/multiwin.rs
```diff
@@ -0,0 +1,118 @@
...
+    let button = Button::new("increment");
+
+    let mut col = Column::new();
+    col.add_child(Align::centered(Padding::uniform(5.0, label)), 1.0);
+    col.add_child(Padding::uniform(5.0, button), 1.0);
+    let wrapper = ActionWrapper::new(col, |data: &mut u32, _env| *data += 1);
+
+    EventInterceptor::new(wrapper, |event, ctx, _data, _env| {
+        if let Event::KeyDown(e) = event {
+            if HotKey::new(SysMods::Cmd, "n").matches(e) {
```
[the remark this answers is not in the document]
cmyr: Hmm, that's confusing, I guess I don't know enough about how `WM_CHAR` and `WM_KEYUP` work.

src/command.rs
cmyr: My intention was that the expected type of the argument is determined by the selector. Also I'm not sure if this needs to be `pub`; it should only be accessed through `get_object`, I think..

src/command.rs
cmyr: Oops, these are unused, I can remove them.

cmyr:
Oops didn't notice this review come in, thanks!

There are definitely some crashes to be teased out, by my inclination here is to merge and continue from there, as long as we think this is a reasonable general approach.

raphlinus:
Yeah, I'm fine with that. I'll mark this as approved.

PR #156, opened 2019-09-16
Menus

cmyr:
This introduces a new platform-agnostic API for menus. It's still a little rough around the edges, but I think it's ready for a look. 

- Menus are assigned `Commands`; when a menu is selected, the corresponding command is delivered to the widget tree as an `Event`.

- The menu can be changed at runtime by sending a 'set menu' command.

- Menus are per-window; on macOS the active window controls the app menu.

- This wires up some standard macOS menu items: new, close window, quit, hide, and hide all are all working.

Questions, notes, concerns:

- This adds a way to create a `Button` that uses a closure to mutate the data, and optionally return an event. This doesn't belong in this PR, but it made it much easier to write the `multiwin` example. Something like this should probably exist?

- It feels like maybe `MenuDesc` should have an `update` fn that determines if we should rebuild the menu? That would make the general case simpler. I've punted on this for now.

- I'd like there to be a better way for us to provide the standard menu items for a given platform.


examples/multiwin.rs
```diff
@@ -69,6 +94,7 @@ fn ui_builder() -> impl Widget<u32> {
 /// and which can either consume events itself or return them to have them
 /// be passed to the inner widget.
 struct EventInterceptor<T> {
+    /// Custom widget-level state
```
cmyr: delete me please

resources/i18n/en-US/builtin.ftl
```diff
@@ -1,3 +1,24 @@
...
+macos-menu-preferences = Preferences...
+macos-menu-hide-app = Hide { -app-name }
+macos-menu-hide-others = Hide Others
+macos-menu-show-all = Show All
+macos-menu-services = Services
+macos-menu-application-menu = { -app-name }
+macos-menu-quit-app = Quit { -app-name }
+
+# macOS 'File' menu
+macos-menu-file-menu = File
```
raphlinus: So to make Alt-navigation work on Windows, this should be `&File`, `&New File`, etc. But in the current code, that causes the literal ampersand to be shown on mac. There are of course a bunch of ways to handle this, but I personally would lean towards having the ampersands in the strings and stripping them on mac. (I'm fairly sure some previous iteration of the code did this).

druid-shell/examples/hello.rs
```diff
@@ -109,10 +109,22 @@ fn main() {
...
+    file_menu.add_item(
+        0x101,
+        "O&pen",
+        Some(&HotKey::new(SysMods::Cmd, "o")),
+        true,
+        false,
+    );
     let mut menubar = Menu::new();
-    menubar.add_dropdown(file_menu, "&File");
+    menubar.add_dropdown(file_menu, "&File", true);
```
raphlinus: For some reason, this menu doesn't seem to be showing up on mac (but is fine on Windows). I haven't investigated more deeply.

src/menu.rs
```diff
@@ -0,0 +1,525 @@
...
+    ///         MenuItem::new(
+    ///             LocalizedString::new("hello-counter").with_arg("count", move |_, _| i.into()),
+    ///             Command::new(MENU_COUNT_ACTION, i),
+    ///        )
+    ///     })
+    /// );
+    ///
+    /// assert_eq!(my_menu.len(), 4);
+    /// ```
+    pub fn append_iter<I: Iterator<Item = MenuItem<T>>>(mut self, f: impl FnOnce() -> I) -> Self {
```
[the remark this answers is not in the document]
cmyr: `extend` is part of the `Extend` trait, so I'm reluctant to implement it directly without us being able to use that trait.

I did try and have this take an iterator directly (instead of a closure that produces an iterator) but there was some problem I ran into. 😕

resources/i18n/en-US/builtin.ftl
cmyr: My plan here is to have different localized strings for windows; this will let us exactly match native localization per-platform.

That said, my intent was that macOS should be stripping `&`, so if it isn't that's a bug.

druid-shell/examples/hello.rs
cmyr: It seems like the first menu item is always given the title of the running application; we would need to add an extra menu item if we wanted a file menu.

resources/i18n/en-US/builtin.ftl
raphlinus: Interesting; I thought about that a bit. My personal feeling is that things should be fanned out cross-platform by default, but there should be the option to override that to make things more platform-native. But if the conventions are consistently different, maybe it's better to explicitly highlight the requirement to do multiple versions.

cmyr: My feeling here is that getting window/app menus right is an important part of feeling native, and the only way to really do that is to declare them per-platform.

For instance: the menu item to quit the application is 'Quit' on mac and 'Exit' on windows. On mac, it's in the 'application' menu; on windows it's in the 'File' menu.

I think it will be hard to have a single 'default menu' that correctly handles this case, and many others like it. Instead what I would like to do is to expose convenience functions that will create individual menus and menu items, per platform, so that it is as easy as possible for the user to create native-feeling menus, but still have the expectation be that you're doing this manually.

PR #163, opened 2019-09-17
Add default logger to app builder

### LINKED ISSUE #162: add option to log to stderr to AppLauncher
This would be a builder style fn, so that you could write:

```rust
        .print_logs(true)
        .launch(State::default())
        .expect("launch failed");
}
```

And then when we start the application we could setup a very barebones log handler that just calls `eprintln`. (@LiHRaM had something like this in an early draft of #135, which we might be able to reuse?)

LiHRaM:
closes #162


src/app.rs
```diff
@@ -48,6 +48,11 @@ impl<T: Data + 'static> AppLauncher<T> {
         }
     }
 
+    pub fn use_logger(self) -> Self {
```
cmyr: This name is slightly misleading; there are many ways to use the a logger (for instance the user can set up logging manually, or in the future we might by default enable logging to the platform logging service) so what we're really doing here is printing logs to stderr. You could say `use_simple_logger` or `print_logs` or `print_log_messages` or something? And I'd also include a doc comment to explain that this is provided for debugging, during development.

Cargo.toml
```diff
@@ -22,10 +22,8 @@ fluent-syntax = "0.9"
 unic-langid = "0.4.1"
 unicode-segmentation = "1.3.0"
 log = "0.4.8"
+simple_logger = "1.3.0"
```
cmyr: cool, I was thinking we might want to avoid the dependency and just do something simple ourselves, but this doesn't seem to have too much overhead, so I think it's fine. If you wanted to play around with implementing an even simpler logger yourself you'd be welcome, but I'm happy with this.

cmyr:
I also think I'd turn off the color output, to keep things as light as possible.

```
[dependencies.simple_logger]
default-features = false
```

LiHRaM:
Thanks for the fast feedback, @cmyr!

I'm happy with just using `simple_logger` as well, I feel like implementing our own might just add maintenance burden atm. 🤷‍♂

PR #165, opened 2019-09-18
Replace mem::uninitialized with mem::MaybeUninit

### LINKED ISSUE #164: Replace mem::uninitialized with mem::MaybeUninit
[See docs for info](https://doc.rust-lang.org/std/mem/fn.uninitialized.html): `mem::uninitialized` is deprecated, and should be replaced by `mem::MaybeUninit`.

LiHRaM:
closes #164 

I ran some examples, and everything _seemed_ to work just the same. Using `::zeroed`, I believe the structs in these cases can handle zero values for all of their fields. I am not familiar with the external functions, however, and don't know if we want to `assume_init` after calling them.

### DIFF
--- a/druid-shell/src/windows/mod.rs
+++ b/druid-shell/src/windows/mod.rs
@@ -390,8 +390,11 @@ impl WndProc for MyWndProc {
                 if let Ok(mut s) = self.state.try_borrow_mut() {
                     let s = s.as_mut().unwrap();
                     if s.dcomp_state.is_some() {
-                        let mut rect = mem::MaybeUninit::<RECT>::zeroed().assume_init();
-                        GetClientRect(hwnd, &mut rect);
+                        let mut rect: RECT = mem::zeroed();
+                        if GetClientRect(hwnd, &mut rect) == 0 {
+                            warn!("GetClientRect failed.");
+                            return None;
+                        }
                         let width = (rect.right - rect.left) as u32;
                         let height = (rect.bottom - rect.top) as u32;
                         let res = (*s.dcomp_state.as_mut().unwrap().swap_chain).ResizeBuffers(
@@ -878,8 +881,13 @@ unsafe fn choose_adapter(factory: *mut IDXGIFactory2) -> *mut IDXGIAdapter {
         if !SUCCEEDED((*factory).EnumAdapters(i, &mut adapter)) {
             break;
         }
-        let mut desc = mem::MaybeUninit::<DXGI_ADAPTER_DESC>::zeroed().assume_init(); // TODO: Confirm that this works correctly
-        (*adapter).GetDesc(&mut desc);
+        let mut desc = mem::MaybeUninit::uninit();
+        let hr = (*adapter).GetDesc(desc.as_mut_ptr());
+        if !SUCCEEDED(hr) {
+            error!("Failed to get adapter description: {:?}", Error::Hr(hr));
+            break;
+        }
+        let mut desc: DXGI_ADAPTER_DESC = desc.assume_init();
         let vram = desc.DedicatedVideoMemory;
         if i == 0 || vram > best_vram {
             best_vram = vram;
--- a/druid-shell/src/windows/mod.rs
+++ b/druid-shell/src/windows/mod.rs
@@ -390,7 +390,7 @@ impl WndProc for MyWndProc {
                 if let Ok(mut s) = self.state.try_borrow_mut() {
                     let s = s.as_mut().unwrap();
                     if s.dcomp_state.is_some() {
-                        let mut rect: RECT = mem::uninitialized();
+                        let mut rect = mem::MaybeUninit::<RECT>::zeroed().assume_init();
                         GetClientRect(hwnd, &mut rect);
                         let width = (rect.right - rect.left) as u32;
                         let height = (rect.bottom - rect.top) as u32;
@@ -878,7 +878,7 @@ unsafe fn choose_adapter(factory: *mut IDXGIFactory2) -> *mut IDXGIAdapter {
         if !SUCCEEDED((*factory).EnumAdapters(i, &mut adapter)) {
             break;
         }
-        let mut desc: DXGI_ADAPTER_DESC = mem::uninitialized();
+        let mut desc = mem::MaybeUninit::<DXGI_ADAPTER_DESC>::zeroed().assume_init(); // TODO: Confirm that this works correctly
         (*adapter).GetDesc(&mut desc);
         let vram = desc.DedicatedVideoMemory;
         if i == 0 || vram > best_vram {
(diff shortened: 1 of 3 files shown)


druid-shell/src/windows/mod.rs
```diff
@@ -899,7 +899,7 @@ unsafe fn choose_adapter(factory: *mut IDXGIFactory2) -> *mut IDXGIAdapter {
         if !SUCCEEDED((*factory).EnumAdapters(i, &mut adapter)) {
             break;
         }
-        let mut desc: DXGI_ADAPTER_DESC = mem::uninitialized();
+        let mut desc = mem::MaybeUninit::<DXGI_ADAPTER_DESC>::zeroed().assume_init(); // TODO: Confirm that this works correctly
```
LiHRaM:
```suggestion
        let mut desc = mem::MaybeUninit::<DXGI_ADAPTER_DESC>::zeroed().assume_init();
```

druid-shell/src/windows/mod.rs
```diff
@@ -391,7 +391,7 @@ impl WndProc for MyWndProc {
                 if let Ok(mut s) = self.state.try_borrow_mut() {
                     let s = s.as_mut().unwrap();
                     if s.dcomp_state.is_some() {
-                        let mut rect: RECT = mem::uninitialized();
```
cmyr: This stuff is new to me too, but reading the docs for `MaybeUninit` it doesn't sound like `assume_init` is what we want. I think we want to just use `uninit()`, and then call `assum_init` after we've called the external function.

Basically what's happening in all these places is we're passing a memory location to the windows API, which is supposed to write something to that location; so the memory is not safe to access until after that write has happened (and I imagine all of these winapi functions return a flag indicating if there is an error or not, and if there's an error we should also assume the memory is uninitialized...)

LiHRaM: Yeah, so the current refactoring initializes the memory to safe values and assumes init, which I guess is only _slightly_ safer than the previous use of `mem::uninitialized` - I was just treading the path of least resistance there. The problem then becomes whether passing `MaybeUninit` to these external functions is the same as passing the pointers themselves, or whether we need to do some black magic there. 🤔I wasn't able to tell from the docs how MaybeUninit translates to C.

cmyr: ah okay, I guess the main question I'd have is whether `zeros` is valid data for all the types used here? If it is then that works for me!

cmyr: and yes, passing a pointer to the `MaybeUninit` is the same as passing a pointer to the value; all `MaybeUninit` does is use the type system to force us to be explicit about initialization on the Rust side.

druid-shell/src/windows/win_main.rs
```diff
@@ -84,7 +84,7 @@ impl RunLoop {
 
                 // Handle windows messages
                 loop {
-                    let mut msg = mem::uninitialized();
+                    let mut msg = mem::MaybeUninit::<MSG>::zeroed().assume_init();
```
LiHRaM: @cmyr Am unable to find `MSG`, which is supposed to be the type for msg here.

`RECT` should be fine to zero, as it just contains four numerical values.
`DXGI_ADAPTER_DESC` is confusing to me. The type declaration is as follows:
```rust
struct DXGI_ADAPTER_DESC {
    Description: [WCHAR; 128],
    VendorId: UINT,
    DeviceId: UINT,
    SubSysId: UINT,
    Revision: UINT,
    DedicatedVideoMemory: SIZE_T,
    DedicatedSystemMemory: SIZE_T,
    SharedSystemMemory: SIZE_T,
    AdapterLuid: LUID,
}
```
My search-foo fails me here, am unable to find out what `LUID` is. On the other hand, I assume that Rust initializes the vector for `Description` within the struct instead of using a pointer to an external vector, which would make zeroing it safe.

druid-shell/src/windows/mod.rs
raphlinus: This is fancier than needed for RECT, you can just do `mem::zeroed` instead of `mem::uninitialized`. It's a tiny optimization not to zero, but when I wrote the code I was doing things "the C way."

druid-shell/src/windows/mod.rs
```diff
@@ -899,7 +899,7 @@ unsafe fn choose_adapter(factory: *mut IDXGIFactory2) -> *mut IDXGIAdapter {
         if !SUCCEEDED((*factory).EnumAdapters(i, &mut adapter)) {
             break;
         }
-        let mut desc: DXGI_ADAPTER_DESC = mem::uninitialized();
+        let mut desc = mem::MaybeUninit::<DXGI_ADAPTER_DESC>::zeroed().assume_init();
```
raphlinus: I think the way I'd write this is:

```rust
    let desc = mem::MaybeUninit::uninit();
    let hr = (*adapter).GetDesc(desc.as_mut_ptr());
    ...check hr...
    let desc = desc.assume_init();
```

I'm noticing that the existing code is not checking the return code. This is UB if the call fails.

druid-shell/src/windows/mod.rs
LiHRaM: Okay, good to know. Now my only concern is with the implications of `GetClientRect` returning false, is that something we want to log and/or return an error?

raphlinus: Yes. This is extremely unlikely, but we want to program defensively in any case.

LiHRaM: Thanks for the feedback, friends! It's updated now. 😄

druid-shell/src/windows/mod.rs
```diff
@@ -878,8 +881,13 @@ unsafe fn choose_adapter(factory: *mut IDXGIFactory2) -> *mut IDXGIAdapter {
...
         }
-        let mut desc: DXGI_ADAPTER_DESC = mem::uninitialized();
-        (*adapter).GetDesc(&mut desc);
+        let mut desc = mem::MaybeUninit::uninit();
+        let hr = (*adapter).GetDesc(desc.as_mut_ptr());
+        if !SUCCEEDED(hr) {
+            error!("Failed to get adapter description: {:?}", Error::Hr(hr));
+            break;
+        }
+        let mut desc: DXGI_ADAPTER_DESC = desc.assume_init();
```
raphlinus: It feels like this doesn't really need to be `mut`. The problem is that our "util" module takes a `LPWSTR`, which is more idiomatically C. I think the solution is to use the [from_wide_ptr_null()](https://docs.rs/wio/0.2.2/i686-pc-windows-msvc/wio/wide/trait.FromWide.html#method.from_wide_ptr_null) method from wio and remove our hand-rolled "util" stuff, but that's arguably a separate issue from the scope of this PR, which is specifically about uninit.

LiHRaM: Sounds good. I'll add an issue for it.

PR #170, opened 2019-09-19
Mouse convenience

cmyr:
These save having to bring `MouseButton` into scope in the common case of just wanting to know if an event is a left or right click.

### DIFF
--- a/druid-shell/src/window.rs
+++ b/druid-shell/src/window.rs
@@ -210,6 +210,20 @@ pub enum MouseButton {
     X2,
 }
 
+impl MouseButton {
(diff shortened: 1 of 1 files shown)


druid-shell/src/window.rs
```diff
@@ -210,6 +210,18 @@ pub enum MouseButton {
     X2,
 }
 
+impl MouseButton {
+    /// Returns `true` if this is the left mouse button.
+    pub fn is_left(self) -> bool {
```
raphlinus: Maybe `#[inline]` because this the comparison is cheaper than the function call?

PR #172, opened 2019-09-19
Strip bidi control characters from localization results

cmyr:
These were being rendered as 'missing glyph' in various places,
and aren't currently buying us anything.

### DIFF
--- a/src/localization.rs
+++ b/src/localization.rs
@@ -269,7 +269,22 @@ impl L10nManager {
         for err in errs {
             warn!("localization error {:?}", err);
         }
-        Some(result.to_string())
+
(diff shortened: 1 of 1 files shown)


cmyr:
The thing that pushed me over the edge is that this didn't work correctly in menus on windows. 😒

raphlinus:
> The thing that pushed me over the edge is that this didn't work correctly in menus on windows. 😒

Ah, that sucks :/ I wonder what's going on there; probably some old code paths that haven't been upgraded to DirectWrite yet. Well, this is probably the simplest approach then.

cmyr:
Yea, happy to think of this as a bandaid, we can be more judicious at some point in the future, if we're seeing real-world bugs.

PR #174, opened 2019-09-20
Support astral plane characters on Windows

chris-morgan:


cmyr:
cool this looks good, will test when I'm next in front of my proper machine. Thanks!

src/widget/textbox.rs
```diff
@@ -431,12 +431,13 @@ impl Widget<String> for TextBoxRaw {
                         self.reset_cursor_blink(ctx);
                     }
                     // Actual typing
-                    k_e if k_e.key_code.is_printable() => {
```
raphlinus: The problem with removing this is that control characters get inserted into the string. I agree with you that we shouldn't be looking at the keycode (that's not really a source of truth), but should have an API indicating whether the keyboard event has printable text.

chris-morgan: In what situations do control characters come through this path? I haven’t tried much, but nothing immediately sprang to mind.

raphlinus: It happens when you press, say, control-A, but the command isn't interpreted.

cmyr:
@chris-morgan do you have any interest in revisiting this?

chris-morgan:
Not personally at present. In a month or six maybe if no one has by then, but not now. This was just a quick proof of concept, someone else can feel free to take it over and do it however they like.

cmyr:
okay, thanks. Going to close for now, we can revisit down the road.

PR #177, opened 2019-09-22
basic copy & paste

### LINKED ISSUE #98: Add copy / paste
As we work toward the "longer term future" as described in https://github.com/xi-editor/druid/issues/16, a pasteboard (the Mac word) / clipboard (the Windows word) would very nice to have!

One possibility is to rely on the [Clipboard crate](https://crates.io/crates/clipboard). In fact, it seems like it's not totally necessary for this functionality to be inside druid-shell or even druid? To my understanding, [the winit project declared clipboard functionality out of scope](https://github.com/rust-windowing/winit/issues/162) for this reason.

In the longer term, I would expect a clipboard that supports data types beyond text, or a non-global pasteboard, will be important for apps like Runebender. But maybe that's running before we can walk.

cmyr:
(This includes #175).

This adds basic clipboard support (text only) on mac and windows. 

The API is a bit funny; I've added an `Event` for paste, and but setting the clipboard is implemented in `platform::Application`, and then hung off of `WinCtx` because we don't necessarily want to expose `platform::Application`?

There's a reasonable case that we could also have `Event::Cut` and `Event::Copy`.

I worry that our API is getting a bit fragmented; when should things be commands, when should they be events, when should they be on `WinCtx` or `WindowHandle` etc; I don't feel like I have much clarity around this stuff.

closes #98

### DIFF
--- a/druid-shell/src/mac/application.rs
+++ b/druid-shell/src/mac/application.rs
@@ -52,7 +52,7 @@ impl Application {
             let data_types: id = msg_send![pasteboard, types];
             let count: usize = msg_send![data_types, count];
 
-            for i in 0..=count {
+            for i in 0..count {
                 let dtype: id = msg_send![data_types, objectAtIndex: i];
                 let is_string: BOOL = msg_send![dtype, isEqualToString: NSPasteboardTypeString];
                 if is_string == YES {
@@ -68,7 +68,7 @@ impl Application {
         }
     }
 
-    /// Sets th contents of the system clipboard.
+    /// Sets the contents of the system clipboard.
     pub fn set_clipboard_contents(item: ClipboardItem) {
         unsafe {
             let nspasteboard = class!(NSPasteboard);
--- a/druid-shell/src/mac/application.rs
+++ b/druid-shell/src/mac/application.rs
@@ -14,8 +14,10 @@
 
 //! macOS implementation of features at the application scope.
 
-use cocoa::appkit::NSApp;
-use cocoa::base::{id, nil};
+use super::util;
+use crate::clipboard::ClipboardItem;
+use cocoa::appkit::{NSApp, NSPasteboardTypeString};
+use cocoa::base::{id, nil, BOOL, YES};
 
 pub struct Application;
 
@@ -39,6 +41,30 @@ impl Application {
             let workspace = class!(NSWorkspace);
             let shared: id = msg_send![workspace, sharedWorkspace];
             msg_send![shared, hideOtherApplications];
+        }
+    }
+
+    /// Returns the contents of the clipboard, if any.
+    pub fn get_clipboard_contents() -> Option<ClipboardItem> {
+        unsafe {
+            let nspasteboard = class!(NSPasteboard);
+            let pasteboard: id = msg_send![nspasteboard, generalPasteboard];
+            let data_types: id = msg_send![pasteboard, types];
+            let count: usize = msg_send![data_types, count];
+
+            for i in 0..=count {
+                let dtype: id = msg_send![data_types, objectAtIndex: i];
+                let is_string: BOOL = msg_send![dtype, isEqualToString: NSPasteboardTypeString];
+                if is_string == YES {
+                    let contents: id = msg_send![pasteboard, stringForType: dtype];
+                    let contents = util::from_nsstring(contents);
+                    return Some(contents.into());
+                } else {
+                    log::info!("unhandled pasteboard type {}", util::from_nsstring(dtype));
+                }
+                //TODO: handle other data types
+            }
+            None
         }
     }
 }
--- a/druid-shell/src/mac/application.rs
+++ b/druid-shell/src/mac/application.rs
@@ -67,4 +67,20 @@ impl Application {
             None
         }
     }
+
+    /// Sets th contents of the system clipboard.
+    pub fn set_clipboard_contents(item: ClipboardItem) {
+        unsafe {
+            let nspasteboard = class!(NSPasteboard);
+            let pasteboard: id = msg_send![nspasteboard, generalPasteboard];
+            match item {
+                ClipboardItem::Text(string) => {
+                    let nsstring = util::make_nsstring(&string);
+                    msg_send![pasteboard, clearContents];
+                    msg_send![pasteboard, setString: nsstring forType: NSPasteboardTypeString];
+                }
+                other => log::warn!("unhandled clipboard data {:?}", other),
+            }
+        }
+    }
 }
--- a/druid-shell/src/window.rs
+++ b/druid-shell/src/window.rs
@@ -17,6 +17,8 @@
 use std::any::Any;
 use std::ops::Deref;
 
+use crate::clipboard::ClipboardItem;
+//TODO: why is this pub?
 pub use crate::keyboard::{KeyEvent, KeyModifiers};
 use crate::kurbo::{Point, Vec2};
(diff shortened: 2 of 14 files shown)


druid-shell/src/mac/application.rs
```diff
@@ -41,4 +43,44 @@ impl Application {
...
+                } else {
+                    log::info!("unhandled pasteboard type {}", util::from_nsstring(dtype));
+                }
+                //TODO: handle other data types
+            }
+            None
+        }
+    }
+
+    /// Sets th contents of the system clipboard.
```
raphlinus: "the"

druid-shell/src/mac/application.rs
```diff
@@ -41,4 +43,44 @@ impl Application {
...
+
+    /// Returns the contents of the clipboard, if any.
+    pub fn get_clipboard_contents() -> Option<ClipboardItem> {
+        unsafe {
+            let nspasteboard = class!(NSPasteboard);
+            let pasteboard: id = msg_send![nspasteboard, generalPasteboard];
+            let data_types: id = msg_send![pasteboard, types];
+            let count: usize = msg_send![data_types, count];
+
+            for i in 0..=count {
```
raphlinus: This seems wrong, I believe we shouldn't include `count` in the iter. That said, tracing the code below, if we get a `nil`, `is_string` will be false for that, so I can see the code silently doing the right thing.

druid-shell/src/window.rs
```diff
@@ -17,6 +17,8 @@
 use std::any::Any;
 use std::ops::Deref;
 
+use crate::clipboard::ClipboardItem;
+//TODO: why is this pub?
```
raphlinus: I think we need to clean this up a bit :/. I don't have a super-clear picture of exactly what these types should be. Seems out of scope for this PR though.

druid-shell/src/windows/application.rs
```diff
@@ -14,10 +14,121 @@
 
 //! Windows implementation of features at the application scope.
 
+use std::ptr;
+
+use winapi::shared::minwindef::FALSE;
+use winapi::shared::minwindef::UINT;
+use winapi::shared::ntdef::{LPWSTR, WCHAR};
+use winapi::shared::winerror::ERROR_SUCCESS;
+use winapi::um::errhandlingapi::*;
```
raphlinus: *Generally* I prefer listing the items actually imported rather than using wildcards, though it can get annoying.

druid-shell/src/windows/application.rs
```diff
@@ -14,10 +14,121 @@
...
+
+    pub fn set_clipboard_contents(item: ClipboardItem) {
+        unsafe {
+            if OpenClipboard(ptr::null_mut()) == FALSE {
+                return;
+            }
+            EmptyClipboard();
+            match item {
+                ClipboardItem::Text(string) => {
+                    let wstr = string.to_wide();
```
raphlinus: This is ok but as I've mentioned elsewhere I'd like to migrate from our util wide methods to those in `wio`.

druid-shell/src/windows/mod.rs
```diff
@@ -1284,6 +1286,11 @@ impl<'a> WinCtx<'a> for WinCtxImpl<'a> {
             .unwrap_or(0);
         TimerToken::new(id)
     }
+
+    fn set_clipboard_contents(&mut self, contents: ClipboardItem) {
+        log::info!("set_clipboard_contents item: {:?}", &contents);
```
raphlinus: Not sure we should be keeping this log, though I certainly see how it's helpful for development.

src/event.rs
```diff
@@ -87,6 +88,8 @@ pub enum Event {
     /// Because of repeat, there may be a number `KeyDown` events before
     /// a corresponding `KeyUp` is sent.
     KeyUp(KeyEvent),
+    /// Called when a paste command is received.
```
raphlinus: Can you briefly explain why this needs to be its own event and not treated as an instance of command? Part of the reason I ask is that it's going to grow variants ("paste without formatting") and so this feels a bit like a layering violation.

cmyr: This was an experiment. `Command` is a funny thing. My gut feeling is that 'command' is the solution for when we need runtime dynamism, and are willing to sacrifice ergonomics. In the case of paste, event flow works and is more explicit.

I'm not sure this is correct, of course.

In the case of 'paste' growing options, I think that's okay; we could have the argument be a `PasteEvent` that has both `ClipboardItem` and `PasteOptions` fields.

### About `Command`s:

`Command` is a bit funny. Per my current understanding, `Command` has two main uses; the first is to allow passing of arbitrary messages around the tree, and the second is to enable certain standard platform actions, usually related to menus.

Here's an example of how this is a bit strange:

When the user selects the "new window" item in a menu, the system sends a `NEW_WINDOW` command up the widget tree. This command doesn't actually create the new window, though, because the menu doesn't know the details of the window the user wants to create; instead the user is expected to receive this command and then submit _another_ command, `CREATE_WINDOW`, which actually has as a payload the window we want to create.

I don't see a good alternative to Commands for the first of these cases, the command we get from the menu. But I _could_ imagine `CREATE_WINDOW` being replaced by some method that we expose on `WinCtx`, or elsewhere.

So this is the general question: what sorts of things should be commands, which should be events, and which should be methods? Is there a good test we can apply to decide?

druid-shell/src/window.rs
cmyr: yea, there's a bunch of related cleanup, I'll add this to the list.

druid-shell/src/windows/application.rs
cmyr: right, I was definitely fumbling in the dark here. Also not totally confident about coping the data between the pointers, although it seems okay?

raphlinus: The copying looks fine to my eyes. You could get away with one fewer copy but I think it's not worth it; better to keep the unsafe block small.

src/event.rs
raphlinus: Hmm. The criterion I was (probably subconsciously) applying is that event is "comes from system" which "create window" unambiguously falls under, while command is "an app concern," which is also related to how open-ended it is. But this might not be the right way to slice it.

I'm willing to accept this PR as "good enough" and try and figure out the details later. It's possible my concerns would be settled by having an extension point in the `Event` enum that could be filled out by the app.

cmyr: So _my_ thinking here is that an event is something that "comes from the framework"; and things like paste are a bit of a gray zone; there are certainly things very similar to paste, like drag and drop, that are unambiguously system events.

As a hiccup for "comes from system": we will want lifecycle events at some point, and while those are related to things happening in the system they're also fairly artificial; I would imagine having an event immediately after a widget is added to the tree, for instance, and that wouldn't be coming from the system at all?

Anyway: this is not to say that this approach is good, just further explication of my thinking.

raphlinus: What you say makes total sense, and by "system" I'm not trying to make a sharp distinction with framework events. For example, we already have "HotChanged" in that category (synthesized by system, happens to be derived from mouse move event from system), and I'm not suggesting moving that out.

Total agreement that drag'n'drop belongs in event. I think we basically agree that "paste" is in the gray zone here, it *could* be modeled purely as an app command, the handling of which would be doing a (synchronous?) query into the platform to retrieve the paste buffer, but it's not necessary to do so.

So basically, yeah, this is fine for now and not urgent to reconsider.

PR #182, opened 2019-09-30
Add procedural macro for deriving Data

ineol:
This PR adds a simple procedural macro to derive the Data trait.

It simply calls the `same` function recursively on each field. It can handle structs (with anonymous fields), and enums (with inline structs).

One thing it does _not_ do is  detect that an enum is C-like, and optimize the `same` implementation to equality.

The code is a bit rough, especially the error messages, but I wanted some feedback.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -170,6 +170,7 @@ dependencies = [
 name = "druid"
 version = "0.3.0"
 dependencies = [
+ "druid-derive-data 0.1.0",
  "druid-shell 0.3.0",
  "fluent-bundle 0.7.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "fluent-locale 0.7.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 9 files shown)


ineol:
Hmmm, strangely enough, the rustfmt check passed on windows but not on mac... I did not change the offending file `src/command.rs`.

cmyr:
Cool, this is exciting! The breakage is due to rust 1.38 just being released, which I guess has a minor change in rustfmt? I've opened #183 for that.

druid-derive-data/Cargo.lock
```diff
@@ -0,0 +1,47 @@
+# This file is automatically @generated by Cargo.
```
cmyr: I don't think we need `Cargo.lock` to be checked in; this is usually only done for binary crates.

ineol: Oups, I did not know!

raphlinus:
> should this be behind a feature, or do we include it always?

I'm ok making the executive decision to always depend on proc macros. Otherwise we risk fragmentation. In druid, we're going to use them for data and lenses, at least, and in Runebender we're going to use them extensively for structured font data.

cmyr:
@ineol you should be able to rebase on origin/master and pass CI now.

cmyr:
Cool, I'm going to merge this now, and I'll open an issue to update the docs. Thanks @ineol!

PR #186, opened 2019-09-30
Add druid-derive-data/Cargo.lock to .gitignore

cmyr:


.gitignore
```diff
@@ -1,4 +1,5 @@
 target
+druid-derive-data/Cargo.lock
```
raphlinus: Is it more idiomatic to put it here, or as a separate .gitignore in the subdir?

cmyr: I'm not sure? I tend to only ever use one `.gitignore`, at the repo root.

PR #191, opened 2019-10-05
Setup ci through github actions

cmyr:
Had to figure out how to do this for my other hacking, so we might as well use it, should be a good speedup.


cmyr:
making an executive decision to merge this.

PR #194, opened 2019-10-06
Fix bloat-cmp workflow

cmyr:
This will now correctly get revisions when started via comment, and
will abort when triggered with identical commits.


cmyr:
I'm going to turn this off for PRs, so it will have to be triggered manually.

PR #200, opened 2019-10-09
Add optimization to derive macro for Data

ineol:
Optimize the case of C-style macros (only tags) to just check for equality instead of matching on both arguments. I checked that the complier does not dot do this optimization itself.

Also some minor cleanup of the code, and an error message if it is used on a `union`.

### DIFF
--- a/druid-derive-data/src/lib.rs
+++ b/druid-derive-data/src/lib.rs
@@ -21,14 +21,19 @@ use syn::{parse_macro_input, spanned::Spanned, Data, DataEnum, DataStruct, Field
 #[proc_macro_derive(Data)]
 pub fn derive(input: TokenStream) -> TokenStream {
     let input = parse_macro_input!(input as syn::DeriveInput);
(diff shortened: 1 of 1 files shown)


ineol:
My reasoning is that we will get the same error message as when one writes `derive(Eq)` without `derive(PartialEq)`, see [this example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7dd0f4b8a8ade026b94214df53643ab4).

As far as I know, this is the only case where using equality is an optimization (?) since the point of `Data` is to be shallower than `Eq`.

raphlinus:
Having thought about it a bit, I think the approach in this PR is probably the best way to go forward. The distinction between Data and PartialEq is subtle, but for no-payload enums it's almost always the case that you want to derive PartialEq, so it seems reasonable to me to special-case it. I definitely understand the appeal of an attribute, but think that's a little less ergonomic.

PR #203, opened 2019-10-09
Implement a simple 'app delegate'

cmyr:
This is an experiment for one possible way to allow more top-level customization. This is motivated by the fact that there's currently no way to modify menus or create and close windows without implementing a custom widget.

The basic idea is to introduce an `AppDelegate` struct, which if used will be called at various points and allowed to handle events and modify state. Initially there is only one customization point, which is the ability to intercept and modify events before they're passed to the rest of the tree.

I can imagine this expanding over time; in particular I think that the app delegate should probably have access to all window state, and might even be the place to implement things like opening and saving files.

This is just a sketch; I'm going to try and use this in Runebender shortly, but I wanted to open a PR for discussion.

### DIFF
--- a/examples/multiwin.rs
+++ b/examples/multiwin.rs
@@ -14,12 +14,11 @@
 
 //! Opening and closing windows and using window and context menus.
 
-use druid::kurbo::Size;
 use druid::menu::{ContextMenu, MenuDesc, MenuItem};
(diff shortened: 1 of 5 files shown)


cmyr:
I've been using this in runebender and I quite like it; in particular it's a nice solid place to put window management code. It's a new pattern, so deserves scrutiny, but I propose merging this.

PR #205, opened 2019-10-10
Removing one unsafe block and denying unsafe code

Duddino:
This PR removes one unsafe block and denies unsafe code, this would make druid 100% safe code.
Denying unsafe code is both cool and describes in one line of code that the project does not contain unsafe code

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -14,7 +14,7 @@
 
 //! Simple data-oriented GUI.
 
-#![deny(intra_doc_link_resolution_failure)]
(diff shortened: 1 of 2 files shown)


raphlinus:
I think a useful heuristic is that, if it needs unsafe, it should be in druid-shell. Because druid is the top-level druid crate and druid-shell, the *actual* safety guarantee is weak (and there is tons of scope for UB to happen because the underlying platform code is so messy), but I think it does help reason about this stuff. So thanks for this.

PR #209, opened 2019-10-18
Allow user to customize the env before app launch

cmyr:
There's currently no way to set custom theme keys/values, or to override existing ones.

This also fixes a few small issues related to themes; `Env::set` had the wrong signature, and we weren't using the theme when setting the background color.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -22,11 +22,15 @@ use crate::shell::window::WindowHandle;
 use crate::shell::{init, runloop, Error as PlatformError, WindowBuilder};
 use crate::win_handler::AppState;
 use crate::window::{Window, WindowId};
-use crate::{theme, Data, DruidHandler, LocalizedString, MenuDesc, Widget};
+use crate::{theme, Data, DruidHandler, Env, LocalizedString, MenuDesc, Widget};
+
+/// A function that modifies the initial environment.
+type EnvSetupFn = dyn FnOnce(&mut Env);
 
 /// Handles initial setup of an application, and starts the runloop.
 pub struct AppLauncher<T> {
     windows: Vec<WindowDesc<T>>,
+    env_setup: Option<Box<EnvSetupFn>>,
 }
 
 /// A function that can create a widget.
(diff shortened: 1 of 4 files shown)


futurepaul:
This all looks good to me. Excited to have this functionality!

Makes me think: should we have some sort of StyleChild widget that accepts a closure like this and sends the mutated env on down to its child?

src/env.rs
```diff
@@ -138,7 +138,7 @@ impl Env {
     ///
     /// Panics if the environment already has a value for the key, but it is
     /// of a different type.
-    pub fn set<'a, V: ValueType<'a>>(mut self, key: Key<V>, value: impl Into<V::Owned>) {
+    pub fn set<'a, V: ValueType<'a>>(&'a mut self, key: Key<V>, value: impl Into<V::Owned>) {
```
raphlinus: Oops :)

src/app.rs
```diff
@@ -22,11 +22,15 @@ use crate::shell::window::WindowHandle;
 use crate::shell::{init, runloop, Error as PlatformError, WindowBuilder};
 use crate::win_handler::AppState;
 use crate::window::{Window, WindowId};
-use crate::{theme, Data, DruidHandler, LocalizedString, MenuDesc, Widget};
+use crate::{theme, Data, DruidHandler, Env, LocalizedString, MenuDesc, Widget};
+
+/// A function that modifies the initial environment.
+type EnvSetupFn = dyn Fn(&mut Env);
```
raphlinus: Would FnOnce be a more precise bound?

raphlinus:
A widget that layers stuff onto the environment would absolutely be appropriate. One of the things that's been holding me back is that it's likely that this would have some performance impact (especially if used very widely). But that's not a good reason not to do it. I think we can figure out the performance stuff later if we get confidence that the patterns are expressive.

src/app.rs
cmyr: Yea I think that's reasonable, I could imagine needing this to be reusable at some point in the future but we can change it back if that happens. :)

src/env.rs
cmyr: 😁

cmyr:
@futurepaul would be happy to see an env-modifying widget. Looking at this stuff again makes me realize there's a whole laundry list of follow-up work and related things to consider: how do we do dark mode, how do we set per-system themes, how do we respect system text size preferences (in thee accessibility sense) etc. etc. Anyway, one thing at a time...

PR #219, opened 2019-10-22
Update deps, address soundness issue in msg_send macro

cmyr:
As described in [rust-objc#62][], the msg_send macro could cause
UB in certain situations when built with a version of rustc that
includes the ! type.

[rust-objc#62]: https://github.com/SSheldon/rust-objc/issues/62

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,10 +2,10 @@
 # It is not intended for manual editing.
 [[package]]
 name = "arrayvec"
-version = "0.4.11"
+version = "0.4.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "nodrop 0.1.13 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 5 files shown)


SSheldon:
@cmyr mentioned over in SSheldon/rust-objc#83:
> In that thread raph makes a good point, which is that reflexively casting unused results to `()` might cause other problems. Is there any room for concern here? I would expect to be able to cast any `id` to a `void*`, but I'm not sure about casting to void. 😕

Definitely not an expert here, but I can say that the objc runtime uses different msg_send primitives on certain architectures between returns for floating point, structs, or other types. These mainly matter so that messaging nil gives a properly zeroed output. But there could be other spooky undefined behavior consequences?

Anyways, although I don't know of any specific problems you'd see, I'd say it is worth giving a quick audit while you're at it to see what the return type should be for the omitted returns. Hopefully not too hard, lots of them are setters which almost always return void in objc.

cmyr:
Thanks @SSheldon!

PR #226, opened 2019-10-23
Add Region and basic minimal drawing

cmyr:
This introduces a `Region` type, and uses it to track the
currently visible region in the PaintCtx.

Widgets can check whether they are inside this region to decide
whether or not they should draw.

----

This uses a closure to generate the child contexts, that is it uses the call-stack as the stack. I tried to just have a fn `child_ctx(new_region: Rect) -> PaintCtx` on `PaintCtx` but was running into some hairy lifetime issues. I'd probably prefer that if it were possible.


src/lib.rs
```diff
@@ -272,6 +272,31 @@ pub struct PaintCtx<'a, 'b: 'a> {
...
+
+impl Region {
+    /// Returns the smallest `Rect` that encloses the entire region.
+    pub fn to_rect(&self) -> Rect {
+        self.0
+    }
+
+    /// Returns `true` if `self` intersects with `other`.
+    #[inline]
+    pub fn intersects(&self, other: Rect) -> bool {
```
raphlinus: I was thinking `impl Into<Region>` but maybe keeping it simpler is better. It's the kind of thing I can imagine wanting specialization when that lands...

src/lib.rs
```diff
@@ -272,6 +272,31 @@ pub struct PaintCtx<'a, 'b: 'a> {
...
+        self.0.intersect(other).area() > 0.
+    }
+}
```
raphlinus: Probably not essential for this PR, but I think implementing Add/Sub<Vec2> would be useful.

Also I think it would be slightly better if `intersect` were on here too, rather than forcing going through `Rect`.

src/lib.rs
```diff
@@ -409,8 +463,19 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
...
         }
-        paint_ctx.transform(Affine::translate(self.state.layout_rect.origin().to_vec2()));
-        self.paint(paint_ctx, data, env);
+        let layout_origin = self.state.layout_rect.origin().to_vec2();
+        paint_ctx.transform(Affine::translate(layout_origin));
+
+        let visible = paint_ctx
+            .region()
+            .to_rect()
+            .intersect(self.state.layout_rect)
```
raphlinus: Hmm. First, we talked about not always culling to the layout rect, as this shouldn't be a hard invariant. Second, if we are intersecting, then this might be a good place to stop the recursion (rather than having the caller check).

Basically what I think we should do here is have two variants, one which does intersect layout rect (and culls recursion), one that just applies `layout_origin`.

src/lib.rs
cmyr: My thinking here was that I would then have to implement intersection for every concrete type that impls `Into<Region>`; but I think we can revisit when we get there?

src/lib.rs
cmyr: What does `intersect` return? Another region? My inclination was to keep this as minimal as possible for now, and then when we want to implement minimal invalidation, we can flesh out exactly what this should look like.

src/lib.rs
cmyr: good point, hadn't thought of that. will take another look.

src/lib.rs
raphlinus: I'm thinking another region, yes. And yeah, we don't need to get too fancy when region is just rect. But I can definitely imagine having more fine-grained regions in the future.

src/lib.rs
cmyr: So thinking about this more, I think the intersection was just a mistake. My intention here was to always make the caller check, if they're interested. I can see the merit of doing the check in the paint call itself, though, I'll experiment with that (as a separate fn)

cmyr: @raphlinus I've updated this to add `paint_with_offset_always`; `paint_with_offset` gets the short-circuiting behaviour. This should mean that most widgets will 'just work', not getting painted if they aren't visible.

I'm not sure what to call the new function. 🤷‍♂

raphlinus: Heh, I think it's ok if it has a somewhat awkward name. We're going to need to rework that section more than a bit when we do fine-grained invalidation.

I'm ok with this going in at this point.

cmyr: I'm surprised you aren't a bit more circumspect about my context closure business.

I would have preferred to have some method on `PaintCtx` that took the new region and returned a new `PaintCtx` but I couldn't figure that out.

PR #231, opened 2019-10-25
Toggle Switch Widget

scholtzan:
I was playing around with widgets a bit after reading [Building a widget for Druid](https://pauljmiller.com/posts/druid-widget-tutorial.html) and implemented a toggle switch:
Screenshot 2019-10-25 at 1 36 12 PM

Any comments are appreciated.

### DIFF
--- a/druid/src/widget/switch.rs
+++ b/druid/src/widget/switch.rs
@@ -21,8 +21,7 @@ use crate::piet::{
 use crate::theme;
 use crate::widget::Align;
 use crate::{
-    BaseState, BoxConstraints, Env, Event, EventCtx, LayoutCtx, LocalizedString, PaintCtx,
-    UpdateCtx, Widget,
+    BaseState, BoxConstraints, Env, Event, EventCtx, LayoutCtx, PaintCtx, UpdateCtx, Widget,
 };
 
 /// A switch that toggles a boolean.
@@ -55,12 +54,8 @@ impl Switch {
         let font_size = env.get(theme::TEXT_SIZE_NORMAL);
         let switch_height = env.get(theme::BORDERED_WIDGET_HEIGHT);
 
-        let label: LocalizedString<&str> = if *data {
-            LocalizedString::new("On")
-        } else {
-            LocalizedString::new("Off")
-        };
-        let localized_label = label.localized_str().to_uppercase();
+        // TODO: use LocalizedString
+        let label = if *data { "ON" } else { "OFF" };
 
         let font = paint_ctx
             .text()
@@ -70,7 +65,7 @@ impl Switch {
 
         let text_layout = paint_ctx
             .text()
-            .new_text_layout(&font, &localized_label)
+            .new_text_layout(&font, label)
             .build()
             .unwrap();
 
--- a/druid/src/widget/switch.rs
+++ b/druid/src/widget/switch.rs
@@ -14,37 +14,32 @@
 
 //! A toggle switch widget.
 
-use crate::kurbo::{Circle, Point, Rect, RoundedRect, Size};
-use crate::piet::{FontBuilder, Text, TextLayout, TextLayoutBuilder};
-use crate::piet::{LinearGradient, RenderContext, UnitPoint};
+use crate::kurbo::{Circle, Point, Rect, RoundedRect, Shape, Size};
+use crate::piet::{
+    FontBuilder, LinearGradient, RenderContext, Text, TextLayout, TextLayoutBuilder, UnitPoint,
+};
 use crate::theme;
 use crate::widget::Align;
 use crate::{
-    BaseState, BoxConstraints, Env, Event, EventCtx, LayoutCtx, PaintCtx, UpdateCtx, Widget,
+    BaseState, BoxConstraints, Env, Event, EventCtx, LayoutCtx, LocalizedString, PaintCtx,
+    UpdateCtx, Widget,
 };
 
-#[derive(Debug, Clone)]
-pub struct Switch;
-
-impl Switch {
-    pub fn new() -> impl Widget<bool> {
-        Align::vertical(UnitPoint::CENTER, SwitchRaw::default())
-    }
-}
-
+/// A switch that toggles a boolean.
 #[derive(Debug, Clone, Default)]
-pub struct SwitchRaw {
+pub struct Switch {
     knob_pos: Point,
     knob_hovered: bool,
 }
 
-impl SwitchRaw {
+impl Switch {
+    pub fn new() -> impl Widget<bool> {
+        Align::vertical(UnitPoint::CENTER, Self::default())
+    }
+
     fn knob_hit_test(&self, knob_width: f64, mouse_pos: Point) -> bool {
         let knob_circle = Circle::new(self.knob_pos, knob_width / 2.);
-        if mouse_pos.distance(knob_circle.center) < knob_circle.radius {
-            return true;
-        }
-        false
+        knob_circle.winding(mouse_pos) > 0
     }
 
     fn paint_label(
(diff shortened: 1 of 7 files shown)


src/widget/switch.rs
```diff
@@ -0,0 +1,242 @@
...
+use crate::kurbo::{Circle, Point, Rect, RoundedRect, Size};
+use crate::piet::{FontBuilder, Text, TextLayout, TextLayoutBuilder};
+use crate::piet::{LinearGradient, RenderContext, UnitPoint};
+use crate::theme;
+use crate::widget::Align;
+use crate::{
+    BaseState, BoxConstraints, Env, Event, EventCtx, LayoutCtx, PaintCtx, UpdateCtx, Widget,
+};
+
+#[derive(Debug, Clone)]
```
futurepaul: We like to include a doc comment on public items.

src/widget/switch.rs
```diff
@@ -0,0 +1,242 @@
...
+            .new_text_layout(&font, label)
+            .unwrap()
+            .build()
+            .unwrap();
+
+        let mut origin = UnitPoint::LEFT.resolve(Rect::from_origin_size(
+            Point::ORIGIN,
+            Size::new(
+                (base_state.size().width - text_layout.width()).max(0.0),
+                base_state.size().height + (font_size * 1.2) / 2.,
```
futurepaul: Currently this widget gets messed up when squished below `BORDERED_WIDGET_HEIGHT`. That's a larger issue we need to deal with generally, but the quick fix here I believe would be to use `BORDERED_WIDGET_HEIGHT` instead of base_state height here.

src/widget/switch.rs
```diff
@@ -0,0 +1,242 @@
...
+        let knob_size = switch_height - 2. * switch_padding;
+
+        let background_rect = RoundedRect::from_origin_size(
+            Point::ORIGIN,
+            Size::new(switch_width, switch_height).to_vec2(),
+            switch_height / 2.,
+        );
+
+        // paint different background for on and off state
+        let background_gradient = if *data {
```
futurepaul: I have an orange "caution" color that none of the widgets have used yet. Could potentially use it here for the off state, but this also looks fine as-is. Orange might be too intense for a typical switch use case? [Here's the figma if you're interested](https://www.figma.com/file/YudE0StDEPdemyIIVdUMzr/Druid-GUI-concept?node-id=0%3A1).

examples/switch.rs
```diff
@@ -0,0 +1,42 @@
...
+
+use druid::widget::{Column, Label, Padding, Row, Switch};
+use druid::{AppLauncher, Data, Lens, LensWrap, Widget, WindowDesc};
+
+#[derive(Clone, Data, Lens)]
+struct DemoState {
+    value: bool,
+}
+
+fn build_widget() -> impl Widget<DemoState> {
```
futurepaul: this might be a bit too minimal of an example. potentially we could put switch inside another example, or expand this example to show off some other complementary aspects of druid. but I defer to @cmyr who has helped keep a handle on this directory.

cmyr: I find that when I'm _using_ sample code, I benefit when examples are more complex, use API in interesting ways, or otherwise show me things I might be less likely to figure out on my own.

So this is okay, but it would also be totally okay to add this to an existing example, to change behaviour in some way.

src/widget/switch.rs
```diff
@@ -0,0 +1,242 @@
...
+pub struct Switch;
+
+impl Switch {
+    pub fn new() -> impl Widget<bool> {
+        Align::vertical(UnitPoint::CENTER, SwitchRaw::default())
+    }
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct SwitchRaw {
```
cmyr: I am not a huge fan of this 'Widget + WidgetRaw' pattern; I don't think it should be the default pick for most widgets.

In this case, your `Switch` statement doesn't have any state; it's really just being used as a marker, and as somewhere to hang the `new` method. I would delete SwitchRaw` and just put that state in `Switch` itself, and make it the widget.

src/widget/switch.rs
```diff
@@ -0,0 +1,242 @@
...
+    knob_pos: Point,
+    knob_hovered: bool,
+}
+
+impl SwitchRaw {
+    fn knob_hit_test(&self, knob_width: f64, mouse_pos: Point) -> bool {
+        let knob_circle = Circle::new(self.knob_pos, knob_width / 2.);
+        if mouse_pos.distance(knob_circle.center) < knob_circle.radius {
```
cmyr: the less obvious but maybe canonical way to do this might be `if knob_circle.winding(mouse_pos) > 0 { .. ` but personally I think that the `Shape` trait should just have a `contains(pt: Point)` type method that calls `winding` under the hood; I think it would be a much simpler API to discover and to use, in many cases.

src/widget/switch.rs
```diff
@@ -0,0 +1,242 @@
...
+        base_state: &BaseState,
+        data: &bool,
+        env: &Env,
+        switch_width: f64,
+        switch_padding: f64,
+    ) {
+        let font_name = env.get(theme::FONT_NAME);
+        let font_size = env.get(theme::TEXT_SIZE_NORMAL);
+
+        let label = if *data { "ON" } else { "OFF" };
```
cmyr: Since these are part of the core UI, these should be `LocalizedString`s.

You can see elsewhere where these are used. You'll also need to add the actual strings to the two localization strings files in `resources` (If you're bored you could even add a `de-DE` locale).

src/widget/switch.rs
cmyr: I think that as future work color should be configurable.

src/widget/switch.rs
```diff
@@ -0,0 +1,242 @@
...
+        match event {
+            Event::MouseDown(_) => {
+                ctx.set_active(true);
+                ctx.invalidate();
+            }
+            Event::MouseUp(_) => {
+                if ctx.is_active() {
+                    ctx.set_active(false);
+                    if ctx.is_hot() {
+                        if *data {
```
cmyr: you could write this as just, `*data = !*data`, maybe?

src/widget/switch.rs
```diff
@@ -0,0 +1,242 @@
...
+                    }
+                    ctx.invalidate();
+                }
+            }
+            Event::MouseMoved(mouse) => {
+                if ctx.is_active() {
+                    // todo: animate dragging of knob
+                }
+                if ctx.is_hot() {
+                    if self.knob_hit_test(knob_size, mouse.pos) {
```
cmyr: again, I think I find `self.knob_hovered = self.knob_hit_test` clearer here?

examples/switch.rs
scholtzan: I couldn't think of how existing examples could be extended in a useful way right now. But maybe there will be some better examples in the future that could replace this one.

druid/src/widget/switch.rs
```diff
@@ -0,0 +1,234 @@
...
+        data: &bool,
+        env: &Env,
+        switch_width: f64,
+        switch_padding: f64,
+    ) {
+        let font_name = env.get(theme::FONT_NAME);
+        let font_size = env.get(theme::TEXT_SIZE_NORMAL);
+        let switch_height = env.get(theme::BORDERED_WIDGET_HEIGHT);
+
+        let label: LocalizedString<&str> = if *data {
```
cmyr: okay, I appreciate the effort but this won't really work. I really need to do a write up about how localization is supposed to function; in the meantime I think you should go back to just using a normal string and add a `TODO` that this should be localized at some point in the future. :)

scholtzan: Okay, done :)

cmyr:
Feel free to merge this @scholtzan!

scholtzan:
I don't have write access to the repo

cmyr:
oops, you do now!

Issue #167, opened 2019-09-18
track active window


cmyr:
As part of #156 I included a new method to WinHandler, `got_focus()`. This should be called whenever focus changes to a different window. This is implemented in mac but needs to be implemented for windows.

Maan2003:
This is implemented for gtk in #1404

PR #235, opened 2019-10-27
Implement window tracking for Windows

LiHRaM:
I've added handling of the `WM_SETFOCUS` message for `WndProc`, which calls `got_focus` whenever the message is received.

See #167 for the motivation.

Note: Do we want to handle `WM_KILLFOCUS` as well? Might be relevant to know when none of the windows is focused.

### DIFF
--- a/druid-shell/src/windows/mod.rs
+++ b/druid-shell/src/windows/mod.rs
@@ -333,6 +333,16 @@ impl WndProc for MyWndProc {
         //println!("wndproc msg: {}", msg);
         match msg {
             WM_ERASEBKGND => Some(0),
+            WM_SETFOCUS => {
+                if let Ok(mut s) = self.state.try_borrow_mut() {
+                    let s = s.as_mut().unwrap();
+                    let mut c = WinCtxOwner::new(self.handle.borrow(), &self.dwrite_factory);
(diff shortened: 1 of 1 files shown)


cmyr:
I think we will want something like `lost_focus` at some point, but I think it's okay to leave that until we know what we're going to do with it. (I think maybe these window events should be handled by a delegate, instead of as an event; they aren't necessarily relevant to the whole widget tree.)

LiHRaM:
> it's okay to leave that until we know what we're going to do with it

Fair enough. 😄

> these window events should be handled by a delegate

You mean like in #203? Is that something you want as a part of this PR or a followup?

cmyr:
Okay so I think before we worry about focus, we should be doing a better job of tracking when windows are opened and closed. If you're interested in a follow up, that might be a good place to start; it's something that clearly belongs in appdelegate (I think), and will inform whether focus belongs there to.

I can open an issue for this, but basically what we want is to have additional closures on `AppDelegate` that will be called when a new window is created, or a window is destroyed. This message should include the window's id.

we'll also want to give `DelegateCtx` access to all windows.

LiHRaM:
Alright, thanks! I am interested in following up on what we've discussed here, I'll see if I can get something working within the next few days.

PR #236, opened 2019-10-28
Update kurbo (and other deps)

cmyr:
This is mostly about updating kurbo (so we can use the insets stuff)
but I bumped everything else at the same time.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -214,7 +214,7 @@ version = "0.1.0"
 dependencies = [
  "proc-macro2 1.0.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "quote 1.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "syn 1.0.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "syn 1.0.7 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


cmyr:
I was thinking about this; the approach you describe was initially what I'd thought, but now I'm not sure.

Basically: clients will ignore druid's `Cargo.lock` when it is used as a library. Since we don't specify any version of `kurbo` in druid, the only requirement that exists is what is declared in piet; currently that's `kurbo = 0.5.1`, which means `kurbo >= 0.5.1 < 0.6`. 

So the only failure case here that I can identify is if an app is using an older version of druid, then manually repins druid, but does not run `cargo update`. A solution to this would be to explicitly make kurbo `0.5.3` a requirement of druid-shell; since that is compatible with the version spec in `piet`, we would still only have one version of `kurbo`, but it would always have to be `>= 0.5.3`.

I think I've rubber-ducked myself into making this change.

cmyr:
(I would really like there to be a way to get kurbo patches into druid without needing to do a whole piet release, and I think this gets us there.)

raphlinus:
Yes, having druid-shell depend on 0.5.3 would also satisfy the correctness concern. I think it might make sense to remove that dep when we rev piet - it seems to me a goal is to DRY the statement of dependencies. But we can cross that bridge when we get to it.

cmyr:
agreed.

PR #237, opened 2019-10-28
Padding with insets

cmyr:
This adds a new constructor to the `Padding` widget to take advantage fo the new `Insets` type in kurbo.

Question: should this just be called `Padding::new`? Should we get rid of `Padding::uniform`, which can be well handled by a single `impl Into<Insets>` argument?

edit: this is based off of #236, which should go in first

### DIFF
--- a/src/widget/padding.rs
+++ b/src/widget/padding.rs
@@ -14,6 +14,7 @@
 
 //! A widget that just adds padding during layout.
 
+use crate::kurbo::Insets;
 use crate::{
     BaseState, BoxConstraints, Data, Env, Event, EventCtx, LayoutCtx, PaintCtx, Point, Rect, Size,
     UpdateCtx, Widget, WidgetPod,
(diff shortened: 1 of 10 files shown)


src/widget/padding.rs
```diff
@@ -40,6 +41,33 @@ impl<T: Data> Padding<T> {
...
+    /// use druid::kurbo::Insets;
+    ///
+    /// let child: Label<()> = Label::new("I need my space!");
+    /// let padding = Padding::with_insets((10.0, 20.0), child);
+    /// // equivalent:
+    /// let moar_padding = Padding::with_insets(Insets::uniform_xy(10.0, 20.0), padding);
+    /// ```
+    ///
+    /// [`kurbo::Insets`]: https://docs.rs/kurbo/0.5.3/kurbo/struct.Insets.html
+    pub fn with_insets(insets: impl Into<Insets>, child: impl Widget<T> + 'static) -> Padding<T> {
```
raphlinus: This makes obvious sense. I'd even be ok with this being called `new`. It should work with a scalar as well, because of the into conversion, no? That's one of the things that makes me think this could be the designated constructor of choice.

futurepaul:
This isn't a new problem of this PR, but it's pretty easy for padding to pass along negative BoxConstraints, especially with the subtraction happening in here. Not sure if this is where we want to take a stand, just wanted to give a heads up.

raphlinus:
Agreed. This can happen in Flex when the non-flex widgets exceed the constraint as well.

src/widget/padding.rs
cmyr: Yea, I think that makes sense. Will update.

Issue #238, opened 2019-10-28
Padding widget can pass negative BoxConstraints


futurepaul:
It's possible to construct invalid BoxConstraints, and Padding can easily do this because it's doing subtraction. An easy fix to Padding is to make sure it never passes a minimum or maximum constraint that's less than zero.

What I'm unsure about: what is the correct behavior if Padding is given a max height of 10, plus a uniform padding of 7? Do you collapse the size of the padding or the child first?

Also, Padding isn't the only culprit irt negative BoxConstraints, just thought it would be a good place to start fighting back.

futurepaul:
Closed by #239

PR #239, opened 2019-10-28
Fix up BoxConstraints from Padding and Flex

futurepaul:
I forgot I had these fixes for Flex leftover. I also did the simplest thing possible to fix Padding (#238).

Additionally, I've added bc.check to the top of every widget's layout method, which should only be called if we're in log level Warn. I ran all the examples and after these changes to Flex and Padding I get no failures of bc.check() when sizing the window. That doesn't mean we're perfect now, but it's a big improvement!

### DIFF
--- a/src/widget/flex.rs
+++ b/src/widget/flex.rs
@@ -141,9 +141,7 @@ impl<T: Data> Widget<T> for Flex<T> {
         data: &T,
         env: &Env,
     ) -> Size {
-        if log::log_enabled!(log::Level::Warn) {
-            bc.check("Flex");
-        }
+        bc.debug_check("Flex");
 
         // Measure non-flex children.
         let mut total_non_flex = 0.0;
@@ -178,11 +176,7 @@ impl<T: Data> Widget<T> for Flex<T> {
             if child.params.flex != 0.0 {
                 let major = remaining * child.params.flex / flex_sum;
 
-                let min_major = if major == std::f64::INFINITY {
-                    0.0
-                } else {
-                    major
-                };
+                let min_major = if major.is_infinite() { 0.0 } else { major };
 
                 let child_bc = match self.direction {
                     Axis::Horizontal => BoxConstraints::new(
@@ -212,10 +206,8 @@ impl<T: Data> Widget<T> for Flex<T> {
             major += self.direction.major(rect.size());
         }
 
-        if log::log_enabled!(log::Level::Warn) {
-            if flex_sum > 0.0 && total_major == std::f64::INFINITY {
-                log::warn!("A child of Flex is flex, but Flex is unbounded.")
-            }
+        if flex_sum > 0.0 && total_major.is_infinite() {
+            log::warn!("A child of Flex is flex, but Flex is unbounded.")
         }
 
         if flex_sum > 0.0 {
--- a/src/widget/flex.rs
+++ b/src/widget/flex.rs
@@ -141,19 +141,23 @@ impl<T: Data> Widget<T> for Flex<T> {
         data: &T,
         env: &Env,
     ) -> Size {
+        if log::log_enabled!(log::Level::Warn) {
+            bc.check("Flex");
+        }
+
         // Measure non-flex children.
         let mut total_non_flex = 0.0;
-        let mut minor = 0.0f64;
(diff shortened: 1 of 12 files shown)


cmyr:
I like the check idea, one of the things we're going to need to put some energy into is exactly what our debugging story is.

src/widget/flex.rs
```diff
@@ -165,21 +169,28 @@ impl<T: Data> Widget<T> for Flex<T> {
...
+        let total_major = self.direction.major(bc.max());
+        let remaining = (total_major - total_non_flex).max(0.0);
         let flex_sum: f64 = self.children.iter().map(|child| child.params.flex).sum();
 
         // Measure flex children.
         for child in &mut self.children {
             if child.params.flex != 0.0 {
                 let major = remaining * child.params.flex / flex_sum;
+
+                let min_major = if major == std::f64::INFINITY {
```
cmyr: you can also write, `if major.is_infinite() { .. }`.

I might rewrite this as,

```rust
let major = if child.params.flex.is_infinite() {
    0.0 
} else {
    remaining * child.params.flex / flex_sum
};
```

but that's just for the sake of conversation, this certainly works and is clear.

src/widget/flex.rs
```diff
@@ -200,9 +211,17 @@ impl<T: Data> Widget<T> for Flex<T> {
             child.widget.set_layout_rect(rect.with_origin(pos));
             major += self.direction.major(rect.size());
         }
+
+        if log::log_enabled!(log::Level::Warn) {
```
cmyr: I appreciate these sorts of diagnostics.

I don't know if it's worth doing the `log_enabled!` check, here; the comparison below is very cheap, but 🤷‍♂

src/widget/radio.rs
```diff
@@ -111,6 +111,10 @@ impl<T: Data + PartialEq> Widget<T> for RadioRaw<T> {
         data: &T,
         env: &Env,
     ) -> Size {
+        if log::log_enabled!(log::Level::Warn) {
```
cmyr: I like these, but I think I would consider trying to make it more concise. Some things to consider:

- is `log_enabled` necessary? how expensive is the check?
- can we do the `log_enabled` check _inside_ the `check()` fn? Maybe we can mark that fn as `#[inline]`?
- is `Warn` the correct level? I might expect `Warn` to be turned on in release builds. If that's the case, it isn't totally clear in what situation this check will fail.

src/widget/flex.rs
```diff
@@ -141,19 +141,23 @@ impl<T: Data> Widget<T> for Flex<T> {
...
         env: &Env,
     ) -> Size {
+        if log::log_enabled!(log::Level::Warn) {
+            bc.check("Flex");
+        }
+
         // Measure non-flex children.
         let mut total_non_flex = 0.0;
-        let mut minor = 0.0f64;
+        let mut minor = self.direction.minor(bc.min());
```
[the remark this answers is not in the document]
raphlinus: Yes. This would return 0 instead of minimum in the case of an empty container.

src/widget/flex.rs
```diff
@@ -165,21 +169,28 @@ impl<T: Data> Widget<T> for Flex<T> {
             }
         }
 
-        let total_major = self.direction.major(bc.max);
-        let remaining = total_major - total_non_flex;
+        let total_major = self.direction.major(bc.max());
+        let remaining = (total_major - total_non_flex).max(0.0);
```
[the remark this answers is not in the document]
raphlinus: It would create constraints of negative size. Often these would go through a `Rect` constructor that would take `abs`, so you'd get this sorta "bounce" effect.

src/widget/radio.rs
cmyr: well logs aren't ever really going to be shown a typical user, because the actual user will be using a packaged platform build, with no console; if logging is enabled it will be writing log files only for the purpose of, say, debugging a crash.

the usefulness of `log_enabled!` is that you can use it to do some expensive calculation that you don't want to do if it isn't necessary.

So what I would probably do is rename `check` to `debug_check` or something, to make it clear that it's a diagnostic, and just put it everywhere without worrying too much about the cost, assuming it isn't doing a bunch of expensive work?

PR #240, opened 2019-10-29
Allow interacting with scrollbars via mouse pointer

ForLoveOfCats:
Closes https://github.com/xi-editor/druid/issues/220
This handles showing the scrollbars on mouse over, not hiding until mouse out, and dragging the bars to scroll. 

In addition to the usual code quality review there is a strange behavior I'd like input on. At smaller window sizes dragging the scrollbar appears to be floaty for some reason (especially on `scroll_colors` but less so on `scroll` for some reason?). I cannot at this time figure out why this is. It does not seem to be the result of the whole number mouse positions as the innacuracy is much larger than a single pixel and swapping out the calls to `scroll` for directly setting to `scroll_offset` does not affect the issue.

Furthermore is there a way to track mouse movement and Up/Down when outside the window bounds? That is a situation where the current logic breaks down slightly (but not significantly).

### DIFF
--- a/druid/src/widget/scroll.rs
+++ b/druid/src/widget/scroll.rs
@@ -305,17 +305,21 @@ impl<T: Data, W: Widget<T>> Widget<T> for Scroll<T, W> {
             match event {
                 Event::MouseMoved(event) => {
                     if self.scroll_bars.vertical_held {
+                        let scale_y = viewport.height() / self.child_size.height;
                         let bounds = self.calc_vertical_bar_bounds(&viewport);
                         let mouse_y = event.pos.y + self.scroll_offset.y;
+                        let delta = mouse_y - bounds.y0 - self.scroll_bars.held_offset;
                         self.scroll(
-                            Vec2::new(0f64, mouse_y - bounds.y0 - self.scroll_bars.held_offset),
+                            Vec2::new(0f64, (delta / scale_y).ceil()),
                             size,
                         );
                     } else if self.scroll_bars.horizontal_held {
+                        let scale_x = viewport.width() / self.child_size.width;
                         let bounds = self.calc_horizontal_bar_bounds(&viewport);
                         let mouse_x = event.pos.x + self.scroll_offset.x;
+                        let delta = mouse_x - bounds.x0 - self.scroll_bars.held_offset;
                         self.scroll(
-                            Vec2::new(mouse_x - bounds.x0 - self.scroll_bars.held_offset, 0f64),
+                            Vec2::new((delta / scale_x).ceil(), 0f64),
                             size,
                         );
                     }
--- a/druid/src/widget/scroll.rs
+++ b/druid/src/widget/scroll.rs
@@ -157,7 +157,7 @@ impl<T: Data, W: Widget<T>> Scroll<T, W> {
         self.scroll_offset
     }
 
-    fn calc_scollbar_bounds(&self, viewport: &Rect) -> Rect {
+    fn calc_scrollbar_screen_edges(&self, viewport: &Rect) -> Rect {
         return Rect::new(
             self.scroll_offset.x + SCROLL_BAR_PAD,
             self.scroll_offset.y + SCROLL_BAR_PAD,
@@ -167,7 +167,7 @@ impl<T: Data, W: Widget<T>> Scroll<T, W> {
     }
 
     fn calc_vertical_bar_bounds(&self, viewport: &Rect) -> Rect {
-        let scrollbar_bounds = self.calc_scollbar_bounds(viewport);
+        let scrollbar_bounds = self.calc_scrollbar_screen_edges(viewport);
 
         let scale_y = viewport.height() / self.child_size.height;
 
@@ -184,7 +184,7 @@ impl<T: Data, W: Widget<T>> Scroll<T, W> {
     }
 
     fn calc_horizontal_bar_bounds(&self, viewport: &Rect) -> Rect {
-        let scrollbar_bounds = self.calc_scollbar_bounds(viewport);
+        let scrollbar_bounds = self.calc_scrollbar_screen_edges(viewport);
 
         let scale_x = viewport.width() / self.child_size.width;
 
--- a/druid/src/widget/scroll.rs
+++ b/druid/src/widget/scroll.rs
@@ -169,14 +169,9 @@ impl<T: Data, W: Widget<T>> Scroll<T, W> {
     fn calc_vertical_bar_bounds(&self, viewport: &Rect) -> Rect {
         let scrollbar_bounds = self.calc_scollbar_bounds(viewport);
 
-        let content_size = Size::new(
-            viewport.width() - 2.0 * SCROLL_BAR_PAD,
-            viewport.height() - 2.0 * SCROLL_BAR_PAD,
-        );
+        let scale_y = viewport.height() / self.child_size.height;
 
-        let scale_y = content_size.height / self.child_size.height;
-
-        let h = (scale_y * content_size.height).ceil();
+        let h = (scale_y * viewport.height()).ceil();
         let dh = (scale_y * self.scroll_offset.y).ceil();
 
         let x0 = scrollbar_bounds.x1 - SCROLL_BAR_WIDTH;
@@ -191,14 +186,9 @@ impl<T: Data, W: Widget<T>> Scroll<T, W> {
     fn calc_horizontal_bar_bounds(&self, viewport: &Rect) -> Rect {
         let scrollbar_bounds = self.calc_scollbar_bounds(viewport);
 
-        let content_size = Size::new(
-            viewport.width() - 2.0 * SCROLL_BAR_PAD,
-            viewport.height() - 2.0 * SCROLL_BAR_PAD,
-        );
+        let scale_x = viewport.width() / self.child_size.width;
 
-        let scale_x = content_size.width / self.child_size.width;
-
-        let w = (scale_x * content_size.width).ceil();
+        let w = (scale_x * viewport.width()).ceil();
         let dw = (scale_x * self.scroll_offset.x).ceil();
 
         let x0 = scrollbar_bounds.x0 + dw;
--- a/druid/src/widget/scroll.rs
+++ b/druid/src/widget/scroll.rs
@@ -55,14 +55,14 @@ impl ScrollDirection {
 
 enum BarHoveredState {
     None,
-    VerticalHovered,
-    HorizontalHovered,
+    Vertical,
+    Horizontal,
 }
 
 impl BarHoveredState {
     fn is_hovered(&self) -> bool {
         match self {
-            BarHoveredState::VerticalHovered | BarHoveredState::HorizontalHovered => true,
+            BarHoveredState::Vertical | BarHoveredState::Horizontal => true,
             _ => false,
         }
     }
@@ -72,10 +72,10 @@ enum BarHeldState {
     None,
     /// Vertical scrollbar is being dragged. Contains an `f64` with
     /// the initial y-offset of the dragging input
-    VerticalHeld(f64),
+    Vertical(f64),
     /// Horizontal scrollbar is being dragged. Contains an `f64` with
     /// the initial x-offset of the dragging input
-    HorizontalHeld(f64),
+    Horizontal(f64),
 }
 
 struct ScrollBarsState {
@@ -309,7 +309,7 @@ impl<T: Data, W: Widget<T>> Widget<T> for Scroll<T, W> {
 
         if match event {
             Event::MouseMoved(_) | Event::MouseUp(_) => match self.scroll_bars.held {
-                BarHeldState::VerticalHeld(_) | BarHeldState::HorizontalHeld(_) => true,
+                BarHeldState::Vertical(_) | BarHeldState::Horizontal(_) => true,
                 _ => false,
             },
             _ => false,
@@ -317,14 +317,14 @@ impl<T: Data, W: Widget<T>> Widget<T> for Scroll<T, W> {
             match event {
                 Event::MouseMoved(event) => {
                     match self.scroll_bars.held {
-                        BarHeldState::VerticalHeld(offset) => {
+                        BarHeldState::Vertical(offset) => {
                             let scale_y = viewport.height() / self.child_size.height;
                             let bounds = self.calc_vertical_bar_bounds(viewport);
                             let mouse_y = event.pos.y + self.scroll_offset.y;
                             let delta = mouse_y - bounds.y0 - offset;
                             self.scroll(Vec2::new(0f64, (delta / scale_y).ceil()), size);
                         }
-                        BarHeldState::HorizontalHeld(offset) => {
+                        BarHeldState::Horizontal(offset) => {
                             let scale_x = viewport.width() / self.child_size.width;
                             let bounds = self.calc_horizontal_bar_bounds(viewport);
                             let mouse_x = event.pos.x + self.scroll_offset.x;
@@ -354,9 +354,9 @@ impl<T: Data, W: Widget<T>> Widget<T> for Scroll<T, W> {
                     let mut transformed_event = event.clone();
                     transformed_event.pos += self.scroll_offset;
                     if self.point_hits_vertical_bar(viewport, transformed_event.pos) {
-                        self.scroll_bars.hovered = BarHoveredState::VerticalHovered;
+                        self.scroll_bars.hovered = BarHoveredState::Vertical;
                     } else {
-                        self.scroll_bars.hovered = BarHoveredState::HorizontalHovered;
+                        self.scroll_bars.hovered = BarHoveredState::Horizontal;
                     }
 
                     self.scroll_bars.opacity = 0.7;
@@ -367,11 +367,11 @@ impl<T: Data, W: Widget<T>> Widget<T> for Scroll<T, W> {
                     let pos = event.pos + self.scroll_offset;
 
                     if self.point_hits_vertical_bar(viewport, pos) {
-                        self.scroll_bars.held = BarHeldState::VerticalHeld(
+                        self.scroll_bars.held = BarHeldState::Vertical(
                             pos.y - self.calc_vertical_bar_bounds(viewport).y0,
                         );
                     } else if self.point_hits_horizontal_bar(viewport, pos) {
-                        self.scroll_bars.held = BarHeldState::HorizontalHeld(
+                        self.scroll_bars.held = BarHeldState::Horizontal(
                             pos.x - self.calc_horizontal_bar_bounds(viewport).x0,
                         );
                     }
--- a/druid/src/widget/scroll.rs
+++ b/druid/src/widget/scroll.rs
@@ -56,6 +56,7 @@ impl ScrollDirection {
 struct ScrollBarsState {
     opacity: f64,
     timer_id: TimerToken,
+    hovered: bool,
 }
 
 impl Default for ScrollBarsState {
@@ -63,6 +64,7 @@ impl Default for ScrollBarsState {
         Self {
             opacity: 0.0,
             timer_id: TimerToken::INVALID,
+            hovered: false,
         }
     }
 }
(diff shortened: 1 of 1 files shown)


cmyr:
Not a thorough review, but I played around a little bit:

- it would be great if a hovered scrollbar was indicated visually. Maybe it's drawn more opaque?
- your bug doesn't seem to be about smaller window sizes as much as it's about small scrollbars? That is, if I increase the number of cells in `scroll_colors` I can see this issue even at large window sizes. Not sure what's going on, though
- mouse movement outside of window bounds is a good question. These _should_ still be delivered to the active widget, I think; not sure what's going on there but maybe open an issue?

ForLoveOfCats:
I was thinking about having the bars be more opaque when hovered but decided to add that after submitting a PR adding the fade out time, rounded edge amount, and max opacity, ect to the theme instead of magic numbers.

So the scrollbar size affects the bug... that explains some things. I'll have to play around with it some more. Any input on this issue is welcome.

Tracking mouse events outside the window bounds should already work? I've been working on this on Linux so I guess that feature should be checked on macOS and Windows. I'll check on Windows in a bit, will open a bug for it based on what I find.

ForLoveOfCats:
I've just fixed the floaty scrollbar dragging bug. I think that this is ready for code quality review!

ForLoveOfCats:
Any update on this review? Any changes I need to make? I have some more stuff I want to work on which is blocked by this.

cmyr:
@ForLoveOfCats oops sorry, just got lost in the stack. Will take a look tomorrow!

src/widget/scroll.rs
```diff
@@ -56,13 +56,21 @@ impl ScrollDirection {
 struct ScrollBarsState {
     opacity: f64,
     timer_id: TimerToken,
+    hovered: bool,
+    vertical_held: bool,
+    horizontal_held: bool,
+    held_offset: f64,
```
cmyr: The thing that jumps out at me in the PR are these four items.

I have a fairly abstract concern/curiosity: is this the best way to encode the state that we're trying to encode? I might be mistaken, but it seems like a bunch of these states are mutually exclusive. That is, it seems like you are _either_ `hovered` _or_ `vertical_held` _or_ `horizontal_held`. And it seems like `held_offset` refers to an offset in the `y` direction *if* `horizontal_held` is `true`, and an offset in the `x` direction _if_ `vertical_held` is `true`; and it's an invariant that only one of these can be true at a time. (I think; I could be wrong!)

Anyway, given these assumptions, I could imagine replacing these four variables with an enum, that might look something like:

```rust
/// The state of the mouse with respect to the scrollbars
enum Mouse {
    /// The mouse is not interacting with a scrollbar
    None,
    /// the mouse is over the vertical scrollbar
    HoverVertical,
    /// the mouse is over the horizontal scrollbar
    HoverHorizontal,
    /// The mouse is dragging the vertical scrollbar. The `f64` field
    /// is the y-offset where the gesture began.
    DownVertical(f64),
    /// The mouse is dragging the horizontal scrollbar. The `f64` field
    /// is the x-offset where the gesture began.
    DownHorizontal(f64),
}
```

With something like this, you could have a single value that represented the current state, and your handling of events could just be a matter of transitioning between states and acting accordingly. As a bonus, this makes illegal states impossible to create; you cannot have `held_vertical` and `held_horizontal` both be true, `held_offset` can't have a value unless something is held, etcetera.

You can also add some simple methods to this enum, like

```rust
impl Mouse {
    fn is_held(self) -> bool {
        match self {
            Mouse::DownVertical(_) | Mouse::DownHorizontal(_) => true,
            _ => false,
        }
    }
}
```

And so it can still act like a `bool` where that makes sense.


Anyway: this is just a sketch, and it may not correctly model the state that you've written, but I suspect that there is _some_ model like this that will be more robust, more maintainable, and more ergonomic than juggling a bunch of individual values.

src/widget/scroll.rs
```diff
@@ -153,57 +204,38 @@ impl<T: Data, W: Widget<T>> Scroll<T, W> {
...
+            let rect = RoundedRect::from_rect(self.calc_vertical_bar_bounds(viewport), 5.0);
             paint_ctx.render_ctx.fill(rect, &brush);
             paint_ctx.render_ctx.stroke(rect, &border_brush, 1.0);
         }
 
         // Horizontal bar
         if viewport.width() < self.child_size.width {
-            let w = (scale.x * content_size.width).ceil();
-            let dw = (scale.x * self.scroll_offset.x).ceil();
+            let rect = RoundedRect::from_rect(self.calc_horizontal_bar_bounds(viewport), 5.0);
```
cmyr: Guilty of this myself all the time, but this `5.0` probably deserves to be `const SCROLLBAR_CORNER_RADIUS: f64 = 5.0` somewhere.

src/widget/scroll.rs
```diff
@@ -153,57 +204,38 @@ impl<T: Data, W: Widget<T>> Scroll<T, W> {
...
-            let y0 = scroll_bar_bounds.y1;
+    fn mouse_over_vertical_bar(&self, viewport: Rect, pos: Point) -> bool {
+        if viewport.height() < self.child_size.height {
+            let bounds = self.calc_vertical_bar_bounds(&viewport);
+            return pos.y > bounds.y0 && pos.y < bounds.y1 && pos.x > bounds.x0;
+        }
 
-            let x1 = (x0 + w).min(scroll_bar_bounds.x1);
-            let y1 = y0 - bar_thickness;
+        return false;
```
cmyr: rust style is to skip `return` at the end of a function; the last expression in a function is returned automatically. (that is, you can replace `return false;` with just `false`).

src/widget/scroll.rs
```diff
@@ -139,6 +147,49 @@ impl<T: Data, W: Widget<T>> Scroll<T, W> {
         self.scroll_offset
     }
 
+    fn calc_scrollbar_screen_edges(&self, viewport: &Rect) -> Rect {
```
cmyr: total nit: you pass `viewport` by reference here, but by value elsewhere. I'd probably just always pass it by value?

src/widget/scroll.rs
```diff
@@ -139,6 +147,49 @@ impl<T: Data, W: Widget<T>> Scroll<T, W> {
...
+        return Rect::new(
+            self.scroll_offset.x + SCROLL_BAR_PAD,
+            self.scroll_offset.y + SCROLL_BAR_PAD,
+            self.scroll_offset.x - SCROLL_BAR_PAD + viewport.width(),
+            self.scroll_offset.y - SCROLL_BAR_PAD + viewport.height(),
+        );
```
cmyr: there is now an [`Insets`] type in kurbo that was intended to be useful in situations like this. (I'm not sure it would be useful here, but it might be worth checking out?)

[`Insets`]: https://docs.rs/kurbo/0.5.4/kurbo/struct.Insets.html

src/widget/scroll.rs
```diff
@@ -153,57 +204,38 @@ impl<T: Data, W: Widget<T>> Scroll<T, W> {
...
-            let dw = (scale.x * self.scroll_offset.x).ceil();
+            let rect = RoundedRect::from_rect(self.calc_horizontal_bar_bounds(viewport), 5.0);
+            paint_ctx.render_ctx.fill(rect, &brush);
+            paint_ctx.render_ctx.stroke(rect, &border_brush, 1.0);
+        }
+    }
 
-            let x0 = scroll_bar_bounds.x0 + dw;
-            let y0 = scroll_bar_bounds.y1;
+    fn mouse_over_vertical_bar(&self, viewport: Rect, pos: Point) -> bool {
```
cmyr: total nit: this name sounds slightly imperative. I might prefer `hit_test_vertical_bar` or `point_hits_vertical_bar`? but not a blocker just an observation.

src/widget/scroll.rs
```diff
@@ -153,57 +204,38 @@ impl<T: Data, W: Widget<T>> Scroll<T, W> {
...
-            let rect = RoundedRect::new(x0, y0, x1, y1, 5.0);
-            paint_ctx.render_ctx.fill(rect, &brush);
-            paint_ctx.render_ctx.stroke(rect, &border_brush, 1.0);
+    fn mouse_over_horizontal_bar(&self, viewport: Rect, pos: Point) -> bool {
+        if viewport.width() < self.child_size.width {
+            let bounds = self.calc_horizontal_bar_bounds(&viewport);
+            return pos.x > bounds.x0 && pos.x < bounds.x1 && pos.y > bounds.y0;
         }
+
+        return false;
```
cmyr: ditto

src/widget/scroll.rs
```diff
@@ -239,30 +271,104 @@ impl<T: Data, W: Widget<T>> Widget<T> for Scroll<T, W> {
...
-                let diff = 2.0 * (*interval as f64) * 1e-9;
-                self.scroll_bars.opacity -= diff;
-                if self.scroll_bars.opacity > 0.0 {
-                    ctx.request_anim_frame();
+
+        if match event {
+            Event::MouseMoved(_) | Event::MouseUp(_) => {
+                self.scroll_bars.vertical_held || self.scroll_bars.horizontal_held
+            }
+            _ => false,
```
cmyr: one of the powers of rust match statements is that you can include conditionals in each arm; arms are evaluated in order, and the first matching arm is followed. So you can match the same item with different conditions, which should let you clean this up quite a bit, like:

```rust
match event {
    Event::MouseMoved(_) if self.scroll_bars.vertical_held || self.scroll_bars.horizontal_held => ...,
    Event::MouseUp(_) => if self.scroll_bars.vertical_held || self.scroll_bars.horizontal_held => ...,
    Event::MouseMoved(_) => ... // only if we didn't match above,
    Event::MouseUp(_) => ... // only if we didn't match above,
    _ => (), // all other events
}

File: src/widget/scroll.rs (hunk 2 above)
Comment (ForLoveOfCats) [reply src/widget/scroll.rs]: The magic number here was inherited, as previously stated I plan on a separate PR adding the rounded edge amount, fade out time, and max opacity, ect to the theme (or something else, I don't know there would have to be an issue first)

File: src/widget/scroll.rs (hunk 4 above)
Comment (ForLoveOfCats) [reply src/widget/scroll.rs]: Ahhh yes, that's a good point

File: src/widget/scroll.rs (hunk 5 above)
Comment (ForLoveOfCats) [reply src/widget/scroll.rs]: This code was also inherited, personally I don't like it but I didn't want to modify the existing way of doing things to much

File: src/widget/scroll.rs (hunk 3 above)
Comment (ForLoveOfCats) [reply src/widget/scroll.rs]: I personally don't like that, can you explain the reasoning why it's a good thing to do? If you would really like I can change this, not my project not my code style, but would prefer not to.

File: src/widget/scroll.rs (hunk 8 above)
Comment (ForLoveOfCats) [reply src/widget/scroll.rs]: Noice :ok_hand:

File: src/widget/scroll.rs (hunk 1 above)
Comment (ForLoveOfCats) [reply src/widget/scroll.rs]: I *completely* forgot about Rust's magical enums. I frequently wish I had sum types in the languages I use on the daily but I thoroughly forgot about them once I was in a language which *does* have them! I'll look into this right away!

File: src/widget/scroll.rs (hunk 3 above)
Comment (cmyr) [reply src/widget/scroll.rs]: I appreciate it can be a bit jarring at first, but it's part of the idea that rust is expression oriented; everything is an expression and every expression explicitly returns. Here's some reasonable answers to [someone else asking the same question](https://stackoverflow.com/questions/27961879/why-is-using-return-as-the-last-statement-in-a-function-considered-bad-style).

Comment (cmyr) [reply src/widget/scroll.rs]: And here's the [relevant section of the book](https://doc.rust-lang.org/book/ch03-03-how-functions-work.html#functions-with-return-values).

File: src/widget/scroll.rs (hunk 8 above)
Comment (ForLoveOfCats) [reply src/widget/scroll.rs]: rustfmt pls
image

Comment (ForLoveOfCats) [reply src/widget/scroll.rs]: I think in this specific spot the original was better, I'll try to find any other match cases I added where it could be applied though

File: src/widget/scroll.rs (hunk 1 above)
Comment (ForLoveOfCats) [reply src/widget/scroll.rs]: I ended up using separate enums for held and hover statuses

Comment (ForLoveOfCats) [reply src/widget/scroll.rs]: I ended up using separate enums for hovered and held statuses

**Comment by ForLoveOfCats** [comment]
I believe I've properly resolved all items requested. Round two?

**Comment by ForLoveOfCats** [comment]
Oop, I forgot a few places where it's still passing `&Rect`

File: src/widget/scroll.rs (hunk 8 above)
Comment (cmyr) [reply src/widget/scroll.rs]: Hmm. My thinking with that initial comment was that you could replace _all_ instances of `if match` with a single match statement that had conditionals on the arms?

This isn't strictly necessary, but I think that `if match()` is hard to follow, in terms of the control flow, because you have to jump from looking at the match statement down to the `{` where the conditional statement is, often jumping back up to look at various arms of the match to see which are evaluating true and false; if this is a single `match` statement you can basically read it top to bottom, following immediately from a match to the branch that will be taken.


... _some time passes_ ...

okay, I wanted to actually see if my ideas made sense, since this does seem like tricky control flow. Here is my best attempt at implementing the `event` fn, which i _think_ has identical behaviour to your implementation; hopefully this can be helpful at least as an illustration. I personally think it's slightly clearer to follow, but not hugely so or anything; the logic here is hard to capture.

```rust
fn event(&mut self, event: &Event, ctx: &mut EventCtx, data: &mut T, env: &Env) {
    let size = ctx.base_state.size();
    let viewport = Rect::from_origin_size(Point::ORIGIN, size);
    
    // first check if this is a mouse even related to the scroll bars,
    // which we will swallow.
    let is_handled = match event {
        Event::MouseMoved(event) if self.scroll_bars.held.is_held() => {
            self.reset_scrollbar_fade(ctx);
            match self.scroll_bars.held {
                BarHeldState::VerticalHeld(offset) => {
                    let scale_y = viewport.height() / self.child_size.height;
                    let bounds = self.calc_vertical_bar_bounds(&viewport);
                    let mouse_y = event.pos.y + self.scroll_offset.y;
                    let delta = mouse_y - bounds.y0 - offset;
                    self.scroll(Vec2::new(0f64, (delta / scale_y).ceil()), size);
                    true
                }
                BarHeldState::HorizontalHeld(offset) => {
                    let scale_x = viewport.width() / self.child_size.width;
                    let bounds = self.calc_horizontal_bar_bounds(&viewport);
                    let mouse_x = event.pos.x + self.scroll_offset.x;
                    let delta = mouse_x - bounds.x0 - offset;
                    self.scroll(Vec2::new((delta / scale_x).ceil(), 0f64), size);
                    true
                }
                _ => unreachable!(),
            }
        }
        Event::MouseUp(_) if self.scroll_bars.held.is_held() => {
            self.scroll_bars.held = BarHeldState::None;
            true
        }
        Event::MouseMoved(event) => {
            let transformed_pos = event.pos + self.scroll_offset;
            if self.point_hits_vertical_bar(viewport, transformed_pos) {
                self.scroll_bars.hovered = BarHoveredState::VerticalHovered;
                self.reset_scrollbar_fade(ctx);
                true
            } else if self.point_hits_horizontal_bar(viewport, transformed_pos) {
                self.scroll_bars.hovered = BarHoveredState::HorizontalHovered;
                self.reset_scrollbar_fade(ctx);
                true
            } else {
                if self.scroll_bars.hovered.is_hovered() {
                    self.scroll_bars.hovered = BarHoveredState::None;
                    self.reset_scrollbar_fade(ctx);
                    // we need to invalidate here manually, because we're returning
                    // false (continuing to propogate this event) _and_ we need to be drawn.
                    ctx.invalidate();
                }
                false
            }
        }
        Event::MouseDown(event) => {
            let transformed_pos = event.pos + self.scroll_offset;
            if self.point_hits_vertical_bar(viewport, transformed_pos) {
                self.scroll_bars.held = BarHeldState::VerticalHeld(
                transformed_pos.y - self.calc_vertical_bar_bounds(&viewport).y0,
                );
                true
            } else if self.point_hits_horizontal_bar(viewport, transformed_pos) {
                self.scroll_bars.held = BarHeldState::HorizontalHeld(
                transformed_pos.x - self.calc_horizontal_bar_bounds(&viewport).x0,
                );
                true
            } else {
                false
            }
        }
        _ => false,
    };
    
    if is_handled {
        ctx.invalidate();
        ctx.is_handled();
        return;
    }
    
    // send the event down to the child
    let child_event = event.transform_scroll(self.scroll_offset, viewport);
    if let Some(child_event) = child_event {
        self.child.event(&child_event, ctx, data, env)
    };
    
    // handle other events after child, as appropriate
    
    match event {
        // Show the scrollbars any time our size changes
        Event::Size(_) => self.reset_scrollbar_fade(ctx),
        // The scroll bars will fade immediately if there's some other widget requesting animation.
        // Guard by the timer id being invalid.
        Event::AnimFrame(interval) if self.scroll_bars.timer_id == TimerToken::INVALID => {
            // Animate scroll bars opacity
            let diff = 2.0 * (*interval as f64) * 1e-9;
            self.scroll_bars.opacity -= diff;
            if self.scroll_bars.opacity > 0.0 {
                ctx.request_anim_frame();
            }
        }
        Event::Timer(id) if *id == self.scroll_bars.timer_id => {
            // Schedule scroll bars animation
            ctx.request_anim_frame();
            self.scroll_bars.timer_id = TimerToken::INVALID;
        }
        // we give the child a chance to handle scroll-wheel events first
        Event::Wheel(wheel) if !ctx.is_handled() => {
            if self.scroll(wheel.delta, size) {
                ctx.invalidate();
                ctx.set_handled();
                self.reset_scrollbar_fade(ctx);
            }
        }
        _ => (),
    }
}
```

So... i dunno. This is definitely a big ole' function, in any case.

src/widget/scroll.rs
```diff
@@ -239,30 +292,115 @@ impl<T: Data, W: Widget<T>> Widget<T> for Scroll<T, W> {
...
+                    let mut transformed_event = event.clone();
+                    transformed_event.pos += self.scroll_offset;
+                    if self.point_hits_vertical_bar(viewport, transformed_event.pos) {
+                        self.scroll_bars.hovered = BarHoveredState::VerticalHovered;
+                    } else {
+                        self.scroll_bars.hovered = BarHoveredState::HorizontalHovered;
+                    }
+
+                    self.scroll_bars.opacity = 0.7;
+                    self.scroll_bars.timer_id = TimerToken::INVALID; // Cancel any fade out in progress
```
cmyr: is there a reason we can't use `reset_scrollbar_fade` here?

ForLoveOfCats: Yes because `reset_scrollbar_fade` starts the timer for a fade out

src/widget/scroll.rs
ForLoveOfCats: :man_shrugging:

src/widget/scroll.rs
```diff
@@ -153,57 +225,38 @@ impl<T: Data, W: Widget<T>> Scroll<T, W> {
...
-            let y0 = scroll_bar_bounds.y1;
+    fn point_hits_vertical_bar(&self, viewport: Rect, pos: Point) -> bool {
+        if viewport.height() < self.child_size.height {
+            let bounds = self.calc_vertical_bar_bounds(&viewport);
+            return pos.y > bounds.y0 && pos.y < bounds.y1 && pos.x > bounds.x0;
+        }
 
-            let x1 = (x0 + w).min(scroll_bar_bounds.x1);
-            let y1 = y0 - bar_thickness;
+        false
```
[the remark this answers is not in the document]
ForLoveOfCats: That `if if` feels less readable and more opaque than the existing function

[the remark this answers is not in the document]
cmyr: oh sorry that extra `if` is a typo 😬

cmyr:
@ForLoveOfCats I'm happy with this if you are.

src/widget/scroll.rs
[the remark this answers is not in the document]
ForLoveOfCats: 😆

ForLoveOfCats:
👍

cmyr:
thanks!

PR #245, opened 2019-10-31
ci: Test druid-derive-data and druid-derive-lens

cmyr:
Noticed that these two crates aren't currently being tested.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,30 @@ jobs:
           command: rustc
           args: --manifest-path=druid-shell/Cargo.toml -- -D warnings
 
+      - name: Run cargo check in druid/druid-derive-data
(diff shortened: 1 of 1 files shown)


cmyr:
There's a reasonable argument that we should combine these two crates into one, you're allowed to have multiple derive impls in a single crate and it would make it easier to reuse code and validate attributes.

raphlinus:
I'm also ok with combining them. At some point, there's probably a good argument for splitting Data and Lens out into their own crate, largely so that fancy datatypes can impl them without taking a dep on druid as a whole. But I think Data and Lens share the same fate here. There's other stuff I can imagine wanting to proc macro (stuff to synthesize property-sheet style UI) that should probably be separate because it *does* bring in more of a heavyweight druid dep.

PR #247, opened 2019-10-31
support ignoring fields when deriving data

cmyr:
This was a bit of a rabbit hole.
tl;dr: you can now write this:

```rust
struct AppState {
    text: String,
    copies: usize,
    #[druid(ignore)]
    user: UserId,
}
```

And the last field will be ignored for the purpose of `same()`.

I ended up making some largish organizational changes as a result of this work.

@futurepaul want to take a look and see if this makes sense to you?

### DIFF
--- a/druid-derive-data/Cargo.toml
+++ b/druid-derive-data/Cargo.toml
@@ -12,3 +12,5 @@ syn = "1.0.5"
 quote = "1.0.2"
 proc-macro2 = "1.0.4"
 
+[dev-dependencies]
+druid = { path = ".." }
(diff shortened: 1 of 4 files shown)


futurepaul:
It makes sense overall to me, can't say I understand or have absorbed all the details. Macro lingo is pretty hard for me to hold onto. The tests are reassuring, I should def add some to Lens!

cmyr:
@futurepaul fair enough, I definitely seem to remember zero from the last time I was playing with proc macros, which wasn't more that 6 months ago. 😒

PR #248, opened 2019-10-31
Allow custom comparison functions for fields when deriving Data

cmyr:
This is includes commits from #247, which should get merged first.

This lets you write,

```rust
#[derive(Clone, Data)]
struct NonBitwiseFloat {
    #[druid(with_same = "PartialEq::eq")]
    inner: f64,
}
```

And you will get an implementation of `Data` that uses `PartialEq::eq` when checking that field.

I don't know what to call this attribute; serde uses `serialize_with = "..."` and `skip_serializing_if = "..."`. Maybe this should be `cmp_fn = "..."` or `same_fn = "..."` or `compare = "..."` or... I'm not sure, really.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -196,7 +196,7 @@ dependencies = [
 name = "druid"
 version = "0.3.0"
 dependencies = [
- "druid-derive-data 0.1.0",
+ "druid-derive-data 0.1.1",
  "druid-derive-lens 0.1.0",
  "druid-shell 0.3.0",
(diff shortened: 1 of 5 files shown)


raphlinus:
I'd personally lean toward "same_fn".

The f64 example makes the math person in me cringe a bit :) I wonder if there's a better example that doesn't have the same failure modes.

cmyr:
The f64 example is purely demonstrative: it's a situation where `Data` and `PartialEq` both exist but are implemented differently, so it's a good way to prove that we're using the one behaviour and not the other. It won't go in any public docs. 😉

PR #254, opened 2019-11-02
ci: Do less work in CI

cmyr:
This reduces the amount of work we will do in CI in two ways:

- it moves the 'deny warnings' check to happen in the same
job as the `cargo test` check, because they both compile
at the same optimization level and we can reuse artifacts.

- it removes the 'cargo check' job, since any failure there
will also be caught in `cargo test`. If jobs were sequential
it might be worth checking first, to fail quickly, but
since jobs are parallel this is just extra work for no
tangible benefit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,66 +5,6 @@ on:
   pull_request:
 
 jobs:
-  check-stable:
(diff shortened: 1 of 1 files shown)


cmyr:
this will conflict with #253, but want to run it now and see if it works, will patch up after.

PR #256, opened 2019-11-02
Move cursor on mouse click

cmyr:
This implements (very basic) handling of cursor movement with the mouse. Clicking in the textbox will set the cursor, and shift+click will modify the selection.

Co-authored-by: Paul Miller <paul@pauljmiller.com>

### DIFF
--- a/druid/src/widget/textbox.rs
+++ b/druid/src/widget/textbox.rs
@@ -150,6 +150,21 @@ impl TextBoxRaw {
         self.selection.end
     }
 
+    /// For a given point, returns the corresponding offset (in bytes) of
+    /// the grapheme cluster closest to that point.
+    fn offset_for_point(
+        &self,
+        point: Point,
+        piet_text: &mut PietText,
+        data: &String,
+        env: &Env,
+    ) -> usize {
+        let layout = self.get_layout(piet_text, env, data);
+        let hscrolled_point = Point::new(point.x + self.hscroll_offset, point.y);
+        let hit_test = layout.hit_test_point(hscrolled_point);
+        hit_test.metrics.text_position
+    }
+
     /// Calculate a stateful scroll offset
     fn update_hscroll(&mut self, rc_text: &mut PietText, env: &Env, data: &String) {
         let cursor_x = self.substring_measurement_hack(rc_text, data, 0, self.cursor(), env);
@@ -324,10 +339,14 @@ impl Widget<String> for TextBoxRaw {
 
     fn event(&mut self, event: &Event, ctx: &mut EventCtx, data: &mut String, env: &Env) {
         match event {
-            Event::MouseDown(_) => {
+            Event::MouseDown(mouse) => {
                 ctx.request_focus();
-                // TODO: hit test and do this for real
-                self.cursor_to(self.selection.end);
+                let cursor_off = self.offset_for_point(mouse.pos, ctx.text(), data, env);
+                if mouse.mods.shift {
+                    self.selection.end = cursor_off;
+                } else {
+                    self.cursor_to(cursor_off);
+                }
                 ctx.invalidate();
                 self.reset_cursor_blink(ctx);
             }


futurepaul:
Right now the resulting cursor position "rounds up" to the end of whatever we're clicking on. Might be something to tweak in Piet.

Also worth testing non-Mac platforms. I can do Windows later today if nobody gets to it.

src/widget/textbox.rs
```diff
@@ -150,6 +150,24 @@ impl TextBoxRaw {
         self.selection.end
     }
 
+    /// For a given point, returns the cooresponding offset (in bytes) of
```
cmyr: 'corresponding' 🙄

src/widget/textbox.rs
```diff
@@ -150,6 +150,24 @@ impl TextBoxRaw {
...
+        piet_text: &mut PietText,
+        data: &String,
+        env: &Env,
+    ) -> usize {
+        let layout = self.get_layout(piet_text, env, data);
+
+        let hscrolled_point = Point::new(point.x + self.hscroll_offset, point.y);
+
+        let hit_test = layout.hit_test_point(hscrolled_point);
+        //TODO: how should we be handling when we're outside the text?
```
cmyr: a question for @hwchen: it wasn't totally clear from the docs in piet how this API is supposed to be used. This might be something to clarify somewhat? In particular, the `is_text` property on `HitTestMetrics` isn't described, and it would be nice if there was a bit more information on what `is_inside` communicates, and how it relates to the returned `text_position`, or how `text_position` is determined for clicks that fall outside the bounds of the text?

hwchen: Sorry for lack of clarity, I should go back and document a little better.

I think my comments here on `is_inside` and `text_position` should be helpful. `is_text` is a bit trickier (I think I messed up), and I will be discussing with @raphlinus how to handle it better. I think that for basic cursor positioning you can safely ignore it.

For `hit_test_point`:
- `is_inside` returns `true` if the point be tested falls within the bounds of the text (basically just the text width).
- some `text_position` will always be returned; if the tested point is inside, it returns the appropriate text position; if it's outside, it will return the nearest text position (either `0` or `text.len()`).

As for `is_text`:
- `is_text` should return true if there's text at that text position.
- You shouldn't have to worry about it for outside-of-text situations. it _should_ return `true` as long as there's text at the `text_position` returned.
- I just dug a little, because of an itch in my head, and it looks like web/cairo can return `false` when the point tested is outside of the text, which is inconsistent with directwrite. This would be a bug in the current implementation.
- `is_text` kind of slipped through from my initial port from directwrite. I'm not sure it's really needed with the current api. An example of usage (to draw selected images): https://github.com/microsoft/Windows-classic-samples/blob/master/Samples/Win7Samples/multimedia/DirectWrite/PadWrite/TextEditor.cpp#L520
- I'll be discussing with @raphlinus soon. Sorry!

cmyr: those are good answers, thanks! Definitely would be nice to get that into the docs at some point, but no sweat. :)

hwchen:
@futurepaul just saw your note about rounding up. I'll look into it, but it should be rounding down if on the left side of the grapheme, and rounding up if on the right side.

hwchen:
@futurepaul I took a look at the rounding behavior in druid. I set the font size really large, and from visual inspection it's doing what I expect; that there's a break halfway through the glyph, hits on left go left, hits on right go right.

Let me know if you can replicate (I'm on linux/cairo, and mac/cairo should have the same logic).

If you can, and it still feels a bit off, we'd have to have some higher-level discussions on whether to change the rounding behavior.

cmyr:
Okay I'm going to merge this shortly. @futurepaul afterwards you'll want to rebase #257 onto `origin/master`, and heads-up that I squashed the two commits here so you might notice an extra commit (with the hscroll changes) when you try to rebase, you should be able to just drop that commit if you do `rebase -i`.

scholtzan:
If the text contains non-ascii characters (eg. `©©` or `ßß`) then moving the cursor results in a panic: 

```
2019-11-08 11:05:12,980 DEBUG [druid::localization] resolved: [en-US]
thread 'main' panicked at 'attempt to subtract with overflow', /Users/anna/.cargo/registry/src/github.com-1ecc6299db9ec823/piet-cairo-0.0.7/src/lib.rs:583:34
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

Not sure if that is a known problem or if a separate issue should be created for this.

raphlinus:
Have you tried whether druid#275 fixes that?

scholtzan:
It doesn't seem to fix it. Still crashing.

raphlinus:
Ah :( Then we have a deeper problem. I didn't carefully verify #275 against everything that can go wrong.

futurepaul:
Which platform? Those characters are working on mac for me.

scholtzan:
It's on mac. Sometimes it takes a couple of clicks and inserting more characters for it to crash.

futurepaul:
Ah yeah I see it now just had to insert more and click more.

hwchen:
I've made a pr for a fix for hanging (infinite loop) in https://github.com/linebender/piet/pull/105 . This fixes the infinite loop issue, but I was not able to reproduce the `subtract with overflow` issue.

https:://github.com/linebender/piet/pull/104 seemed to have fixed both of them; I just wasn't sure if fixing the infinite loop will also solve the `subtract with overflow` issue.

@scholtzan and @futurepaul can you give me an exact sequence to reproduce `subtract with overflow`?

PR #257, opened 2019-11-02
Measure cursor position using Piet hit testing

futurepaul:
Builds on #256, should be merged after.

This removes the substring measurement hack we've been relying on to position the cursor and selection box when drawing. I also did some cleanup to the usage of `get_layout`, but I couldn't quite figure out how to cache layout, due to lifetime issues.

As a bonus, I also added PR to this click-and-drag selection because it was an easy fix!

### DIFF
--- a/druid/src/widget/textbox.rs
+++ b/druid/src/widget/textbox.rs
@@ -118,12 +118,16 @@ impl TextBox {
         }
     }
 
-    fn get_layout(&self, t: &mut PietText, env: &Env, data: &String) -> PietTextLayout {
+    fn get_layout(&self, piet_text: &mut PietText, data: &String, env: &Env) -> PietTextLayout {
         let font_name = env.get(theme::FONT_NAME);
         let font_size = env.get(theme::TEXT_SIZE_NORMAL);
         // TODO: caching of both the format and the layout
-        let font = t.new_font_by_name(font_name, font_size).build().unwrap();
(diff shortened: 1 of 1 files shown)


futurepaul:
I believe I have accomplished rebase.

druid/src/widget/textbox.rs
```diff
@@ -265,20 +255,21 @@ impl Widget<String> for TextBox {
...
-                    let selection_width =
-                        self.substring_measurement_hack(rc.text(), data, left, right, env);
+                    let selection_width = self.x_for_offset(&text_layout, right)
+                        - self.x_for_offset(&text_layout, left);
 
                     let selection_pos = Point::new(
-                        self.substring_measurement_hack(rc.text(), data, 0, left, env)
-                            + PADDING_LEFT
-                            - 1.,
+                        self.x_for_offset(&text_layout, left) + PADDING_LEFT - 1.,
```
cmyr: total nit but since we use this twice I would probably just do a `let left_offset = self.x_for_offset(&text_layout, left)` above, and use both places, which probably let this fit on one line. 🤷‍♂

futurepaul:
Ah, that was my confusion, I was leaving that in, thinking it was my "starting point". Makes sense that if it's already in master it doesn't need to be in my branch.

I _believe_ I now have accomplished rebase?

cmyr:
yep looks good!

PR #258, opened 2019-11-03
Use a cargo workspace

cmyr:
I'm not sure if there was a reason we hadn't done this earlier? In any case I wanted to see if it would be much hassle, and it wasn't.

This moves to using a cargo workspace, which should improve compilation
times and simplify CI.

- as a result of this, we should no longer have targets that share a name;
as a result druid-shell/examples/hello.rs is now 'shello.rs'.


druid-shell/Cargo.toml
```diff
@@ -42,7 +42,6 @@ features = ["d2d1_1", "dwrite", "winbase", "libloaderapi", "errhandlingapi", "wi
 [target.'cfg(target_os="macos")'.dependencies]
 cocoa = "0.19.0"
 objc = "0.2.5"
-core-graphics = "0.18.0"
```
LiHRaM: Just out of curiosity, does the workspace feature somehow make this unnecessary / did you put this somewhere else?

cmyr: Good question, and no, this is an unrelated change I'm sneaking in.

I noticed while moving to the workspace that we were importing two versions of this, one directly and one via the `cocoa` crate. After a bit of digging it turned out that we weren't actually using this import, so I took the opportunity to delete.

PR #259, opened 2019-11-04
Split panel

Zarenor:
Here's a initial iteration of a panel to split an area in two, either vertically or horizontally.

I followed the pattern as in `Flex` w.r.t creating public convenience impls which are the only (current) ways to construct the type, and also used the same `Axis` enum for direction

I've got the basic setup here, but I'm still out on a couple of questions:

- Should we refactor `Axis` to be in a more-obvious location to be shared, or should I duplicate it here?
- Do we want this to be a drag-adjustable split? Or is that a separate, similar control? Maybe controlled by a field?
- I'm not sure how best to handle unbounded size

A couple of other notes:
- I haven't created any tests or many docs yet - I'm planning to review the practice across the other widgets and meet at least the standard there. Any specific points I need to cover are welcome!

### DIFF
--- a/druid/src/widget/split.rs
+++ b/druid/src/widget/split.rs
@@ -1,0 +1,310 @@
+// Copyright 2019 The xi-editor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! A widget which splits an area in two, with a set ratio.
+
+use crate::kurbo::{Line, Point, Rect, Size};
+use crate::widget::flex::Axis;
+use crate::{
+    theme, BaseState, BoxConstraints, Cursor, Data, Env, Event, EventCtx, LayoutCtx, PaintCtx,
+    RenderContext, UpdateCtx, Widget, WidgetPod,
+};
+
+///A container containing two other widgets, splitting the area either horizontally or vertically.
+pub struct Split<T: Data> {
+    split_direction: Axis,
+    draggable: bool,
+    split_point: f64,
+    splitter_size: f64,
+    child1: WidgetPod<T, Box<dyn Widget<T>>>,
+    child2: WidgetPod<T, Box<dyn Widget<T>>>,
+}
+
+impl<T: Data> Split<T> {
+    ///Create a new split panel.
+    fn new(
+        split_direction: Axis,
+        child1: impl Widget<T> + 'static,
+        child2: impl Widget<T> + 'static,
+    ) -> Self {
+        Split {
+            split_direction,
+            split_point: 0.5,
+            splitter_size: 10.0,
+            draggable: false,
+            child1: WidgetPod::new(child1).boxed(),
+            child2: WidgetPod::new(child2).boxed(),
+        }
+    }
+    /// Create a new split panel, splitting the vertical dimension between two children.
+    pub fn vertical(child1: impl Widget<T> + 'static, child2: impl Widget<T> + 'static) -> Self {
+        Self::new(Axis::Vertical, child1, child2)
+    }
+    /// Create a new split panel, splitting the horizontal dimension between two children.
+    pub fn horizontal(child1: impl Widget<T> + 'static, child2: impl Widget<T> + 'static) -> Self {
+        Self::new(Axis::Horizontal, child1, child2)
+    }
+    /// Set container's split point as a fraction of the split dimension
+    /// The value must be between 0.0 and 1.0, exclusive
+    pub fn split_point(mut self, split_point: f64) -> Self {
+        assert!(
+            split_point > 0.0 && split_point < 1.0,
+            "split_point must be between 0.0 and 1.0!"
+        );
+        self.split_point = split_point;
+        self
+    }
+    /// Set the width of the splitter bar, in pixels
+    /// The value must be positive or zero
+    pub fn splitter_size(mut self, splitter_size: f64) -> Self {
+        assert!(
+            splitter_size >= 0.0,
+            "splitter_width must be 0.0 or greater!"
+        );
+        self.splitter_size = splitter_size;
+        self
+    }
+    /// Set whether the splitter's split point can be changed by dragging.
+    pub fn draggable(mut self, draggable: bool) -> Self {
+        self.draggable = draggable;
+        self
+    }
+    fn splitter_hit_test(&self, size: Size, mouse_pos: Point) -> bool {
+        match self.split_direction {
+            Axis::Horizontal => {
+                let center = size.width * self.split_point;
+                (center - mouse_pos.x).abs() < self.splitter_size.min(5.0) / 2.0
+            }
+            Axis::Vertical => {
+                let center = size.height * self.split_point;
+                (center - mouse_pos.y).abs() < self.splitter_size.min(5.0) / 2.0
+            }
+        }
+    }
+    fn update_splitter(&mut self, size: Size, mouse_pos: Point) {
+        self.split_point = match self.split_direction {
+            Axis::Horizontal => {
+                let max_limit = size.width - (self.splitter_size * 0.5).min(5.0);
+                let min_limit = (self.splitter_size * 0.5).min(5.0);
+                let max_split = max_limit / size.width;
+                let min_split = min_limit / size.width;
+                if mouse_pos.x > max_limit {
+                    max_split
+                } else if mouse_pos.x < min_limit {
+                    min_split
+                } else {
+                    mouse_pos.x / size.width
+                }
+            }
+            Axis::Vertical => {
+                let max_limit = size.height - (self.splitter_size * 0.5).min(5.0);
+                let min_limit = (self.splitter_size * 0.5).min(5.0);
+                let max_split = max_limit / size.height;
+                let min_split = min_limit / size.height;
+                if mouse_pos.y > max_limit {
+                    max_split
+                } else if mouse_pos.y < min_limit {
+                    min_split
+                } else {
+                    mouse_pos.y / size.height
+                }
+            }
+        }
+    }
+}
+impl<T: Data> Widget<T> for Split<T> {
+    fn paint(&mut self, paint_ctx: &mut PaintCtx, base_state: &BaseState, data: &T, env: &Env) {
+        let size = base_state.size();
+        //third, because we're putting the lines at roughly third points.
+        //small, because we floor, to give the extra pixel (roughly) to the middle.
+        let small_third = (self.splitter_size / 3.0).floor();
+        let (line1, line2) = match self.split_direction {
+            Axis::Horizontal => {
+                let reduced_width = size.width - self.splitter_size;
+                let edge1 = reduced_width * self.split_point;
+                let edge2 = edge1 + self.splitter_size;
+                (
+                    Line::new(
+                        Point::new(edge1 + small_third, 0.0),
+                        Point::new(edge1 + small_third, size.height),
+                    ),
+                    Line::new(
+                        Point::new(edge2 - small_third, 0.0),
+                        Point::new(edge2 - small_third, size.height),
+                    ),
+                )
+            }
+            Axis::Vertical => {
+                let reduced_height = size.height - self.splitter_size;
+                let edge1 = reduced_height * self.split_point;
+                let edge2 = edge1 + self.splitter_size;
+                (
+                    Line::new(
+                        Point::new(0.0, edge1 + small_third),
+                        Point::new(size.width, edge1 + small_third),
+                    ),
+                    Line::new(
+                        Point::new(0.0, edge2 - small_third),
+                        Point::new(size.width, edge2 - small_third),
+                    ),
+                )
+            }
+        };
+        let line_color = if self.draggable {
+            env.get(theme::BORDER_LIGHT)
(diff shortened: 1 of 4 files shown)


futurepaul:
* I could see there being some sort of util.rs in widgets for shared widget stuff like this. Not sure the best way to organize it. There is a lot of overlap with Flex here though which makes me wonder if this could be implemented on top of Flex instead of duplicating? I'm not sure of all the implications of that.
* Drag-adjustable seems a good default for this widget imo. At least that's what I'd like to use it for!
* We just merged some improvements to Flex's handling of unbounded, maybe that could help? I'm trying to think of all the scenarios for split... most of the widgets shrink down to a minimum size (sometimes hardcoded) when handed an unbounded max constraint. I could see split giving the "first" item of the split a fixed size and then handing all remaining space to the "second" item.

I wonder if most of this could be accomplished with Flex and SizedBox: instead of having the two panes always split proportionally, you could have one with a width or height in pixels and a sized box, set Flex 0.0, while the other pane gets Flex 1.0. This would cover a lot of traditional split pane use cases like a text editor with file list on the left or a chat window with usernames on the left. And then another version could make both panes Flex 1.0. Not sure what to call these different modes, or if I'm making sense!

* layout example sounds correct to me, but feel free to make the split more prominent.
* The big minimum requirement is to document public structs / methods.

futurepaul:
Oh, and if you do drag-adjustable you get to use a custom cursor!
```
Cursor::ResizeLeftRight / Cursor::ResizeUpDown
```

Zarenor:
>     * We just merged some improvements to Flex's handling of unbounded, [...]
I'll certainly want to pull down the latest changes to Flex, and see how it's handling it. I think it's something of a weakness of a single-layout-pass single-size-returned layouting system that you can bump into this where you maybe want more information than you're provided, and can't anticipate another pass in which to resolve those concerns.  I think the strength here is the simplicity and speed.
If we can agree to some consistent way for layout widgets to handle unbounded size, I think we'll be okay.

> I wonder if most of this could be accomplished with Flex and SizedBox [...]

I think that's certainly a way to handle it. I think the idea behind having a more restrictive widget is having less algorithmic 'weight'. The algorithm here is pretty dead simple, and doesn't have to handle many edge cases (aside for the aforementioned unbounded-size in the split direction).

I think you could get fixed-split with some setups for flex, but I don't think it would be too simple to implement draggable resize with flex. Maybe possible, leveraging some SizedBoxes, sure.

> I could see there being some sort of util.rs in widgets for shared widget stuff like this. [...]
If I was going to begin advocating for something like this, it would be perhaps a widget::layout mod, to more granularly organize the growing number of widgets. I'm not sure when is the point to do that, or if that's actually something anyone else is interested in. But it would be my first instinct.

Zarenor:
After some discussions on zulip with @cmyr, I'm going to move this out of draft without the draggability implemented. I hope to write that up and PR it tomorrow, but we'll see how it goes.

cmyr:
Just a preliminary comment: something that's really missing in this PR is an example of the widget in use. I think including an example with  a new widget is important because it makes it much easier to actually see & evaluate the behaviour. 😎

druid/src/widget/split.rs
```diff
@@ -0,0 +1,220 @@
...
+}
+
+/// A builder for a widget which divides space horizontally
+/// The splitter bar will run top-to-bottom
+///
+/// The actual widget it implemented by [`Split`], but this builder
+/// is provided for convenience.
+///
+/// [`Split`]: struct.Split.html
+pub struct HorizontalSplit;
```
cmyr: I'm not sure I love this pattern of having 'marker' widgets that delegate to another widget under the hood.

To me it might be easier to just have separate `Split::vertical()` and `Split::horizontal()` constructors on `Split` itself.

But this is mostly my preference; if other folks prefer it this way I'm not hugely bothered.

druid/src/widget/split.rs
```diff
@@ -0,0 +1,220 @@
...
+            child1: WidgetPod::new(child1).boxed(),
+            child2: WidgetPod::new(child2).boxed(),
+        }
+    }
+}
+impl<T: Data> Widget<T> for Split<T> {
+    fn paint(&mut self, paint_ctx: &mut PaintCtx, base_state: &BaseState, data: &T, env: &Env) {
+        let size = base_state.size();
+        let line1;
+        let line2;
```
cmyr: this doesn't feel super 'rusty'. Unless there's a clear reason why it's difficult, I would prefer this in the form, `let (line1, line2) = match self.split_direction { ...`, and then just have each branch of the match statement resolve to a 2-tuple of lines.

cmyr: (to clarify: this isn't blocking, I'm just stating my preferences)

druid/src/widget/split.rs
Zarenor: I'm not sure how I liked the abstraction, but I copied the idea directly from `Flex`. I'm open to going with the `::vertical()` and `::horizontal()` constructors.

I'd like to pick a style and stick with it. I'm fairly serious about continuing to build layout widgets, and I don't want our style to be all over the place, if we can avoid it. 😊

druid/src/widget/split.rs
Zarenor: I'm more than open to this kind of style nit - I've done a lot of study and only a little hobby work in rust. I'll see if I can work this the way you suggest.

druid/src/widget/split.rs
cmyr: yep I agree consistency would be good, maybe something to discuss on zulip.

druid/src/widget/split.rs
Zarenor: I made this change, here and in several other (child1_xxx, child2_xxx) locations. Definitely feels better.

druid/src/widget/split.rs
```diff
@@ -0,0 +1,327 @@
...
+            child1,
+            child2,
+        )
+    }
+}
+
+impl<T: Data> Split<T> {
+    ///Create a new split panel.
+    ///Note that split_point must be between 0 and 1.
+    fn new(
```
cmyr: Okay, my only main nit now is around constructors, where I have a few thoughts.

- Per the discussion in zulip, I think that `Split::vertical` and `Split::horizontal` constructors make sense. If we want to keep the `VerticalSplit` and `HorizontalSplit` structs, their `new` fns can just call the corresponding constructors on `Split`.
- With widgets that can take a number of arguments at init time, I think a builder pattern makes sense. For an example of this, see [`SizedBox`](https://github.com/xi-editor/druid/blob/master/druid/src/widget/sized_box.rs#L59). In this case, I think that `draggable` and `splitter_size` probably make sense as optional builder-style methods.
- `splitter_size` is an interesting one; this might actually be something that we want to put in the `env`? Or we might want to put the default value in the `env`? This comes down to whether or not we expect to be tuning this per-platform. I'm okay with punting on this for now.

druid/src/widget/split.rs
```diff
@@ -0,0 +1,327 @@
...
+            Axis::Vertical => {
+                let center = size.height * self.split_point;
+                (center - mouse_pos.y).abs() < self.splitter_size / 2.0
+            }
+        }
+    }
+    fn update_splitter(&mut self, size: Size, mouse_pos: Point) {
+        self.split_point = match self.split_direction {
+            Axis::Horizontal => {
+                let max_limit = size.width - (self.splitter_size * 0.5);
```
cmyr: If I make the splitter large enough I can end up with constraints that are negative sizes, which is undesirable. Not sure the best approach, but we should be limiting this somehow.

druid/src/widget/split.rs
```diff
@@ -0,0 +1,327 @@
...
+            }
+        };
+        self.child1.set_layout_rect(child1_rect);
+        self.child2.set_layout_rect(child2_rect);
+        my_size
+    }
+
+    fn event(&mut self, event: &Event, ctx: &mut EventCtx, data: &mut T, env: &Env) {
+        if self.draggable && self.splitter_size > 0.0 {
+            //TODO: Draggable splitter
```
cmyr: can get rid of this todo?

druid/src/widget/split.rs
```diff
@@ -0,0 +1,327 @@
...
+                    }
+                }
+                Event::MouseUp(mouse) => {
+                    if mouse.button.is_left() && ctx.is_active() {
+                        ctx.set_active(false);
+                        self.update_splitter(ctx.base_state.size(), mouse.pos);
+                        ctx.invalidate();
+                    }
+                }
+                Event::MouseMoved(mouse) => {
```
cmyr: One thing missing here is changing the cursor to indicate that we can resize, when we can resize.

The current mechanism for this is a bit iffy (I suspect it will change) but it's easy enough in this case; any time we get `MouseMoved`, if we are overtop of the split we should call `ctx.set_cursor(Cursor::ResizeLeftRight/ResizeUpDown)`.

druid/src/widget/split.rs
```diff
@@ -0,0 +1,327 @@
...
+        Split {
+            split_point,
+            split_direction,
+            splitter_size,
+            draggable,
+            child1: WidgetPod::new(child1).boxed(),
+            child2: WidgetPod::new(child2).boxed(),
+        }
+    }
+    fn splitter_hit_test(&self, size: Size, mouse_pos: Point) -> bool {
```
cmyr: Something this is making me realize: we're conflating the _visual_ size of the splitter with its size for the purpose of hit-testing, and I think these should be different; or at the very least we should have a minimum size that we use for the purpose of hit-testing.

Zarenor: That makes sense to me, but I do worry a bit about 'absorbing' input over the children. or both the splitter and the child reacting. Currently, I mark the mousedown handled, so the absorption case is what would happen.

I also ignore 'draggable' if the splitter size is 0.

If we do go for 'fuzzy', do we just want to add a fuzz-constant (2? 3?) to the hit-test and still mark events handled? Or something more nuanced?

druid/src/widget/split.rs
Zarenor: I'd thought about this and then forgot to implement it. Will do.

druid/src/widget/split.rs
Zarenor: 😅 It shouldn't have made it to this commit in the first place.

druid/src/widget/split.rs
cmyr: What we might want to do is check the children and see if either is active, and if they're active then we send them the event first.

druid/src/widget/split.rs
Zarenor: I may clamp-to-0 this out, but I suspect I need to do a little better check in my extrema-limiting in `adjust_splitter`. Probably some floating-point error allowing that to happen. I can probably mitigate it with some more 'fuzz', similar to what we might do in the hit test

druid/src/widget/split.rs
Zarenor: I had punted on adjusting the types-versus-multiple-constructors based on where we'd landed in zulip - I thought you and I agreed, but the consensus was 🤷. I still do like the constructors method, so I can refactor that way.

I can move to builder-style functions here, if that's preferable. Do we want to allow for not setting children? My reasoning for requiring them in `::new()` was that you can always put some blank widget as a child, if you want a blank space, but it ups the complexity significantly to handle the `None` case in each of our methods.

I definitely have no objection to `splitter_size` and `draggable` as builder methods.

I think some more discussion on if/how we scope `Env` and what goes in it is a good idea. I can see something like a `size::PADDING` (default_padding? border_width?), and the splitter would presumably be two of those wide/tall.

cmyr: I think we definitely always want both children to be specified. It's possible that we don't always want `splitter_position`, (and it's possible we want to allow specifying it in pixels) but I'd been assuming we would always take it, as well; and then we would just have the two 'more optional' arguments be builder methods.

Zarenor: Refactored for `::vertical` and `::horizontal` constructors.
Added builder-style methods for `split_point`, `splitter_width`, `draggable`, with defaults of 0.5, 10.0, and false, respectively. Punted on setting the split point in pixels from an edge - unless we're in a `SizedBox`, we don't know our size at construction time (and even then, we could end up in a smaller size)

druid/src/widget/split.rs
Zarenor: I went with the check-for-active. I'm not sure if that's the nicest way to do it. But it seemed more-sane to me? I also check for marked-resolved, and return in that case. Maybe we don't want to do that?

druid/src/widget/split.rs
Zarenor: I went with more of a fuzz approach by attaching a minimum size 'gap' that can be provided by doing `(splitter_size*0.5).min(5.0)`. 

Gotta run for now, but will test this tomorrow to see if I need to increase/decrease that number. Otherwise, I'll adjust `layout` to clamp the numbers provided to the BoxConstraints to 0 when propagating them down (not that I like that option - it feels like a kludge for a failure in this algorithm).

Zarenor:
I pushed the changes up here tonight, to get them in the cloud. However, I haven't specifically tested the negative-box-constraints problem - I'm going to try to get some more in-depth testing on that tomorrow.

Feel free to review any of the other nits, or if you have a suggestion for how better to handle the issue.

cmyr:
I'm getting a bunch of warnings and nothing drawn when I run the `split_demo` as of 9699dd8. 😢

druid/src/widget/split.rs
```diff
@@ -0,0 +1,311 @@
...
+                    }
+                }
+                Event::MouseMoved(mouse) => {
+                    if ctx.is_active() {
+                        if !ctx.is_hot() {
+                            ctx.set_active(false);
+                        }
+                        self.update_splitter(ctx.base_state.size(), mouse.pos);
+                        ctx.invalidate();
+                    } else if ctx.is_hot()
```
cmyr: I think we should be setting the resize cursor while a drag is active, as well as when inactive but where we are in the hit-area of a split.

PR #264, opened 2019-11-05
remove "Raw" indirection from basic widgets

futurepaul:
This wasn't obvious to me when I did it, but we don't need an actual distinct "Raw" version of each struct when we're wrapping a widget in Align. This PR brings the basic widgets in line with the style of the new Switch widget.

The drawback is that when the new() method on a widget returns a Align-wrapped widget, you lose access to methods on that struct. I solved that in the case of TextBox's tests by adding a raw() method. Maybe that's a better pattern? Also, perhaps core layout widgets could have a way of routing method calls to a child? Don't know if that's a can of worms.

Also used this opportunity to use winding hit testing in Slider and the newfound ability to get width off of the EventCtx instead of storing it as state!

(I'll probably have to rebase because this touches textbox?)

### DIFF
--- a/druid/src/widget/checkbox.rs
+++ b/druid/src/widget/checkbox.rs
@@ -23,19 +23,16 @@ use crate::{
 };
 
 /// A checkbox that toggles a boolean
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Default)]
 pub struct Checkbox;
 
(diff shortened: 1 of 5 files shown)


zaynetro:
Wow! This somehow wasn't obvious to me either.

PR #268, opened 2019-11-06
Add closures to AppDelegate

### LINKED ISSUE #242: Track window creation & deletion in AppDelegate
We currently don't have any way to respond to window creation or deletion.

It may be that we want an `Event` for this, but as a first attempt I'd like to try handling this via the `AppDelegate`.

I imagine `AppDelegate` getting two new optional 'methods', for `window_added` and `window_removed`. These work like the current `event` method, except that the first argument is a `WindowId` and not a an `Event`.

LiHRaM:
closes #242 

I've added two optional closures to AppDelegate, which can be used to track window creation and deletion.
I was unable to pass WinCtx along from `connect` and `remove_window`, and after consulting with @cmyr we decided to separate DelegateCtx from WinCtx. Other calls have been refactored to reflect this.

In my testing, I noticed that closing the window normally in `multiwin` does not emit the println, whereas using the close menu command does.

Any help/comments appreciated!

### DIFF
--- a/druid/src/app_delegate.rs
+++ b/druid/src/app_delegate.rs
@@ -19,20 +19,12 @@ use std::collections::VecDeque;
 use crate::{Command, Data, Env, Event, WinCtx, WindowId};
 
 /// A context passed in to [`AppDelegate`] functions.
-pub struct DelegateCtx<'a, 'b> {
+pub struct DelegateCtx<'a> {
     pub(crate) source_id: WindowId,
     pub(crate) command_queue: &'a mut VecDeque<(WindowId, Command)>,
-    pub(crate) win_ctx: &'a mut dyn WinCtx<'b>,
 }
 
-impl<'a, 'b> DelegateCtx<'a, 'b> {
-    /// Get the [`WinCtx`].
-    ///
-    /// [`WinCtx`] trait.WinCtx.html
-    pub fn win_ctx(&self) -> &dyn WinCtx<'b> {
-        self.win_ctx
-    }
-
+impl<'a> DelegateCtx<'a> {
     /// Submit a [`Command`] to be run after this event is handled.
     ///
     /// Commands are run in the order they are submitted; all commands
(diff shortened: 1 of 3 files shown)


druid/src/win_handler.rs
```diff
@@ -306,14 +306,41 @@ impl<T: Data + 'static> AppState<T> {
 
     fn connect(&mut self, id: WindowId, handle: WindowHandle) {
         self.windows.connect(id, handle);
+        if let Some(delegate) = &mut self.delegate {
```
cmyr: So these 10 lines or so are the main place here where I wonder if we can't reduce boilerplate. What I am imagining is a function on `self` with the signature,

```rust
fn with_delegate(&mut self, f: impl Fn(&mut AppDelegate, &mut DelegateCtx));
```

which internally does this whole dance of destructing self and creating the `DelegateCtx`, and then calls the passed closure with them.

If we had this, we could replace these 12 lines with,

```rust
self.with_delegate(|del, ctx| del.window_added(id, data, env, ctx));
```

Does that make sense?

druid/src/app_delegate.rs
```diff
@@ -56,13 +48,24 @@ impl<'a, 'b> DelegateCtx<'a, 'b> {
 ///
 /// You customize the `AppDelegate` by passing closures during creation.
 pub struct AppDelegate<T> {
-    event_fn: Option<Box<dyn Fn(Event, &mut T, &Env, &mut DelegateCtx) -> Option<Event> + 'static>>,
+    event_fn: Option<
```
cmyr: I agree that this complexity isn't worth it, and we should just back up and make `AppDelegate` a trait.

If you'd like to try that, it's probably a separate PR, and then we could redo this on top of that? Or I could make that change, whichever is easiest.

druid/src/app_delegate.rs
```diff
@@ -73,22 +76,63 @@ impl<T: Data> AppDelegate<T> {
...
     pub fn event_handler<F>(mut self, f: F) -> Self
     where
-        F: Fn(Event, &mut T, &Env, &mut DelegateCtx) -> Option<Event> + 'static,
+        F: Fn(Event, &mut T, &Env, &mut DelegateCtx, &mut dyn WinCtx) -> Option<Event> + 'static,
     {
         self.event_fn = Some(Box::new(f));
         self
     }
 
+    pub fn window_added_handler<F>(mut self, f: F) -> Self
```
cmyr: The one thing missing in this PR is docs for the public API.

druid/src/app_delegate.rs
LiHRaM: Yeah, I'd like to try my hand at that. I'll leave this as-is for now then.

LiHRaM:
closed in favor of #286

PR #270, opened 2019-11-06
Add WidgetPod::widget/mut_widget & Scroll::child/mut_child

cmyr:
These provide direct access to the underlying widgets.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -194,7 +194,7 @@ dependencies = [
 
 [[package]]
 name = "druid"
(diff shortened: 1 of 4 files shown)


cmyr:
@raphlinus can you clarify your concerns? I don't think I understand.

raphlinus:
Just that it might make it harder to, eg, refactor Scroll into separate container and clip view. But we can cross that bridge if and when we get to it. I'm fine with this going in as-is.

PR #271, opened 2019-11-07
Fix a bunch of clippy lints

cmyr:
Had let this get a bit out of hand; I think we will want to turn clippy on in CI at some point, so it's worth keeping on top of this stuff.

This also caught at least two bugs where I had the same statement in both branches of an `if { } else { }`. 😬


druid-shell/src/mac/mod.rs
```diff
@@ -13,7 +13,9 @@
 // limitations under the License.
 
 //! macOS implementation of window creation.
-#![allow(non_snake_case)]
+
+#![allow(non_snake_case,
+    clippy::let_unit_value,)]
```
raphlinus: This is a pretty classic example of why I haven't been a fan of clippy. In this case, it's directly telling us to do the wrong thing, because it doesn't have domain understanding of objC.

druid-shell/src/mac/mod.rs
```diff
@@ -484,7 +495,7 @@ extern "C" fn mods_changed(this: &mut Object, _: Sel, nsevent: id) {
     if down {
         (*view_state).handler.key_down(event, &mut ctx);
     } else {
-        (*view_state).handler.key_down(event, &mut ctx);
+        (*view_state).handler.key_up(event, &mut ctx);
```
raphlinus: Good catch! This is actually very strong evidence in favor of clippy.

druid-shell/src/mac/menu.rs
```diff
@@ -104,6 +104,12 @@ impl Menu {
     }
 }
 
+impl std::default::Default for Menu {
```
raphlinus: I'm slightly confused by these. First, we don't need `std::default::`, it's already in the prelude. Second, is this saying we should prefer `default` to `new` in the simple case?

cmyr: Ah, this is an editor macro I have to make writing these impls easier, i didn't realize `Default` was in the prelude.

The text for this lint is basically "You have a zero-argument fn `new()`; people may expect to be able to use `default` here. In the one place where I strongly disagreed (around `WindowId`) I changed the name of the fn from `new` to `next`.

My feeling on these others is basically: the impls aren't _bad_, per-se, and I prefer adding them to adding a blanket `allow(...)` statement.

druid-shell/src/mac/mod.rs
cmyr: Agreed, but also fair enough. `objc` is a weird case.

PR #273, opened 2019-11-07
Fix scrollbar bounds calculation & prevent scrollbar overlap

ForLoveOfCats:
When I first started on scrollbar mouse interaction this code perplexed me endlessly but I didn't want to change it any more than I had to in order to stay within the scope of that PR. However when I went to prevent the bars from overlapping at the bottom I realized that there was at least one bug hiding in there. Specifically when the bars reached their maximums they would often times be shorter then when not at their maximums. As it turns out the calculations it was using for figuring out the bar bounds was incorrect which was being hid with a `.min` hack.

Here I've fixed the calculations, improved readability, reduced code length, and prevented the bars from overlapping.

### DIFF
--- a/druid/src/widget/scroll.rs
+++ b/druid/src/widget/scroll.rs
@@ -178,45 +178,32 @@ impl<T: Data, W: Widget<T>> Scroll<T, W> {
         self.scroll_offset
     }
 
-    fn calc_scrollbar_screen_edges(&self, viewport: Rect) -> Rect {
-        Rect::new(
-            self.scroll_offset.x + SCROLL_BAR_PAD,
-            self.scroll_offset.y + SCROLL_BAR_PAD,
-            self.scroll_offset.x - SCROLL_BAR_PAD + viewport.width(),
-            self.scroll_offset.y - SCROLL_BAR_PAD + viewport.height(),
-        )
-    }
-
     fn calc_vertical_bar_bounds(&self, viewport: Rect) -> Rect {
-        let scrollbar_bounds = self.calc_scrollbar_screen_edges(viewport);
-
         let scale_y = viewport.height() / self.child_size.height;
 
-        let h = (scale_y * viewport.height()).ceil();
-        let dh = (scale_y * self.scroll_offset.y).ceil();
+        let top_y_offset = (scale_y * self.scroll_offset.y).ceil();
+        let bottom_y_offset = (scale_y * viewport.height()).ceil() + top_y_offset;
 
-        let x0 = scrollbar_bounds.x1 - SCROLL_BAR_WIDTH;
-        let y0 = scrollbar_bounds.y0 + dh;
+        let x0 = self.scroll_offset.x + viewport.width() - SCROLL_BAR_WIDTH - SCROLL_BAR_PAD;
+        let y0 = self.scroll_offset.y + top_y_offset + SCROLL_BAR_PAD;
 
-        let x1 = scrollbar_bounds.x1;
-        let y1 = (y0 + h).min(scrollbar_bounds.y1);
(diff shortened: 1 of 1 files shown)


ForLoveOfCats:
I think that in the long run it will be good to track which scrollbars are visible and use that to conditionally add the extra space in the lower right hand corner. However I figured that always having that space would be a good thing to have until then.

cmyr:
Thanks, will take a look when I get a chance, but always appreciate cleanup. I'm using this widget extensively in runebender and I'm bumping into lots of little issues; we may need to do some fairly significant rethinking of bits of this.

ForLoveOfCats:
Can you open a metabug of all the issues you have? I'd like to see what I can think of

zaynetro:
The reason scroll bars were overlapping is that that is the behaviour on macos. I am personally very used to it.

scroll-bars-overlap

Ref: https://jsfiddle.net/yb8chzxq/

ForLoveOfCats:
That's very interesting, TIL. I guess that is something that needs to be discussed then? To remove the overlap prevention is a 38 character change so it's not a big deal. If it is decided that they should continue to overlap I can make that change and force push.

The more important part is that the original calculations were buggy and that this patch greatly improves the simplicity and conciseness while fixing the bug which matters much more to me than the scrollbars overlapping.

ForLoveOfCats:
I think that part of why the overlapping scrollbars look decent in the macOS example is that they seem to be handling the borders differently.

Whereas in Druid:
image
image

So if the decision is to still have them overlap I think the default border style will have to change. To what I don't know.

zaynetro:
Thanks for fixing the calculations! So far it seems like everyone was just introducing things to their taste. I would like to see some kind of default design guidelines but this could be to big of a task.

cmyr:
I think that we're probably a little ways away from the 'design guidelines' stage; my feeling right now is that druid is an experiment, we're all trying things out, and that if things are a bit of a mess sometimes that's a reasonable price to pay for letting people move quickly and have a high level of autonomy. If druid starts to gain real ground, I expect we will end up rewriting tons of it based on the lessons we learn in the meantime. :)

For this particular concern: it's a great example of the way different platforms have different expectations. In the long-term, we are going to want the default styles to be per-platform. In the near-term, a reasonable solution might be the overlapping behaviour be determined by a flag, and have it default to `true` on macos and `false` elsewhere.

My own particular headache with the `Scroll` widget right now is around what the behaviour should be when the content of the scroll widget is _smaller_ than the scroll widget's parent; currently this gets weird.

ForLoveOfCats:
I'd like to take a look at that after this gets merged

cmyr:
Okay, I'm happy to merge this with the caveat that I would like the overlapping behaviour to come back as an option at some point.

ForLoveOfCats:
How about a bool in the theme? Something else? I'm waiting for this to be merged before adding the scroll stuff to the theme so this seems like the perfect opportunity.

cmyr:
A bool that can be set when the widget is constructed would be fine, I think.

ForLoveOfCats:
I was thinking that putting it in the theme would allow for the default to vary on different platforms, say having overlap on macOS but not on Windows/Linux

cmyr:
Yes, that's an option. Realistically, there's a much larger difference between scroll ui on windows and macOS than just the overlapping behaviour; windows doesn't (as far as I know, in standard apps) do floating sliders at all, and has the clickable buttons at edges that manually adjust the bar.

cmyr:
my gut feeling is that if scrollbars float they should be able to overlap. If anyone knows examples of floating scrollbars that don't overlap I would be curious to see them. (I haven't looked too hard, maybe they're all over the place!)

Issue #260, opened 2019-11-04
default window size is really small


cmyr:
In gtk we open very small windows by default. Window size should be a launch option (#232) and additionally we should default to the size of the last manually sized & positioned window; but in any case we want default windows bigger than they are now.

runiq:
Triage: Looks like the #274 merge didn't close this, but it should have.

cmyr:
thanks!

PR #274, opened 2019-11-08
Set window size in builder

### LINKED ISSUE #232: Support setting window size during creation
Currently window sizes are hard-coded.

This project has two components: the first is updating the various `WindowBuilder`s in druid-shell to get `set_size()` fns, and then using those sizes when building windows. The second part is exposing this in druid; this should be similar, adding a field and a builder method to `WindowDesc` in `app.rs`.

Although this will touch a number of different files, the actual work should be straight-forward, and would be a good issue for someone curious about Druid.

8176135:
Closes #232 and #260 

Couple of things to note:

- The Mac implementation has not been tested as I don't have Mac OS installed anywhere
- My Mac implementation doesn't take in to account DPI as the previous implementation didn't, not sure if it automatically does it, as I can't test it.
- Windows and Linux GTK has both been tested with different screen DPIs (Windows 10 and xfce Arch Linux)
- The DPI scaling on linux doesn't seem to work as expected. The window size is scaled up, but the contents are not. Other linux applications don't actually scale their window size up, but just scale the contents. I have no idea how to go about scaling up the contents.
- It might be better to remove the window size dpi scaling on linux for now.
- The current implementation allows for different default screen sizes per platform.


druid-shell/src/gtk/mod.rs
```diff
@@ -87,6 +87,8 @@ pub struct WindowBuilder {
     handler: Option<Box<dyn WinHandler>>,
     title: String,
     menu: Option<menu::Menu>,
+    width: f32,
```
cmyr: I would use the `kurbo::Size` type for this; it's the type we use to represent sizes elsewhere, and it's the type we will use in the public api in druid. 👌

druid/src/app.rs
```diff
@@ -44,6 +44,7 @@ type WidgetBuilderFn<T> = dyn Fn() -> Box<dyn Widget<T>> + 'static;
 pub struct WindowDesc<T> {
     pub(crate) root_builder: Arc<WidgetBuilderFn<T>>,
     pub(crate) title: Option<LocalizedString<T>>,
+    pub(crate) size: Option<(f32, f32)>,
```
cmyr: ditto, this can just be `Option<Size>`.

druid/src/app.rs
```diff
@@ -138,6 +140,12 @@ impl<T: Data + 'static> WindowDesc<T> {
         self
     }
 
+    /// Set the window size at creation
+    pub fn window_size(mut self, width: f32, height: f32) -> Self {
```
8176135: Should this take width and height as it is, and then call Size::new() inside, or just take in a Size?

8176135: Its more flexible to take in a Size, but when using a library it's slightly annoying to have to dig up more types to do something simple.

cmyr: I'm sympathetic to that. One work around we have is that there's a `From<(f64, f64)>` for `Size`, so the actual argument here can be `size: impl Into<Size>`, which will let people just pass a tuple.

`kurbo` is a foundational crate for `druid`, and is used all over the place. If this gets too painful we can consider a prelude at some point, but I tend to prefer being explicit about imports.

druid-shell/src/gtk/mod.rs
```diff
@@ -35,6 +35,7 @@ use win_main::with_application;
 use crate::clipboard::ClipboardItem;
 use crate::dialog::FileDialogOptions;
 use crate::keyboard;
+use crate::kurbo::Size;
```
cmyr: you can combine this with the other kurbo import on the next line. 👌

druid-shell/src/mac/mod.rs
```diff
@@ -53,6 +53,7 @@ use piet_common::{Piet, RenderContext};
 use crate::clipboard::ClipboardItem;
 use crate::dialog::{FileDialogOptions, FileDialogType};
 use crate::keyboard::{KeyCode, KeyEvent, KeyModifiers};
+use crate::kurbo::Size;
```
cmyr: ditto, there's already a kurbo import on L50.

druid-shell/src/windows/mod.rs
```diff
@@ -64,6 +64,7 @@ use crate::application::Application;
 use crate::clipboard::ClipboardItem;
 use crate::dialog::{FileDialogOptions, FileDialogType};
 use crate::keyboard::{KeyCode, KeyEvent, KeyModifiers};
+use crate::kurbo::Size;
```
cmyr: 😁

druid/examples/multiwin.rs
```diff
@@ -85,7 +85,9 @@ fn make_delegate() -> AppDelegate<State> {
     AppDelegate::new().event_handler(|event, data, _env, ctx| {
         match event {
             Event::Command(ref cmd) if cmd.selector == druid::command::sys::NEW_FILE => {
-                let new_win = WindowDesc::new(ui_builder).menu(make_menu(data));
+                let new_win = WindowDesc::new(ui_builder)
+                    .menu(make_menu(data))
+                    .window_size(data.selected as f64 * 100.0 + 300.0, 500.0);
```
cmyr: cool, thanks. This makes it much easier to test this PR. :)

8176135:
While testing the multiwin on linux, it actually seems like the File menu is missing? I'll open a new issue for this

druid/src/app.rs
```diff
@@ -138,6 +141,17 @@ impl<T: Data + 'static> WindowDesc<T> {
         self
     }
 
+    /// Set the window size at creation
```
cmyr: Thanks for the docs!

PR #275, opened 2019-11-08
Fix crash when typing on Windows

raphlinus:
When inserting text, the `update_scroll` method is called to measure the
new cursor position, but with a stale text layout. This patch rebuilds
the text layout when the text is changed.

I'm not sure if this is 100% correct, but fixes the panic.


cmyr:
@raphlinus ping?

raphlinus:
Yeah, I think this should be merged. I seem to remember there was another issue around multibyte characters, but I think that was a different issue.

PR #278, opened 2019-11-08
Add List Widget

zaynetro:
**TODO:**

* [x] Unify `Widget` implementations by having a list iteration trait

Will make a list after the weekend.


zaynetro:
OK. I have commited the changes to unify List implementations using the trait. I am not sure if this is approach that Raph was suggesting in the comment. So I am open to any feedback :)

The current solution doesn't seem to be that flexible. I couldn't even completely implement the trait for list with shared data...

druid/src/widget/list.rs
```diff
@@ -0,0 +1,204 @@
...
+        }
+    }
+}
+
+/// This iterator enables writing List widget for any `Data`.
+pub trait ListIter<T: Data>: Data {
+    /// Iterate over each data child.
+    fn for_each(&self, cb: impl FnMut(&T, usize));
+
+    /// Iterate over each data child. Keep track of changed data and update self.
```
raphlinus: I think `for_each_mut` might be a better name for this.

druid/examples/list.rs
```diff
@@ -0,0 +1,66 @@
...
+        0.0,
+    );
+
+    root.add_child(
+        Scroll::new(List::new(|| {
+            Box::new(
+                SizedBox::new(
+                    Container::new(Padding::new(
+                        10.0,
+                        DynLabel::new(|d, _| format!("Hello world #{}", d)),
```
raphlinus: Probably "List item #{}" is a better string for this.

druid/src/widget/list.rs
```diff
@@ -0,0 +1,204 @@
...
+use std::sync::Arc;
+
+use crate::kurbo::{Point, Rect, Size};
+
+use crate::{
+    BaseState, BoxConstraints, Data, Env, Event, EventCtx, LayoutCtx, PaintCtx, UpdateCtx, Widget,
+    WidgetPod,
+};
+
+pub struct List<T: Data> {
```
raphlinus: This needs a one-line docstring. I suggest:

```rust
/// A list widget for a variable-size collection of items.
```

zaynetro:
Thanks! I will merge this tomorrow.

I agree that having at least this version of List widget in druid is important.

PR #283, opened 2019-11-09
Refactor AppDelegate into a trait

LiHRaM:
While working on #268, @cmyr and I decided it might be worthwhile to implement `AppDelegate` as a trait rather than an object storing closures.
This is the initial attempt, and it feels nice and simple.

I implemented the trait for an empty struct in `multiwin`, which smells a bit.
This approach makes it possible for the AppDelegate to maintain state separately from the data which is passed to it via events, any thoughts on how that could be useful? Otherwise this approach might need some tweaking.

### DIFF
--- a/druid/examples/multiwin.rs
+++ b/druid/examples/multiwin.rs
@@ -34,9 +34,8 @@ struct State {
 fn main() {
     simple_logger::init().unwrap();
     let main_window = WindowDesc::new(ui_builder).menu(make_menu(&State::default()));
-    let delegate = Box::new(MultiwinDelegate {});
     AppLauncher::with_window(main_window)
-        .delegate(delegate)
+        .delegate(Delegate)
         .launch(State::default())
         .expect("launch failed");
 }
@@ -82,9 +81,9 @@ fn ui_builder() -> impl Widget<State> {
     col
 }
 
-struct MultiwinDelegate {}
+struct Delegate;
 
-impl AppDelegate<State> for MultiwinDelegate {
+impl AppDelegate<State> for Delegate {
     fn event(
         &mut self,
         event: Event,
--- a/druid/examples/multiwin.rs
+++ b/druid/examples/multiwin.rs
@@ -17,7 +17,7 @@
 use druid::menu::{ContextMenu, MenuDesc, MenuItem};
 use druid::widget::{Align, Button, Column, Label, Padding, Row};
 use druid::{
-    AppDelegate, AppLauncher, Command, Data, DelegateCtx, Event, EventCtx, LocalizedString,
+    AppDelegate, AppLauncher, Command, Data, DelegateCtx, Env, Event, EventCtx, LocalizedString,
     Selector, Widget, WindowDesc,
 };
 
@@ -34,8 +34,9 @@ struct State {
 fn main() {
     simple_logger::init().unwrap();
     let main_window = WindowDesc::new(ui_builder).menu(make_menu(&State::default()));
+    let delegate = Box::new(MultiwinDelegate {});
     AppLauncher::with_window(main_window)
-        .delegate(make_delegate())
+        .delegate(delegate)
         .launch(State::default())
         .expect("launch failed");
 }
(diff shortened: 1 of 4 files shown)


cmyr:
An impl for an empty struct is totally reasonable.

The 'should the app delegate have secret state' question is a deep one. I really don't know; the main justification for the previous design was to make this impossible, but I didn't have a good justification for wanting this to be impossible. 

More and more, lately, one of my main design principles for druid is that we should make it easy for framework users to implement the hacks they need to get their apps working, and a stateful app delegate is another tool in that drawer.

druid/examples/multiwin.rs
```diff
@@ -81,8 +82,16 @@ fn ui_builder() -> impl Widget<State> {
     col
 }
 
-fn make_delegate() -> AppDelegate<State> {
-    AppDelegate::new().event_handler(|event, data, _env, ctx| {
+struct MultiwinDelegate {}
```
cmyr: For a struct with no fields I prefer the 'unit struct' syntax, e.g. you can just write,

```rust
struct Delegate;
```

druid/src/app.rs
```diff
@@ -70,7 +70,7 @@ impl<T: Data + 'static> AppLauncher<T> {
     /// Set the [`AppDelegate`].
     ///
     /// [`AppDelegate`]: struct.AppDelegate.html
-    pub fn delegate(mut self, delegate: AppDelegate<T>) -> Self {
+    pub fn delegate(mut self, delegate: Box<dyn AppDelegate<T>>) -> Self {
```
cmyr: I would make this argument be, `impl AppDelegate<T>`, and then we can box inside the function; this just saves the caller from the hassle of having to box before they call.

(you might also need a `+ 'static` after the type)

druid/src/app_delegate.rs
```diff
@@ -57,46 +55,18 @@ type EventFn<T> = dyn Fn(Event, &mut T, &Env, &mut DelegateCtx) -> Option<Event>
...
-            None => Some(event),
-        }
-    }
-}
-
-impl<T: Data> Default for AppDelegate<T> {
-    fn default() -> Self {
-        AppDelegate::new()
-    }
+    ) -> Option<Event>;
```
cmyr: I would give this a default impl, which which is just `Some(event)`, e.g. it just passes through the event unchanged.

My thinking is that, on the app delegate, all methods are optional.

LiHRaM: Ah, yes, I had this thought too, but I've just iterated through several redesigns and totally forgot it. I'll add that one for sure. :+1:

LiHRaM: So the compiler complains about unused variables, but I don't want to underscore the names, as they'll be part of the docs. Ergo the current approach, as suggested in https://github.com/rust-lang/rust/issues/26487.

druid/src/app_delegate.rs
```diff
@@ -57,46 +55,23 @@ type EventFn<T> = dyn Fn(Event, &mut T, &Env, &mut DelegateCtx) -> Option<Event>
...
-        }
-    }
-}
-
-impl<T: Data> Default for AppDelegate<T> {
-    fn default() -> Self {
-        AppDelegate::new()
+        let _ = data;
+        let _ = env;
+        let _ = ctx;
```
cmyr: what's this about?

cmyr: oh, unused warnings? I would annotate this function `#[allow(unused)]`.

LiHRaM: Yeah, I was following the discussion mentioned in https://github.com/xi-editor/druid/pull/283#discussion_r344462460, and it seemed to be the most correct solution, but the attribute is also fine with me. 😄

LiHRaM: Done!

cmyr: ah gotcha, missed that comment. We have adopted the pattern of using `#[allow]` on each function with this problem, so let's just keep doing that. 🤷‍♂

cmyr:
feel free to merge. :)

PR #285, opened 2019-11-09
Add scrollbar constants and magic numbers to theme

ForLoveOfCats:
This *should be* (:crossed_fingers:) a short and sweet patch which adds a number of scrollbar constants to the theme. I didn't add an option to enable/disable scrollbar overlap as it seemed like more discussion is required about how to handle overlap, floating, and default style on different platforms.

### DIFF
--- a/druid/src/env.rs
+++ b/druid/src/env.rs
@@ -69,6 +69,7 @@ pub enum Value {
     Color(Color),
     LinearGradient(Arc<LinearGradient>),
     Float(f64),
+    UnsignedInt(u64),
     String(String),
 }
 
@@ -173,6 +174,7 @@ impl Debug for Value {
             // TODO: make PaintBrush impl debug?
             Value::LinearGradient(g) => write!(f, "LinearGradient {:?}", g),
             Value::Float(x) => write!(f, "Float {}", x),
+            Value::UnsignedInt(x) => write!(f, "UnsignedInt {}", x),
             Value::String(s) => write!(f, "String {:?}", s),
         }
     }
@@ -221,6 +223,7 @@ impl Value {
             (Color(_), Color(_)) => true,
             (LinearGradient(_), LinearGradient(_)) => true,
             (Float(_), Float(_)) => true,
+            (UnsignedInt(_), UnsignedInt(_)) => true,
             (String(_), String(_)) => true,
             _ => false,
         }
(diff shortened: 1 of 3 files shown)


druid/src/env.rs
```diff
@@ -69,6 +69,7 @@ pub enum Value {
     Color(Color),
     LinearGradient(Arc<LinearGradient>),
     Float(f64),
+    UnsignedFloat(u64),
```
cmyr: this seems a little odd. Maybe not meant to be committed? It doesn't seem to be used anywhere, and the internal type is not a float. It's possible we will want an `Integer` type here at some point (and maybe even an `UnsignedInteger`) but if we can get by with just `Float` for now that's great.

edit: okay, I see this is used for duration. so `UnsignedInt`?

druid/src/theme.rs
```diff
@@ -42,6 +42,12 @@ pub const BORDERED_WIDGET_HEIGHT: Key<f64> = Key::new("bordered_widget_height");
 
 pub const SCROLL_BAR_COLOR: Key<Color> = Key::new("scroll_bar_color");
 pub const SCROLL_BAR_BORDER_COLOR: Key<Color> = Key::new("scroll_bar_border_color");
+pub const SCROLL_BAR_MAX_OPACITY: Key<f64> = Key::new("scroll_bar_max_opacity");
+pub const SCROLL_BAR_FADE_DELAY: Key<u64> = Key::new("scroll_bar_fade_time");
```
cmyr: these first two items are borderline, to me, in terms of whether they should be in a theme. This basically touches on some very foundational ideas about what a theme is supposed to be and what it's supposed to do. Do we want to be tuning these things per-platform, or to generally let the user tune them? It's possible that we do, but if we really want to be 'native' feeling per-platform it's likely going to take more than this.

Anyway, I'm rambling. I don't think it's bad for these to be here, per-se, I just think it's not quite clear exactly what the scope of a theme is, and this is something we're going to have to figure out. 👌

ForLoveOfCats: I placed these here as someone who has plans to use Druid in an application I know that these are values I want control over without modifying Druid itself as a dep for said application

druid/src/env.rs
ForLoveOfCats: Ooooooh that is such a goof, I was under the impression that `u64` was an unsigned 64-bit float not an int :man_facepalming: I'll correct this and force push

ForLoveOfCats: Annnnd corrected

ForLoveOfCats:
Wow... I am honored. I guess I'll have to be a lot more careful as to what remote I push to eek!

PR #286, opened 2019-11-09
Track window creation and deletion: Take 2

### LINKED ISSUE #242: Track window creation & deletion in AppDelegate

LiHRaM:
closes #242 

This is the second attempt at adding event handlers for tracking window creation and deletion, this time using the trait implemented in #283.

### DIFF
--- a/druid/src/win_handler.rs
+++ b/druid/src/win_handler.rs
@@ -308,8 +308,32 @@ impl<T: Data + 'static> AppState<T> {
             .and_then(|w| w.get_menu_cmd(cmd_id))
     }
 
+    fn with_delegate(
+        &mut self,
+        id: WindowId,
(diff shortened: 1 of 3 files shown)


druid/src/win_handler.rs
```diff
@@ -308,16 +308,44 @@ impl<T: Data + 'static> AppState<T> {
             .and_then(|w| w.get_menu_cmd(cmd_id))
     }
 
+    fn with_delegate(
```
cmyr: cool, glad this worked 👍

PR #298, opened 2019-11-12
EditableText trait

futurepaul:
This is work in the general direction of making the textbox generic on an `EditableText` trait. This is to serve the longterm goal of being able to swap-in Xi's Rope or another data structure to back an editable text field. One of the perks is we finally have some serious backspace logic!

I need to do a ton of cleanup and documenting. Also a lot of the methods aren't in use, but were implemented just because I didn't know which specific ones I was going to use. But wanted to check if I'm headed in the right direction before I wade through all that. 

I hope I'm not cargo-culting Xi-Rope too much. I believe I've gotten rid of everything Rope-related, but it's quite possible I have some methods in here that won't do me much good in String land.


cmyr:
Cool, very happy to see you taking this on.

I think the major question is, "where does this live"? That is to say, I could see an argument for it being in a separate repo, or in the xi repo, or.. those are the two main options.

So maybe having it here for now makes sense, and we can revisit if someone else wants to use it?

raphlinus:
I vote for druid for now, factor it out when that starts looking like it makes sense. Even if there is an edit-engine crate (which I've actually started prototyping), I can see it taking an optional dep on druid just like norad does now, though I see value in a different structure.

(Haven't dug in to the code yet, juggling too much other stuff right now, but looks fun)

druid/src/widget/textbox/editable_text.rs
```diff
@@ -0,0 +1,212 @@
...
+    // pub fn at_or_next_codepoint_boundary(&self, offset: usize) -> Option<usize>;
+
+    // pub fn at_or_prev_codepoint_boundary(&self, offset: usize) -> Option<usize>;
+
+    fn prev_grapheme_offset(&self, offset: usize) -> Option<usize>;
+
+    fn next_grapheme_offset(&self, offset: usize) -> Option<usize>;
+}
+
+impl EditableText for &str {
```
cmyr: thinking about this more, I don't know if makes sense for us to think of a string slice as 'editable', since it generally isn't. (For instance, you can't change the size of a string slice).

In fact, I think we can get away with having this only impl'd for `String`. Because the trait defines functions as taking `&self` and `&mut self`, it doesn't make much sense to impl it for `&String`; in that case the `self` type in the methods would be `&&String` and `&mut &String`.

druid/src/widget/textbox/editable_text.rs
```diff
@@ -0,0 +1,212 @@
...
+    fn set(&mut self, position: usize);
+
+    // get cursor position
+    fn pos(&self) -> usize;
+
+    fn is_boundary(&self) -> bool;
+
+    // moves cursor to previous boundary if exists
+    // else becomes invalid cursor
+    fn prev(&mut self) -> Option<(usize)>;
```
cmyr: This is returning a tuple?

druid/src/widget/textbox/editable_text.rs
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    /// Gets the previous grapheme from the given index.
+    fn prev_grapheme_offset(&self, from: usize) -> Option<usize> {
+        let mut c = GraphemeCursor::new(from, self.len(), true);
+        c.prev_boundary(self, 0).unwrap()
+    }
+}
+
+pub trait EditableTextCursor<'a> {
```
cmyr: To check my understanding: each type that implements `EditableText` will have a corresponding cursor type?

In this case, we may want to reflect this relationship in the type system, where possible. In particular, the constructor for the cursor should take the type of the corresponding `EditableText`.

This is a bit tricky. One option would be to have the `EditableText` type by an associated type of `EditableTextCursor`, so you'd have something like

```rust
pub trait EditableTextCursor<'a> {
    type Text: EditableText;
    
    fn new(s: &'a Self::Text, position: usize) -> Self;

    // ...
}
```

but it might be simpler to go the other way, and have the cursor be the associated type of the text, e.g.

```rust
pub trait EditableText {
   type Cursor: EditableTextCursor<Self>;

   fn cursor(&self, position: usize) -> Self::Cursor;
}
```

Which of these makes the most sense probably comes down to how the API is used, which type gets passed around more, etc. What we want to be able to do is to remove references to concrete types like `str` & `String` where possible, and doing this requires making the types 'aware' of each other, somehow.

futurepaul:
Okay some pretty big progress: TextBox is now generic over EditableText.

My big remaining question before I clean up and document this: what methods belong in EditableText, and what methods belong in EditableTextCursor?

Alternatively: is EditableTextCursor even useful as a separate trait? Could we just fold its methods into EditableText? The only place it's really being used right now is the backspace logic and I don't think it would be difficult to change that.

druid/src/widget/textbox/editable_text.rs
futurepaul: I went with this latter version, which makes sense to me. But also: we might not even need the separate cursor type?

futurepaul:
This is feeling pretty good to me. I went ahead and kept `EditableTextCursor` separate, I don't know if I can exactly explain the distinction but it feels distinct enough to me. There are still some todos but this might be a good checkpoint if it looks correct to y'all?

I didn't mention before, but this PR adds a dependency for backspace: `xi-unicode = "0.2.0"` Guessing the bloat check won't be too happy about this.

druid-shell/src/platform/mac/clipboard.rs
```diff
@@ -34,6 +34,7 @@ impl Clipboard {
             unsafe {
                 let nsstring = util::make_nsstring(s);
                 let pasteboard: id = msg_send![class!(NSPasteboard), generalPasteboard];
+                let _: NSInteger = msg_send![pasteboard, clearContents];
```
cmyr: it might be worth making this a separate patch? it's a bug unrelated to this PR, and its fate isn't tied to this work.

druid/src/widget/textbox/editable_text.rs
```diff
@@ -0,0 +1,326 @@
...
+    fn prev_codepoint_offset(&self, offset: usize) -> Option<usize>;
+
+    /// Get the previous codepoint offset from the given offset, if it exists.
+    fn next_codepoint_offset(&self, offset: usize) -> Option<usize>;
+}
+
+impl EditableText for String {
+    type Cursor = StringCursor;
+
+    //TODO: this can panic so maybe we should return a Result?
```
cmyr: this should not be able to panic. I think you're thinking of the case where the new string is larger than the provided interval? In this case you would split self at the start of the interval, append the new string, and then append the remaining old string on that.

druid/src/widget/textbox/editable_text.rs
```diff
@@ -0,0 +1,326 @@
...
+
+/// An EditableText trait.
+pub trait EditableText: Sized {
+    type Cursor: EditableTextCursor<Self>;
+
+    /// Create a cursor with a copy of the text and a offset position.
+    fn cursor(&self, position: usize) -> Self::Cursor;
+
+    /// Replace range with new text.
+    fn edit(&mut self, iv: Range<usize>, new: impl Into<String>);
```
cmyr: It might make more sense for this to not take `Into<String>` but instead take `Into<Self>`?

druid/src/widget/textbox/editable_text.rs
```diff
@@ -0,0 +1,326 @@
...
+
+    /// Get slice of text at range.
+    fn slice(&self, iv: Range<usize>) -> Option<&str>;
```
cmyr: making this return a `&str` requires self to basically always store its contents as a contiguous sequence of bytes, in other words to be a `String`. It isn't possible for a rope to necessarily return its contents as string slices, because it stores the text in chunks, and a slice could span multiple leaves.

A simple alternative would be to return a `Cow<str>`; this would be a slice in the general case, but could be a new string if necessary.

druid/src/widget/textbox/editable_text.rs
```diff
@@ -0,0 +1,326 @@
...
+    fn slice(&self, iv: Range<usize>) -> Option<&str> {
+        self.get(iv)
+    }
+
+    fn len(&self) -> usize {
+        self.len()
+    }
+
+    fn cursor(&self, position: usize) -> Self::Cursor {
+        // TODO: can we get rid of this .to_string()?
```
cmyr: this cursor should borrow self.

druid/src/widget/textbox/editable_text.rs
```diff
@@ -0,0 +1,326 @@
...
+        // TODO: what happens if this position isn't at a valid offset?
```
cmyr: We have options, I think.

The first option I see is to always advance or retreat to the first valid offset when creating a cursor, to ensure that `cursor.pos()` always returns a valid boundary; in this case we would want to document this explicitly.

The second option is to not validate the initial position, but to explicitly state that a cursor's position is not guaranteed to be valid until one of the cursor movement functions has been called.

So the first is 'easier' but might lead to unexpected behaviour; do we want to move forward or backwards? The second is clearer, but easier to misuse.

druid/src/widget/textbox/editable_text.rs
```diff
@@ -0,0 +1,326 @@
...
+
+    /// Return current offset if it's a boundary, else next.
+    fn at_or_next(&mut self) -> Option<usize>;
+
+    /// Return current offset if it's a boundary, else previous.
+    fn at_or_prev(&mut self) -> Option<usize>;
+}
+
+#[derive(Debug, Clone)]
+pub struct StringCursor {
```
cmyr: yea, you want something like,

```rust
pub struct StringCursor<'a> {
    text: &'a str,
    position: usize,
}
```

futurepaul:
This still needs work but I finally made some forward progress on the lifetimes stuff (by cheating). Maybe we can find some time to chat next week?

druid/src/widget/textbox/editable_text.rs
futurepaul: I ended up returning an Option to give the consumer a chance to react to a cursor being bad.

futurepaul:
Okay, I made a few tweaks, rebased, and added some documentation.

As per a conversation with @cmyr, I'm punting on making this truly generic due to a hairy lifetimes issue. If anyone knows how to make the EditableText trait not specific about StringCursor and yet not having the StringCursor lifetimes leak up uncontrollably, let me know!

I'm pretty sure this will need to evolve some more before we can be confident it's the right abstraction, but I feel like this could be a good checkpoint. TextBox is a hot commodity and don't want to get too out of step.

futurepaul:
Refactored as discussed on zulip (move the text editing logic to its own module).

Only new thing is that I had to add type hints to the tests in `widget_ext`, likely due to something weird I'm doing with types. Any and all suggestions are welcome to make TextBox have all the wins of the EditableText trait but none of the downsides of not being explicitly `String`.

druid/src/widget/textbox.rs
```diff
@@ -32,103 +29,54 @@ use crate::piet::{
...
     width: f64,
     hscroll_offset: f64,
     selection: Selection,
     cursor_timer: TimerToken,
     cursor_on: bool,
+    phantom: std::marker::PhantomData<E>,
 }
 
-impl TextBox {
+impl<E: 'static + EditableText + Data + std::string::ToString> TextBox<E> {
```
cmyr: I think since we're somewhat compromising on the genericness of `EditableText`, we can get away with just keeping this as `TextBox<String>` for now.

Issue #246, opened 2019-10-31
derive-data does not work when a named or tuple struct has no fields


cmyr:
e.g,

```rust
#[derive(Clone, Data)]
struct Named {}

#[derive(Clone, Data)]
struct Tuple ();
```

will both produce a fairly unhelpful compiler error.

I believe that these should get implementations that always return `true'; alternatively if we don't want to allow this we should return a compiler error that communicates the issue.

PR #303, opened 2019-11-12
Allow #[derive(Data)] on empty structs

theduke:
Detect structs with no fields and generate no-op
compare function accordingly.

### DIFF
--- a/druid-derive/src/attr.rs
+++ b/druid-derive/src/attr.rs
@@ -69,6 +69,10 @@ impl Fields {
         Ok(Fields { kind, fields })
     }
 
+    pub fn len(&self) -> usize {
+        self.fields.len()
+    }
+
     pub fn iter(&self) -> impl Iterator<Item = &Field> {
         self.fields.iter()
     }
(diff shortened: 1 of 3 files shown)


theduke:
Useful while iterating. The compile errors are quite off-putting otherwise.

cmyr:
yea this was an open bug, #246. Thanks!

theduke:
Oh, I didn't even see the bug. In that case: 

Fixes #246

cmyr:
I'd love to see tests for this; they can go anywhere in `druid-derive/tests`. The following should all compile:

```rust
#[derive(Clone, Data)]
struct Empty ();

#[derive(Clone, Data)]
struct Empty2 {}

#[derive(Clone, Data)]
enum Items {
One,
Two(),
Three{},
}
```

theduke:
I actually was about to suggest that. 

Do you have a preference for where to put lens tests? `src/lens/test_derive.rs`?

theduke:
Done

theduke:
Ah no I wasn't done yet, still some work to do on enums.

cmyr:
haha sorry. looked like a good checkpoint in any case, so a follow up sounds good.

PR #304, opened 2019-11-12
Toggle switch widget animations

scholtzan:
I played around a bit and added animations to the toggle switch widget:

out


raphlinus:
> This looks amazing.

Yes, I'm really impressed.

PR #301, opened 2019-11-12
Readme 2.0

futurepaul:
It should be pretty obvious where this is rough (I'm leaning heavily on TKTK). Please fill in the blanks / rewrite at will!

Concepts section might be too much info and code? Also wasn't sure how to order them. What is the logical progression of understanding druid?

PR #306, opened 2019-11-13
Fix clippy

scholtzan:
I forgot to push a fix for a clippy complaint in the toggle switch widget.


raphlinus:
There's also a fix for this in #301. It might be cleaner to get this in and take the change out of #301, but I don't have a strong feeling either way.

scholtzan:
Oh, didn't see it there.

PR #307, opened 2019-11-15
Change Split to keep dragging until MouseUp

kbalt:
I'd like to make a small change to the new Split Widget.
Currently when the MouseMoved Event is handled there is a if branch checking if the widget is no longer hot.
When that check succeeds the dragging stops.
This feels unintuitive and is inconsistent with most common UIs.

This PR removes that block, causing the the drag to keep happening until the MouseUp event is triggered.

### DIFF
--- a/druid/src/widget/split.rs
+++ b/druid/src/widget/split.rs
@@ -279,12 +279,10 @@ impl<T: Data> Widget<T> for Split<T> {
                 }
                 Event::MouseMoved(mouse) => {
                     if ctx.is_active() {
-                        if !ctx.is_hot() {
-                            ctx.set_active(false);
-                        }
                         self.update_splitter(ctx.base_state.size(), mouse.pos);
                         ctx.invalidate();
                     }
+
                     if ctx.is_hot() && self.splitter_hit_test(ctx.base_state.size(), mouse.pos) {
                         match self.split_direction {
                             Axis::Vertical => ctx.set_cursor(&Cursor::ResizeUpDown),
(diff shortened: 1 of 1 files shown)


Zarenor:
I had wondered a bit about opinions on this behavior when I wrote the behavior the way I did. I will admit it's a little... clunky, as now implemented. 
However, the problems I was guarding against is some uncertainty in how the events would be received, and if the simple algorithm I wrote for laying out the widget and it's children would react sanely to drags outside the possible bounds. I do have some mandatory padding in the `adjust_splitter` method that may keep us out of the woods here. 

The situation I wanted to be careful we didn't end up in is:
- User begins dragging splitter
- User drags outside window, releases the mouse
- Our state is still 'splitter is active', and we need another `MouseUp` event with the left button to un-stick

If this doesn't happen - or we'd consider it a bug in the -shell libraries for us to not receive that `MouseUp`, then I support this change.

kbalt:
I was only able to really test this on Linux - it worked there and didn't seem to cause any problems in that regard.

The `MouseUp` event should always be is delivered because the window is still the 'active' one even if the cursor is not inside the bounds of it.

And as you said inside your `adjust_splitter` method you limit the min/max position of the splitter which nicely positions it at the start or end of the split widget when the cursor is outside the bounds of that axis.

futurepaul:
I'm not at a computer but I'm pretty sure Slider works off window.

cmyr:
What's the status of this?

kbalt:
That depends on how you guys want the behavior to be.
For me personally the changes make the widget a lot nicer to use.

cmyr:
@kbalt okay, sorry for the delay, this just got lost in the shuffle.

I agree that this feels much better, so happy to merge. If mouse up events are not delivered when the mouse is released outside of the view, I would consider that a bug in druid-shell, and we can deal with that then.

Thanks for the patch, and sorry again for letting it sit around, it's much appreciated!

PR #308, opened 2019-11-15
Stepper widget

scholtzan:
I did some further experimenting and implemented a stepper widget:
Screenshot 2019-11-15 at 10 18 55 AM

I tried to hook it up to a textbox but for it to work textbox would need some kind of `on_change` callback to update the stepper value.

### DIFF
--- a/druid/examples/switch.rs
+++ b/druid/examples/switch.rs
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use druid::widget::{Flex, Switch, DynLabel, Label, Padding, Row, Stepper, Switch};
-use druid::{AppLauncher, Data, Lens, LensWrap, Widget, WindowDesc};
+use druid::widget::{Flex, Label, Padding, Parse, Stepper, Switch, TextBox};
+use druid::{AppLauncher, Data, Lens, LensExt, LensWrap, Widget, WindowDesc};
 
 #[derive(Clone, Data, Lens)]
 struct DemoState {
@@ -30,22 +30,30 @@ fn build_widget() -> impl Widget<DemoState> {
     row.add_child(Padding::new(5.0, switch_label), 0.0);
     row.add_child(Padding::new(5.0, switch), 0.0);
 
-    let label_stepper = LensWrap::new(
+    let stepper = LensWrap::new(
         Stepper::new(0.0, 10.0, 0.25, true, |_ctx, _data, _env| {}),
-        lenses::demo_state::stepper_value,
+        DemoState::stepper_value,
     );
 
-    let mut stepper_row = Row::new();
+    let mut textbox_row = Flex::row();
+    let textbox = LensWrap::new(
+        Parse::new(TextBox::new()),
+        DemoState::stepper_value.map(|x| Some(*x), |x, y| *x = y.unwrap_or(0.0)),
+    );
+    textbox_row.add_child(Padding::new(5.0, textbox), 0.0);
+    textbox_row.add_child(Padding::new(5.0, stepper), 0.0);
 
-    let label = DynLabel::new(|data: &DemoState, _env| {
+    let mut label_row = Flex::row();
+
+    let label = Label::new(|data: &DemoState, _env: &_| {
         format!("Stepper value: {0:.2}", data.stepper_value)
     });
 
-    stepper_row.add_child(Padding::new(5.0, label), 0.0);
-    stepper_row.add_child(Padding::new(5.0, label_stepper), 0.0);
+    label_row.add_child(Padding::new(5.0, label), 0.0);
 
     col.add_child(Padding::new(5.0, row), 1.0);
-    col.add_child(Padding::new(5.0, stepper_row), 1.0);
+    col.add_child(Padding::new(5.0, textbox_row), 1.0);
+    col.add_child(Padding::new(5.0, label_row), 1.0);
     col
 }
 
--- a/druid/examples/switch.rs
+++ b/druid/examples/switch.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use druid::widget::{Align, Flex, Label, Padding, Parse, Stepper, Switch, TextBox};
+use druid::widget::{Flex, Label, Padding, Parse, Stepper, Switch, TextBox, WidgetExt};
 use druid::{AppLauncher, Data, Lens, LensExt, LensWrap, Widget, WindowDesc};
 
 #[derive(Clone, Data, Lens)]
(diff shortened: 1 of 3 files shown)


cmyr:
For this to work we need some sort of 'validating textbox' that validates its inputs, and can then have it's `Data` be a numerical type; when an edit happens we check that the new content of the textbox is valid data; if it is we mutate the data and if it isn't we discard the edit.

cmyr:
Or rather: that's the mechanism I think this should be built on. We will need data-validation for text entry at some point anyway, so it makes sense to reuse that for this widget.

Ralith:
Currently exploring such a mechanism in #346.

cmyr:
@scholtzan sorry to let this sit around for so long, it does look like a solid patch. Maybe it's worth seeing if the basic input validation stuff in #346 helps you at all?

scholtzan:
Thanks! No worries, I've been quite busy the past couple of weeks anyway. Probably will pick it up again after Christmas.

scholtzan:
Alright, I updated the stepper widget example and got it work with textboxes thanks to `Parse`.

druid/src/widget/stepper.rs
```diff
@@ -0,0 +1,208 @@
...
+
+                if mouse.pos.y > height / 2. {
+                    self.decrease_active = true;
+                } else {
+                    self.increase_active = true;
+                }
+
+                self.change_value(ctx, data, env);
+
+                let delay = Instant::now() + Duration::from_millis(500);
```
cmyr: maybe make these `Duration`s be `consts` with descriptive names?

druid/src/widget/stepper.rs
```diff
@@ -0,0 +1,208 @@
...
+                self.change_value(ctx, data, env);
+                let delay = Instant::now() + Duration::from_millis(200);
+                self.timer_id = ctx.request_timer(delay);
+            }
+            _ => (),
+        }
+    }
+
+    fn update(&mut self, ctx: &mut UpdateCtx, _old_data: Option<&f64>, _data: &f64, _env: &Env) {
+        ctx.invalidate();
```
cmyr: in general in `update` we want to only invalidate if the data has changed. The idiomatic way of writing this would be,

```rust
if !Some(data).same(old_data) {
    ctx.invalidate()
}
```

druid/src/widget/stepper.rs
```diff
@@ -0,0 +1,208 @@
...
+        };
+
+        if self.decrease_active {
+            paint_ctx.fill(decrease_button_rect, &active_gradient);
+        } else {
+            paint_ctx.fill(decrease_button_rect, &inactive_gradient);
+        };
+
+        // draw up and down triangles
+        let mut increase_button_arrow = BezPath::new();
```
cmyr: You should be able to have both arrows in one path; after `close_path` you can do `move_path` again.

druid/src/widget/stepper.rs
```diff
@@ -0,0 +1,208 @@
...
+    /// A closure that will be invoked when the value changed.
+    value_changed: Box<dyn Fn(&mut EventCtx, &mut f64, &Env)>,
+    /// Keeps track of which button is currently triggered.
+    increase_active: bool,
+    decrease_active: bool,
+    timer_id: TimerToken,
+}
+
+impl Stepper {
+    pub fn new(
```
cmyr: I might consider splitting some of these fields into builder methods. By default I might only require the `value_changed` argument; other arguments could be optional? I think that reasonable defaults exists (`std::f64::MIN`, `std::f64::MAX`, `1.0`, `false`)?

druid/src/widget/stepper.rs
cmyr: if #454 lands this snippet won't work anymore, my alternative would be:

```rust
if old.map(|old| !old.same(new)).unwrap_or(true) { ... }
```

scholtzan:
Ready for another look

druid/src/widget/stepper.rs
```diff
@@ -42,26 +46,37 @@ pub struct Stepper {
...
-                min,
-                step,
-                wrap,
-                value_changed: Box::new(value_changed),
-                increase_active: false,
-                decrease_active: false,
-                timer_id: TimerToken::INVALID,
-            },
-        )
+    pub fn new(value_changed: impl Fn(&mut EventCtx, &mut f64, &Env) + 'static) -> Self {
```
cmyr: looking at this and at your example, i'm wondering if we even need _this_ argument? Unlike a `Button`, `Stepper` is mutating the data, and so this function feels like something that would only be used in special cases?

It's also possible that we don't need it at all?

druid/src/widget/stepper.rs
```diff
@@ -42,26 +46,37 @@ pub struct Stepper {
...
+            step: 1.0,
+            wrap: false,
+            value_changed: Box::new(value_changed),
+            increase_active: false,
+            decrease_active: false,
+            timer_id: TimerToken::INVALID,
+        }
+    }
+
+    pub fn max(mut self, max: f64) -> Self {
```
cmyr: I'd like docs on public methods please :)

druid/examples/switch.rs
```diff
@@ -41,7 +45,7 @@ fn build_widget() -> impl Widget<DemoState> {
         DemoState::stepper_value.map(|x| Some(*x), |x, y| *x = y.unwrap_or(0.0)),
     );
     textbox_row.add_child(Padding::new(5.0, textbox), 0.0);
-    textbox_row.add_child(Padding::new(5.0, stepper), 0.0);
+    textbox_row.add_child(Padding::new(5.0, Align::centered(stepper)), 0.0);
```
cmyr: not a necessary change for this patch, but I think in general I'd like the `WidgetExt` methods to be the default way we demonstrate building up combined widgets, so this can be,

```rust
stepper.center().padding(5.0)
```

druid/src/widget/stepper.rs
scholtzan: Right, it doesn't really seem necessary

scholtzan:
Thanks for the comments. Should be ready for another look.

druid/examples/switch.rs
```diff
@@ -29,14 +30,40 @@ fn build_widget() -> impl Widget<DemoState> {
...
+        DemoState::stepper_value,
+    );
+
+    let mut textbox_row = Flex::row();
+    let textbox = LensWrap::new(
+        Parse::new(TextBox::new()),
+        DemoState::stepper_value.map(|x| Some(*x), |x, y| *x = y.unwrap_or(0.0)),
+    );
+    textbox_row.add_child(Padding::new(5.0, textbox), 0.0);
+    textbox_row.add_child(Padding::new(5.0, stepper.center().padding(5.0)), 0.0);
```
cmyr: I'm sorry, I've led you astray. This is now adding two separate padding widgets, let's just stick with the old style and be consistent. 😅 

```suggestion
    textbox_row.add_child(Padding::new(5.0, stepper.center()), 0.0);
```

scholtzan: whoops

PR #309, opened 2019-11-15
druid-shell/src/mac: call finishLaunching, set window level to 

### LINKED ISSUE #43: Make mac Window creation smoother

ratmice:
CGWindowLevelkey normalWindow.

I can't test this not having a mac, but this is an attempt to fix #43, 
So, i would appreciate it if someone with a mac could test it.

I believe it should work because finishLaunching should initiate the connection between the app and the window server allowing the setWindowLevel_ (called before run) to be received by the window sever.

As it is the app -> window server event mechanism isn't being registered until run(), at which point setWindowLevel_ has already been called, causing the window server not to receive the windowLevel.

This is similar to [futurepaul PR 54](https://github.com/xi-editor/druid/pull/54), but also adds the call to finishLaunching.

I had initially put this into it's own function `startup_ritual`, but it seemed better to just put it into `RunLoop::new` unless there is a purpose for being able to make some call in between `new()` and `startup_ritual` (or whatever good name for it would be).

cc @futurepaul

### DIFF
--- a/druid-shell/src/mac/mod.rs
+++ b/druid-shell/src/mac/mod.rs
@@ -161,6 +161,7 @@ impl WindowBuilder {
                 NO,
             );
 
+            window.setLevel_(4);
             window.cascadeTopLeftFromPoint_(NSPoint::new(20.0, 20.0));
             window.setTitle_(make_nsstring(&self.title));
             // TODO: this should probably be a tracking area instead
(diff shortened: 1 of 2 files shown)


cmyr:
Cool, thanks for this, will have a chance to test it soon. 👍

PR #313, opened 2019-11-16
set is_hot upon MouseDown if the target is not already hot.

ratmice:
I'm not sure about this, it seemed tricky to get right in examples/slider
is it correct that we may change child_ctxt but shouldn't recurse?

Anyhow the behavior this patch tries to change is that on at least gtk, if you 
start the application, and the mouse is already hovering over a button, and click it without moving the mouse.

Basically because the mouse never moved, the widget is not hot, and it gets ignored by button.
I tried to fix this by setting is_hot, if it isn't already.

### DIFF
--- a/druid/src/lib.rs
+++ b/druid/src/lib.rs
@@ -567,7 +567,13 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
                 Event::Size(*size)
             }
             Event::MouseDown(mouse_event) => {
-                recurse = had_active || !ctx.had_active && rect.winding(mouse_event.pos) != 0;
+                let had_hot = child_ctx.base_state.is_hot;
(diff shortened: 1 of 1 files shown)


ratmice:
Alright, I think that is as clear as I can make the code reflect the intent.

The other alternate way to do this would be to add something to the platform backend druid-shell to get the current mouse position on startup, and make the widget under the cursor hot there, instead of doing so in the event loop, this might be better visually as the widget would then be bordered immediately, but did seem like a lot more work I don't know how to do on most platforms.

cmyr:
@raphlinus can you take a peek at this?

ratmice:
@cmyr should I rebase it onto master?  There shouldn't be any conflicts as is, but either way let me know.

cmyr:
Nope this is fine, I can do a one-click 'rebase and merge'. Thanks!

Issue #310, opened 2019-11-15
A new pattern for platform-specific types


cmyr:
I had an idea earlier about a way we can handle the problem of backend-specific types.

### Background

 there are a bunch of types that are exposed in druid that are defined in each backend. Some examples are `Menu`, `Application`, and `WindowBuilder`. So each backend has a distinct implementation of these, which an identical API.

This can be annoying because it means if you change API on one of them you have to change API on all of them, and you don't get compiler warnings or errors for backends you are not configured to compile.

Another downside of this is that there's no 'single source of truth' for documentation; even though these types are expected to have a uniform API, each one has to be documented individually in order for those docs to be always available to the caller.

Finally, this has the problem of making the API be "by convention". Essentially there's no 'source of truth'; we just have to be careful and make sure that the API has the same signatures in all places. This has an additional problem: because we don't necessarily *use* the entire API in tests or from druid, it's possible to miss type-mismatches and other bugs.

### use a trait?

At times we have discussed using traits to describe the interfaces for these types; that would let us have default impls (so you wouldn't need to always update all backends when adding a method) and would also provide a single place to put docs, but it means having to import all the traits in addition to the concrete impls, or else using `dyn Trait` everywhere.

We've tried this for `WinCtx` and it works okay, but hasn't entirely convinced.

### tldr; wrapper structs

I think a good compromise would be to settle on wrapper structs.

Basically: every type that is public API in druid-shell would be declared as a newtype struct in the 'main' druid-shell module, that is, not in `platform`.

These structs would declare the public interfaces, and include all documentation. They would wrap the platform-specific types, and forward calls to them. 

This has a few major advantages:

- every bit of public API on every backend would be used, which would make it much easier to catch little API mismatches
- there would be a single place where everything would be documented, and it would only have to be documented once
- implementing a new backend would be a process of chasing the compiler warnings: your root `mod` for the platform would have to export the correct types, which the correct signatures, in order for the crate to compile.

### conclusion

I'm pretty convinced this is a good idea, and I'm probably going to go ahead and start work on it shortly, unless anyone can identify anything I've missed. 😎

raphlinus:
This is similar to what we have for `WinHandler`, yes?

In any case, I totally agree with the goals. Making things more consistent from platform to platform is a good thing.

cmyr:
it's like `WinHandler` but without any deref; we want the wrappers to be explicit about the API, and ensure it is used by the inner types.

cmyr:
This was implemented in #314

PR #314, opened 2019-11-16
Wrapper types

cmyr:
This moves to the pattern described in #310, where every public
type in druid-shell is defined as a newtype in the main
module, wrapping a concrete type on each platform.

There's a bunch of cleanup work to do, for which I have left
copious TODOs; but I would like to get this up as a proof-of-concept
before making more substantial changes.

-----

this includes #312, which should go in first

### DIFF
--- a/druid-shell/src/platform/windows/mod.rs
+++ b/druid-shell/src/platform/windows/mod.rs
@@ -24,9 +24,9 @@ pub mod dcomp;
 pub mod dialog;
 pub mod menu;
 pub mod paint;
+pub mod runloop;
 mod timers;
 pub mod util;
-pub mod win_main;
 
 use std::any::Any;
 use std::cell::{Cell, RefCell};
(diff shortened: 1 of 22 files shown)


druid-shell/src/platform/windows/mod.rs
```diff
@@ -135,11 +135,11 @@ pub struct IdleHandle {
 }
 
 trait IdleCallback: Send {
-    fn call(self: Box<Self>, a: &mut dyn Any);
+    fn call(self: Box<Self>, a: &dyn Any);
```
raphlinus: Hmm, I think we might want to standardize on `&mut` for this. That's basically saying that it will be called single-threaded. On the other hand, changing to that later is easier than the other way around.

cmyr: Yea, that was my thinking; making this non-mut didn't break anything, and we can change it later. In any case this gets moved again in #322, so I'd like to punt on this for now?

Issue #154, opened 2019-09-15
argument ordering of Widget methods is inconsistent


jjl:
`event`, unlike the other methods, takes the event before the context, which sticks out a bit. It should use the same order as the others for consistency

raphlinus:
This feels minor, but is one of those things that's probably easier to fix now than later. I'll accept a PR, though it'll cause merge conflicts with other in-flight PR's, notably multi-window and gtk.

jjl:
> On 15 Sep 2019, at 18:47, Raph Levien <notifications@github.com> wrote:
> This feels minor, but is one of those things that's probably easier to fix now than later. I'll accept a PR, though it'll cause merge conflicts with other in-flight PR's, notably multi-window and gtk.
> 
I’m unlikely to get to it before the multi-window is merged (at least I hope that’ll be done before it), at which point I’ll update it along with the web backend.

/j

cmyr:
fixed in #317

PR #317, opened 2019-11-17
Make Widget trait event fn conform to argument order of other fn's.

ratmice:
Fixes issue #154

I'm not sure when a good time to merge this would be, the things noted in the issue seem to have been merged (multiwin), but if necessary feel free to ping me at that time and I can deal with any fallout that happens between now and then.


cmyr:
#308 will need to be updated.

PR #324, opened 2019-11-18
[shell] Sketch out platform error types

cmyr:
This moves windows specific error stuff to a windows-specific
error types, and adds skeleton errors for mac and GTK.

### DIFF
--- a/druid-shell/src/window.rs
+++ b/druid-shell/src/window.rs
@@ -173,7 +173,7 @@ impl WindowBuilder {
     ///
     /// If this fails, your application should exit.
     pub fn build(self) -> Result<WindowHandle, Error> {
-        self.0.build().map(WindowHandle)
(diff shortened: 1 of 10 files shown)


druid-shell/src/window.rs
```diff
@@ -173,7 +173,7 @@ impl WindowBuilder {
     ///
     /// If this fails, your application should exit.
     pub fn build(self) -> Result<WindowHandle, Error> {
-        self.0.build().map(WindowHandle)
+        self.0.build().map(WindowHandle).map_err(Into::into)
```
raphlinus: I wonder if this would be slightly cleaner with a `?` formulation. But that's strictly a style question.

cmyr: Good question, I don't have strong feelings. This feels less magic and more functional, `Ok(thing()?)` always seems a bit awkward to me.

PR #325, opened 2019-11-19
Fix bug in Split widget layouting when child has a minimum width/height

kbalt:
Me again messing with the `Split` widget.

This PR changes the way the layout is calculated.

The implementation is able to deal with widgets that have a minimum size, while the current has some minor problems.

In this example a `Container` with some Text inside a `Padding` which enforces a minimum width.
old

The new behavior would be blocking the splitter's position:
new

This is obviously only one of many solutions to this problem,
we could for example hide one widget if its layout is computed larger than the panel.
But I think that this would be the simplest and most commonly used behavior.

There is only one problem I sadly could not get rid of.
In some cases the layout of the children need to be computed twice per call which kind of goes against the convention of calling each layout method only once, but I wasn't able to figure out how to do it without.

### DIFF
--- a/druid/src/widget/split.rs
+++ b/druid/src/widget/split.rs
@@ -177,73 +177,118 @@ impl<T: Data> Widget<T> for Split<T> {
         bc.debug_check("Split");
 
         let mut my_size = bc.max();
-        let reduced_width = bc.max.width - self.splitter_size;
-        let reduced_height = bc.max.height - self.splitter_size;
-        let (child1_bc, child2_bc) = match self.split_direction {
+
+        match self.split_direction {
             Axis::Horizontal => {
                 if !bc.is_width_bounded() {
                     log::warn!("A Split widget was given an unbounded width to split.")
                 }
+                let reduced_width = bc.max.width - self.splitter_size;
+
                 let child1_width = (reduced_width * self.split_point).max(0.0);
-                let child2_width = (reduced_width - child1_width).max(0.0);
-                (
-                    BoxConstraints::new(
-                        Size::new(child1_width, bc.min.height),
-                        Size::new(child1_width, bc.max.height),
-                    ),
-                    BoxConstraints::new(
+                let mut child1_bc = BoxConstraints::new(
+                    Size::new(child1_width, bc.min.height),
+                    Size::new(child1_width, bc.max.height),
+                );
+
+                let mut looped = false;
+
+                let (child1_size, child2_size) = loop {
+                    let child1_size = self.child1.layout(ctx, &child1_bc, &data, env);
+
+                    let child2_width = (reduced_width - child1_size.width).max(0.0);
+                    let child2_bc = BoxConstraints::new(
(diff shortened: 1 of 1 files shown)


Zarenor:
Thanks for continuing to look at the issues with `Split`! We discussed this bug [here](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/Split.20widget/near/180690295) in the project's Zulip chat - I'm not sure if you're aware we do a lot of design discussion there. We'd love to have you join us there, if you can. 

Currently, I think our opinion was to treat this issue as a bug in Padding, rather than Split, since Padding is returning a Rect which violates the BoxConstraints it was given. The assumption that Split made (and which I think a few other widgets make) is that the child widgets will not violate the BoxConstraints they're given, and Padding is violating that contract.

I'm okay if we decide to go this way instead of - or in addition to - fixing the issue in Padding. A part of the discussion we've been having is whether it's acceptable for children to 'push back' by having a minimum size, as Padding does here. I don't think we've resolved whether we will support minimum sizes pushed 'up' from children - or whether we will support minimum sizes at all.

kbalt:
I have quickly read the comments on zulip and this seems to be a bit more complex design problem.

My PR is a good example why a "pushback" for minimum size is probably not the best solution

worst case example in my implementation:

- splitter is moved and `Split::layout` is called
- child1 layout fn is called and returns normally
- child2 layout fn is called and it pushes back a minimum size
- child1 BoxConstraint will be adjusted to fit child2's minimum size
- child1's layout function gets called again
- now child1's layout fn might return a minimum size because it was downsized and `Split` will violate its BoxConstraints because it wants to fit both widgets
- child2's layout fn gets called again and we assume it will be now return the same size as before

Thats twice as many calls to `layout` as there should be since it always needs to be checked that all widgets are satisfied with their new BoxConstraints.

While I agree that in this case it would be the right thing to fix the `Padding` related bug, there should still be some way for widgets to force a minimum/static size.
Or alternatively give widgets like `Split` an option to limit the size it can minimize its child to, which might be good enough for some use cases.

Zarenor:
Exactly. I think we're on the same page. Minimum sizes set on the parent container (like `Split`) was a discussion we had, I think a bit earlier in the thread. I certainly agree that it's pretty suboptimal to loop in `layout` - and it breaks our exactly-once-per-pass design.

I agree we need more ways to allow constraining the layout here, and in other layout widgets in any case.

Another concern we had was the overflow of children outside their parents. The solution I think we were considering was a `ClipRect` widget, which does what it says - clips it's children. This follows Flutter's design, where druid draws a lot of inspiriation from.

cmyr:
Sorry, I let this one slip by. @kbalt @Zarenor what's the next step here?

kbalt:
Well, since the behaviour is caused by a bug in ´Padding´ and not ´Split´ we should probably close this?

cmyr:
That is an entirely reasonable next step. :)

PR #326, opened 2019-11-19
Don't expose platform::util at top level

cmyr:
This moves various utility methods into `Application`, following the
new wrapper pattern.

It also cleans up imports a bit, because I have ADHD

-----

functions in `util.rs` were the only remaining place where we were directly exposing a platform implementation in the public API. this moves the two relevant functions in `Application`.

### DIFF
--- a/druid-shell/src/application.rs
+++ b/druid-shell/src/application.rs
@@ -23,6 +23,11 @@ use crate::platform::application as platform;
 pub struct Application;
 
 impl Application {
+    /// Initialize the app. At the moment, this is mostly needed for hi-dpi.
+    pub fn init() {
+        platform::Application::init()
+    }
+
     /// Terminate the application.
     pub fn quit() {
         platform::Application::quit()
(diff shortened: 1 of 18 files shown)


druid-shell/src/application.rs
```diff
@@ -23,6 +23,11 @@ use crate::platform::application as platform;
 pub struct Application;
 
 impl Application {
+    /// Initialize the app. At the moment, this is mostly needed for hi-dpi.
```
raphlinus: Longer term I think we might want to look at getting rid of this, as it's a small source of noise, but I think appropriate to consider this separately from this refactor.

druid-shell/src/platform/gtk/application.rs
```diff
@@ -37,4 +41,9 @@ impl Application {
     pub fn set_clipboard_contents(_item: ClipboardItem) {
         log::warn!("set_clipboard_contents is unimplemented on GTK");
     }
+
+    pub fn get_locale() -> String {
+        //TODO ahem
```
raphlinus: :)

druid-shell/src/platform/mac/application.rs
```diff
@@ -85,4 +87,17 @@ impl Application {
...
         }
     }
+
+    pub fn get_locale() -> String {
+        unsafe {
+            let nslocale_class = class!(NSLocale);
+            let locale: id = msg_send![nslocale_class, currentLocale];
+            let ident: id = msg_send![locale, localeIdentifier];
+            let mut locale = util::from_nsstring(ident);
+            if let Some(idx) = locale.chars().position(|c| c == '@') {
```
raphlinus: This is not Unicode-correct, the result of position is a code point count, the truncation expects utf-8 code units. But the locale string itself will always be ascii, so this is a nit.

cmyr: yes this was a total hack when it was written, just trying to get something working.

PR #327, opened 2019-11-19
Move platform keycode code into platform modules

cmyr:
This resolves a few things: a long-standing TODO where `keycodes.rs` didn't actually have the keycodes declared in it, and that `keyboard.rs` was filled with platform code.


druid-shell/src/platform/mac/keycodes.rs
```diff
@@ -0,0 +1,169 @@
...
+            0x2b => KeyCode::Comma,
+            0x2c => KeyCode::Slash,
+            0x2d => KeyCode::KeyN,
+            0x2e => KeyCode::KeyM,
+            0x2f => KeyCode::Period,
+            0x30 => KeyCode::Tab,
+            0x31 => KeyCode::Space,
+            0x32 => KeyCode::Backtick,
+            0x33 => KeyCode::Backspace,
+            //0x34 => unkown,
```
raphlinus: These should be "unknown". I know that's pre-existing, but as long as the diff is so large we might as well fix it :)

PR #329, opened 2019-11-20
[RFC] Add keyboard events to calc.

ratmice:
This is I guess an RFC, In general i tend to be duplicating char's in quite a few places.
Most of the code here (I.e. the key_map field of the KeyMap widget), is not actually used.

The unused code is intended to support animating the button gradients like mouse press does when performing a key press.  There are some comments there in the `fn event(...)`, 
and prototype code along with the comments.  However i'm unsure of how to route to the individual button widgets through the flex at the moment.

However i'm wondering if this ECS-esque method of attempting to index into a flex via the HashMap is worth further exploration, or perhaps even whether I am missing something obvious .

The code would certainly be a lot shorter if I got rid of all that and only routed state updates through the `CalcState`, anyhow it seemed worth attempting.

### DIFF
--- a/druid/examples/calc.rs
+++ b/druid/examples/calc.rs
@@ -14,9 +14,12 @@
 
 //! Simple calculator.
 
+use druid::widget::{Button, Column, DynLabel, Flex, Padding, Row};
 use druid::{AppLauncher, Data, Lens, LensWrap, Widget, WindowDesc};
-
-use druid::widget::{Button, Column, DynLabel, Padding, Row};
+use druid::{BaseState, BoxConstraints, Env, EventCtx, LayoutCtx, PaintCtx, UpdateCtx};
+use druid::{Event, KeyCode, KeyEvent, KeyModifiers};
+use druid_shell::kurbo::Size;
+use std::collections::HashMap;
 
 #[derive(Clone, Data, Lens)]
 struct CalcState {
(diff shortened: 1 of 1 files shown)


druid/examples/calc.rs
```diff
@@ -112,95 +115,239 @@ impl CalcState {
...
+
+impl KeyMap {
+    fn new(flex: Flex<CalcState>, key_map: HashMap<ButtonId, ButtonInfo>) -> KeyMap {
+        KeyMap { key_map, flex }
+    }
+}
+
+fn key_event_to_button_id(event: &KeyEvent) -> Option<ButtonId> {
+    use crate::ButtonId::*;
+    match event.mods {
```
cmyr: for this sort of thing, look at the `HotKey` type; basically you can describe a key + mods, and then see if an event matches it. For an example, see how we handle keys in [the textbox widget](https://github.com/xi-editor/druid/blob/master/druid/src/widget/textbox.rs#L380)

druid/examples/calc.rs
```diff
@@ -112,95 +115,239 @@ impl CalcState {
...
+                match kc {
+                    None => (),
+                    Some(button_id) => {
+                        let ButtonInfo {
+                            row_id: _row_id,
+                            col_id: _col_id,
+                        } = self.key_map.get(&button_id).unwrap();
+                        // In theory if there was a mechanism to get random access into
+                        // self.flex.children we could remove the above match,
+                        // and do something fancy with the buttons here instead.
```
cmyr: there's a maybe-hacky-maybe-just-how-we-do-things way to do this (although it would require some new logic in button): basically you would define a `Command` for each button, and when the hotkey was triggered you would send the appropriate `Command`; each button would be listening for its own command, and when it was received it would trigger the button's action and do the animation, etc.

druid/examples/calc.rs
```diff
@@ -112,95 +115,239 @@ impl CalcState {
...
-fn op_button_label(op: char, label: String) -> impl Widget<CalcState> {
+fn op_button_label(op: char, label: String) -> Padding<CalcState> {
     pad(Button::new(
         label,
         move |_ctx, data: &mut CalcState, _env| data.op(op),
     ))
 }
 
-fn op_button(op: char) -> impl Widget<CalcState> {
+fn op_button(op: char) -> Padding<CalcState> {
```
cmyr: Curious if this was necessary, or just a style decision?

druid/examples/calc.rs
```diff
@@ -112,95 +115,239 @@ impl CalcState {
...
-            op_button('±'),
-            digit_button(0),
-            op_button('.'),
-            op_button('='),
-        ),
-        1.0,
-    );
-    column
+    let mut key_map = HashMap::new();
+    for (col_id, buttons) in button_rows.iter().enumerate() {
```
cmyr: to check my understanding: this key map isn't actually used, it's just sort of showing how we _could_ address buttons?

druid/examples/calc.rs
ratmice: It was necessary, because lines [223-225](https://github.com/xi-editor/druid/pull/329/commits/3ecf4d4ce7104f903cd21fdb0d6390e6877aa2f3#diff-40cfb56c817860b90b6a65ec02961f03R223-R225)

The `impl Widget<CalcState>` erases the `Padding` type, and so the compiler could no longer distinguish that the match expressions results were of uniform type.

druid/examples/calc.rs
ratmice: Yes, exactly

ratmice:
Thanks for the review, I will look into these things.

PR #330, opened 2019-11-22
Multi-format copy and paste

cmyr:
This was very frustrating! I spent a lot of time trying to come up with a more "just works" API, but it ended up getting hairy and ultimately I ripped it out and tried to just do the simplest API I could.

In any case, this is working on mac, windows and gtk. The only bug I'm aware of is on GTK; if you put multiple formats on the clipboard, and one of them is text, the clipboard doesn't seem able to retrieve the text. I'm not totally sure what's going on there, but I'm not going to fight it right now, it's a pretty marginal case all in all.

Anyway: this covers the common case (text) and the special case required by runebender (multi-format vector data) comfortably. There may be future limitations; in particular I think doing copy/paste for images is going to be a headache and a half, but we can worry about that at some future date.

### DIFF
--- a/druid-shell/src/clipboard.rs
+++ b/druid-shell/src/clipboard.rs
@@ -13,28 +13,210 @@
 // limitations under the License.
 
 //! Interacting with the system pasteboard/clipboard.
+pub use crate::platform::clipboard as platform;
 
-/// An item on the system clipboard.
+/// A handle to the system clipboard.
+///
+/// To get access to the global clipboard, call [`Application::clipboard()`].
+///
+///
+/// # Working with text
+///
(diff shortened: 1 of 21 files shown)


druid-shell/src/clipboard.rs
```diff
@@ -13,28 +13,217 @@
...
+
+/// Data coupled with a type identifier.
+#[derive(Debug, Clone)]
+pub struct ClipboardFormat {
+    pub(crate) identifier: FormatId,
+    pub(crate) data: Vec<u8>,
+}
+
+impl ClipboardFormat {
+    /// Create a new `ClipboardFormat` with the given `FormatId` and byes.
```
raphlinus: bytes

PR #331, opened 2019-11-22
s/Meta/Super/ in GTK

Ralith:
On Linux, the windows key is Super; Meta is an ancient and unused relic.

There's a case for renaming the druid-shell keycode, but that would be substantially more invasive and I don't personally care what it's called so long as it's not spitting errors into the console.

### DIFF
--- a/druid-shell/src/platform/gtk/keycodes.rs
+++ b/druid-shell/src/platform/gtk/keycodes.rs
@@ -85,8 +85,8 @@ impl From<u32> for KeyCode {
             Alt_R => KeyCode::RightAlt,
             Shift_L => KeyCode::LeftShift,
             Shift_R => KeyCode::RightShift,
-            Meta_L => KeyCode::LeftMeta,
-            Meta_R => KeyCode::RightMeta,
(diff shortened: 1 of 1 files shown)


cmyr:
Thanks! if someone who uses gtk more than me would like to sign off on this that would be great. @ratmice? @mrandri19?

ForLoveOfCats:
I've not built anything with GTK before but I can confirm that the Windows key is called Super on Linux. In fact the "Meta" key is frequently the Alt key. It seems from a little research that the Windows key has been historically used to emulate both Super and Meta from older keyboards but Super is much more prevalent.

Some links:
https://unix.stackexchange.com/questions/28993/what-is-bashs-meta-key
https://en.wikipedia.org/wiki/Meta_key

ratmice:
@cmyr It looks good to me, I generally run with `setxkbmap -option "compose:lwin"`, which tricks
it into thinking it's actually [Multi_key](https://en.wikipedia.org/wiki/Compose_key), but disabling that nonstandard configuration, the patch worked fine.

cmyr:
Okay, thanks all!

PR #332, opened 2019-11-22
Allow lens input/output types to be unsized

Ralith:
They're only handled by reference, so this seems like a free genericity win.


cmyr:
Cool, I can definitely imagine cases where we would want this.

Ralith:
> I'm assuming you found a use for this?

I was experimenting with implementing a lens combinator for indexable containers and found I had to specify `<T as Index>::Target: Sized` to get anywhere, which mainly just seemed unnecessary, though it's possible to imagine cases where it would have been a real obstacle.

Issue #318, opened 2019-11-18
gtk::make_key_event numpad keys


ratmice:
These are mostly overly detailed notes regarding an issue with GTK numpad events,
I will try and come up with a patch based on them in the morning.
Currently on gtk backend Numpad keys 0-9 and Dot/Del when numpad is either on or off will emit KeyCode::Unknown(65429)` through `KeyCode::Unknown(65439)`.

These numbers 65429-65439 correlate to the following [GDK keysyms](https://gitlab.gnome.org/GNOME/gtk/blob/master/gdk/gdkkeysyms.h#L107-119)

```
#define GDK_KEY_KP_Home 0xff95
#define GDK_KEY_KP_Left 0xff96
#define GDK_KEY_KP_Up 0xff97
#define GDK_KEY_KP_Right 0xff98
#define GDK_KEY_KP_Down 0xff99
#define GDK_KEY_KP_Page_Up 0xff9a
#define GDK_KEY_KP_Page_Down 0xff9b
#define GDK_KEY_KP_Begin 0xff9d
#define GDK_KEY_KP_End 0xff9c
#define GDK_KEY_KP_Insert 0xff9e
#define GDK_KEY_KP_Delete 0xff9f
```

KeyCode as it is doesn't have numpad variations for these,
The issue seems to stem from the function [make_key_event](https://github.com/xi-editor/druid/blob/6ec1de4660785f939bb7202b708e7dae4dbb4782/druid-shell/src/gtk/mod.rs#L722-L726)

Interestingly the keyval code is correct, mapping to
`#define GDK_KEY_KP_0 0xffb0` when numlock is on, and `#define GDK_KEY_KP_End 0xff9c` when numlock is off.

It's once it goes through `hardware_keycode` and `hardware_keycode_to_keyval` where it loses track of the difference.  As far as I can tell `get_modifiers` doesn't treat numlock as any form of modifier, there is a `LOCK_MASK`, but that is only documented as pertaining to [`CAPS_LOCK`](https://gitlab.gnome.org/GNOME/gtk/blob/master/gdk/gdktypes.h#L167).

Many of the above keys could be (but are currently not mapped to their non keypad equivalents) with the only exception I believe being KP_Begin (5 on the keypad).

I'm think we should just add the key Numpad variations of these keys though for consistency with e.g. NumpadEnter.

## Things to do:
1:  Add missing KeyCode variants for Numpad, or alternately map them to non-Numpad variants.

[...]

PR #334, opened 2019-11-22
[gtk] Fix numpad events, since the hardware mapping treats numlock as off

ratmice:
This should fix issue #318 -- to some extent I believe the only thing that is left is:

It doesn't have a mechanism to tell whether numlock is on or off from the `KeyModifiers` passed in with the `KeyEvent`.  I'm not sure if that is consistent with other platforms.

While this is a bit uglier than matching `KP_0`, it seemed better than hacking `platforms/gtk/window.rs:make_key_event()` to actually return the `keyval` for this specific subset of key codes.

Edit: Sorry, I just after rereading the commit message didn't think it was clear, thats the only thing that changed.


cmyr:
Ouf this gets metaphysical quick.

Okay looking at this and then reading #318, I think that we are wrong to prefer the hardware keycode over the returned keycode, at least in this range?

At a high level: I think we want to be able to distinguish between events with and without numlock. I don't think we want to throw this away.

the reality of writing backends, especially for stuff like this, is that you have to hack around to provide the expected interface. Having weird special cases is unavoidable.

so my gut, here, is that we should do what we have to in order to respect numlock, and not go straight to the hardware key, at least for this set of affected keys.

Does that seem reasonable? It's very possible I'm overlooking something, but this is what I see from here.

ratmice:
I think one consideration in this is that we would need to add a whole bunch of new keycodes to the platform neutral keycodes.rs in order to have both the numlock and other non-numlock key variant.
alternately we could add a numlock modifier.

numlock isn't typically in the list of modifiers given with a keypress, there is a function to get if from the keyboard state I believe, or we could infer it from the range of keys.

So I think the choices are
* Add a numlock modifier, get the state from the keyboard.
* Add a numlock modifier, infer it from the key range.
* Add a bunch of new keycodes for the non-numlock case infer it from the keycode range.

The 2nd option seems most reasonable to me, either way we go is going to require some modification that affects all platforms. This seemed stand-alone enough that it at least got them working.

Hopefully i'm not missing anything, which actually allows other platforms to differentiate these keys.

ratmice:
Superseded by #342

PR #335, opened 2019-11-22
Cleanup druid exports

cmyr:
This is a bunch of cleanup around what we export from druid.

- We no longer expose `druid_shell`, instead selectively exposing
the types that that are intended to be used directly.
- We are more selective about what modules we expose at the top
level, trying only to do this for modules that expose constants
like `theme`.
- For submodules that expose constants, like `command::sys` and
`menu::sys` we reexport these directly at the top level.
- Other types and reexports that were unnecessarily public are removed.

### DIFF
--- a/druid/examples/layout.rs
+++ b/druid/examples/layout.rs
@@ -14,7 +14,7 @@
 
 //! This example shows how to construct a basic layout.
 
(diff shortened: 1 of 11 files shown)


cmyr:
It's been that kind of week. 😬

PR #336, opened 2019-11-22
Add druid-shell readme, fixup main readme, add readmes to manifests

cmyr:
More bookkeeping and release prep.

### DIFF
--- a/druid/README.md
+++ b/druid/README.md
@@ -1,0 +1,1 @@
+../README.md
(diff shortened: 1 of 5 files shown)


druid/README.md
```diff
@@ -0,0 +1 @@
+../README.md
```
raphlinus: Does this work? If it's a one-line path reference, it works like a symlink?

cmyr: this *is* a symlink, it's just showing up funny in the diff? Unless I really messed something up

raphlinus: Sounds like it will be exciting then, to see if everything works when the crate is published :)

PR #339, opened 2019-11-23
Change Lens to Lens2 in-place

raphlinus:
The Lens2 version is more flexible, as it allows synthesizing data. This patch updates the Lens trait to that.


Ralith:
I was just thinking about how this was probably necessary!

Ralith:
It seems a bit inconsistent to have `get` and `with_mut`. Maybe standardize on `with`/`with_mut` or `get`/`get_mut`? I'm weakly in favor of the former.

raphlinus:
Yeah, I think "with" is appropriate. I'll make the change.

zaynetro:
Perfect timing!

PR #341, opened 2019-11-23
Declarative lens macro

Ralith:
This makes it possible to access fields of foreign types with lenses, and reduces magic and namespace clutter. The main downside is that the resulting types aren't nameable (at least until type alias impl trait comes along); if that's important we might want to retain the derive for those cases.

### DIFF
--- a/druid/src/lib.rs
+++ b/druid/src/lib.rs
@@ -57,7 +57,7 @@ pub use command::{sys as commands, Command, Selector};
 pub use data::Data;
 pub use env::{Env, Key, Value};
 pub use event::{Event, WheelEvent};
-pub use lens::{Lens, LensWrap};
+pub use lens::{Field, Lens, LensWrap};
 pub use localization::LocalizedString;
 pub use menu::{sys as platform_menus, ContextMenu, MenuDesc, MenuItem};
 pub use mouse::MouseEvent;
--- a/README.md
+++ b/README.md
@@ -206,6 +206,15 @@ CamelCase to snake_case):
 LensWrap::new(WidgetThatExpectsf64::new(), AppState::value);
 ```
 
+Alternatively, lenses for structs, tuples, and indexable containers can be
+constructed on-demand with the `lens` macro:
+
+```rust
+LensWrap::new(WidgetThatExpectsf64::new(), lens!(AppState, value));
+```
+
+This is particularly useful when working with types defined in another crate.
+
 ## Using druid
 
 An explicit goal of druid is to be easy to build, so please open an issue if you
--- a/druid/src/lens.rs
+++ b/druid/src/lens.rs
@@ -141,3 +141,62 @@ where
         }
     }
 }
+
+/// Lens accessing a member of some type using accessor functions
+///
+/// See also the `lens` macro.
+///
+/// ```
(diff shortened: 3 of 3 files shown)


Ralith:
Rebased.

cmyr:
Interesting!

I'm going to defer to raph on this. I think this totally makes sense given how lens is currently used, but I suspect many real-world use cases are going to be more complicated, and then derive gives us a lot more flexibility. But I'm not _sure_ that's the case, and it may be that in the complicated cases we'll end up just wanting to hand-roll the implementations?

raphlinus:
I took a quick look at this and think it might well be the way to go. Part of the original motivation was to have one lens rather than create a new one on each use, but in hindsight that might not actually be much of an advantage.

I'll spend a little time today wrapping my mind around lens stuff, and do a proper review as part of that.

Ralith:
We can always retain the derive (or bring it back in the future) if more complex behavior is desired, but I think this is a major win for simple cases at the very least. It's unfortunate that there's no way to have constants that involve anonymous types, but another simple declarative macro could be introduced to declare lenses that look/work like the proc macro's output if desired.

cmyr:
I'm inclined to agree, simple is better and we can always get fancy later if we need to.

Ralith:
Rebased and removed the wholesale deletion of the derive impl since we're making some progress on improving it.

cmyr:
I like this and am broadly happy to merge.

The thing that is niggling at me a little bit is how we expose it. `field` as a top level export seems like it might have discoverability problems? I might be overthinking this. I could imagine `field_lens!` or `lens_field!`, or maybe in an ideal world we would make the `lens` module public and then this would be used as `lens::field!`?

All the logic looks good, idea is good, just feeling a bit unsure about the ergonomics.

Ralith:
Added support for indexing, renamed to `lens`, and unclobbered the examples in light of the associated constants change.

druid/src/lib.rs
```diff
@@ -57,7 +57,7 @@ pub use command::{sys as commands, Command, Selector};
 pub use data::Data;
 pub use env::{Env, Key, Value};
 pub use event::{Event, WheelEvent};
-pub use lens::{Lens, LensWrap};
+pub use lens::{Field, Lens, LensWrap};
```
cmyr: do we need to export the `lens!` macro itself? I'm not sure how macro visibility works now. 😕

cmyr: also do we expect `Field` to ever be used directly, or is it an implementation detail of `lens!`? If the latter I'd probably not reexport here.

Ralith: > do we need to export the lens! macro itself? I'm not sure how macro visibility works now.

Declarative macros annotated with `#[macro_export]` are always implicitly exported from the crate root.

> also do we expect Field to ever be used directly, or is it an implementation detail of lens!?

It might be a useful building block for things that are difficult or impossible to express with `lens!` for some reason, e.g. something that implements copy-on-write semantics in the mutable getter. In any case, macro_rules can't reference things that aren't public, so the best we could only hide it in the docs.

cmyr: Okay, thanks for the explanations.

PR #342, opened 2019-11-23
Add UnknownKeyMap trait, Fix numpad events on gtk

ratmice:
Since the hardware mapping treats numlock as off
This patch makes make_key_event return KP_0..KP_9 when numlock is on on the
gtk platform,

When Numlock is off, Numpad events will result in Unknown(...),

Further, it adds a new trait UnknownKeyMap to druid_shell which allows you to
map from Unknown(...) KeyCodes with the to_logical() and to_physical().

These functions can be used to map unknown KeyCodes e.g:
Unknown(Home).to_logical() -> Home,
Unknown(Home).to_physical() -> Numpad7,

With these in place it is possible to differentiate the different keys
without having to add any new KeyCodes or modifiers.

These functions are intended to loosely correlate to the 'key', and 'code' fields
of the KeyEvent structure of the WebAPI.

### DIFF
--- a/druid-shell/src/keyboard.rs
+++ b/druid-shell/src/keyboard.rs
@@ -84,6 +84,18 @@ impl KeyEvent {
         }
     }
 
+    /// Map a `KeyEvent`'s `key_code` to by calling f.
+    /// The `KeyEvent` returned leaves all other fields unmodified.
+    pub fn map_keycode(&self, f: fn(KeyCode) -> KeyCode) -> Self {
+        KeyEvent {
+            is_repeat: self.is_repeat,
+            mods: self.mods,
+            key_code: f(self.key_code),
+            text: self.text,
+            unmodified_text: self.unmodified_text,
+        }
+    }
+
     /// For creating `KeyEvent`s during testing.
     #[doc(hidden)]
     pub fn for_test(mods: impl Into<KeyModifiers>, text: &'static str, code: KeyCode) -> Self {
--- a/druid-shell/src/keyboard.rs
+++ b/druid-shell/src/keyboard.rs
@@ -84,18 +84,6 @@ impl KeyEvent {
         }
     }
 
-    /// Map a `KeyEvent`'s `key_code` to by calling f.
-    /// The `KeyEvent` returned leaves all other fields unmodified.
-    pub fn map_keycode(&self, f: fn(KeyCode) -> KeyCode) -> Self {
-        KeyEvent {
-            is_repeat: self.is_repeat,
-            mods: self.mods,
-            key_code: f(self.key_code),
-            text: self.text,
-            unmodified_text: self.unmodified_text,
-        }
-    }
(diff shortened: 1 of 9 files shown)


ratmice:
Just a note that I'm not entirely sure If this is trait impl is the right way, in regards to #310,
With the thing being that here we have a platform specific trait on a platform neutral type.

With the question being whether the platform should re-export the platform-neutral KeyCode type.
And we should wrap and provide a default impl on that.

I'm still coming to grips with all the cross platform trait stuff.

druid-shell/src/keyboard.rs
```diff
@@ -84,6 +84,18 @@ impl KeyEvent {
...
+    pub fn map_keycode(&self, f: fn(KeyCode) -> KeyCode) -> Self {
+        KeyEvent {
+            is_repeat: self.is_repeat,
+            mods: self.mods,
+            key_code: f(self.key_code),
+            text: self.text,
+            unmodified_text: self.unmodified_text,
+        }
+    }
+
```
ratmice: Based on your feedback from to #329,  I tried to include the update event handling stuff into calc,
since this patch is pretty painful to test without it.

doing brought to my attention an ergonomic issue when trying to use this API with `HotKey::matches`.
This was the best thing I could come up with to resolve that, however leaving `text` and `unmodified_text` untouched seems pretty terrible.

druid/examples/calc.rs
```diff
@@ -148,6 +166,116 @@ fn flex_row<T: Data>(
...
+        use druid::KeyCode::*;
+        use druid::RawMods;
+        use druid::RawMods::Shift;
+        use druid_shell::UnknownKeyMap;
+        use CalcInput::*;
+
+        match event {
+            Event::KeyDown(key_event) => {
+                // map to physical keys since we want numbers with numlock both on and off.
+                let key_event = key_event.map_keycode(UnknownKeyMap::to_physical);
```
ratmice: I think the thing to do here is to just avoid routing through key_event for this and avoid all the pain of `text` and `unmodified_text`.

We could perhaps implement `Eq` for `HotKey`

druid/examples/calc.rs
```diff
@@ -148,6 +166,116 @@ fn flex_row<T: Data>(
...
+                if key_event.is_repeat
+                    || !key_event.key_code.is_printable()
+                    || key_event.key_code == Space
+                {
+                    return;
+                }
+
+                let calc_input = match key_event {
+                    k_e if (HotKey::new(Shift, Key8).matches(k_e)
+                        || HotKey::new(None, NumpadMultiply).matches(k_e)) =>
```
ratmice: So these become `HotKey::new(Shift, Key8) == HotKey::new(k_e.mods, k_e.key_code)

ratmice:
Of course, closing for now can always revisit later if necessary.

PR #343, opened 2019-11-23
fix a regression where apps won't release focus caused by 4e29054

ratmice:
I believe this should fix the regression, and hopefully leave the fix for issue #43 intact

### DIFF
--- a/druid-shell/src/platform/mac/window.rs
+++ b/druid-shell/src/platform/mac/window.rs
@@ -142,7 +142,12 @@ impl WindowBuilder {
                 NO,
             );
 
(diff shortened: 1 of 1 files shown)


cmyr:
Okay, this works and I think it's fine to use `0` directly, although you could expose `CGWindowLevelForKey` with,

```rust
extern "system" {
    pub fn CGWindowLevelForKey(key: int32_t) -> int32_t;
}
```

but I'm not sure this is worth it.

cmyr:
in any case I'm going to merge this now to unbreak things. thanks @ratmice!

PR #344, opened 2019-11-23
Implement lens composition

Ralith:


Ralith:
Rebased and made independent.

Ralith:
Rebased back onto #341 for the sake of tests, renamed `Pipe` to `Then`.

druid/src/lens.rs
```diff
@@ -141,3 +142,218 @@ where
...
+macro_rules! lens {
+    ($ty:ty, [$index:expr]) => {
+        $crate::lens::Field::new::<$ty, _>(|x| &x[$index], |x| &mut x[$index])
+    };
+    ($ty:ty, $field:tt) => {
+        $crate::lens::Field::new::<$ty, _>(|x| &x.$field, |x| &mut x.$field)
+    };
+}
+
+/// `Lens` composed of two lenses joined together
```
raphlinus: This is definitely one of the most important combinators, thanks! I think I'm good with "Then", but was also thinking about names more reminiscent of functional composition.

druid/src/lens.rs
```diff
@@ -141,3 +142,218 @@ where
...
+pub struct Then<T, U, B: ?Sized> {
+    left: T,
+    right: U,
+    _marker: PhantomData<B>,
+}
+
+impl<T, U, B: ?Sized> Then<T, U, B> {
+    pub fn new<A: ?Sized, C: ?Sized>(left: T, right: U) -> Self
```
raphlinus: The docstring should probably point to the `then` method. Also, I'm not sure "left" and "right" are clear names, though I can't think of more authoritative ones off the top of my head. (I feel like I should read a primer on lenses or compositional mathematic operations)

I know in Haskell it's just function composition, but that doesn't seem available here.

druid/src/lens.rs
```diff
@@ -141,3 +142,218 @@ where
...
+        self.left.with(data, |b| self.right.with(b, f))
+    }
+
+    fn with_mut<V, F: FnOnce(&mut C) -> V>(&self, data: &mut A, f: F) -> V {
+        self.left.with_mut(data, |b| self.right.with_mut(b, f))
+    }
+}
+
+/// `Lens` for indexing containers
+pub struct Index<I> {
```
raphlinus: I'm impressed we can use the Rust `Index` trait - I was kinda thinking we'd need separate ones for Vec, etc. I am wondering, though, what we need to punch through the `Arc` in the common `Arc<Vec>` combination.

druid/src/lib.rs
```diff
@@ -26,7 +26,8 @@ mod command;
 mod data;
 mod env;
 mod event;
-mod lens;
+pub mod lens;
```
raphlinus: I think we don't want the modules in the namespace, just the items such as `Lens` and so on. I can see some possible issues with the intermediate structs, in which case we should probably be using `pub(crate)` if `impl Lens` will work as a type, or failing that, `#[doc(hidden)]`.

druid/src/lens.rs
Ralith: I think a `deref` combinator is possible. Alternatively, for better ergonomics maybe `Deref` can be baked directly into this? I'll experiment some tonight.

druid/src/lens.rs
```diff
@@ -141,3 +142,218 @@ where
...
+
+    fn with_mut<V, F: FnOnce(&mut O) -> V>(&self, data: &mut T, f: F) -> V {
+        let mut temp = &*data * &self.factor;
+        let x = f(&mut temp);
+        *data = temp / &self.factor;
+        x
+    }
+}
+
+/// `Lens` for an offset value
```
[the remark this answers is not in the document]
Ralith: These are motivated by my own real-world use, where I want a slider to control a value whose range is not 0-1. They're particularly convenient because they bake in the inverse operation, which is an error/correctness risk otherwise. I'll see if I can maybe expand `Iso` into something more convenient for this case while retaining generality.

druid/src/lens.rs
Ralith: I'm open to suggestions! This has some precedent in std, with [`Option::and_then`](https://doc.rust-lang.org/core/option/enum.Option.html#method.and_then) and moreso [`FutureExt::then`](https://docs.rs/futures/0.3.1/futures/future/trait.FutureExt.html#method.then).

druid/src/lens.rs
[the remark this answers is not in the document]
raphlinus: Ahhh right. I wonder whether longer term we want to do it this way (emphasizing the simplicity of individual components and composing them), or make the slider more configurable. But yes, that is an additional use case, and argues in favor of including them.

druid/src/lib.rs
Ralith: `impl Lens` is unfortunately impossible in an extension trait for the time being. I'm not sure how I feel about going out of our way to hide these types; what's the upside?

raphlinus: It's not so much that we want to hide the types, as that we want to avoid modules being exported directly into the namespace. This is recent reorg work by @cmyr and maybe he has stronger feelings about it.

Ralith: Personally, I like `pub mod foo { struct Bar; }` when I would otherwise have to do `struct FooBar;`, which I think is the case here; the lens names are pretty vague out of context.

cmyr: I'm 100% okay with _some_ public modules in the case where there is a clear argument for them, e.g. they have tons of exports and it is clarifying to keep them separate, and this sounds like a reasonable case for that.

I mostly just don't want public modules to be the default choice.

druid/src/lens.rs
Ralith: Neither of those approaches admit a reasonable way to inject the `Arc::make_mut`, so probably a different tactic is needed for that.

Ralith:
Stripped this down to the obviously correct portion; will continue to explore more advanced stuff in one or more new PRs.

Ralith:
Renamed the `set` helper to `put` for greater edit distance.

druid/src/lens_ext.rs
```diff
@@ -0,0 +1,40 @@
+use crate::lens::Then;
```
cmyr: might it make sense to put this trait in `lens.rs` itself?

Another nit is that, in general, we try to have a docstring for each module. I think in druid we also include the license in each module file? I may have skipped this in an earlier review.

Ralith: Flattened into lens.rs.

PR #350, opened 2019-11-24
Take a closure returning a concrete widget in List

Ralith:
This is a marginal ergonomics improvement consistent with existing conventions (no need for the caller to box explicitly) and clarifies the intent of the interface, assuming it is actually the intention of the interface that lists be homogeneous.

### DIFF
--- a/druid/examples/list.rs
+++ b/druid/examples/list.rs
@@ -46,17 +46,15 @@ fn ui_builder() -> impl Widget<AppData> {
 
     root.add_child(
         Scroll::new(List::new(|| {
-            Box::new(
-                SizedBox::new(
-                    Container::new(Padding::new(
-                        10.0,
-                        DynLabel::new(|d, _| format!("List item #{}", d)),
-                    ))
-                    .background(Color::rgb(0.5, 0.5, 0.5)),
-                )
-                .expand()
-                .height(50.0),
+            SizedBox::new(
+                Container::new(Padding::new(
+                    10.0,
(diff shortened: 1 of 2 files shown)


cmyr:
okay so if I understand correctly this would still let the widget type be `Box<Widget<T>>` if desired, but that would need to be explicit; and then we would also support other widget types as well?

I didn't participate in the list widget stuff much initially but this makes sense to me.

Ralith:
> this would still let the widget type be Box<Widget<T>> if desired

I don't think `Box<dyn Widget<T>>` implements `Widget<T>`, strictly speaking, so the user explicitly returning that wouldn't work. If that's a desired use-case then this PR should be closed, but it's difficult for me to imagine how you'd reasonably make use of that since the closure is `Fn`, i.e. cannot have a mutable environment, so you'd have to go well out of your way to actually do something different in each invocation.

cmyr:
It does: https://github.com/xi-editor/druid/blob/master/druid/src/lib.rs#L241

cmyr:
so I think this would work in both cases?

Ralith:
Somehow I missed that in my hurry this morning. Yep, should be fine!

Ralith:
Of course, redundant boxing is kind of ugly, but maybe less ugly than requiring explicit boxing from all users.

cmyr:
yea, I think we'll just have to swallow that.

cmyr:
Thanks! you're opening these faster than I can merge them. 😉

PR #351, opened 2019-11-24
Helpers to construct Flexes directly from their contents

Ralith:
The somewhat-cheeky definition of `Column` and `Row` as facades for `Flex` prevents us from actually implementing `FromIterator`.

### DIFF
--- a/druid/src/widget/flex.rs
+++ b/druid/src/widget/flex.rs
@@ -92,10 +92,24 @@ impl Row {
             children: Vec::new(),
         }
     }
+
+    /// Create a row widget from a set of widgets and flex values.
+    ///
+    /// See `add_child` for details on how the `f64` is interpreted.
+    #[allow(clippy::should_implement_trait)]
+    pub fn from_iter<T: Data>(
+        iter: impl IntoIterator<Item = (impl Widget<T> + 'static, f64)>,
+    ) -> Flex<T> {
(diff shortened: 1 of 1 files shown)


cmyr:
Hmm. The problem with this as I see it is that it requires already having a collection of homogenous widgets to iterate over, which will be a hassle to construct in practice; it would likely end up requiring a `Vec` of `Box<Widget<T>>`, where each widget is manually boxed.

When I was thinking about similar things, the approach that made sense to me was just a macro; either a `widgets!` macro that worked like `vec!` but created a vector of boxed widgets, or else separate `flex_column!` and `flex_row!` macros.

I'm not strictly opposed to the approach here, but I would want to see an example of how it would be used in practice that demonstrates some reasonable ergonomics improvement over the current approach?

Ralith:
Ah, the homogeneity problem is a good point. The `widgets` macro strikes me as a pleasantly general-purpose solution.

cmyr:
the one problem with the `widgets!` macro is that we don't expose a constructor on `Flex` that just takes a `Vec<Widget<T>>`. But I mean that's a very solvable problem.

cmyr:
@Ralith should we close this, and consider any future macro version new work?

Ralith:
Sure, I'll make a new PR when I've got something ready.

PR #352, opened 2019-11-25
Expose derived lenses as associated constants

Ralith:
Generates lens type definitions inside a private module with an unlikely-to-collide name, then uses those to populate associated constants on the type for extremely ergonomic access. The main drawback here is that the private `__Foo_druid_lenses` module name pops up in error messages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,7 +188,8 @@ struct AppState {
 ### lens
 
 The [Lens datatype] gives access to a part of a larger data structure. Like
-`Data`, This can be derived.
+`Data`, this can be derived. Derived lenses are accessed as associated constants
+with the same name as the field.
 
 ```rust
 #[derive(Clone, Data, Lens)]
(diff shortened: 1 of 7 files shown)


Ralith:
Updated to define the lenses in a public module with some documentation. This makes things a little prettier; it's not perfect but I think it's pretty good.

Example error from using a lens of the wrong type:
```
error[E0277]: the trait bound `impl druid::Widget<f64>: druid::Widget<bool>` is not satisfied
  --> druid/examples/slider.rs:53:37
   |
53 |     col.add_child(Padding::new(5.0, slider), 1.0);
   |                                     ^^^^^^ the trait `druid::Widget<bool>` is not implemented for `impl druid::Widget<f64>`
   |
   = note: required because of the requirements on the impl of `druid::Widget<DemoState>` for `druid::lens::LensWrap<bool, demo_state_lenses::double, impl druid::Widget<f64>>`
   = note: required by `druid::widget::padding::Padding::<T>::new`
```

cmyr:
My only nit might be to add `generated` or `derived` to the name of the module; so something like `demo_state_generated_lenses` or `demo_state_derived_lenses`?

Ralith:
We'll be able to get rid of the public module when https://github.com/rust-lang/rust/issues/63065 comes along.

cmyr:
want to make that name tweak? then I'm happy to merge, this is a solid improvement.

Ralith:
Done, plus some brief documentation.

PR #353, opened 2019-11-25
SVG widget

futurepaul:
I was correct, this was a lot of fun. The downside: it's a bloat monster because I'm relying on `usvg` to parse and simplify the incoming SVG for me. So perhaps it should be behind a feature flag?

One other caveat: clippy doesn't like the explicit `return` keyword in `get_size()`, but if I don't do that then the borrow checker gets mad about how long the `*root.borrow()` might live for.


raphlinus:
Very cool! Initial thought is that for stuff that increases bloat but which not all clients will use, a feature flag is appropriate, though I'm sure we'll use this in Runebender.

I would file an issue against clippy. It feels like making a recommendation which breaks the code should be considered a bug. In the meantime, of course suppress the warning.

druid/src/widget/svg.rs
```diff
@@ -0,0 +1,222 @@
...
+
+fn color_from_fill(fill: &Fill) -> Color {
+    match fill.paint {
+        Paint::Color(c) => {
+            let alpha = (fill.opacity.value() * 255.0) as u8;
+            Color::rgba8(c.red, c.green, c.blue, alpha)
+        }
+        _ => {
+            //TODO: figure this out!
+            error!("I don't know what a Paint::Link is so here's some pink.");
```
Ralith: This is presumably for values like `url(#myGradient)`, referencing some `<linearGradient id="myGradient">`.

druid/src/widget/svg.rs
```diff
@@ -0,0 +1,222 @@
...
+
+    fn event(&mut self, _ctx: &mut EventCtx, _event: &Event, _data: &mut T, _env: &Env) {}
+
+    fn update(&mut self, _ctx: &mut UpdateCtx, _old_data: Option<&T>, _data: &T, _env: &Env) {}
+}
+
+fn color_from_fill(fill: &Fill) -> Color {
+    match fill.paint {
+        Paint::Color(c) => {
+            let alpha = (fill.opacity.value() * 255.0) as u8;
```
raphlinus: This might be ever so slightly more idiomatic as `Color::rgb8(c.red, c.green, c.blue).with_alpha(fill.opacity.value())`. But that's entirely a style question, this is fine.

futurepaul:
> Could this perhaps be a standalone crate that implements SVG drawing for piet

Happy to make a standalone crate if that's what people prefer. In the meantime I could refactor this so the hardworking svg-to-piet functions are available without using the widget?

Ralith:
Sounds like a good idea to me!

futurepaul:
Turns out this specific borrow situation is a known problem with clippy: https://rust-lang.github.io/rust-clippy/master/index.html#needless_return

druid/src/widget/svg.rs
```diff
@@ -0,0 +1,284 @@
...
+            Color::rgb8(255, 192, 203)
+        }
+    }
+}
+
+#[cfg(feature = "svg")]
+fn color_from_fill(fill: &usvg::Fill) -> Color {
+    match fill.paint {
+        usvg::Paint::Color(c) => {
+            Color::rgb8(c.red, c.green, c.blue).with_alpha(fill.opacity.value())
```
cmyr: these two impls share the bulk of their code, might be room for a helper function?

I also might consider making these impls of the `From` trait, but I totally don't mind doing it like this for one-offs.

druid/src/widget/svg.rs
```diff
@@ -0,0 +1,284 @@
...
+
+#[cfg(feature = "svg")]
+use crate::{
+    kurbo::BezPath,
+    piet::{Color, RenderContext},
+    Point,
+};
+
+#[cfg(feature = "svg")]
+pub use usvg;
```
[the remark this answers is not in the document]
futurepaul: Reexporting is to cover the use case of someone wanting do something custom with SVGs while relying on my svg-to-piet function (which expects a usvg tree). If people have to import usvg in their own cargo.toml to do stuff that could cause conflicts yes?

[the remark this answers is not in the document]
cmyr: it would only cause conflicts if they were using a specific version of `usvg` that was different from the one that we support. I think it's okay to not re-export; presumably somebody using this functionality can control the version of `usvg` that they support, and if there are future major releases it shouldn't be hard for us to update.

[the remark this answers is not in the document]
Ralith: I think reexporting crates that appear in the public API is reasonably good practice to reduce breakage. Better still might be hiding the external dep entirely with a newtype, though.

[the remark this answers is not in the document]
futurepaul: Okay I did the obvious-in-hindsight thing and just exported the functions we need to make a tree out of a str or a path.

futurepaul:
I think everything is addressed. Only issue is that CI doesn't know about this feature flag. If it did it would notice those two (unrelated) clippy lints that show up when running with this flag.

cmyr:
hmm, I guess we can't do `cargo test --all-features`?

futurepaul:
I think it tries to run all the platform-specific stuff, so it has a ton of errors on my system.

druid/examples/svg.rs
```diff
@@ -0,0 +1,183 @@
...
+
+//! This example shows how to draw an SVG.
+//!
+//! Requires the non-default "svg" feature to be enabled:
+//! `cargo run --example svg --features "svg"`
+
+use druid::widget::{Flex, Svg, WidgetExt};
+use druid::{AppLauncher, Widget, WindowDesc};
+
+const TIGER: &str = r###"
```
Ralith: I think this would be better off as an `include_str`.

druid/src/widget/svg.rs
```diff
@@ -0,0 +1,294 @@
...
+
+    fn event(&mut self, _ctx: &mut EventCtx, _event: &Event, _data: &mut T, _env: &Env) {}
+
+    fn update(&mut self, _ctx: &mut UpdateCtx, _old_data: Option<&T>, _data: &T, _env: &Env) {}
+}
+
+// Create a raw usvg tree from an SVG string.
+// Pass this tree to `svg_to_piet()` to draw it.
+#[cfg(feature = "svg")]
+pub fn svg_tree_from_str(svg_str: &str) -> usvg::Tree {
```
Ralith: Consider wrapping this in a newtype (and then perhaps posing this function as a constructor for the newtype) so we're not exposing an external crate's type in our API. That way a user can't inadvertently come to rely on it.

druid/src/widget/svg.rs
```diff
@@ -0,0 +1,294 @@
...
+        ..usvg::Options::default()
+    };
+
+    match usvg::Tree::from_str(svg_str, &re_opt) {
+        Ok(tree) => tree,
+        Err(err) => {
+            error!("{}", err);
+            usvg::Tree::from_str(EMPTY_SVG, &re_opt).unwrap()
+        }
+    }
```
Ralith: This needs to return a `Result` rather than trying to handle the error internally.

druid/src/widget/svg.rs
```diff
@@ -0,0 +1,294 @@
...
+}
+
+// Create a raw usvg tree from a path to an SVG.
+// Pass this tree to `svg_to_piet()` to draw it.
+#[cfg(feature = "svg")]
+pub fn svg_tree_from_path(path: impl AsRef<std::path::Path>) -> usvg::Tree {
```
Ralith: Given how convenient it is to call [`fs::read_to_string`](https://doc.rust-lang.org/std/fs/fn.read_to_string.html), I don't think it's worth expanding our API for this.

druid/src/widget/mod.rs
```diff
@@ -70,3 +70,10 @@ pub use widget_ext::WidgetExt;
 
 mod list;
 pub use crate::widget::list::{List, ListIter};
+
+mod svg;
+#[cfg(feature = "svg")]
+pub use crate::widget::svg::{svg_to_piet, svg_tree_from_path, svg_tree_from_str, Svg};
+
+#[cfg(not(feature = "svg"))]
+pub use crate::widget::svg::FakeSvg as Svg;
```
Ralith: I'd rather see the module feature-guarded in totality rather than playing error-prone and surprising tricks like this, especially since this doesn't even cover the full API.

cmyr: the discussion was that sometimes that leads to confusing errors, but I do agree that it's not a great feeling to have all this extra code for not a lot of tangible benefit.

druid/examples/svg.rs
cmyr: oops yes I missed this file, that's a good suggestion.

druid/src/widget/mod.rs
Ralith: I think having the API silently do nothing (in the very likely event that you're not seeing `log` messages), or change shape in arbitrary ways (the current state, where not all items are stubbed) is strictly worse than having things missing, particularly given that rustdoc can be manipulated to prominently display feature guards.

cmyr: cool, didn't know about that you could show guards in rustdoc.

I'm happy to have this removed, it does make things clearer. Maybe another way to salve this wound would be to turn the `svg` feature on by default?

Ralith: See [tokio](https://docs.rs/tokio/0.2.2/tokio/) for an example of this. Note that it requires nightly rustdoc.

I think enabling it by default would somewhat defeat the point of keeping druid slim.

raphlinus: Agree, experience in the Rust ecosystem is that it's *much* harder to turn off default features than it is to opt in.

druid/src/lib.rs
```diff
@@ -16,6 +16,7 @@
 
 #![deny(intra_doc_link_resolution_failure, unsafe_code)]
 #![allow(clippy::new_ret_no_self)]
+#![feature(doc_cfg)]
```
Ralith: This needs to be `#![cfg_attr(docsrs, feature(doc_cfg))]`.

druid/src/widget/mod.rs
```diff
@@ -70,3 +70,10 @@ pub use widget_ext::WidgetExt;
 
 mod list;
 pub use crate::widget::list::{List, ListIter};
+
+#[cfg(any(feature = "svg", rustdoc))]
+#[doc(cfg(feature = "svg"))]
+mod svg;
+#[cfg(any(feature = "svg", rustdoc))]
+#[doc(cfg(feature = "svg"))]
```
Ralith: What happened to the `cfg_attr` guards?

futurepaul: They weren't doing anything that I could detect. When I run it like this I see `This is supported on feature="svg" only.` messages under the type's methods which isn't amazing.

futurepaul: Okay I switched back to cfg_attr, can you confirm this looks good?

Ralith: Looks reasonable; can't test personally right now but it fixes CI and we can always tweak it.

druid/src/lib.rs
```diff
@@ -16,6 +16,7 @@
 
 #![deny(intra_doc_link_resolution_failure, unsafe_code)]
 #![allow(clippy::new_ret_no_self)]
+#![cfg_attr(docsrs, feature(doc_cfg))]
```
cmyr: what does this do?

druid/src/widget/svg.rs
```diff
@@ -0,0 +1,230 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! An SVG widget.
+
+use crate::{
```
cmyr: not a blocker, but I generally try to `std` imports first, then extern items, then crate items.

druid/src/widget/svg.rs
```diff
@@ -0,0 +1,230 @@
...
+    pub fn new(svg_str: &str) -> impl Widget<T> {
+        let svg_data = SvgData::from_str(svg_str);
+
+        match svg_data {
+            Ok(data) => Svg {
+                svg_data: data,
+                phantom: Default::default(),
+            },
+            Err(err) => {
+                error!("{}", err);
```
cmyr: I think this is reasonable until we figure out a better error handling story. 🤷‍♀

druid/src/widget/svg.rs
```diff
@@ -0,0 +1,230 @@
...
+    phantom: PhantomData<T>,
+}
+
+impl<T: Data> Svg<T> {
+    /// Create an SVG-drawing widget from a valid SVG string literal.
+    ///
+    /// The SVG will scale to fit its box constraints.
+    /// If SVG is invalid a blank SVG will be rendered instead.
+    pub fn new(svg_str: &str) -> impl Widget<T> {
+        let svg_data = SvgData::from_str(svg_str);
```
cmyr: you know the trick with `FromStr`? that it gives you a new fn on string, `parse()`, so this could also be written,

```rust
let svg_data = svg_str.parse::<SvgData>();
```

(this isn't a suggested change, I'm just pointing out this feature)

druid/src/lib.rs
futurepaul: I'm fairly certain this is saying: if compiling for docs.rs, do this fancy doc_cfg thing.

druid/src/widget/svg.rs
Ralith: Is there a good reason not to return `Result`?

druid/src/widget/svg.rs
```diff
@@ -0,0 +1,230 @@
...
+    svg_data: SvgData,
+    phantom: PhantomData<T>,
+}
+
+impl<T: Data> Svg<T> {
+    /// Create an SVG-drawing widget from a valid SVG string literal.
+    ///
+    /// The SVG will scale to fit its box constraints.
+    /// If SVG is invalid a blank SVG will be rendered instead.
+    pub fn new(svg_str: &str) -> impl Widget<T> {
```
Ralith: If this took `SvgData` rather than `str`, then redundant work could be avoided in constructing multiple widgets with the same SVG, and there'd be no need to deal with errors internally.

druid/src/widget/svg.rs
cmyr: not a great reason, but my thinking right now is that no other widget constructors are failable, and it feels weird to start here.

My long-term vision for this is that you shouldn't be loading data directly from the file system; there should be a 'resource bundle' that druid exposes that contains various types of resources (image files, font files, localization files) and that is validated at compile time, so that at runtime getting a resource looks something like,

```rust
let my_svg: SvgData = env.get_bundle_resource(MY_SVG_IDENTIFIER);
```

or maybe even in a very fancy world there's a build phase that generates per-app marker 'keys' (like `env::Key`) for each of the resources in this app's bundle?

druid/src/widget/svg.rs
cmyr: This might be a good compromise around the error handling; you are expected to check errors when constructing `SvgData`, but then you're free to do what you want.

We could also consider implementing `Default` for `SvgData`, which would do what `Svg::empty()` does? I'm not sure if this is helpful or not.

If we expect `SvgData` to be shared, it might be worth `Arc`ing the tree?

druid/src/widget/svg.rs
futurepaul: I can open an issue for this if you'd like?

druid/src/widget/svg.rs
futurepaul: Okay how does this look?

druid/src/widget/svg.rs
```diff
@@ -0,0 +1,229 @@
...
+use crate::{
+    kurbo::BezPath,
+    piet::{Color, RenderContext},
+    BaseState, BoxConstraints, Data, Env, Event, EventCtx, LayoutCtx, PaintCtx, Point, Size,
+    UpdateCtx, Widget,
+};
+
+/// A widget that renders a SVG
+pub struct Svg<T> {
+    // On construction the SVG string is parsed into a SvgData.
```
Ralith: Outdated comment

futurepaul: thank you!

druid/src/widget/svg.rs
cmyr: sure!

cmyr:
this looks good to me. @Ralith this is blocked on your requested changes, so if you want to take another look then we can merge on your ✅

druid/src/widget/svg.rs
```diff
@@ -0,0 +1,228 @@
...
+    /// The SVG will scale to fit its box constraints.
+    pub fn new(svg_data: SvgData) -> impl Widget<T> {
+        Svg {
+            svg_data,
+            phantom: Default::default(),
+        }
+    }
+
+    /// Measure the SVG's size
+    #[allow(clippy::needless_return)]
```
Ralith: Why not use conventional style here?

druid/src/widget/svg.rs
```diff
@@ -0,0 +1,228 @@
...
+
+        match usvg::Tree::from_str(svg_str, &re_opt) {
+            Ok(tree) => Ok(SvgData {
+                tree: Arc::new(tree),
+            }),
+            Err(err) => Err(err.into()),
+        }
+    }
+}
+fn color_from_usvg(paint: &usvg::Paint, opacity: usvg::Opacity) -> Color {
```
Ralith: Missing empty line before this

druid/src/widget/svg.rs
futurepaul: The borrow checker disagrees with clippy on the "needless" aspect: https://rust-lang.github.io/rust-clippy/master/index.html#needless_return

RazrFalcon:
>The downside: it's a bloat monster because I'm relying on usvg

Sad, but reasonable. I just wanted to note that `usvg` is absolutely stripped down. There are nothing to remove. The bloat comes mainly from harfbuzz bindings and there is nothing I can do about it. SVG is hard, text is hard.

futurepaul:
I didn't mean this as a criticism of usvg! Just in comparison to the average PR. Thank you for making this widget so straightforward to implement!

PR #355, opened 2019-11-26
Convenience macros for Flex construction

Ralith:
These allow more concise construction of simple layouts.

Followup to #351. Syntax bikeshedding welcome. My first try looked like `row![(widget, flex), ...]` but this seems more readable.

### DIFF
--- a/druid/examples/layout.rs
+++ b/druid/examples/layout.rs
@@ -16,26 +16,24 @@
 
 use druid::piet::Color;
 use druid::widget::{Button, Flex, Label, SizedBox, WidgetExt};
-use druid::{AppLauncher, Widget, WindowDesc};
+use druid::{row, AppLauncher, Widget, WindowDesc};
 
 fn build_app() -> impl Widget<u32> {
     // Begin construction of vertical layout
     let mut col = Flex::column();
 
     // Construct a horizontal layout.
-    let mut header = Flex::row();
-    header.add_child(
-        Label::new("One")
+    let header = row! {
+        0.0 => Label::new("One")
             .center()
             .fix_width(60.0)
             .background(Color::rgb8(0x77, 0x77, 0))
             .border(Color::WHITE, 3.0),
-        0.0,
-    );
+        1.0 => SizedBox::empty().expand(),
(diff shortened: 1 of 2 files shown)


cmyr:
Having to specify the flex factor really makes it hard to come up with a good API for this.

Looking over the macro, it really feels like... a less idiomatic version of a builder. That is, the code:

```rust
let row = row! {
    0.0 => Label::new("foo"),
    1.0 => ProgressBar::new(),
};
```

seems like it should just be (names tbd):

```rust
let row = Flex::row()
    .with_child(Label::new("foo"), 0.0),
    .with_child(ProgressBar::new(), 1.0);
```

We don't currently do this (have builder methods on `Flex`) but there's no reason why not, and I think this would give us everything the macro does?

Similarly, I'm really not sure of the advantage of the `from_iter` methods. I could _kind of_ imagine the advantage of actually implementing `FromIterator<(Box<dyn Widget<T>>, f64)>`, so that you could, say, work with `collect()` but I don't really think this works in practice either, unless we made `Axis` a generic parameter of `Flex`.

I'm curious if there's some scenario I'm missing, where this approach has clear advantages? I'm very open to that being the case, but from my current perspective I'm just not sure this is adding enough functionality to justify the additional code.

raphlinus:
As a general philosophical thought, I have very mixed feelings about macros. I certainly understand the desire to make things look more declarative, but I think there are downsides. The biggest is that it obscures what the code really does. Certainly this is my reaction when I look at an unfamiliar codebase that's heavily reliant on macros. And I can think of a number of projects that used to rely on macros and are moving away (like nom).

Another concern (and this is shared by the `WidgetExt` methods) is that there is more than one way to do things, which creates confusion about which way is preferred. That's not to say I'm against using macros to make code more concise and cleaner, but I want to be careful.

Just for contrast, an example where I *do* think macros are important is deriving lenses and the `Data` trait. I have lots of experience now of forgetting to update the `same` method when adding fields, and using a macro really helps with that.

Ralith:
I think a builder is a much better idea.

PR #356, opened 2019-11-26
Lens derive fixes

Ralith:
Fixes odd corner-cases of name resolution, e.g. when inside a function.


Ralith:
Basically, `super` doesn't work the way you'd expect for modules defined
inside a function, with regard to types also defined in the function, which
comes up a lot in doctests if nowhere else.

PR #358, opened 2019-11-27
Flex::with_child

Ralith:
Straightforward ergonomic improvement. I didn't update all the examples because it was getting tedious and didn't seem necessary.

### DIFF
--- a/druid/src/widget/flex.rs
+++ b/druid/src/widget/flex.rs
@@ -121,6 +121,14 @@ impl<T: Data> Flex<T> {
         }
     }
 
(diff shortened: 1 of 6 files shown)


druid/src/widget/flex.rs
```diff
@@ -138,6 +140,14 @@ impl<T: Data> Flex<T> {
         };
         self.children.push(child);
     }
+
```
cmyr: biggest nit in the world: I put builder functions right below constructors and before all other methods.

Ralith: Done.

PR #359, opened 2019-11-28
[mac] Don't manually call finishLaunching()

cmyr:
According to the NSApplication docs, this is called automatically
when the `run()` method is called.

@raphlinus I might be missing something here, but it this call seems out of place, and nothing seems broken with it removed.

### DIFF
--- a/druid-shell/src/platform/mac/runloop.rs
+++ b/druid-shell/src/platform/mac/runloop.rs
@@ -32,7 +32,6 @@ impl RunLoop {
 
             let ns_app = NSApp();
             ns_app.setActivationPolicy_(NSApplicationActivationPolicyRegular);
-            ns_app.finishLaunching();
             RunLoop { ns_app }
         }
     }
--- a/druid-shell/src/platform/mac/window.rs
+++ b/druid-shell/src/platform/mac/window.rs
@@ -142,12 +142,6 @@ impl WindowBuilder {
                 NO,
             );
 
-            // TODO:
-            // I'm not actually sure if this call is necessary.
-            // I believe this should really be
-            // window.setLevel_(CGWindowLevelForKey(CGWindowLevelKey::normalWindow));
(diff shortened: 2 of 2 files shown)


ratmice:
@raphlinus that was part of the attempt to fix #43 
the thing that `finishLaunching` is supposed to be doing side-effects:
causing creation of the connection between NSApplication and the window server, 
registering that connection with the NSRunLoop,

While `NSApplication run` *will* call `finishLaunching` if it hasn't already,

We in fact call a bunch of `NSWindow` methods such as ordering in, in between `sharedApplication` and `run`. see [druid/src/app.rs](https://github.com/xi-editor/druid/blob/master/druid/src/app.rs#L97-L110), removing the `finishlaunching` call effectively moves the time between its call from the first line highlighted to the last line highlighted.

In a typical macOS program all of this ordering will happen e.g. during .nib file loading called from some method which has been called by or after `finishLaunching`. But creating this delegate and passing it all the state it needs to do the ordering makes memory management from within that call is tricky to do with rust.

I would certainly be curious about the effect on #43 after removing this method call, because the only thing left of the fix for that, is the setting of the `windowLevel` to what I believe might be the default window level.  My thinking is that the `finishLaunching` call might be more important than that `setWindowLevel_(0)` to the fix.

I cannot test macOS however, if moving the call from one place to the other has no effect i'd certainly remove it.

cmyr:
@ratmice thanks for the explanation, that makes sense. I'll read through that article and see if I get a better understanding of what's going on.

cmyr:
Okay, well, there went an afternoon.

The actual bug here was subtle, and is fixed in #360. I've updated this PR to remove both the `finishLaunching` and the `setLevel` calls, neither of which seems necessary.

cmyr:
#360 fixes the window level issue; this code has no apparent effect, and is not justified by my understanding of the application launch flow.

PR #360, opened 2019-11-28
Fix #43, again (🤞)

### LINKED ISSUE #43: Make mac Window creation smoother

cmyr:
My main strategy here was to create a minimal single-window objective-c
application (based on [1]) and then to trace our executation
and compare that against the execution of the objective-c app.

There were a few places where our behaviour differed, but changing
those had no apparent effect on application behaviour; but in
the course of that work I found the actual bug, which was much sneaker.

TL;DR: We were using [NSRunningApplication activateWithOptions:],
when we should have been using [NSApplication  activateIgnoringOtherApps:].
The NSRunningApplication API is intended for controlling other running
applications; we were likely sending the msg to nil.

[1]: https://www.cocoawithlove.com/2010/09/minimalist-cocoa-programming.html

### DIFF
--- a/druid-shell/src/platform/mac/runloop.rs
+++ b/druid-shell/src/platform/mac/runloop.rs
@@ -15,7 +15,7 @@
 //! macOS implementation of runloop.
 
 use cocoa::appkit::{NSApp, NSApplication, NSApplicationActivationPolicyRegular};
-use cocoa::base::{id, nil};
+use cocoa::base::{id, nil, YES};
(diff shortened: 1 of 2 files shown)


ratmice:
Aha, so because the NSApplication hadn't `run()` yet when `show()` is called `[NSRunningApplication currentApplication]` returned nil, that makes a lot of sense and seems entirely plausible to me. good sleuthing!

PR #361, opened 2019-11-28
[mac] Setup window delegate in build(), not show()

cmyr:
Minor cleanup, but this feels more correct.


cmyr:
I'm going to revisit this when #362 is merged; I don't need to use NSNotificationCenter for this if I'm implementing NSWindowDelegate.

Issue #114, opened 2019-08-09
Add some kind of top-level application type


cmyr:
I would like some object at a higher level than a window that can serve as a general application coordinator. This will be responsible for handling top-level lifecycle and system events, and managing windows.

A basic question (with a macOS bias): if I have no windows open, who handles a File > Open event?

ratmice:
IIUC this is partially there with AppDelegate<T + Data> and Commands,
With the missing bits being in `platform/`, having a mechanism for the backend to send Commands or Events like OpenFile.  AppDelegate<T> doesn't by itself get passed down to platform and runloop, and in fact Data doesn't exist at the `druid-shell` level.

As far as I can tell, all events currently in the macOS platform currently come from an `ivar` held within an `NSView`.  One thing worth noting is that on macOS there is *always* a window.  If there is no window on screen, there is still the application icon.  Well i guess it isn't actually a window, but it does have a view, and afaict that is the main thing that is needed.

The `NSApplication` class has a method `-dockTile`, which has a method `contentView` which you can use to set/get the AppIcon's view.

File > Open outside of dialog boxes on macOS happens through the NSApplicationDelegate class, which will receive a message `application:openFile:`, the first argument being the application from where you can take get the `dockTile` from.  Plus we get to draw on it!

Is there anything that I am missing already which allows us to send events to AppDelegate<T> events without a window?  
Would making an icon window (window-like-thing?) a parameter to AppLauncher be a viable solution?

ratmice:
IIUC this is partially there with AppDelegate<T + Data> and Commands,
With the missing bits being in `platform/`, having a mechanism for the backend to send Commands or Events like OpenFile.  AppDelegate<T> doesn't by itself get passed down to platform and runloop, and in fact Data doesn't exist at the `druid-shell` level.

As far as I can tell, all events currently in the macOS platform currently come from an `ivar` held within an `NSView`.  One thing worth noting is that on macOS there is *always* a window.  If there is no window on screen, there is still the application icon.  Well i guess it isn't actually a window, but it does have a view, and afaict that is the main thing that is needed.

The `NSApplication` class has a method `-dockTile`, which has a method `contentView` which you can use to set/get the AppIcon's view.

File > Open outside of dialog boxes on macOS happens through the NSApplicationDelegate class, which will receive a message `application:openFile:`, the first argument being the application from where you can take get the `dockTile` from.  Plus we get to draw on it!

[...]

PR #362, opened 2019-11-28
Standardize window closing API

cmyr:
We want to be notified when a window closes. For this to work we need
to handle the destroy() method in WinHandler.

After making this change, we ended up with two code paths for closing
windows: one for when we were closing the window ourselves (for
instance, when manually handling a menu command) and a second for when
we were closed by the system, such as when the user clicked on the
window's 'close' button. (all examples macOS).

This unifies these two cases; when we want to close a window we
call the WindowHandle's close method, which begins the platform
close routine; as part of this we are called back (at destroy())
and then we have a single place to handle cleaning up state
and posting the appropriate events.

-----
- this was motivated by runebender stuff; I only want a glyph open in a single window at once, which means i need to accurately know when a window has closed.

- This will likely require work on win and gtk; I'm not sure if either of those platforms is correctly posting `destroy()`.

### DIFF
--- a/druid-shell/src/platform/mac/window.rs
+++ b/druid-shell/src/platform/mac/window.rs
@@ -158,6 +158,8 @@ impl WindowBuilder {
             let frame = NSView::frame(content_view);
             view.initWithFrame_(frame);
 
+            let () = msg_send![window, setDelegate: view];
+
             if let Some(menu) = self.menu {
                 NSApp().setMainMenu_(menu.menu);
             }
@@ -277,6 +279,10 @@ lazy_static! {
         decl.add_method(
             sel!(showContextMenu:),
             show_context_menu as extern "C" fn(&mut Object, Sel, id),
+        );
+        decl.add_method(
+            sel!(windowWillClose:),
+            window_will_close as extern "C" fn(&mut Object, Sel, id),
         );
         ViewClass(decl.register())
     };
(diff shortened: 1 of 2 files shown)


xStrom:
I'm just starting my journey to understand druid internals, but I think the Windows platform code [already triggers the destroying message](https://github.com/xi-editor/druid/blob/ca0aa8dde1e80bd0b088b31b067591e3e62575a6/druid-shell/src/platform/windows/window.rs#L1076-L1083).

I've been looking for window closing related functionality in druid but as best as I can make out, significant parts of it are not implemented yet. Standardizing on one single place to handle cleanup is definitely a good step forward. I have some additional thoughts here but my knowledge is mostly Windows based and I may be lacking understanding of druid given that I just started learning it. Also I don't want to balloon the scope of this PR any more than necessary, but for now this seems like an on-topic place to comment.

### Exiting the `RunLoop` when all windows have been destroyed
On the surface this seems like a good idea to me. Currently the process keeps on running even after all the windows have been destroyed. Is there some specific reason why the runloop keeps going or is this just a matter of not having been implemented yet?

### Handling user requested window close events, e.g. `WM_CLOSE` on Windows
Currently `WM_CLOSE` is handled by `DefWindowProc` which just destroys the window. To enable *Exit without saving?* type confirmations we should handle the `WM_CLOSE` event. My initial thought is that this should be translated into some sort of platform independent event. If the client application handles the event, then its their job to eventually destroy the window, or not. If the event is not handled then druid can fall back to destroying the window.

cmyr:
The reason a bunch of this stuff hasn't been implemented is because there are significant differences in expected behaviour on different platforms.

on windows, yes, we should terminate the application when the last window closes. We could implement this today, specifically for windows, with little trouble.

on mac, an application is not terminated by closing its windows; it can still be interacted with the menubar or the dock icon (and various other ways). Making this work correctly is what motivates #114.

I'm not sure about other platforms, e.g. GTK.

I agree that we want some sort of `ShouldCloseWindow` event. This is definitely in the "just not implemented yet" category.

cmyr:
self approving this, would like to clear up the queue

PR #367, opened 2019-12-01
Deref, Index, and InArc lenses

Ralith:
Builds on #344.

Note that this does not provide a way to index into an `Arc<Vec<T>>`. That's intentional, as any such operation would cause a deep copy in every `Widget::event`.

### DIFF
--- a/druid/src/lens.rs
+++ b/druid/src/lens.rs
@@ -119,8 +119,7 @@ pub trait LensExt<A: ?Sized, B: ?Sized>: Lens<A, B> {
     ///
     /// ```
     /// # use druid::*;
-    /// let lens = lens!((bool, Box<u32>), 1).deref();
-    /// assert_eq!(lens.get(&(false, Box::new(42))), 42);
+    /// assert_eq!(lens::Id.deref().get(&Box::new(42)), 42);
     /// ```
     fn deref(self) -> Then<Self, Deref, B>
     where
@@ -134,8 +133,7 @@ pub trait LensExt<A: ?Sized, B: ?Sized>: Lens<A, B> {
     ///
     /// ```
     /// # use druid::*;
-    /// let lens = lens!((bool, Vec<u8>), 1).deref().index(2);
-    /// assert_eq!(lens.get(&(false, vec![0, 1, 2, 3])), 2);
+    /// assert_eq!(lens::Id.index(2).get(&vec![0u32, 1, 2, 3]), 2);
     /// ```
     fn index<I>(self, index: I) -> Then<Self, Index<I>, B>
     where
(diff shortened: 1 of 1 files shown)


druid/src/lens.rs
```diff
@@ -194,9 +196,174 @@ where
...
+    fn with_mut<V, F: FnOnce(&mut A) -> V>(&self, data: &mut A, f: F) -> V {
+        f(data)
+    }
+}
+
+/// A `Lens` that exposes data within an `Arc` with copy-on-write semantics
+///
+/// A copy is only made in the event that a different value is written.
+#[derive(Debug, Copy, Clone)]
+pub struct MakeMut<L, M, B> {
```
Ralith: Bikeshedding very welcome on this name.

druid/src/lens.rs
```diff
@@ -91,6 +93,64 @@ pub trait LensExt<A: ?Sized, B: ?Sized>: Lens<A, B> {
...
+    /// let original = x.clone();
+    /// assert_eq!(lens.get(&x), 2);
+    /// lens.put(&mut x, 2);
+    /// assert!(Arc::ptr_eq(&original, &x), "no-op writes don't cause a deep copy");
+    /// lens.put(&mut x, 42);
+    /// assert_eq!(&*x, &[0, 1, 42, 3]);
+    /// ```
+    fn make_mut<L, C, D>(self, lens: L) -> MakeMut<Self, L, C>
+    where
+        Self: Sized + Lens<A, Arc<C>>,
```
Ralith: This feels like a hack, but it works.

druid/src/lens.rs
```diff
@@ -91,6 +93,59 @@ pub trait LensExt<A: ?Sized, B: ?Sized>: Lens<A, B> {
...
+    /// let lens = lens::Id.index(2).in_arc();
+    /// let mut x = Arc::new(vec![0, 1, 2, 3]);
+    /// let original = x.clone();
+    /// assert_eq!(lens.get(&x), 2);
+    /// lens.put(&mut x, 2);
+    /// assert!(Arc::ptr_eq(&original, &x), "no-op writes don't cause a deep copy");
+    /// lens.put(&mut x, 42);
+    /// assert_eq!(&*x, &[0, 1, 42, 3]);
+    /// ```
+    fn in_arc(self) -> InArc<Self>
```
Ralith: Bikeshedding still welcome, though I think this is probably adequate.

druid/src/lens.rs
```diff
@@ -290,3 +343,104 @@ impl<T: Clone, U: Clone, B> Clone for Then<T, U, B> {
         }
     }
 }
+
+#[derive(Debug, Copy, Clone)]
+pub struct Deref;
```
cmyr: if this can appear in documentation I would give it a doc string.

druid/src/lens.rs
```diff
@@ -290,3 +343,104 @@ impl<T: Clone, U: Clone, B> Clone for Then<T, U, B> {
...
+    A: Clone,
+    B: Data,
+    L: Lens<A, B>,
+{
+    fn with<V, F: FnOnce(&B) -> V>(&self, data: &Arc<A>, f: F) -> V {
+        self.inner.with(data, f)
+    }
+
+    fn with_mut<V, F: FnOnce(&mut B) -> V>(&self, data: &mut Arc<A>, f: F) -> V {
+        let mut temp = self.inner.with(data, |x| x.clone());
```
cmyr: clever solution. :)

druid/src/lens.rs
Ralith: Good catch!

druid/src/lens.rs
cmyr: I can't think of anything better? maybe just `arc`, but either or.

PR #368, opened 2019-12-01
Helper lens for simple transformations

Ralith:
Builds on #344.

I experimented with a number of different approaches, and settled on something similar to the existing `Field` lens, but handling the target type by value so as to permit computation. This is minimally opinionated, while still making it easy to e.g. convert units or adjust slider ranges.

### DIFF
--- a/druid/src/lens.rs
+++ b/druid/src/lens.rs
@@ -91,6 +91,28 @@ pub trait LensExt<A: ?Sized, B: ?Sized>: Lens<A, B> {
     {
         Then::new(self, other)
     }
+
+    /// Combine a `Lens<A, B>` with a function that can transform a `B` and its inverse.
+    ///
+    /// Useful for cases where the desired value doesn't physically exist in `A`, but can be
+    /// computed. For example, a lens like the following might be used to adapt a value with the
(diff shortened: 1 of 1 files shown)


druid/src/lens.rs
```diff
@@ -91,6 +91,22 @@ pub trait LensExt<A: ?Sized, B: ?Sized>: Lens<A, B> {
     {
         Then::new(self, other)
     }
+
+    /// Combine a `Lens<A, B>` with getter/setter functions to produce a `Lens<A, C>`
+    ///
+    /// ```
+    /// # use druid::*;
```
cmyr: Reading through these PRs, I am consistently wishing that there was more substantial documentation; ideally this would take the form of illustrating how this particular lens is useful in some hypothetical but realistic seeming scenario.

It's possible that this example could be that, but it would be nice to have a bit more hand holding, like "say you have data of type $type and a widget that expects a $something, using $this_lens, you can ...."

cmyr: basically: I think you're way down there in the lens mines working on your own, and those of us puttering around up here on the surface need a bit of help to understand the particular motivations for various designs. :)

Ralith: Added some more discussion.

PR #369, opened 2019-12-02
Handle home/end in text box

Ralith:


druid/src/widget/textbox.rs
```diff
@@ -381,12 +381,16 @@ impl Widget<String> for TextBox {
                         self.selection = Selection::new(0, data.len());
                     }
                     // Jump left (Ctrl+ArrowLeft || Cmd+ArrowLeft)
-                    k_e if (HotKey::new(SysMods::Cmd, KeyCode::ArrowLeft)).matches(k_e) => {
+                    k_e if HotKey::new(SysMods::Cmd, KeyCode::ArrowLeft).matches(k_e)
```
cmyr: Is ctrl + arrow jump left on all platforms?

We don't currently have an easy way to b fancier than this (that's part of the motivation for the `EditableText` trait work; we can have platform-specific mappings from keys to various edit events. This is an improvement in the meantime, though.

Ralith: > Is ctrl + arrow jump left on all platforms?

I have no idea, but that's pre-existing behavior.

PR #370, opened 2019-12-02
Implement KeyMatcher trait

raphlinus:
Adds a KeyMatcher trait, which is inhabited by nice, ergonomic combinations
of modifier specs and keys.

This is work in progress, it should be unified more with HotKeys.

### DIFF
--- a/druid-shell/src/keyboard.rs
+++ b/druid-shell/src/keyboard.rs
@@ -532,12 +532,12 @@ impl KeyMatcher for char {
         // We allow either text or unmodified text to match.
         let text = key_event.text().unwrap_or("");
         let mut chars = text.chars();
-        if chars.next() == Some(*self) && !chars.next().is_some() {
+        if chars.next() == Some(*self) && chars.next().is_none() {
             return true;
         }
         let unmod_text = key_event.unmod_text().unwrap_or("");
         let mut chars = unmod_text.chars();
-        if chars.next() == Some(*self) && !chars.next().is_some() {
+        if chars.next() == Some(*self) && chars.next().is_none() {
             return true;
         }
         false
--- a/druid-shell/src/keyboard.rs
+++ b/druid-shell/src/keyboard.rs
@@ -37,6 +37,37 @@ pub struct KeyEvent {
     //TODO: add time
 }
 
+/// A modifier, for the purpose of matching key events.
+pub enum Modifier {
+    Alt,
+    Ctrl,
+    Meta,
+    Shift,
+    // The command key; maps to Meta on macOS, Ctrl otherwise.
+    Cmd,
+    NoMods,
+}
+
+pub trait KeyMatcher {
+    fn matches(&self, key_event: &KeyEvent) -> bool {
+        self.matches_internal(key_event, KeyModifiers::default())
+    }
+
+    #[doc(hidden)]
+    #[allow(unused)]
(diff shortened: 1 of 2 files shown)


druid-shell/src/keyboard.rs
```diff
@@ -37,6 +37,37 @@ pub struct KeyEvent {
     //TODO: add time
 }
 
+/// A modifier, for the purpose of matching key events.
+pub enum Modifier {
```
cmyr: This has the same issue, I think, that is addressed by the `SysMods` and `RawMods` types in `hotkey.rs`: if we want to match 'undo' we need some key that is 'cmd' on macOS and 'ctrl' elsewhere.

druid-shell/src/keyboard.rs
```diff
@@ -84,6 +115,19 @@ impl KeyEvent {
...
     }
 
+    /// Test whether the key matches the given pattern.
+    ///
+    /// # Examples
+    /// ```
+    /// use druid_shell::{KeyCode, Modifier};
+    /// # let key_event = druid_shell::KeyEvent::for_test(Modifier::NoMods, "a", KeyCode::KeyA);
+    /// let is_undo = key_event.matches(Modifier::Ctrl + 'z');
+    /// let is_reset = key_event.matches(Modifier::Ctrl + Modifier::Alt + KeyCode::Delete);
```
cmyr: In general, I think it's important in our example code to show how things are supposed to work cross-platform; the first of these examples excludes the mac, and the second is windows only?

druid-shell/src/keyboard.rs
```diff
@@ -271,4 +582,26 @@ mod tests {
         assert!(!s_16.is_char_boundary(15));
         let _too_big = TinyStr::new("😍🥰😘😗");
     }
+
+    #[test]
+    fn key_matching() {
+        use super::Modifier::*;
+        let key = KeyEvent::for_test(NoMods, "a", KeyCode::KeyA);
+        assert!(key.matches('a'));
```
cmyr: following macOS, we allow a key's text equivalent to be multiple chars. I don't know of an actual keyboard layout where this is exercised, but I could certainly imagine it with indic languages?

druid-shell/src/keyboard.rs
raphlinus: Oh, I did the "Cmd" thing after writing the docstring. That does the platform-specific behavior you asked.

raphlinus:
Quick notes, but basically I don't have time right now to follow up; this should just be considered a source of ideas and I'll almost certainly close the PR.

I think it's possible to address the duplication with HotKey by using this as an ergonomic syntax for converting Into.

I didn't to multi code point string matching, but that's easy, just do an impl for `&'static string`. There's potentially a *lot* of complexity around shift and case matching.

druid-shell/src/keyboard.rs
raphlinus: At that point, we shouldn't be using key events but IME.

cmyr:
I think we should just skip shift magic when dealing with multi code point values, I don't think we even really know the expected behaviour in those cases and we're more likely to make things worse than to improve them? 😬

raphlinus:
Actually I don't know of any way to get a multi code point key in macOS. So we should probably do some research to determine if that's even a thing, before we put too much implementation work into it.

In any case, I'm closing this now. Keyboard events need to be reworked, as there's a lot of stuff that just doesn't work at all. But I do not have the bandwidth to get into those deep waters.

PR #371, opened 2019-12-02
Update README.md

sombrastudios:
I think the mentiioned conversion is missing in the docs


cmyr:
I _believe_ the docs are up to date with master, but perhaps are behind the last published version? We will be publishing a new version in the next week or so.

PR #373, opened 2019-12-02
Numpad keycodes take3

ratmice:
This is undoubtedly not going to pass CI without work on the windows and Mac backends,

I'm not too keen on the way this platform/gtk/window.rs stuff was handled.
But I couldn't come up with another solution that actually worked.

There are some comments inline about Text and whitespace/control characters, for now I tried to match the macOS backend which emits text for these.

There are some interim back and forth changes here which could use a squashing,
Where I had added some characters with Text, to the Special enum.

### DIFF
--- a/druid-shell/src/platform/gtk/keycodes.rs
+++ b/druid-shell/src/platform/gtk/keycodes.rs
@@ -314,3 +314,34 @@ impl From<KeyCode> for Special {
         }
     }
 }
+
+// I Don't think we actually want this, as it must make a choice to map to one of the
+// Left or Right keys in many cases, don't see much of a solution besides
+// Special::Control(keycode)?.
+//
+// But it's currently used in `menu::register_accelerator`.
+impl From<Special> for u32 {
+    #[allow(non_upper_case_globals)]
+    fn from(src: Special) -> u32 {
+        match src {
+            Special::Home => Home,
+            Special::Left => Left,
+            Special::Up => Up,
+            Special::Right => Right,
+            Special::Down => Down,
+            Special::PageUp => Page_Up,
+            Special::PageDown => Page_Down,
+            Special::End => End,
+            Special::Insert => Insert,
+            Special::Delete => Delete,
+            Special::Control => Control_R,
+            Special::Alt => Alt_R,
+            Special::Shift => Shift_R,
+            Special::Meta => Super_R,
+            Special::Unidentified => unreachable!(
+                "Unreachable: converting unknown Special {:?} to a keyval",
+                src
+            ),
+        }
+    }
+}
--- a/druid-shell/src/platform/gtk/keycodes.rs
+++ b/druid-shell/src/platform/gtk/keycodes.rs
@@ -16,7 +16,7 @@
 
 use gdk::enums::key::*;
 
-use crate::keycodes::KeyCode;
+use crate::keycodes::{KeyCode, Special};
 
 pub type RawKeyCode = u32;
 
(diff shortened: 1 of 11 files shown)


ratmice:
I should probably just leave the windows backend alone here, as it isn't clear to me where to use new_text vs new_special.

druid-shell/src/platform/gtk/window.rs
```diff
@@ -695,14 +696,91 @@ fn get_modifiers(modifiers: gdk::ModifierType) -> keyboard::KeyModifiers {
...
+    // and '\r' for Return.
+    //
+    // https://stackoverflow.com/questions/14705240/
+    //
+    // for gtk this currently would return '\r' for Return
+    // and NUL for KP_Enter.
+    //
+    // That leads to Text for Return and Special for Enter.
+    // Probably not the right thing.
+    // It seems like the w3c Web API doesn't actually give text for
```
ratmice: 1/2: If I understood the discussion on zulip about is_printchar()

druid-shell/src/platform/gtk/window.rs
```diff
@@ -695,14 +696,91 @@ fn get_modifiers(modifiers: gdk::ModifierType) -> keyboard::KeyModifiers {
...
+    } else {
+        // Although it isn't documented according from it's implementation
+        // in the gdk rust bindings, `keyval_to_unicode` will not return Some('\0').
+        // We rely on this to some extent.
+        gdk::keyval_to_unicode(keyval)
+    };
+
+    // [control/whitespace]
+    // we could filter out whitespace here by matching (What about Space? It seems fine to leave.)
+    // Some(c) if !(c.is_control() || c.is_whitespace()) => {
```
ratmice: 2/2: We want to not return Text for these, and use Special?

druid-shell/src/platform/gtk/window.rs
```diff
@@ -695,14 +696,91 @@ fn get_modifiers(modifiers: gdk::ModifierType) -> keyboard::KeyModifiers {
...
+    // In these cases, we want to map to a key which is *not* the lowest level key
+    // On PC Keyboards the lower level keys represent NumPadHome and friends.
+    // which don't exist in `keycodes::KeyCode` so we special case these here.
+    //
+    // We can't just do this in hardware_keycode_to_keyval by forcing an interpretation
+    // with Numlock on there since this is unlikely to work laptop keyboards.
+    //
+    // So we do 2 things:
+    // * If the keyval is a NumPad[0-9], decimal or separator, avoid getting the hardware keyval.
+    // * If the keyval is a NumPad{Home,...} We translate it to a NumPad[0-9].
```
ratmice: I find this pretty terrible, since it is essentially hard coding the correlation between KP_0, and KP_Home...
But haven't managed to successfully do anything else with the gtk backend.

druid-shell/src/platform/gtk/keycodes.rs
```diff
@@ -273,3 +273,44 @@ impl From<KeyCode> for u32 {
...
+            KeyCode::LeftShift
+            | KeyCode::RightShift
+            | KeyCode::Unknown(Shift_L)
+            | KeyCode::Unknown(Shift_R) => Special::Shift,
+            KeyCode::LeftMeta
+            | KeyCode::RightMeta
+            | KeyCode::Unknown(Super_L)
+            | KeyCode::Unknown(Super_R) => Special::Meta,
+            _ => {
+                log::warn!("Warning: unknown special keycode {:?}", src);
```
ratmice: I think this warning might need to go.

cmyr:
@ratmice where did we end up with this? Are we putting this on hold for the time being?

ratmice:
@cmyr Sorry for the lack of activity on my end,
It basically (for better or worse) implements what we discussed for the gtk backend.
My comments are mostly nitpicking where a decision needed to be made,

And one important thing pointing out places where the design ended up forcing a hard coded correlation between keys.  I'm at a loss for how to fix it given that gtk doesn't appear to know which modifier even corresponds to the state of the numlock key.

Nor does it provde any mechanism for converting between keys given a numlock modifier state.
(What essentially we want to do here is treat the numlock modifier as On for KeyCode, and it's actual state for Special, lacking any mechanism I could find to do so I had to hard code this).

So it's kind of ready for review in the sense that decisions need to be made about the nitpicks,
and whether the implementation is *too terrible* but I don't actually anticipate being able to do anything about the terribleness without e.g. treating each modified KeyCode as a distinct key gtk does, or maybe there is some API I am missing to avoid this hard coding.

So you could say I have hesitations about this implementation, but no path towards fixing those,
and it works somewhat as intended.

cmyr:
I am going to reluctantly declare bankruptcy on this issue. The work is appreciated, but I think the underlying architecture for `KeyEvent` needs to be overhauled, and it's hard to review this without addressing that.

Thanks and sorry @ratmice!

ratmice:
Totally understandable, and I basically agree -- but at this point I have run out of ideas for the right way to fix it.

cmyr:
Yea, it needs a careful reading of a bunch of existing APIs and then about a week of staring out into space. 😒

PR #378, opened 2019-12-04
Put Widget methods in lifecycle order

### LINKED ISSUE #377: Sort Widget methods to represent their lifecycle order
It's probably a super-minor issue, sorry was the noise if it is.

When implementing Widget trait for my widgets, I put methods in their lifecycle order (event -> update -> layout -> paint), so it's easier for me to understand how they influence each other.

Also, I decided to give a go CLion with Rust plugin. It has diagnostic which triggers on my impl with the message "Different impl member order from the trait". I guess it's reasonable as having methods in the same order for every impl makes skimming through the code easier.

I wonder if both things sound reasonable and it makes sense to change the order of method declaration in the trait for some gain in ergonomics?

ul:
Fix #377

### DIFF
--- a/druid/examples/anim.rs
+++ b/druid/examples/anim.rs
@@ -28,28 +28,6 @@ struct AnimWidget {
 }
 
 impl Widget<u32> for AnimWidget {
-    fn paint(
-        &mut self,
-        paint_ctx: &mut PaintCtx,
-        _base_state: &BaseState,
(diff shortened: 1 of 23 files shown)


cmyr:
Okay I couldn't get CI to restart, but I've run the tests on windows manually and I'm satisfied. Thanks!

PR #379, opened 2019-12-04
Add list with shared state example

zaynetro:
This is an attempt to document how to use List widget with shared data.

I don't know if I am using it correctly but it works! :) I am primarily interested to know if I am initiating and mutating the list with shared data correctly.

### DIFF
--- a/druid/examples/list.rs
+++ b/druid/examples/list.rs
@@ -12,19 +12,29 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::ops::Deref;
+//! Demos basic list widget and list manipulations.
+
 use std::sync::Arc;
 
 use druid::piet::Color;
 
-use druid::widget::{Button, Container, DynLabel, Flex, List, Padding, Scroll, SizedBox};
-use druid::{AppLauncher, Widget, WindowDesc};
+use druid::lens::{self, LensExt};
+use druid::widget::{Button, DynLabel, Flex, List, Scroll, WidgetExt};
+use druid::{AppLauncher, Data, Lens, LensWrap, Widget, WindowDesc};
 
-type AppData = Arc<Vec<u32>>;
+#[derive(Clone, Data, Lens)]
+struct AppData {
+    left: Arc<Vec<u32>>,
+    right: Arc<Vec<u32>>,
+}
 
 fn main() {
     let main_window = WindowDesc::new(ui_builder);
-    let data = Arc::new(vec![1, 2, 3, 4]);
+    // Set our initial data
+    let data = AppData {
+        left: Arc::new(vec![1, 2]),
+        right: Arc::new(vec![1, 2, 3]),
+    };
     AppLauncher::with_window(main_window)
         .use_simple_logger()
         .launch(data)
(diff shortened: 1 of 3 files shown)


druid/examples/list.rs
```diff
@@ -34,31 +44,79 @@ fn main() {
...
+                )
+                .padding(10.0)
+                .background(Color::rgb(0.5, 0.0, 0.5))
+                .fix_height(50.0)
+        }))
+        .vertical()
+        .lens(lens::Id.map(
+            // Expose shared data with children data
+            |d: &AppData| (d.right.clone(), d.right.clone()),
+            |d, x| {
```
MarcoPolo: I think it might be helpful to write the types here for the reader. i.e. `d: &mut AppData, x: Arc<Vec<u32>>, Arc<Vec<u32>>` (are those the right types?)

druid/examples/list.rs
```diff
@@ -34,31 +44,79 @@ fn main() {
...
-                    DynLabel::new(|d, _| format!("List item #{}", d)),
-                ))
-                .background(Color::rgb(0.5, 0.5, 0.5)),
-            )
-            .expand()
-            .height(50.0)
+            DynLabel::new(|item: &u32, _| format!("List item #{}", item))
+                .padding(10.0)
+                .expand()
+                .height(50.0)
```
cmyr: this is interesting; it's (maybe) a fluke that this works, because `height()` is not a fn on WidgetExt, but only on SizedBox; there's `fix_height(5.0)` on WidgetExt, but I don't love that name. Thinking about this a bit more, if the `WidgetExt` method was *also* called `height` this would probably just work? 🤔

druid/examples/list.rs
```diff
@@ -34,31 +44,79 @@ fn main() {
...
+                    0.0,
+                )
+                .padding(10.0)
+                .background(Color::rgb(0.5, 0.0, 0.5))
+                .fix_height(50.0)
+        }))
+        .vertical()
+        .lens(lens::Id.map(
+            // Expose shared data with children data
+            |d: &AppData| (d.right.clone(), d.right.clone()),
```
cmyr: maybe there should be a `Cloned` lens?

druid/examples/list.rs
```diff
@@ -34,31 +44,79 @@ fn main() {
...
+            |d, x| {
+                // If shared data was changed reflect the changes in our AppData
+                if !x.0.same(&d.right) {
+                    d.right = x.0
+                }
```
Ralith: I don't think this conditional is necessary; if `make_mut` was never called then it'll be a harmless noop.

druid/examples/list.rs
Ralith: Seems pretty niche.

cmyr:
Oh one other thing I noticed while playing with this is that (on a mac) scroll events are not always delivered to the view that is under the cursor?

druid/examples/list.rs
zaynetro: I think that we should unify the names. If we have `width` and `height` on `SizedBox` then we should have the same names on `WidgetExt` or vice versa. 
What do you think?

druid/examples/list.rs
zaynetro: I really don't like that we need to pass the same data two times. I hope that this is just a first step and later we can improve by wrapping it into some helper struct or using another data structure (what Raph was proposing).

zaynetro:
Yep, I have noticed that as well. Sometimes trying to scroll left list scrolls the right one. I haven't yet looked into what is happening.

druid/examples/list.rs
```diff
@@ -34,31 +44,79 @@ fn main() {
...
+                    }),
+                    1.0,
+                )
+                .with_child(
+                    Button::sized(
+                        "Delete",
+                        |_ctx, (shared, item): &mut (Arc<Vec<u32>>, u32), _env| {
+                            // We have access to both child's data and shared data.
+                            // Remove element from right list.
+                            Arc::make_mut(shared).retain(|v| v != item);
```
zaynetro: One more issue that we currently have is that we do not expose the index of the current element. If we have multiple children with the same value then this will remove them all at once.

Shall we pass an index/id in the list constructor? Or can I somehow compare by pointer equality?

druid/examples/list.rs
zaynetro: You mean dropping if condition and keeping the assignment all the time?

Ralith: Yes.

druid/examples/list.rs
cmyr: yea, that's an interesting question. This might be another situation where some sort of widget identity makes sense? I think it's okay to punt on this for now, but it's a question we'll revisit at some point.

druid/examples/list.rs
cmyr: yea, I think that makes sense, I think I hesitated because those two felt names overly generic, but I agree that would be an improvement.

zaynetro:
Thanks! There are definitely many unanswered questions.

PR #381, opened 2019-12-05
Fix fragility in TextBox under external data changes

Ralith:
For example, a wrapper widget that truncates the `String` could easily produce an out-of-bounds selection.

### DIFF
--- a/druid/src/widget/textbox.rs
+++ b/druid/src/widget/textbox.rs
@@ -132,12 +132,8 @@ impl TextBox {
     fn insert(&mut self, src: &mut String, new: &str) {
         // TODO: handle incomplete graphemes
 
-        // replace_range will panic if selection is greater than src length hence we try to constrain it.
-        // This is especially needed when data was modified externally.
-        let selection = self.selection.constrain_to(src);
-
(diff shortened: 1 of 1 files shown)


druid/src/widget/textbox.rs
```diff
@@ -219,6 +215,9 @@ impl TextBox {
 
 impl Widget<String> for TextBox {
     fn event(&mut self, ctx: &mut EventCtx, event: &Event, data: &mut String, env: &Env) {
+        // Guard against external changes in data
+        self.selection = self.selection.constrain_to(data);
```
cmyr: hmm, can we put this in `update`?

Ralith: That was the first thing I tried, but when you have a `TextBox` inside a `WidgetPod` inside a widget that modifies data after `event`, and it happens to modify it back to the data that it was before an event, `WidgetPod` would skip the `update` call as redundant and the internal state would get stuck out of sync.

cmyr: Interesting. good explanation, thanks.

PR #384, opened 2019-12-07
[gtk] sketch out a few ways to bypass a segfault on startup

ratmice:
This is all a bit ad-hoc since the API's lay across crate boundaries, and is mostly meant to be
informative rather than any actual API proposal.

when running multiple instances of druid apps,
since currently they share the same name, GTK will try to connect to the already running
instance, when it does so it never fully initializes the connection to the window system.

The remote application is already "activated", so the local application never "activated".
We sense this via the is_remote_connection() function.

In this patch there also a mechanism to "force" gtk to ignore any already running application.
This is the RunFlags::MultipleInstances flag which can be set on launch.
By setting this flag the is_remote_connection() flag never returns true.

So after this patch the behavioral changes under the assumption that
there is already a running instance of (currently any) druid program:

if RunFlags == MultipleInstances {
  both the existing and the new instance of the program will be launched.
}

if is_remote_connection() {
  exit the new instance we are in the process of launching.
}

### DIFF
--- a/druid-shell/src/lib.rs
+++ b/druid-shell/src/lib.rs
@@ -61,5 +61,5 @@ pub use keyboard::{KeyEvent, KeyModifiers};
 pub use keycodes::KeyCode;
 pub use menu::Menu;
 pub use mouse::{Cursor, MouseButton, MouseEvent};
-pub use runloop::RunLoop;
+pub use runloop::{RunFlags, RunLoop};
 pub use window::{Text, TimerToken, WinCtx, WinHandler, WindowBuilder, WindowHandle};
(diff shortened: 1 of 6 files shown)


Ralith:
I think silently dying if another instance is already open is surprising and not particularly useful. In lieu of first-class support for communicating with an existing instance (e.g. to tell it to open another window), IMO we should *always* provide the "multiple instances" behavior rather than letting it be a footgunny flag.

ratmice:
Totally agree, I just wanted to post this so we could take it into account when figuring out a design for #114 since just setting the flag doesn't really show the cross-crate logic.

PR #385, opened 2019-12-08
[gtk] set the gtk NON_UNIQUE flag on startup, to avoid connecting to a running instance

ratmice:
This is pulled out of my pull request #384 as its the least astonishing/zero additional API means to fix the segfault referenced there, which I would propose for committing as a temporary fix.
While #114 is not yet, place fixing that may impact the design of the API which was sketched out in #384 and the "exit if already running" behavior seen there isn't useful until that time.

This forces the behavior that each druid application can be run multiple times, and each instance creates windows, essentially disabling the gtk remote application behavior.


cmyr:
@ratmice thoughts on this? I'm happy to defer to @Ralith's approval, so merge if you like!

ratmice:
@cmyr thanks & sorry for the holiday delay, was holidaying.

PR #387, opened 2019-12-11
Add default file extension for the file dialog in Windows.

xStrom:
This PR adds the `default_type: Option<FileSpec>` property to the `FileDialogOptions` struct. If set to a `FileSpec` which matches one in `allowed_types` then that specific allowed type is chosen as the default in the file dialog. If it's `None` or a `FileSpec` which isn't actually in `allowed_types` then the default is the first item in `allowed_types`, like before.

As I currently only have easy access to a Windows machine I only implemented it for Windows. I'm not sure what the situation is for other platforms. Probably a good idea to create an issue to track that if this PR gets accepted.

Regarding the Windows implementation, choosing the default primarily depends on [IFileDialog::SetFileTypeIndex](https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-ifiledialog-setfiletypeindex) which was previously hardcoded to the first entry but now respects `default_type`. However it's crucial to also call [IFileDialog::SetDefaultExtension](https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-ifiledialog-setdefaultextension), at least on Windows 7, because otherwise later on [IShellItem::GetDisplayName](https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-ishellitem-getdisplayname) won't contain any extension unless the user manually entered it. Currently when a user types *myfile* there won't be any extension added by Windows, regardless of the choice of allowed type in the dialog. With this PR an extension gets added by Windows, based on the choice, i.e. *myfile.ext*.

Interestingly the actual string passed to [IFileDialog::SetDefaultExtension](https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-ifiledialog-setdefaultextension) doesn't seem to have any effect. It can be *foobar*, some extension present in `allowed_types`, weirdly formatted garbage, or even an empty string! The actual default choice is determined by [IFileDialog::SetFileTypeIndex](https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-ifiledialog-setfiletypeindex). However unless [IFileDialog::SetDefaultExtension](https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-ifiledialog-setdefaultextension) gets called, even if with an empty string, Windows 7 won't append the extensions.

I'm not sure how universal the irrelevance of the [IFileDialog::SetDefaultExtension](https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-ifiledialog-setdefaultextension) string is, so this PR implements it according to spec and uses only the first extension of `default_type`.

### DIFF
--- a/druid-shell/src/platform/windows/dialog.rs
+++ b/druid-shell/src/platform/windows/dialog.rs
@@ -39,6 +39,8 @@ use crate::dialog::{FileDialogOptions, FileDialogType, FileSpec};
 use std::ffi::OsString;
 use std::ptr::null_mut;
 
+use std::convert::TryInto;
+
 // TODO: remove these when they get added to winapi
 DEFINE_GUID! {CLSID_FileOpenDialog,
 0xDC1C_5A9C, 0xE88A, 0x4DDE, 0xA5, 0xA1, 0x60, 0xF8, 0x2A, 0x20, 0xAE, 0xF7}
@@ -117,9 +119,24 @@ pub(crate) unsafe fn get_file_dialog_path(
             .collect::<Vec<_>>()
     });
 
+    // Having Some in raw_spec means that we can safely unwrap options.allowed_types.
+    // Be careful when moving the unwraps out of this block or changing how raw_spec is made.
     if let Some(spec) = &raw_spec {
-        as_result(file_dialog.SetFileTypes(spec.len() as u32, spec.as_ptr()))?;
-        as_result(file_dialog.SetFileTypeIndex(1))?;
+        // We also want to make sure that options.allowed_types wasn't Some(zero-length vector)
+        if !spec.is_empty() {
+            as_result(file_dialog.SetFileTypes(spec.len() as u32, spec.as_ptr()))?;
+
+            let index = match (&options.default_type, &options.allowed_types) {
+                (Some(typ), Some(typs)) => typs.iter().position(|t| t == typ).unwrap_or(0),
+                _ => 0,
+            };
+
+            let default_allowed_type = options.allowed_types.as_ref().unwrap().get(index).unwrap();
+            let default_extension = default_allowed_type.extensions.first().unwrap_or(&"");
+
+            as_result(file_dialog.SetFileTypeIndex((index + 1).try_into().unwrap()))?;
+            as_result(file_dialog.SetDefaultExtension(default_extension.to_wide().as_ptr()))?;
+        }
     }
 
     as_result(file_dialog.SetOptions(flags))?;
--- a/druid-shell/src/dialog.rs
+++ b/druid-shell/src/dialog.rs
@@ -35,6 +35,7 @@ pub enum FileDialogType {
 pub struct FileDialogOptions {
     pub show_hidden: bool,
     pub allowed_types: Option<Vec<FileSpec>>,
+    pub default_type: Option<FileSpec>,
     // we don't want a library user to be able to construct this type directly
     __non_exhaustive: (),
     // multi selection
@@ -49,7 +50,7 @@ pub struct FileDialogOptions {
 /// struct.
 ///
 /// [`COMDLG_FILTERSPEC`]: https://docs.microsoft.com/en-ca/windows/win32/api/shtypes/ns-shtypes-comdlg_filterspec
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq)]
 pub struct FileSpec {
     /// A human readable name, describing this filetype.
     ///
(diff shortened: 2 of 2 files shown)


druid-shell/src/platform/windows/dialog.rs
```diff
@@ -119,7 +121,27 @@ pub(crate) unsafe fn get_file_dialog_path(
...
+        let index = match &options.default_type {
+            Some(default_type) => options
+                .allowed_types
+                .as_ref()
+                .unwrap()
+                .iter()
+                .position(|allowed_type| allowed_type == default_type)
+                .or(Some(0))
+                .unwrap(),
+            None => 0,
```
cmyr: I think this should be `1`, and the position above should be incremented by 1? I might be wrong, i'm just basing this off the [`IFileDialog::SetFileTypeIndex` docs](https://docs.microsoft.com/en-us/previous-versions/windows/desktop/legacy/bb775978(v%3Dvs.85)#parameters).

druid-shell/src/platform/windows/dialog.rs
```diff
@@ -119,7 +121,27 @@ pub(crate) unsafe fn get_file_dialog_path(
...
+        };
+        let default_allowed_type = options.allowed_types.as_ref().unwrap().get(index).unwrap();
+        let default_extension = default_allowed_type
+            .extensions
+            .first()
+            .or(Some(&""))
```
cmyr: I would write this as `unwrap_or("")`.

druid-shell/src/platform/windows/dialog.rs
```diff
@@ -119,7 +121,27 @@ pub(crate) unsafe fn get_file_dialog_path(
 
     if let Some(spec) = &raw_spec {
         as_result(file_dialog.SetFileTypes(spec.len() as u32, spec.as_ptr()))?;
-        as_result(file_dialog.SetFileTypeIndex(1))?;
+
+        let index = match &options.default_type {
+            Some(default_type) => options
+                .allowed_types
+                .as_ref()
+                .unwrap()
```
cmyr: this will panic if `allowed_types` is None, which is possible I think?

druid-shell/src/platform/windows/dialog.rs
```diff
@@ -119,7 +121,27 @@ pub(crate) unsafe fn get_file_dialog_path(
...
+                .iter()
+                .position(|allowed_type| allowed_type == default_type)
+                .or(Some(0))
```
cmyr: `unwrap_or(0)` works here I think.

druid-shell/src/platform/windows/dialog.rs
```diff
@@ -119,7 +121,27 @@ pub(crate) unsafe fn get_file_dialog_path(
 
     if let Some(spec) = &raw_spec {
         as_result(file_dialog.SetFileTypes(spec.len() as u32, spec.as_ptr()))?;
-        as_result(file_dialog.SetFileTypeIndex(1))?;
+
+        let index = match &options.default_type {
+            Some(default_type) => options
```
cmyr: code golf: I might write this as,

```rust
let index = match (&options.default_type, &options.allow_types) {
    (Some(typ), Some(typs)) => typs.iter().position(|t| t == typ).unwrap_or(0),
    _ => 0,
};
```

druid-shell/src/platform/windows/dialog.rs
xStrom: `SetFileTypeIndex` wants 1-based indeed and it gets that +1 when it is called. The `index` variable otherwise refers to the Rust vector index of `options.allowed_types`. I think it makes sense to keep the index 0-based as that is what the druid code uses and then do the +1 when doing the external `SetFileTypeIndex` call. Especially because `index` is also used where it needs to be 0-based.

druid-shell/src/platform/windows/dialog.rs
xStrom: `allowed_types` is never `None` in this section. However it can be `Some(vec![])` for which I added an additional check. The previous code with the hardcoded `SetFileTypeIndex(1)` call also failed to account for the empty vector case.

This is an interesting issue in a more broad sense. Using `unwrap` in the current situation is fine and doing all the checks would result in a bunch of bloat that makes the code harder to read and maintain. However someone else might modify this function some years down the line and not fully realize the assumptions made here about `None` not being a possibility.

For now I added some comments to clarify the assumptions and care that should be taken with future edits.

druid-shell/src/platform/windows/dialog.rs
xStrom: I like it!

xStrom:
I made some changes and fixed a bug with `Some(vec![])`. Ready for another look!

cmyr:
I'm going to merge this now, if any other issues come up they can be follow-on work. :)

PR #388, opened 2019-12-12
Add a WINDOW_CREATED event command.

### LINKED ISSUE #103: Add 'Started' event
This is an event that is sent to widgets when the application first runs.

We likely want a fairly full suite of lifecycle events, but for the time being I think we should just add them as needed. This is an obvious candidate.

xStrom:
Having a reliable early event to bootstrap animation etc is useful. This PR adds a simple `WINDOW_CREATED` event which can be used for such a purpose.

Longer term it probably makes sense to migrate this into some sort of `Event::Lifecycle` collection instead. However I think there's significant value in having a stopgap event in the meanwhile, which this PR brings.

I think this fixes #103 for now.

### DIFF
--- a/druid/src/command.rs
+++ b/druid/src/command.rs
@@ -48,6 +48,11 @@ pub struct Command {
 pub mod sys {
     use super::Selector;
 
+    /// Druid sends this when a window is created, e.g. at startup.
+    ///
+    /// Note that this will likely get refactored into a full lifecycle event suite later.
+    pub const WINDOW_CREATED: Selector = Selector::new("druid-builtin.window-created");
+
     /// Quit the running application. This command is handled by the druid library.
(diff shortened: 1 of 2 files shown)


cmyr:
This motivated me to finally go and do a first pass of adding lifecycle events, in #389. I think that will cover your needs? Are there any other events here you think would be helpful immediately?

xStrom:
Yes #389 will cover my existing use case.

I could also immediately use some events regarding a Widget being attached/detached from the widget tree. However I'm not sure that's a simple addition and will likely require cooperation from the library user. It's also easy enough for me to handle myself with project specific `Event::Command` right now.

cmyr:
Yea, that is a more difficult bit of work, but it's definitely planned; it's on the list of things raph and I are going to look at in the next couple of weeks.

PR #389, opened 2019-12-12
Add basic life cycle events

cmyr:
This adds Event::LifeCycle, and the LifeCycle enum, which for now
has two variants: ApplicationStarted and WindowConnected.

The first of these is sent only once, when the application starts;
the second is sent to a window each time a window is added.

ApplicationLaunched is kind of fiddly; if we want to send that message
to a window, we need to track which windows exist at launch time,
and then send it to each of those windows when they actually connect.

It may be that we only really want to send this message to the top-level
application delegate, which would require tracking less state and
be generally cleaner.

### DIFF
--- a/druid/src/event.rs
+++ b/druid/src/event.rs
@@ -46,6 +46,12 @@ use crate::Command;
 /// [`WidgetPod`]: struct.WidgetPod.html
 #[derive(Debug, Clone)]
 pub enum Event {
+    /// Called when certain application and window lifecycle events occur.
+    ///
+    /// For the various possible events, see the [`LifeCycle`] enum.
+    ///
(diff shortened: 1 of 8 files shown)


druid/src/event.rs
```diff
@@ -131,6 +137,17 @@ pub enum Event {
     Command(Command),
 }
 
+/// Application life cycle events.
+#[derive(Debug, Clone, Copy)]
+pub enum LifeCycle {
+    /// Sent to the root widget of a window when that window is first instantiated.
```
raphlinus: This comment doesn't seem to quite match the event propagation, I believe it's delivered to all widgets (which I think is more useful).

PR #391, opened 2019-12-14
Support multiline text in Label

zaynetro:


zaynetro:
This support will be implemented in piet: https://github.com/linebender/piet/pull/103

PR #396, opened 2019-12-16
Propagate mouse down event to hot widgets.

xStrom:
### Summary

This PR removes a check that prevented mouse down events propagating to hot widgets if some other widget was active.

### Event flow change

Consider a scenario where we have a window with two buttons A & B.

* Click on button A and its MouseDown handler sets the button as active and opens a modal dialog.
* The MouseUp event gets consumed by the new modal dialog and never reaches druid.
* When the modal is finally closed, click on button B. Button B never gets the event because button A is still active. Button A gets the MouseDown event instead and sets itself active once again, plus it opens the modal dialog again which prevents the MouseUp event from being sent.

After clicking button A, whatever you click later will act as a click on button A.

A workaround would be to never do anything in a MouseDown handler that could impact OS event flow. That seems like a big ask though and when you do, it's not clear what's going on, as I was confused by this during my testing. 

It's not just about what the druid app is doing either. There may be other apps running on the computer that suddenly steal focus, just after a user clicked the mouse but before they released it. It would end up in the same scenario.

Thus to make druid more robust, this PR propagates the MouseDown event to hot widgets regardless of whether any other widget is active.

### Widget improvements

I also improved the widgets to work better under circumstances where every MouseDown event doesn't have a buddy MouseUp event.

The button widget now realizes it's no longer active if a MouseDown event happens when the button isn't hot. This allows it to restore it's non-active state sooner.

Similarly the switch widget now animates to its final state already during a MouseDown elsewhere, without waiting for a MouseUp, which may never come.

### DIFF
--- a/druid/src/lib.rs
+++ b/druid/src/lib.rs
@@ -569,7 +569,9 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
                     child_ctx.base_state.is_hot = true;
                     hot_changed = Some(true);
                 }
-                recurse = had_active || !ctx.had_active && now_hot;
+                // Always recurse when hot to prevent active widgets from dominating.
+                // See: https://github.com/xi-editor/druid/pull/396
+                recurse = had_active || now_hot;
                 let mut mouse_event = mouse_event.clone();
                 mouse_event.pos -= rect.origin().to_vec2();
                 Event::MouseDown(mouse_event)
--- a/druid/src/widget/button.rs
+++ b/druid/src/widget/button.rs
@@ -80,8 +80,10 @@ impl<T: Data> Widget<T> for Button<T> {
     fn event(&mut self, ctx: &mut EventCtx, event: &Event, data: &mut T, env: &Env) {
         match event {
             Event::MouseDown(_) => {
-                ctx.set_active(true);
-                ctx.invalidate();
+                if ctx.is_active() != ctx.is_hot() {
+                    ctx.set_active(ctx.is_hot());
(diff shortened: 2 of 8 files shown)


cmyr:
Thanks, will get to this later this evening or tomorrow!

cmyr:
Haven't read the code yet, but reading your description of the problem makes me think of a general solution that you often need in situations like this, which is some sort of 'Cancel' or 'LostFocus' event. This might fit into the lifecycle model (#404) quite well, although the precise logic around dispatching that correctly could be tricky, I would hope it isn't _too_ tricky; any event that would cause a global focus change would need to check if any widgets currently have focus and then dispatch a focus change event.

cmyr:
What's the status of this?

xStrom:
This is on pause waiting for the alerts work. The reason is that [you said](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/druid.23394.20(mousedown.20fixes)) it would be a good idea to have a sample to play around with that shows the issue. That's completely reasonable but I can't provide that until the alerts work is more ready.

I may also adjust this solution somewhat based on some other issues that came up while implementing mouse capture. At least I would document the issue more broadly here. Shouldn't be too long until I can cycle back to this.

cmyr:
okay cool. If you want to talk about modals and alerts at some point let me know, I've done some thinking there

rjwittams:
It feels like this PR is old enough it should be closed and resubmitted if its worked on

cmyr:
that sounds reasonable.

PR #406, opened 2019-12-20
Fixup clippy lints for 1.40

cmyr:


druid/src/widget/list.rs
```diff
@@ -141,19 +141,20 @@ impl<C: Data, T: ListIter<C>> Widget<T> for List<C> {
...
-                if i < len {
-                    return;
-                }
-
-                let mut child = WidgetPod::new((self.closure)());
-                child.update(ctx, child_data, env);
-                self.children.push(child);
-            });
+        match self.children.len() {
+            len if len > data.data_len() => self.children.truncate(data.data_len()),
```
raphlinus: As I said in person, I think the clippy lint is wrong here, and I would push back against it. That said, this isn't *horrible.*

druid-shell/src/platform/windows/mod.rs
```diff
@@ -14,6 +14,8 @@
 
 //! Windows implementation of druid-shell.
 
+#![allow(clippy::missing_safety_doc)]
```
raphlinus: This was left in by mistake; it's clean without.

PR #407, opened 2019-12-21
Move code out of lib.rs

cmyr:
This cleans up lib.rs, moving code to `core.rs`, `box_constraints.rs`. I had intended to have `widget_pod.rs` but that logic is too tightly coupled to the various `Ctx`s, so it felt better to leave them together; otherwise I would've had to make more fields on `BasteState` be `pub(crate)`, which is counter to part of the motivation for this PR.

As a consequence of this work I had to also clean up imports in a bunch of widgets that were using types that just happened to have previously been visible from `lib.rs`.


cmyr:
yea I think that makes sense. At some point we might want a better high-level overview at the top of the docs.

PR #409, opened 2019-12-21
Fix command::CLOSE_WINDOW not triggering `Delegate::window_removed`

edwin0cheng:
This PR try to fix a bug which make `command::CLOSE_WINDOW` not triggering `Delegate::window_removed` in Windows platform.

The root cause of this bug :

A reference of `WindowState` will be hold when we call `DestroyWindow` in our main loop. It is because by design WM_DESTROY message [will not passed through message queue][1] such that `borrow` of WindowState for calling back of handler will always fail.

The fix:
We use an custom message to let message loop itself to call `DestroyWindow` directly.

[1]: https://stackoverflow.com/a/18828222/1378155

### DIFF
--- a/druid-shell/src/platform/windows/window.rs
+++ b/druid-shell/src/platform/windows/window.rs
@@ -192,18 +192,16 @@ struct DCompState {
 /// Message indicating there are idle tasks to run.
 const XI_RUN_IDLE: UINT = WM_USER;
 
-/// Message indicating explit destroy windows
+/// Message relaying a request to destroy the window
 ///
-/// A referene of `WindowState` will be hold when we
-/// call `DestroyWindow` in our main loop. It is because
-/// by design WM_DESTROY messsage will not passed through
-/// message queue such that `borrow` of WindowState for
-/// calling back of handler will always fail.
+/// Calling `DestroyWindow` from inside the handler is problematic
+/// because it will recursively cause a `WM_DESTROY` message to be
+/// sent to the window procedure, even while the handler is borrowed.
+/// Thus, the message is dropped and the handler doesn't run.
 ///
-/// We use an custom message to let message loop itself to
-/// call `DestroyWindow`
-///
-/// see also: https://stackoverflow.com/a/18828222/1378155
+/// As a solution, instead of immediately calling `DestroyWindow`, we
+/// send this message to request destroying the window, so that at the
+/// time it is handled, we can successfully borrow the handler.
 const XI_REQUEST_DESTROY: UINT = WM_USER + 1;
 
 impl Default for PresentStrategy {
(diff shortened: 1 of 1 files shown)


raphlinus:
I am able to reproduce the issue but am not yet convinced that the analysis is correct - I believe our window procedure is getting the `WM_DESTROY` message but it is being dropped because the window state is already borrowed. Are you seeing a "dropped message 0x2" in the logging?

It's entirely possible that the patch is correct, that we should defer destroying the window, but I do want to understand it a bit more clearly before merging a fix.

edwin0cheng:
> I am able to reproduce the issue but am not yet convinced that the analysis is correct - I believe our window procedure is getting the WM_DESTROY message but it is being dropped because the window state is already borrowed. Are you seeing a "dropped message 0x2" in the logging?

Oh, sorry, I didn't explain it well. Let me try to provide more details:

Our window procedure do get the `WM_DESTROY` message and it is being dropped because the window state is already borrowed. But the reason behind that is, when `DestroyWindow` is called, `WM_DESTROY` will not show up in `PeekMessage`, but Windows kernel will call the `WindProc` **directly**. 

You can observe this behavior by inserting debug log to `fn close()`:

```rust
pub fn close(&self) {
    if let Some(w) = self.state.upgrade() {
        let hwnd = w.hwnd.get();
        unsafe {
            dbg!("before DestroyWindow");   <------ A
            DestroyWindow(hwnd);
            dbg!("after DestroyWindow");  <------- B
        }
    }
}
```

When a window is closed,  the logging will be shown as:
```
...
before DestroyWindow
dropped message 0x2
after DestroyWindow
...
```

druid-shell/src/platform/windows/window.rs
```diff
@@ -192,6 +192,20 @@ struct DCompState {
 /// Message indicating there are idle tasks to run.
 const XI_RUN_IDLE: UINT = WM_USER;
 
+/// Message indicating explit destroy windows
```
raphlinus: The change is good but I'd like to rewrite the comment to explain more clearly what's happening. I think the issue is related to the stackoverflow thread, but different - we're not relying on the message being available in `PeekMessage`, this is really more about reentrancy.

I wanted to make this as a [suggestion](https://github.com/isaacs/github/issues/1452) but apparently that only supports a single line.

```
/// Message relaying a request to destroy the window
///
/// Calling `DestroyWindow` from inside the handler is problematic
/// because it will recursively cause a `WM_DESTROY` message to be
/// sent to the window procedure, even while the handler is borrowed.
/// Thus, the message is dropped and the handler doesn't run.
///
/// As a solution, instead of immediately calling `DestroyWindow`, we
/// send this message to request destroying the window, so that at the
/// time it is handled, we can successfully borrow the handler.
```

edwin0cheng: Fixed :)

PR #411, opened 2019-12-22
Implement basic traits for derived_lens

edwin0cheng:
This PR try to add basic traits (Debug, Copy, Clone) for proc-derived field lens as other `LensEx` helpers are also implemented.  (e.g. `Then`, and `Id`)


cmyr:
thanks!

PR #412, opened 2019-12-22
Add druid::prelude

cmyr:
This is mostly a glob re-export of `kurbo` and `piet` types.

I think it's a pretty big papercut that you need to A) know that these types are in a weird place and B) import them  all individually. I don't love preludes, but this feels like an improvement? I guess another option would be to export them all into the base namespace, but that would get cluttered quickly. Another option might be to export them into a number of submodules like `geometry`, `text`, `paint` etcetera, but that also feels too clever.

Very open to thoughts from others on this, doesn't feel like there's an ideal solution.

### DIFF
--- a/druid/src/lib.rs
+++ b/druid/src/lib.rs
@@ -62,3 +62,21 @@ pub use mouse::MouseEvent;
 pub use widget::Widget;
 pub use win_handler::DruidHandler;
 pub use window::{Window, WindowId};
+
+/// The druid prelude imports geometric types from [`kurbo`], drawing types from [`piet`],
+/// deriveable traits ([`Data`] & [`Lens`]), and the extension traits
+/// ([`WidgetExt`] and [`LensExt`]).
+///
+/// [`kurbo`]: https://crates.io/crates/kurbo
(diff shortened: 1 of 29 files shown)


druid/src/lib.rs
```diff
@@ -62,3 +62,21 @@ pub use mouse::MouseEvent;
...
+///
+/// [`kurbo`]: https://crates.io/crates/kurbo
+/// [`piet`]: https://crates.io/crates/piet
+/// [`Data`]: trait.Data.html
+/// [`Lens`]: lens/trait.Lens.html
+/// [`LensExt`]: trait.LensExt.html
+/// [`WidgetExt`]: trait.WidgetExt.html
+pub mod prelude {
+    pub use crate::data::Data;
+    pub use crate::kurbo::*;
```
raphlinus: I'm going to suggest this is: `{Affine, Circle, Insets, Line, Point, Rect, RoundedRect, Size, Vec2}`.

cmyr:
After discussion I'm no longer convinced a prelude makes sense; it feels better to just reexport the types that are exposed in public API at the top level of the crate.

PR #419, opened 2019-12-27
Add menu hotkey support in Windows

edwin0cheng:


cmyr:
Sorry for letting this sit, I won't be able to test windows code until after january 8th.

edwin0cheng:
Thanks for the update and don't worry about it. 👍

cmyr:
@edwin0cheng are you still interested in this patch? I'd be happy to get it merged, just a few lingering questions.

edwin0cheng:
@cmyr Sorry for the late reply, Please don't merge it now. I am a little busy recently and would like to fix your comments in this week.

druid-shell/src/platform/windows/accels.rs
```diff
@@ -0,0 +1,58 @@
+use winapi::ctypes::c_int;
```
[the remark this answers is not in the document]
edwin0cheng: Sure, Added.

druid-shell/src/platform/windows/menu.rs
```diff
@@ -108,4 +125,38 @@ impl Menu {
...
+            return None;
+        }
+        Some(self.accels.values().cloned().collect())
+    }
+}
+
+fn convert_hotkey(id: u32, key: &HotKey) -> Option<ACCEL> {
+    let code = match key.key {
+        KeyCompare::Code(code) => code,
+        _ => return None,
```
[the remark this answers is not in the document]
edwin0cheng: Yes, there is an api for that https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-vkkeyscana. I added this api to handle the `KeyCompare::Text` case.

edwin0cheng:
> This is still an option, but it always felt a little awkward to me. I'm not sure what a good alternative is; a simple one might be to have separate localized strings for windows menu titles, and only do this trick there?

I think a separate localized strings is a reasonable choice here, But I would like to separate to another PR.

edwin0cheng:
I just fixed some issues mentions in comments, please have a look again 👍

druid-shell/src/platform/windows/menu.rs
```diff
@@ -108,4 +125,51 @@ impl Menu {
...
+    }
+    if key_mods.shift {
+        virt_key |= FSHIFT;
+    }
+
+    let raw_key = match key.key {
+        KeyCompare::Code(code) => code.to_i32()?,
+        KeyCompare::Text(text) => {
+            // See https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-vkkeyscana
+            let code = unsafe { VkKeyScanA(text.chars().next().unwrap() as i8) };
```
cmyr: I'm really not sure about this.

It might be fine. But it certainly could be incorrect; the cast to `i8` will be lossy for many possible `char`s. 

What values do we expect  to find here? Does this text have to be ascii? 

I think it's probably worth being explicit and defensive, unless we have a better idea what's going on; so I would A) not unwrap, and B) verify that the char is ascii before calling `VkKeyScanA`. I might also consider logging an error if it isn't?

druid-shell/src/platform/windows/accels.rs
```diff
@@ -0,0 +1,74 @@
...
+use std::sync::{Arc, Mutex};
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+struct WindowHandle(HWND);
+unsafe impl Send for WindowHandle {}
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+struct AccelHandle(HACCEL);
+unsafe impl Send for AccelHandle {}
+unsafe impl Sync for AccelHandle {}
```
cmyr: Do we know that these pointers are thread safe? Can we include a link to some reference or something?

druid-shell/src/platform/windows/menu.rs
edwin0cheng: Agree, and actually I should use should use `VkKeyScanW` and `encode_utf8` instead too :)

druid-shell/src/platform/windows/accels.rs
edwin0cheng: Yeah, if I understand correctly, all handles you obtain from functions in Kernel32 are thread-safe, unless the MSDN Library article for the function explicitly mentions it is not.

See https://docs.microsoft.com/en-us/windows/win32/wsw/thread-safety

And I added a comment for that.

Issue #311, opened 2019-11-15
"improve" error handling in druid-shell


cmyr:
Error handling in druid shell is currently bad. We would like it to be good! This is a solvable problem!

### Background

druid-shell started out as a windows-specific proof of concept. The existing error handling is left-over from that period. As druid-shell was adapted to be multi-backend, error handling was ignored, and now it's basically useless.

### how it should work

Each platform should expose a platform-specific error type that mimics as closely as possible the  system error type on that platform; so on macOS it should look like [`NSError`][], on windows it should be the system error code from [`GetLastError`][] and the message returned by [`FormatMessage`][].

druid-shell's `error` module should have an `Error` enum that includes a `Platform(_)` variant for the platform specific error, and then normal variants for other conditions related to invalid inputs, io failure, etcetera.

We should return meaningful errors whenever possible.

[`GetLastError`]: https://docs.microsoft.com/en-us/windows/win32/api/errhandlingapi/nf-errhandlingapi-getlasterror
[`FormatMessage`]: https://docs.microsoft.com/en-us/windows/desktop/api/WinBase/nf-winbase-formatmessage
[`NSError`]: https://developer.apple.com/documentation/foundation/nserror?language=objc

PR #422, opened 2019-12-28
Expose PlatformError type

mendelt:
I added shell::Error to the types that are re-exported from lib.rs. I think this is sensible because this is the error type that is returned from AppLauncher::launch which is also exported from lib.rs. This makes handling errors in a fn main easier for druid apps.
`Error` is not a really descriptive name so I used `PlatformError` to be consitent with its use in `AppLauncher`
I updated the hello.rs example to show how this might be used for easier error handling.

### DIFF
--- a/druid/examples/hello.rs
+++ b/druid/examples/hello.rs
@@ -13,15 +13,16 @@
 // limitations under the License.
 
 use druid::widget::{Align, Button, Flex, Label, Padding};
-use druid::{AppLauncher, LocalizedString, Widget, WindowDesc};
+use druid::{AppLauncher, LocalizedString, PlatformError, Widget, WindowDesc};
 
-fn main() {
+fn main() -> Result<(), PlatformError> {
     let main_window = WindowDesc::new(ui_builder);
     let data = 0_u32;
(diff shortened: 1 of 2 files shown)


mendelt:
Thank you for looking at this. I agree that this is a bandaid and that should be replaced by Druids own error type. This was meant as a quick fix.
I could make this a bit more comprehensive and create a DruidError newtype around the PlatformError. Or even a DruidError enum with one value wrapping the PlatformError? LaunchError does not really sound right to me. Also because launch is not the best method name. I guess any unhandled error in a druid application will eventually end up here.
I can add an issue about error handling. Maybe we can get some discussion going.

cmyr:
I was thinking that `LaunchError` would be a specific error type that would only be used in this situation; it would not be recoverable etc. A general `Error` enum will cover many cases that _are_ recoverable. 

It's also possible  that we want to behave more like `std`, where we have specific error types for specific failure cases.

So: I think that this _is_ the best bandaid, and the longer-term solution needs some consideration. There is #311, but that is about druid-shell specifically, and may be partially out of date. If you wanted to open a discussion issue that would be welcome,  but no pressure!

PR #423, opened 2019-12-28
Expose types from kurbo & piet at top level

cmyr:
This feels like a cleaner and simpler alternative to having a prelude.

I've limited this to only those types that are used by public API.

### DIFF
--- a/druid/examples/anim.rs
+++ b/druid/examples/anim.rs
@@ -16,11 +16,10 @@
 
 use std::f64::consts::PI;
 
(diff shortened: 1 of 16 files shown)


cmyr:
ugh actually I should update other code to use these probably 😒

cmyr:
This updates some imports to use the newly reexported types; it does this for all of the examples, but only some places elsewhere, because I'm lazy and it doesn't feel super important. 🤷‍♀

PR #425, opened 2019-12-28
Update readme for 0.4.0

cmyr:


README.md
```diff
@@ -127,7 +126,7 @@ includes the current theme parameters (colors, dimensions, etc.).
 
 ```rust
 impl<T: Data> Widget<T> for Button<T> {
-    fn paint(&mut self, paint_ctx: &mut PaintCtx, base_state: &BaseState, data: &T, env: &Env) {
+    fn paint(&mut self, paint_ctx: &mut PaintCtx, data: &T, env: &Env) {
```
ratmice: Just below here, It looks like I missed switching the order of `event: &Event` and `ctx: &mut EventCtx` in pr #317

PR #435, opened 2019-12-31
Pass app state to the EnvScope closure

### LINKED ISSUE #415: EnvScope doesn't have access to the application state
The closure that gets passed to `EnvScope::new()` only has access to the application's `Env`. It has no way to access the application's state, which means it's not possible to style widgets based on app state.

For context, I'm building a todo list application in which each todo item is represented by a `CheckBox` and a `Label`. When I click a `CheckBox` to mark a todo as completed, I want the `CheckBox` and `Label` to be rendered in a lighter gray. I can't accomplish this without having access to the application state inside the `EnvWrap` that wraps the `Label` and `CheckBox`.

s3thi:
This fixes #415.

### DIFF
--- a/druid/examples/textbox.rs
+++ b/druid/examples/textbox.rs
@@ -35,7 +35,7 @@ fn main() {
 fn build_widget() -> impl Widget<String> {
     let textbox = TextBox::new();
     let textbox_2 = EnvScope::new(
-        |env| {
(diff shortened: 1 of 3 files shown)


s3thi:
Looks like I missed updating some tests and docs on this PR. Fixing those and updating this.

PR #440, opened 2020-01-01
Fix SVG widget for 0.4.0

crsaracco:
The SVG feature broke in 0.4.0 because some of the trait method signatures changed.

I'm not familiar enough with github workflows to edit them, but IMO they should be updated to test against different feature flags as well.

### DIFF
--- a/druid/src/widget/svg.rs
+++ b/druid/src/widget/svg.rs
@@ -83,7 +83,7 @@ impl<T: Data> Widget<T> for Svg<T> {
             bc.constrain(self.get_size())
         }
     }
-    fn paint(&mut self, paint_ctx: &mut PaintCtx, data: &T, env: &Env) {
+    fn paint(&mut self, paint_ctx: &mut PaintCtx, _data: &T, _env: &Env) {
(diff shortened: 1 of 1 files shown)


druid/src/widget/svg.rs
```diff
@@ -83,12 +83,12 @@ impl<T: Data> Widget<T> for Svg<T> {
             bc.constrain(self.get_size())
         }
     }
-    fn paint(&mut self, paint_ctx: &mut PaintCtx, base_state: &BaseState, _data: &T, _env: &Env) {
+    fn paint(&mut self, paint_ctx: &mut PaintCtx, data: &T, env: &Env) {
```
raphlinus: I think you still want the underscores here to prevent warnings, as the method doesn't use either data or env.

crsaracco: Oh weird, I didn't get warnings for those. Fixed.

raphlinus: Well, not surprising we didn't get warnings in CI, anyway :). Thanks!

PR #441, opened 2020-01-01
simplify container and add corner radius

mendelt:
I added an option for rounded corners to the container. And a panels example that I used myself for testing this out. Not sure if that should be included in the project. I can take it out if needed.

### DIFF
--- a/druid/examples/panels.rs
+++ b/druid/examples/panels.rs
@@ -1,0 +1,71 @@
+// Copyright 2019 The xi-editor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! This example shows how to construct a basic layout.
+
+use druid::widget::{Flex, Label, WidgetExt};
+use druid::{AppLauncher, Color, LinearGradient, PlatformError, UnitPoint, Widget, WindowDesc};
+
+fn build_app() -> impl Widget<()> {
+    let solid = Color::rgb8(0x3a, 0x3a, 0x3a);
+    let gradient = LinearGradient::new(
+        UnitPoint::TOP_LEFT,
+        UnitPoint::BOTTOM_RIGHT,
+        (Color::rgb8(0x3a, 0x3a, 0x3a), Color::rgb8(0x31, 0x31, 0x31)),
+    );
+
+    Flex::column()
+        .with_child(
+            Flex::row()
+                .with_child(
+                    Label::new("top left")
+                        .border(gradient.clone(), 4.0)
+                        .padding(10.0),
+                    1.0,
+                )
+                .with_child(
+                    Label::new("bottom left")
+                        .background(solid.clone())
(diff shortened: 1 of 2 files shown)


druid/src/widget/container.rs
```diff
@@ -12,54 +12,59 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! A convenience widget that combines common styling and positioning widgets.
+//! Draws a border and a background around a widget or a widget tree.
```
cmyr: This leaves out that this widget can have rounded corners. I don't think the previous docstring is too bad in spirit, just in that it incorrectly describes the implementation (it describes a previous implementation). I might just go with "A widget that provides simple visual styling options to a child."? Or something like this?

druid/src/widget/container.rs
```diff
@@ -12,54 +12,59 @@
...
 
-#[derive(Default)]
-struct ContainerStyle {
+/// Draws a border and a background around a widget or a widget tree.
+pub struct Container<T: Data> {
     background: Option<PaintBrush>,
-    border: Option<BorderState>,
-}
+    border: Option<BorderStyle>,
+    corner_radius: f64,
```
cmyr: my inclination is that this should be `Option<f64>` and we should have different code paths (`Rect` vs `RoundedRect`) based on whether the option is set. But I don't have a good rationale for this, it just feels more 'correct' to me, and I'm willing to change my mind. I'd be curious for @raphlinus's opinion.

druid/examples/panels.rs
```diff
@@ -0,0 +1,71 @@
...
+
+use druid::widget::{Flex, Label, WidgetExt};
+use druid::{AppLauncher, Color, LinearGradient, PlatformError, UnitPoint, Widget, WindowDesc};
+
+fn build_app() -> impl Widget<()> {
+    let solid = Color::rgb8(0x3a, 0x3a, 0x3a);
+    let gradient = LinearGradient::new(
+        UnitPoint::TOP_LEFT,
+        UnitPoint::BOTTOM_RIGHT,
+        (Color::rgb8(0x3a, 0x3a, 0x3a), Color::rgb8(0x31, 0x31, 0x31)),
```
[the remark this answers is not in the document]
mendelt: Ah, stupid of me! I wanted to do that but forgot.

druid/src/widget/container.rs
mendelt: I originally went that direction but that made the `paint` function twice as long with mostly duplicate code. It might be worth it if `Rect` is way faster than `RoundedRect` though.

druid/examples/panels.rs
```diff
@@ -0,0 +1,71 @@
...
+        .with_child(
+            Flex::row()
+                .with_child(
+                    Label::new("top left")
+                        .border(gradient.clone(), 4.0)
+                        .padding(10.0),
+                    1.0,
+                )
+                .with_child(
+                    Label::new("bottom left")
```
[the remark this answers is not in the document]
mendelt: Oops!

druid/src/widget/container.rs
mendelt: Yeah, the previous docstring looks is very similar to the Flutter container docstring. But the implementation is quite different. I like your suggestion. Short and to the point.
When things are more stable I would like to see much more extensive docstrings for all widgets to be honest. This is an important part of our public API so it should be well documented.

druid/src/widget/container.rs
raphlinus: One way to solve this is to have the implementation of `RoundedRect` in piet go to the rect case when the radius is 0. We don't *currently* have that implemented, but I don't see why not, so I'm fine with simpler code.

mendelt: Sounds great. I can even look at Piet this weekend to see if I can figure out how to do that.

PR #445, opened 2020-01-04
reimplement Button::sized in WidgetExt

mendelt:
The Button::sized constructor might be useful for other widgets too so I re implemented it as a WidgetExt method.
I had to add the T: 'sized lifetime constraint to widgetext to allow sized to be wrapped in a SizedBox<T> as well an Allign<T>. I don't think this will cause issues but I would like a second opinion on this.

### DIFF
--- a/druid/src/widget/widget_ext.rs
+++ b/druid/src/widget/widget_ext.rs
@@ -21,7 +21,7 @@ use super::{Align, Container, EnvScope, Padding, Parse, SizedBox};
 use crate::{Data, Env, Lens, LensWrap, Widget};
 
 /// A trait that provides extra methods for combining `Widget`s.
-pub trait WidgetExt<T: Data>: Widget<T> + Sized + 'static {
+pub trait WidgetExt<T: Data + 'static>: Widget<T> + Sized + 'static {
     /// Wrap this widget in a [`Padding`] widget with the given [`Insets`].
     ///
     /// [`Padding`]: struct.Padding.html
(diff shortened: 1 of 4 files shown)


druid/src/widget/widget_ext.rs
```diff
@@ -79,6 +79,17 @@ pub trait WidgetExt<T: Data>: Widget<T> + Sized + 'static {
...
 
+    /// Wrap a widget in an [`Align`] and a [`SizedBox`] to make it a fixed size
+    ///
+    /// [`Align`]: struct.Align.html
+    /// [`SizedBox`]: struct.SizedBox.html
+    fn sized(self, width: f64, height: f64) -> Align<T> {
+        SizedBox::new(self)
+            .width(width)
+            .height(height)
+            .align_vertical(UnitPoint::CENTER)
```
cmyr: I would prefer that we make the user explicitly call this. One of the goals of `WidgetExt` is to make it very easy to write out these sorts of compositions, and my current inclination is to try to avoid hidden composition where possible.

cmyr:
Also: this PR has motivated me to do #451, because I was getting confused by the method names. 😁

mendelt:
I was mostly trying to get sized out of button. I wanted to try out some similar ideas as https://github.com/xi-editor/druid/pull/446 and this was mostl.y meant as cleanup in preparation for that. I agree that sized in widget_ext clashes a bit with the fix_width and fix_height functionality so I would also be in favor of just removing it.

cmyr:
okay, why don't we just remove it for now?

mendelt:
I replaced sized with a fix_size that fits a bit better with the existing fix_width and fix_height but does both at the same time. This allows a user to do exactly the same thing as sized by combining fix_size and allign_vertical.

Issue #447, opened 2020-01-08
AnimFrame Event do not trigger update


edwin0cheng:
I found out when modifying `data` in  `AnimFrame`, the `Widget::update` will not run. 

For example, I have a widget which use `AnimFrame` event to poll some thread results, and then it will change a boolean flag in my data which is controlling a `Either` widget. But it do not work as expect and my `Either` widget do not switch.

I think the reason of that is :

https://github.com/xi-editor/druid/blob/cbc31d86fe0548be507116f543b0f2efa700bdf8/druid/src/win_handler.rs#L144-L165

If I understand correctly, It should call update **after** `AnimFrame` event but current implementation did not.

cmyr:
Yea, `AnimFrame` is strange. It is a goal to move `AnimFrame` out of `Event` and make it part of the new life cycle system; the `lifecycle` method on `Widget` will be similar to update, and you won't be able to mutate data; the only state that should be modified during animation should be state that is internal to the widget. If you need to do something fancier I would send a `Command`.

Also just to clarify: it should not be expected that `update` is called after *every* event, but only that it is called before each `layout` call. So you can receive multiple events before a single update event. (This isn't a problem here, but wanted to mention it because I don't think it's written down anywhere)

edwin0cheng:
Close this one , in flavor of #455

Issue #404, opened 2019-12-19
Split life cycle events into their own method


cmyr:
Over a few days of thinking about life cycle events, we have come around to the idea that there are a number of issues which make them a poor match for the current 'event' system.

## Summary

We propose adding a new widget method, with a signature that looks something like,

```rust
pub trait Widget<T> {
    // ...
    fn lifecycle(&mut self, ctx: &mut LifeCycleCtx, event: &LifeCycle, data: &T, env: &Env) {}
}
```
    
As a consequence of this, we will be able to change the signature of `update`, so that the type of the previous data goes from being `Option<T>` to being just `T`; the case where there is no data will be handled by a life cycle event.

In functionality, `lifecycle` is much closer to `update` than to `event`; it is principally an opportunity for the widget to do internal setup and to prepare for drawing.

## Rationale

There are a number of motivations for this change.

### More controlled mutation

The first of these issues is conceptual, and relates to when we should be mutating application state. It is currently only possible to mutate the app's data during  the `event` method on Widget. This is enforced by the type system; while the data is provided to the other widget methods (`update`, `layout`, and `paint`) it is only provided behind an immutable reference.
            
Are all events expected to mutate application state? We have come to the conclusion that no, they are not: certain events are intended as hooks for widgets to update internal state or to prepare for drawing. These include some events that are currently part of the `Event` enum, such as `AnimFrame`, `HotChanged`, and `FocusChanged`; and it also includes most of the events we've been thinking about as 'lifeycle' events, such as the current `WindowConnected` event, as well as things like `WidgetCreated`, `WidgetHidden`, and `WidgetShown`.

### Life cycle events can occur in a subset of the tree

With normal events, propogation begins at the root of the tree and then flows down the branches. This is not necessarily the case for certain life cycle events. Some examples are `WidgetCreated`, which is sent when a widget is added to the tree, as well as `WidgetHidden` which is sent when a widget is made inactive but not destroyed, such as when it is the inactive branch of an `Either` widget. In both cases, the event should start at some non-root node in the widget graph, and propagate from there.

### Life cycle events can occur outside normal event flow

[...]

PR #448, opened 2020-01-08
Call update after AnimFrame event

edwin0cheng:
This PR try to add back missing update call after `AnimFrame` Event fired.

See: #447

### DIFF
--- a/druid/src/win_handler.rs
+++ b/druid/src/win_handler.rs
@@ -155,12 +155,24 @@ impl<'a, T: Data + 'static> SingleWindowState<'a, T> {
             0
         };
         let anim_frame_event = Event::AnimFrame(interval);
-        let (_, _, request_anim) = self.do_event_inner(anim_frame_event, ctx);
+        let (_, event_needs_inval, request_anim) = self.do_event_inner(anim_frame_event, ctx);
         let prev = if request_anim {
             Some(this_paint_time)
         } else {
             None
         };
+
+        let mut update_ctx = UpdateCtx {
+            text_factory: ctx.text_factory(),
+            window: &self.state.handle,
+            needs_inval: false,
+            window_id: self.window_id,
+        };
+        self.window.update(&mut update_ctx, self.data, self.env);
+        if update_ctx.needs_inval || event_needs_inval {
+            update_ctx.window.invalidate();
+        }
+
         self.state.prev_paint_time = prev;
         request_anim
     }


raphlinus:
I agree with @cmyr's analysis here. One thing is that data changes in response to an animation frame *might* need to propagate to other windows, and I believe that the patch here only does the update for the window receiving the animation frame event. I am hopeful that in our lifecycle cleanup we'll pay attention to this and be able to provide a guarantee that update is called when needed.

edwin0cheng:
> I think this is a bandaid, and probably not worth it: in the next week or two I would like to do the life cycle work (#404) and that will break this. If you need to update the Data and receive update calls, I would suggest using EventCtx::submit_command for now, which will trigger a new event cycle.

@cmyr It sound reasonable. However, I did try to use `EventCtx::submit_command` before, but without success. It is because in `do_anim_frame`, it do not call `process_commands` such that no command submitted inside `do_event_inner` will be processed.

cmyr:
ugh yea, you're right.

This sort of highlights the problem: `AnimFrame` is sent _after_ `Update`, so handling further events here would break our invariant that you will always receive an `update` call between when `Data` changes and when layout/paint happen. This would be a problem if some other widget had setup some drawing state in `update` based on the `Data`, and then the `Data` gets blown away afterwards, when someone handles `AnimFrame`.

The best way to handle this right now would be to schedule an idle call with the runloop if there are commands to process after painting is finished, that would trigger a new event/update pass.

This is something we will need to do anyway, because `lifecycle` will work like `update` and you will be able to submit commands there, but we won't run them until after this paint cycle.

If you're interested in implementing this, you would be welcome to; it would work like `timer` does, where you request a callback from the `WindowHandle` or `WinCtx`, and then I would probably add a `fn idle(&mut self)` method to `druid-shell::WinHandler`, with an implementation in `druid::win_handler` that just processes commands?

Also the mechanism for scheduling idle tasks exists, but is a bit confusing: basically you call `WindowHandle::get_idle_handle()`, which returns an `Option<IdleHandle>`; then you call `add_idle` on that, passing a closure that will be called with an `&dyn Any` that is actually your `WindowHandler` type; you can downcast that to the correct concrete type and do stuff with it.

I'm not totally sure what the rationale for this approach is; the basic idea is that you can do arbitrary things with your `WindowHandler` in that closure, like create and dispatch events, but it's definitely an awkward API.

cmyr:
going to close this, I'll hopefully have this moved to lifecycle by the end of the week.

PR #449, opened 2020-01-09
Add widget identity

raphlinus:
This patch should be considered work in progress, as the code is written
but it hasn't been tested in any way.

### DIFF
--- a/druid/src/core.rs
+++ b/druid/src/core.rs
@@ -23,8 +23,8 @@ use log;
 use crate::kurbo::{Affine, Rect, Shape, Size};
 use crate::piet::{Piet, RenderContext};
 use crate::{
-    BoxConstraints, Command, Cursor, Data, Env, Event, Text, TimerToken, Widget, WinCtx,
-    WindowHandle, WindowId,
+    BoxConstraints, Command, Cursor, Data, Env, Event, IntoOptTarget, Target, Text, TimerToken,
+    Widget, WidgetId, WinCtx, WindowHandle, WindowId,
 };
 
 /// Convenience type for dynamic boxed widget.
@@ -47,6 +47,7 @@ pub struct WidgetPod<T: Data, W: Widget<T>> {
     old_data: Option<T>,
     env: Option<Env>,
     inner: W,
+    id: WidgetId,
 }
 
 /// Generic state for all widgets in the hierarchy.
@@ -103,11 +104,13 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
     /// so it can participate in layout and event flow. The process of
     /// adding a child widget to a container should call this method.
     pub fn new(inner: W) -> WidgetPod<T, W> {
+        let id = inner.get_id().unwrap_or_else(WidgetId::next);
         WidgetPod {
             state: Default::default(),
             old_data: None,
             env: None,
             inner,
+            id,
         }
     }
 
@@ -134,6 +137,10 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
     /// Return a mutable reference to the inner widget.
     pub fn widget_mut(&mut self) -> &mut W {
         &mut self.inner
+    }
+    /// Get the identity of the widget.
+    pub fn id(&self) -> WidgetId {
+        self.id
     }
 
     /// Set layout rectangle.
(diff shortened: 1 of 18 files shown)


druid/src/widget_id.rs
```diff
@@ -0,0 +1,130 @@
...
+    ///
+    /// Do note that if we create 4 billion widgets there may be a collision.
+    pub(crate) fn next() -> WidgetId {
+        let id = WIDGET_ID_COUNTER.fetch_add(1, Ordering::Relaxed);
+        WidgetId(id)
+    }
+}
+
+/// A wrapper that adds an identity to an otherwise anonymous widget.
+pub struct IdentityWrapper<W> {
```
cmyr: Refresh my memory: is the sole purpose of this widget so that we can get access to the `WidgetId` for other purposes? (Say, stashing it in some other widget that is a companion?)

Reading through this I'm noticing a potential gotcha: this will only work if the `IdentityWrapper` itself has a pod, so for instance if you do like,

```rust
let (id, widget) = IdentityWrapper::wrap(TextBox::new());
row.add_child(widget.padding(5.0));
```

You will get assigned a new id?

druid/src/core.rs
```diff
@@ -338,6 +345,14 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
                 Event::Timer(*id)
             }
             Event::Command(cmd) => Event::Command(cmd.clone()),
+            Event::TargetedCommand(target, cmd) => {
+                if *target == Target::Widget(self.id) || target.is_window() {
+                    Event::Command(cmd.clone())
+                } else {
+                    // TODO: here's where Bloom filter magic would happen
```
cmyr: what is the mechanism by which children register their ids with parents? I could imagine this potentially being part of `LifeCycle`?

cmyr:
Oh, one other question: I was thinking it might make sense to include command propagation code in `IdentityWrapper`? i.e. that widget will turn `TargetedCommand` into `Command`, if the ids match?

druid/src/widget_id.rs
raphlinus: The `Padding` widget gets a new id, yes. If you have a wrapper widget that *doesn't* allocate a new WidgetPod (like `EnvScope` and likely lens stuff), then it needs to forward the `get_id` request. That is maybe an argument for raising it to public visibility.

druid/src/core.rs
raphlinus: Yes, there's got to be a call on `EventCtx` and `UpdateCtx` (contexts in which a widget can dynamically add children) to register the child.

raphlinus:
> Oh, one other question: I was thinking it might make sense to include command propagation code in `IdentityWrapper`? i.e. that widget will turn `TargetedCommand` into `Command`, if the ids match?

Interesting point. I was thinking all widgets have identity and can be targets, regardless of whether the id is exposed through the client API. An example would be an async image loading widget, which might query its id from its enclosing pod (through a context). So this wouldn't need an `IdentityWrapper`.

But another approach is that identity is opt-in (this would improve the effectiveness of the Bloom filters, not counting widgets which would never be the targets). Requiring an `IdentityWrapper` might be a good way to indicate that. But it's a slightly different design than what I had in mind.

druid/src/widget_id.rs
cmyr: hmm, this makes me think we might want some sort of general purpose `child(&self) -> Option<&dyn Widget>` method on `Widget`; and if we do that it would be really nice if we had `as_any` so we could downcast to concrete widgets. 😬

raphlinus: I've experimented with that but have always run into problems of trying to be too fancy with the Rust type system, like having object safety problems. But if you can make it work, I agree it would be helpful!

cmyr:
gotcha. I _was_ imagining widget identity being opt-in; your async image loading widget would know its id, for instance; if it's in the default widget set it might not even use `IdentityWrapper` for that, and just generate it itself.

cmyr:
I've done a bit of cleanup and added some testing; in particular a large example that seems to be working, and a single 6 line unit test.

I think this is pretty close to ready, at least as a checkpoint.

Some todos:

- [ ] ability for a widget to get it's id from the contexts
- [ ] discuss naming of `get_id`; `get_` doesn't feel very rusty? Alternative would be `id()` or `widget_id()`.
- [ ] figure out some features that this enables; a simple one might be the ability to set focus on a widget (with some command that is defined in druid, and which we handle in `TextBox` et al.)

In general, though, my inclination is to switch over to doing the life cycle stuff; I think that the next step for this work will be clearer when that is done.

PR #451, opened 2020-01-09
Rename label::align -> label::text_alignment

cmyr:
'align' is ambiguous; it could refer to either the alignment of
the whole widget or the alignment of the text.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -280,7 +280,7 @@ dependencies = [
 
 [[package]]
 name = "druid"
-version = "0.4.0"
+version = "0.4.1"
 dependencies = [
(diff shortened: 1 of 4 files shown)


s3thi:
This is a very small nitpick (and comes down to personal preference), but would it make sense to call this `text_align` instead of `text_alignment`?

- It would be similar to the `align` calls elsewhere in the codebase
- It would use the imperative mood, which is more in line with other layout-related calls such as `fix_height` and `expand`.
- It would be similar to the `text-align` property from CSS (probably reaching here)

But `text_alignment` looks good, too :)

cmyr:
Agree that's better, thanks!

PR #452, opened 2020-01-10
[RFC] Widget::as_any + Widget::child (access to concrete children)

cmyr:
*this is a proof of concept, intended for discussion*

## Summary

This PR demonstrates one possible way to retrieve a particular concrete widget that may be several layers down in the tree, and in particular may be behind one or more `Box<dyn Widget>`s.

The basic idea is to add add two methods to `Widget`, that look like,

```rust
trait Widget {
    // ...
    fn as_any(&self) -> &dyn std::any::Any;
    fn child(&self) -> Option<&dyn Widget<T>> {
        None
    }
}
```

The first of these is, (annoyingly) required; all widgets will have to have an impl, and they will all look like,

```rust
fn as_any(&self) -> &dyn std::any::Any {
        *&self
    }
```

I don't think there is any getting around this; there's no way of having a default impl that doesn't either break object safety or break because the `TypeId` of `dyn Widget<T>` is not the same as the `TypeId` of the concrete type.

The second of these _does_ have a default impl, because not all widgets have a single child, which is the case that we're motivated by, here; widgets that _do_ are required to override this method to return their child.

## Motivation

- At various times I've wanted to be able to reach past multiple `Align`, `Padding`, or `Scroll` widgets to access some property on an inner widget. This might be an anti-pattern! But it has been something I've wanted to do. 😈 

- widget identity: this is what I was initially curious about; whether we could drill down into a stack of widgets and figure out if one of them had an identity. This wouldn't work with exactly this code, but would work similarly.

## ugly

This requires a new `'static` bound on all `Data`. We might be able to get away with this by adding that bound to `Data` itself (maybe by adding a `Same` trait for just the `same()` fn, and making `Data` a marker trait?)

To be properly useful we would also need to have like a `child_mut` fn; this isn't so so bad since we would only expect to need to impl these on library types, but it's still starting to get ugly;

... and it will get even more ugly when we inevitably decide we want to be able to recursively iterate _all children_, and we end up with like `iter_children` and `iter_children_mut`...

## the end

This is mostly interesting to me because it wasn't clear that we could make it work. It's interesting to know that we can. It might be something we will want to look at sometime, but I don't _think_ we'll need to do that now? But I thought it would be worth sharing to see what other folks thought.

### DIFF
--- a/druid/examples/anim.rs
+++ b/druid/examples/anim.rs
@@ -63,6 +63,10 @@ impl Widget<u32> for AnimWidget {
         let ambit = center + 45.0 * Vec2::from_angle((0.75 + self.t) * 2.0 * PI);
         paint_ctx.stroke(Line::new(center, ambit), &Color::WHITE, 1.0);
     }
+
+    fn as_any(&self) -> &dyn std::any::Any {
+        *&self
+    }
 }
 
 fn main() {
--- a/druid/examples/anim.rs
+++ b/druid/examples/anim.rs
@@ -65,7 +65,7 @@ impl Widget<u32> for AnimWidget {
     }
 
     fn as_any(&self) -> &dyn std::any::Any {
-        *&self
+        self
     }
 }
 
--- a/druid/examples/calc.rs
+++ b/druid/examples/calc.rs
@@ -112,7 +112,7 @@ impl CalcState {
     }
 }
 
-fn pad<T: Data>(inner: impl Widget<T> + 'static) -> impl Widget<T> {
+fn pad<T: Data + 'static>(inner: impl Widget<T> + 'static) -> impl Widget<T> {
     Padding::new(5.0, inner)
 }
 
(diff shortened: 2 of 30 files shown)


Zarenor:
I think this is really interesting. I'm a little surprised at how infectious the *'static* bound is.
I definitely can see the desire to reach down into the widget tree, though I'm also not sure if that's going to be wise or not.
I do think it's pretty impressive that this is something we can already do fairly easily.

cmyr:
`'static` ends up everywhere because it's a bound on `Any`, and we now require all widgets to be `Any`. I think we could maybe hide that a bit, though?

_edit_ yea, we can actually clean up things tremendously by making `'static'` a bound on `Data`. I think this is probably okay? I don't really think it makes a ton of sense for `Data` to hold references.

cmyr:
An interesting thing about `as_any` is that it is trivially derivable. I think at some point in the future we might want to use derive to implement a bunch of helper stuff in widgets, such as code for tracing & debugging, so that could be a chance to think about this again.

Something else to consider is that it looks like adding a `'static` bound to `Data` actually cleans up code in a bunch of other places, and may be worth it just for that?

cmyr:
closing this, may revisit at some point down the road when I presume we will want to be deriving a bunch of stuff on widgets for debugging etc.

PR #458, opened 2020-01-10
rename container to panel

mendelt:
I think there is some agreement that Container isn't the best name for this widget and I think good names are important. I think `Panel` is a good name for what this widget does.


mendelt:
Great idea! I like these kinds of tricks to make the UX of an API better.

zaynetro:
`Container` name comes from Flutter's Container class: https://api.flutter.dev/flutter/widgets/Container-class.html

cmyr:
@zaynetro That was the initial inspiration, but I agree that 'container' feels overloaded; we use the term 'container widget' to refer to a widget that manages children, so it makes sense to me to try and keep that word for that concept.

mendelt:
The Flutter `Container` does a lot more than ours. Its a combination of our `Align`, `Padding` and `Container` and it adds the ability to do linear transformations on it's contents. It seems a bit arbitrary that we picked one of these behaviors and called it `Container` while the other ones have descriptive names.
It would be great to have a `Transform` widget too. I haven't played with Piet enough to know if that would be something we can easilly do.

zaynetro:
I was just trying to explain original reasoning which was already a few months ago. 
Nevermind, just found a thread in zulip chat.

👍

mendelt:
Panel will be used for a different widget, container function and name is still under discussion. There might be changes to container but it won't be this.

PR #460, opened 2020-01-11
Extend Doc Comments for Lens and include Example of lens usage

nilsmartel:
English isn't my native tongue, so don't hesitate to point out flaws, it only helps me.

I hope this helps to clear things up for future users of druid. The idea to extend the comments came because of [this](https://github.com/xi-editor/druid/issues/453) issue.

If another example is dragging down the development, it's completely reasonable to exclude it from this PR.


druid/src/widget/widget_ext.rs
```diff
@@ -125,8 +125,38 @@ pub trait WidgetExt<T: Data>: Widget<T> + Sized + 'static {
 
     /// Wrap this widget in a [`LensWrap`] widget for the provided [`Lens`].
     ///
+    /// Lenses are useful whenever a widget only needs access to a subfield of a larger struct.
+    /// For Example: If one wants to embed a [`TextBox`] in a Widget of a Data type different from
```
Ralith:
```suggestion
    /// For example: If one wants to embed a [`TextBox`] in a Widget of a Data type different from
```

druid/src/widget/widget_ext.rs
```diff
@@ -125,8 +125,38 @@ pub trait WidgetExt<T: Data>: Widget<T> + Sized + 'static {
 
     /// Wrap this widget in a [`LensWrap`] widget for the provided [`Lens`].
     ///
+    /// Lenses are useful whenever a widget only needs access to a subfield of a larger struct.
```
Ralith: I think this is a bit narrow; lenses can be useful for accessing the interior of things other than structs as well. Perhaps "access to part of a larger value"?

druid/src/widget/widget_ext.rs
```diff
@@ -125,8 +125,38 @@ pub trait WidgetExt<T: Data>: Widget<T> + Sized + 'static {
...
+    /// For Example: If one wants to embed a [`TextBox`] in a Widget of a Data type different from
+    /// `Widget<String>` they need to specify, which part of the Data type the `TextBox` is
+    /// supposed to access.
```
Ralith: Maybe more broadly, "specify how to access a `String` within the `Data`"?

druid/src/widget/widget_ext.rs
```diff
@@ -125,8 +125,38 @@ pub trait WidgetExt<T: Data>: Widget<T> + Sized + 'static {
...
+    ///
     /// [`LensWrap`]: ../struct.LensWrap.html
     /// [`Lens`]: ../trait.Lens.html
+    /// [`TextBox`]: ../TextBox.html
+    ///
+    ///
+    /// ```
+    /// // Sample use case of lenses
```
Ralith: I don't think this comment is needed.

druid/src/widget/widget_ext.rs
```diff
@@ -125,8 +125,38 @@ pub trait WidgetExt<T: Data>: Widget<T> + Sized + 'static {
...
+    ///
+    ///
+    /// fn my_sidebar() -> impl Widget<MyState> {
+    ///     // `TextBox` is of type `Widget<String>`
+    ///     // via `.lens` we get it to be of type `Widget<MyState>`
+    ///     let searchbar = TextBox::new().lens(MyState::search_term);
+    ///
+    ///     // ...
+    ///
+    ///     // We can now use or `searchbar` just like any other Element of type `Widget<MyState>`
```
Ralith:
```suggestion
    ///     // We can now use `searchbar` just like any other Element of type `Widget<MyState>`
```

I'm not sure what "Element" refers to.

druid/src/widget/widget_ext.rs
nilsmartel: You're right. I now included both statements, as I think the concept is hard to get at first (for me it was) I tried to be as concrete here as possible.

druid/src/widget/widget_ext.rs
nilsmartel: Applied

druid/src/widget/widget_ext.rs
nilsmartel: Applied this

druid/src/widget/widget_ext.rs
nilsmartel: There's some redundancy in that sentence, I agree

nilsmartel:
Update:
I ensured that the doc comment compiles and the code complies with druids test pipeline

cmyr:
This is good, but it doesn't address one of my concerns, which is discoverability. Basically: I expect people who are wondering how to use lenses to end up reading the docs for `lens.rs`, and so it would be good if most of this information was in there.

When I actually go and look at those docs, they're sparser than I expected; but maybe we should move these docs you've written over there, as module docs? (I mean: in the `//! Top section`).

druid/src/widget/widget_ext.rs
```diff
@@ -125,8 +125,38 @@ pub trait WidgetExt<T: Data>: Widget<T> + Sized + 'static {
 
     /// Wrap this widget in a [`LensWrap`] widget for the provided [`Lens`].
     ///
+    /// Lenses are useful whenever a widget only needs access to a subfield of a larger struct or
+    /// generally access to part of a larger value.
+    /// For example: If one wants to embed a [`TextBox`] in a Widget of a Data type different from
```
cmyr:
```suggestion
    /// For example: If one wants to embed a [`TextBox`] in a widget with a `Data` type
```

druid/src/widget/widget_ext.rs
```diff
@@ -125,8 +125,38 @@ pub trait WidgetExt<T: Data>: Widget<T> + Sized + 'static {
...
+    /// `Widget<String>` they need to specify how to access a String within the Data.
```
cmyr:
```suggestion
    ///   that is not `String`, they need to specify how to access a `String` from within the `Data`.
```

druid/src/widget/widget_ext.rs
```diff
@@ -125,8 +125,38 @@ pub trait WidgetExt<T: Data>: Widget<T> + Sized + 'static {
...
+    ///     search_term: String,
+    ///     scale: f64,
+    ///     // ...
+    /// }
+    ///
+    ///
+    /// fn my_sidebar() -> impl Widget<MyState> {
+    ///     // `TextBox` is of type `Widget<String>`
+    ///     // via `.lens` we get it to be of type `Widget<MyState>`
+    ///     let searchbar = TextBox::new().lens(MyState::search_term);
```
cmyr: I would add a comment here that says something like,

"`MyState::search_term` is a lens generated by the `derive(Lens)` macro, that provides access to the `search_term` field."

nilsmartel: Sure, I overlooked this, but it's the most important part of the example. I extended the comment

nilsmartel:
> This is good, but it doesn't address one of my concerns, which is discoverability. Basically: I expect people who are wondering how to use lenses to end up reading the docs for `lens.rs`, and so it would be good if most of this information was in there.
> 
> When I actually go and look at those docs, they're sparser than I expected; but maybe we should move these docs you've written over there, as module docs? (I mean: in the `//! Top section`).

I moved the whole description into the top section of `lens.rs`.
the `.lens` method won't need such a large description then (We shouldn't explain what a structure is doing, every time it is mentioned).

This should have been the way to go from the start. Thanks for the kind comments

PR #461, opened 2020-01-11
Change IdleHandler signuture

edwin0cheng:
This PR change the definition for `IdleHandle` from using `std::Any` to :

```rust
impl IdleHandle {
    pub fn add_idle<F>(&self, callback: F)
    where
        F: FnOnce(&mut dyn WinHandler, &mut dyn WinCtx) + Send + 'static,
    {
        ...
    }
}
```

### DIFF
--- a/druid-shell/src/window.rs
+++ b/druid-shell/src/window.rs
@@ -77,6 +77,13 @@ impl IdleHandle {
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Hash)]
 pub struct IdleToken(usize);
 
+impl IdleToken {
+    /// Create a new `IdleToken` with the given raw `usize` id.
+    pub const fn new(raw: usize) -> IdleToken {
+        IdleToken(raw)
+    }
+}
+
 /// A handle to a platform window object.
 #[derive(Clone, Default)]
 pub struct WindowHandle(platform::WindowHandle);
--- a/druid-shell/src/window.rs
+++ b/druid-shell/src/window.rs
@@ -65,9 +65,17 @@ impl IdleHandle {
     where
         F: FnOnce(&dyn Any) + Send + 'static,
     {
-        self.0.add_idle(callback)
+        self.0.add_idle_callback(callback)
+    }
+
+    pub fn schedule_idle(&mut self, token: IdleToken) {
+        self.0.add_idle_token(token)
     }
 }
+
+/// A token that uniquely identifies a idle schedule.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Hash)]
+pub struct IdleToken(usize);
 
 /// A handle to a platform window object.
 #[derive(Clone, Default)]
@@ -308,6 +316,10 @@ pub trait WinHandler {
     /// WM_NCDESTROY).
     #[allow(unused_variables)]
     fn destroy(&mut self, ctx: &mut dyn WinCtx) {}
+
+    /// Called when a idle token is requested by [`IdleHandle::schedule_idle()`] call.
+    #[allow(unused_variables)]
+    fn idle(&mut self, token: IdleToken, ctx: &mut dyn WinCtx) {}
 
     /// Get a reference to the handler state. Used mostly by idle handlers.
     fn as_any(&mut self) -> &mut dyn Any;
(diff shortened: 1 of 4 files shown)


cmyr:
Interesting. I don't love the current signature, but it does allow you to do a very specific thing; it lets you, in the callback, to downcast to the _concrete_ type of a _particular_ `WinHandler`; so for `druid` that means that we can get a `DruidHandler`., and then we can call arbitrary methods on that that aren't part of the `WinHandler` trait.

I'm not sure if this is worth it, but I'm also not sure what motivated the original design; this code predates druid being druid, and was part of the old `xi-win` experiment. Maybe @raphlinus can chime in?

Something I'd want to think about, now, is what exactly this API is for. For _our_ purposes, all we really want is to get poked, and we don't need any of this fancy stuff; we could just have a `fn idle(&mut self, token: IdleToken)` on the `WinHandler` trait, and then this closure takes no arguments; we just get called back with a token that we provide, and we can choose what to do based on that?

Ralith:
`WinHandler` could implement `downcast` itself without much effort, as done with e.g. [downcast-rs](https://github.com/marcianx/downcast-rs).

cmyr:
oh interesting, I hadn't seen that crate, and I'd been muddling around doing something similar; it's interesting to see it generalized like that.

edwin0cheng:
> Interesting. I don't love the current signature, but it does allow you to do a very specific thing; it lets you, in the callback, to downcast to the concrete type of a particular WinHandler; so for druid that means that we can get a DruidHandler., and then we can call arbitrary methods on that that aren't part of the WinHandler trait.

One of the reason I change to this signuature is , we need 'WinCtx' for post-paint event `process_command`. (And basically all useful cases) But I agreed we could just add a `idle` handler for `WinHandler` and call it directly.

cmyr:
Okay, well with this in mind I would go with a simple token based idle system; you would call `schedule_idle(token: IdleToken)` and then there would be a method on `WinHandler`, `idle(&mut self, ctx: &WinCtx, token: IdleToken)`.

edwin0cheng:
@cmyr I changed the api as you recommended 👍

druid-shell/src/window.rs
```diff
@@ -65,10 +65,18 @@ impl IdleHandle {
     where
         F: FnOnce(&dyn Any) + Send + 'static,
     {
-        self.0.add_idle(callback)
+        self.0.add_idle_callback(callback)
+    }
+
+    pub fn schedule_idle(&mut self, token: IdleToken) {
```
cmyr: This should have some basic doc string, maybe:

"Request a callback from the runloop. Your [`WinHander::idle`](trait.WinHandler.html#tymethod.idle) method will be called with the `token` that was passed in."

druid-shell/src/window.rs
```diff
@@ -65,10 +65,18 @@ impl IdleHandle {
     where
         F: FnOnce(&dyn Any) + Send + 'static,
     {
-        self.0.add_idle(callback)
+        self.0.add_idle_callback(callback)
```
cmyr: I wonder if we even need to keep the closure based API?

We could even get fancy and have a single method that has the signature, `schedule_idle(&self, idle: impl Into<IdleKind>)` but i think that's too fancy for now.

I think what makes the most sense is that we merge this as-is, and we can worry about `add_idle` at some later point.

druid-shell/src/window.rs
```diff
@@ -65,10 +65,18 @@ impl IdleHandle {
...
+    }
+
+    pub fn schedule_idle(&mut self, token: IdleToken) {
+        self.0.add_idle_token(token)
     }
 }
 
+/// A token that uniquely identifies a idle schedule.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Hash)]
+pub struct IdleToken(usize);
```
cmyr: This needs to have some public constructor, probably just `const fn new(raw: usize)`?

druid-shell/src/window.rs
edwin0cheng: Agreed. Let review it in future after we collect more data point of usage of these idle api.

edwin0cheng:
Fixed 👍

PR #462, opened 2020-01-12
Fix some bugs in svg widget

fishrockz:
The svg widget can currently spill out of its own area, this is fixed with a clip function in the first commit.

The svg widget also dose not respect the transforms property which is needed for some svg's, for me the svg's created by plotlib were not rendering correctly, this is fixed in the second commit.

### DIFF
--- a/druid/src/widget/svg.rs
+++ b/druid/src/widget/svg.rs
@@ -24,8 +24,8 @@ use log::error;
 use usvg;
 
 use crate::{
-    kurbo::BezPath, BoxConstraints, Color, Data, Env, Event, EventCtx, LayoutCtx, PaintCtx, Point,
-    RenderContext, Size, UpdateCtx, Widget,
+    kurbo::BezPath, Affine, BoxConstraints, Color, Data, Env, Event, EventCtx, LayoutCtx, PaintCtx,
+    Point, Rect, RenderContext, Size, UpdateCtx, Widget,
 };
 
 /// A widget that renders a SVG
@@ -85,12 +85,19 @@ impl<T: Data> Widget<T> for Svg<T> {
     }
     fn paint(&mut self, paint_ctx: &mut PaintCtx, _data: &T, _env: &Env) {
         //TODO: options for aspect ratio or scaling based on height
-        let scale = paint_ctx.size().width / self.get_size().width;
+        let scalex = paint_ctx.size().width / self.get_size().width;
+        let scaley = paint_ctx.size().height / self.get_size().height;
+        let scale = scalex.min(scaley);
 
         let origin_x = (paint_ctx.size().width - (self.get_size().width * scale)) / 2.0;
         let origin_y = (paint_ctx.size().height - (self.get_size().height * scale)) / 2.0;
         let origin = Point::new(origin_x, origin_y);
 
+        let clip_rect = Rect::ZERO.with_size(paint_ctx.size());
+
+        // The SvgData's to_piet function dose not clip to the svg's size
+        // CairoRenderContext is very like druids but with some extra goodies like clip
+        paint_ctx.clip(clip_rect);
         self.svg_data.to_piet(scale, origin, paint_ctx);
     }
 }
--- a/druid/src/widget/svg.rs
+++ b/druid/src/widget/svg.rs
@@ -132,7 +132,7 @@ impl SvgData {
     /// Convert SvgData into Piet draw instructions
     pub fn to_piet(&self, scale: f64, offset: Point, paint_ctx: &mut PaintCtx) {
         let root = self.tree.root();
-
+        let offset_matrix = Affine::new([scale, 0., 0., scale, offset.x, offset.y]);
         for n in root.children() {
             match *n.borrow() {
                 usvg::NodeKind::Path(ref p) => {
@@ -140,13 +140,9 @@ impl SvgData {
                     for segment in p.data.iter() {
                         match *segment {
                             usvg::PathSegment::MoveTo { x, y } => {
-                                let x = (x * scale) + offset.x;
-                                let y = (y * scale) + offset.y;
                                 path.move_to((x, y));
                             }
                             usvg::PathSegment::LineTo { x, y } => {
-                                let x = (x * scale) + offset.x;
-                                let y = (y * scale) + offset.y;
                                 path.line_to((x, y));
                             }
                             usvg::PathSegment::CurveTo {
@@ -157,13 +153,6 @@ impl SvgData {
                                 x,
                                 y,
                             } => {
-                                let x1 = (x1 * scale) + offset.x;
-                                let y1 = (y1 * scale) + offset.y;
-                                let x2 = (x2 * scale) + offset.x;
-                                let y2 = (y2 * scale) + offset.y;
-                                let x = (x * scale) + offset.x;
-                                let y = (y * scale) + offset.y;
-
                                 path.curve_to((x1, y1), (x2, y2), (x, y));
                             }
                             usvg::PathSegment::ClosePath => {
@@ -171,6 +160,17 @@ impl SvgData {
                             }
                         }
                     }
+
+                    path.apply_affine(Affine::new([
+                        p.transform.a,
+                        p.transform.b,
+                        p.transform.c,
+                        p.transform.d,
+                        p.transform.e,
+                        p.transform.f,
+                    ]));
+                    path.apply_affine(offset_matrix);
+
                     match &p.fill {
                         Some(fill) => {
                             let brush = color_from_usvg(&fill.paint, fill.opacity);


druid/src/widget/svg.rs
```diff
@@ -85,13 +88,29 @@ impl<T: Data> Widget<T> for Svg<T> {
...
-        self.svg_data.to_piet(scale, origin, paint_ctx);
+        let clip_rect = RoundedRect::from_origin_size(
+            Point::ORIGIN,
+            Size::new(paint_ctx.size().width, paint_ctx.size().height).to_vec2(),
+            2.,
+        );
+
+        // The SvgData's to_piet function dose not clip to the svg's size
+        // CairoRenderContext is very like druids but with some extra goodies like clip
+        paint_ctx
```
futurepaul: You shouldn't need to import piet-cairo, the whole point of piet is to abstract over backends like this. If you use a backend explicitly then druid becomes non-portable.

In this case I believe `paint_ctx.clip(clip_rect);` is what you want.

druid/src/widget/svg.rs
```diff
@@ -85,13 +88,29 @@ impl<T: Data> Widget<T> for Svg<T> {
...
+        let scalex = paint_ctx.size().width / self.get_size().width;
+        let scaley = paint_ctx.size().height / self.get_size().height;
+        let scale = scalex.min(scaley);
 
         let origin_x = (paint_ctx.size().width - (self.get_size().width * scale)) / 2.0;
         let origin_y = (paint_ctx.size().height - (self.get_size().height * scale)) / 2.0;
         let origin = Point::new(origin_x, origin_y);
 
-        self.svg_data.to_piet(scale, origin, paint_ctx);
+        let clip_rect = RoundedRect::from_origin_size(
```
futurepaul: Is there a reason this is a `RoundedRect` instead of a plain `Rect`?

fishrockz: Nope, I copy and pasted most of this from one of the other widgets, ill fix this up

druid/src/widget/svg.rs
fishrockz: Ah brill I'll see if i can fix this up

fishrockz:
The current Imp dose not use the transform property so svg's that use it like the plotlib svgs were rendering with things in the wrong place so i needed the first apply_Affine to fix that but as matrix multiplications must be done in order i move the offest and scale multiplication in to a second matrix so they could be applied after the offset, this could be achieved in other ways but this seemed the neatest.

druid/src/widget/svg.rs
```diff
@@ -24,8 +24,9 @@ use log::error;
 use usvg;
 
 use crate::{
-    kurbo::BezPath, BoxConstraints, Color, Data, Env, Event, EventCtx, LayoutCtx, PaintCtx, Point,
-    RenderContext, Size, UpdateCtx, Widget,
+    kurbo::{Affine, BezPath, Rect},
```
cmyr: more a note than anything: `Rect` and `Affine` are both reexported directly by druid, so you don't technically need to reach into kurbo to get them. `BezPath` isn't reexported because it isn't used in any of druid's API.

druid/src/widget/svg.rs
```diff
@@ -85,12 +86,22 @@ impl<T: Data> Widget<T> for Svg<T> {
...
+        let scaley = paint_ctx.size().height / self.get_size().height;
+        let scale = scalex.min(scaley);
 
         let origin_x = (paint_ctx.size().width - (self.get_size().width * scale)) / 2.0;
         let origin_y = (paint_ctx.size().height - (self.get_size().height * scale)) / 2.0;
         let origin = Point::new(origin_x, origin_y);
 
+        let clip_rect = Rect::from_origin_size(
+            Point::ORIGIN,
+            Size::new(paint_ctx.size().width, paint_ctx.size().height),
```
cmyr: This can just be `paint_ctx.size()` directly, that returns a `Size` struct.

A sort-of-sneaky way I sometimes would write this is,

```rust
let clip_rect = Rect::ZERO.with_size(paint_ctx.size());
```

`with_size` is a method on `Rect` that returns a new `Rect` with the same origin as `self`, but with a new size. 🤷‍♂

druid/src/widget/svg.rs
```diff
@@ -150,20 +156,24 @@ impl SvgData {
...
-
                                 path.curve_to((x1, y1), (x2, y2), (x, y));
                             }
                             usvg::PathSegment::ClosePath => {
                                 path.close_path();
                             }
                         }
                     }
+
+                    path.apply_affine(Affine::new([
```
cmyr: annoying that this conversion isn't easier to do, especially since `usvg` _does_ have a dependency on kurbo..

druid/src/widget/svg.rs
```diff
@@ -150,20 +156,24 @@ impl SvgData {
...
+                        p.transform.a,
+                        p.transform.b,
+                        p.transform.c,
+                        p.transform.d,
+                        p.transform.e,
+                        p.transform.f,
+                    ]));
+                    path.apply_affine(offset_matrix);
```
cmyr: i'm really scratching my head here. where does `offset_matrix` come from?

druid/src/widget/svg.rs
```diff
@@ -85,12 +86,22 @@ impl<T: Data> Widget<T> for Svg<T> {
...
         let origin = Point::new(origin_x, origin_y);
 
+        let clip_rect = Rect::from_origin_size(
+            Point::ORIGIN,
+            Size::new(paint_ctx.size().width, paint_ctx.size().height),
+        );
+
+        // The SvgData's to_piet function dose not clip to the svg's size
+        // CairoRenderContext is very like druids but with some extra goodies like clip
+        paint_ctx.clip(clip_rect);
```
cmyr: @raphlinus any reason we wouldn't want to be doing this?

druid/src/widget/svg.rs
cmyr: @razrfalcon do you have any interest in using kurbo for more of the basic geom stuff in usvg?  No worries either way, but if you're already bringing it in for other stuff it might let you delete some code.

RazrFalcon:
>I'm also not positive we even need a clip? If we know the size of the svg and we know the box constraint, can't we always scale the svg to fit inside that constraint?

SVG can have objects outside the viewbox. So clip is important.

druid/src/widget/svg.rs
RazrFalcon: >I do assume there's a good reason you want your own impls

Yes, in my case, my geometry primitives are immutable and always valid. Which simplify things a lot. I plan to move them into a separate crate eventually.

Also, I will move back to `lyon` when `raqote` backend became default, since it already used it.

cmyr: sounds good, thanks!

druid/src/widget/svg.rs
fishrockz: The second commit applies the transform matrix but it also refractors the old offset and scale transforms to happen after the transform matrix has been applied as matrix multiplication is not like scaler multiplication so needs to happen in the right order, this seemed like a much more succinct way of doing it.

maybe scale_offset_matrix may be better?

I think i could fudge the maths so that i don't need to move the old scale and offset transforms and instead apply a matrix to the transform matrix before applying it bu this seems much nicer.

fishrockz: oh i see what you mean! i had messed up a rebase to fix some other comments :( my very bad!

druid/src/widget/svg.rs
fishrockz: I think I have applied this suggestion

druid/src/widget/svg.rs
fishrockz: Applied, thanks

fishrockz:
I think I have ether applied the requested changes or @RazrFalcon has confirmed what I thought about needing the clip.

Could @cmyr and @futurepaul  confirm I have addressed there issues?

I also made my demo app a bit more fun, the slider now moves a the phase of a load of complex numbers and then has a fft and then a peak that lines up with the slider https://gitlab.com/pointswaves/rust-plot-gui/blob/master/src/main.rs I thought going via svg was not such a good idea but i have been impressed at how responsive this is, I would like to come up with a more direct approach however Im impressed with all these libs working together.

druid/src/widget/svg.rs
```diff
@@ -85,12 +85,19 @@ impl<T: Data> Widget<T> for Svg<T> {
...
+        let scaley = paint_ctx.size().height / self.get_size().height;
+        let scale = scalex.min(scaley);
 
         let origin_x = (paint_ctx.size().width - (self.get_size().width * scale)) / 2.0;
         let origin_y = (paint_ctx.size().height - (self.get_size().height * scale)) / 2.0;
         let origin = Point::new(origin_x, origin_y);
 
+        let clip_rect = Rect::ZERO.with_size(paint_ctx.size());
+
+        // The SvgData's to_piet function dose not clip to the svg's size
```
futurepaul: Can get rid of this comment, otherwise looks good to me!

cmyr: I'm happy enough to merge with this here, I think it's a reasonable observation.

cmyr:
Thanks @fishrockz!

Issue #457, opened 2020-01-10
Capture mouse on click


raphlinus:
This is branched from #450, but has come up repeatedly, so I felt it needed its own issue.

Mouse move and up events are dropped when the mouse is clicked in window then dragged outside. This causes all kinds of problems, as widgets (reasonably) expect an eventual mouse up after a mouse down.

I think it needs to be solved at the druid-shell level. On Windows, we should be calling [SetCapture](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setcapture) (and ReleaseCapture). I'm not sure what the counterparts are on other systems; on macOS I think this may be handled by the platform.

When should the app grab mouse capture? One possibility is to *always* do this on mouse click. Another is to have it plumbed down from set_active (and there is a [TODO](https://github.com/xi-editor/druid/blob/master/druid/src/core.rs#L617) in the source for this). If the behavior on macOS is to always capture, then I think the former route is more developer-friendly, as there's less risk of strange behavior when set_active is not called, and of course less platform variation.

mendelt:
We could even send a mouse-up when the mouse leaves the window. That might be a solution for platforms that don't support mouse capture?

cmyr:
Are there platforms that don't support mouse capture? Sending a mouse up sounds like it would cause other problems; what happens if the user drags briefly out off a button out of the window, then back onto the button and actually does mouse up?

justinwilaby:
Hello - 

After reading thought the linked documents, `SetCapture` looks most appropriate for certain circumstances. e.g. scroll thumb press and drag (since scrolling should still occur even if the mouse leaves the window after pressing the scroll thumb). If the plan is to implement drag and drop operations natively on widgets identified as drag or drop targets, [DragDetect](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-dragdetect) is promising.

For all other cases, this is what I've seen in the past:

1. User presses the left mouse down on a widget - the widget receives a `MouseDown` as per usual.
2. User drags the widget - if it's a drag target, a drag operation is initiated and `DragMove` events are triggered until the mouse is released regardless of where the release occurs.
3. If it is not a drag target and the mouse leaves the window and then is released sometime later, a `MouseUpSomewhere` event is dispatched, or some other event that distinguishes between a normal `MouseUp`. This prevents the normal `MouseUp` behavior but allows things like `MouseMove` and maybe the `down` or `focused` state to continue until the user releases the mouse.
[...]

PR #463, opened 2020-01-13
[WIP] VirtualList and decoupled Scrollbar widgets

justinwilaby:
Hi -

Cool project...
I'm a 12 year UI dev and would like to contribute. Here are a few items I feel this project could benefit from and would love your feedback. I'm a relative newcomer to Rust and Druid but UI has been my focus for some time and the patterns used in Druid are very familiar to me.

## This PR is a WIP and contains: 
1. Virtual List widget - Tested using 1 million rows with no performance cost. It can take many, many more rows with zero performance impact. Works with both Vertical and Horizontal axis.
2. A Decoupled `Scrollbar` widget - Basically a slider that interpolates between a given Range instead of a viewport boundary. This decoupling allows the most flexible usage for views that virtualize layouts.
3. `Event::Size` propagates to children - this was necessary to adjust the scroll position bindings and to calculate a new `max_scroll_position` value. 

## To Test
run the `virtual_list.rs` in `examples/`

### VirtualList
- [ ] Keyboard navigation / accessibility
- [x] Determine if reusing popped widgets provides advantages and implement if yes. ~~(tabled for later)~~ Implemented
- [x] Variable size rows/cols - Potentially impactful with performance since all items would need to be measured in order to calculate the `max_scroll_position`.
- [ ] Smooth scroll - A fairly straightforward implementation since the VirtualList design keeps this in mind. 
- [ ] Test thoroughly 
### Scrollbar
- [x] Apply animation to scrollbar - this will likely be a copy/paste from the original `Scroll` widget
- [x] Page the scroll position up or down depending on where a mouse event occurs in relation to the scroll thumb along the scroll track.
- [ ] Wire in whatever solution is found for #457 
- [x] Test thoroughly

I have a few questions also: 

1. The `update` flow seems to aggressively clone the entire data structure even when a single field is all that has changed. This still seems to happen when manually implementing the `Data` trait also. Is there a way to avoid this? If no, are there plans to address this? e.g. Observables 
2. I found the `Data` requirement to be cumbersome. Why was this chosen over just `PartialEq`? I see they are somewhat synonymous `#[druid(same_fn = "PartialEq::eq")]`
3. `Event::MouseMoved` requires a delta to be calculated manually. Are there plans to include the delta similar the `Event::Scroll`?

I hope you find these widgets useful. I certainly had fun creating them. 

Thank you for your time and feedback!

Justin

### DIFF
--- a/druid/src/core.rs
+++ b/druid/src/core.rs
@@ -275,7 +275,6 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
         let child_event = match event {
             Event::LifeCycle(event) => Event::LifeCycle(*event),
             Event::Size(size) => {
-                recurse = ctx.is_root;
                 Event::Size(*size)
             }
             Event::MouseDown(mouse_event) => {
--- a/druid/src/core.rs
+++ b/druid/src/core.rs
@@ -266,7 +266,6 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
             base_state: &mut self.state,
             had_active,
             is_handled: false,
-            is_root: false,
         };
         let rect = child_ctx.base_state.layout_rect;
         // Note: could also represent this as `Option<Event>`.
@@ -563,7 +562,6 @@ pub struct EventCtx<'a, 'b> {
     pub(crate) base_state: &'a mut BaseState,
     pub(crate) had_active: bool,
     pub(crate) is_handled: bool,
-    pub(crate) is_root: bool,
     pub(crate) widget_id: WidgetId,
 }
 
--- a/druid/src/core.rs
+++ b/druid/src/core.rs
@@ -273,9 +273,7 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
         let mut hot_changed = None;
         let child_event = match event {
             Event::LifeCycle(event) => Event::LifeCycle(*event),
-            Event::Size(size) => {
-                Event::Size(*size)
-            }
+            Event::Size(size) => Event::Size(*size),
             Event::MouseDown(mouse_event) => {
                 let had_hot = child_ctx.base_state.is_hot;
                 let now_hot = rect.winding(mouse_event.pos) != 0;
--- a/druid/src/theme.rs
+++ b/druid/src/theme.rs
@@ -50,6 +50,8 @@ pub const SCROLL_BAR_PAD: Key<f64> = Key::new("scroll_bar_pad");
 pub const SCROLL_BAR_RADIUS: Key<f64> = Key::new("scroll_bar_radius");
 pub const SCROLL_BAR_EDGE_WIDTH: Key<f64> = Key::new("scroll_bar_edge_width");
 
+pub const DEFAULT_ANIMATION_DURATION: Key<u64> = Key::new("default_animation_duration");
+
 /// An initial theme.
 pub fn init() -> Env {
     let mut env = Env::default()
@@ -78,7 +80,8 @@ pub fn init() -> Env {
         .adding(SCROLL_BAR_WIDTH, 8.)
         .adding(SCROLL_BAR_PAD, 2.)
         .adding(SCROLL_BAR_RADIUS, 5.)
-        .adding(SCROLL_BAR_EDGE_WIDTH, 1.);
+        .adding(SCROLL_BAR_EDGE_WIDTH, 1.)
+        .adding(DEFAULT_ANIMATION_DURATION, 250_000_000u64);
 
     #[cfg(target_os = "windows")]
     {
--- a/druid/src/widget/scrollbar.rs
+++ b/druid/src/widget/scrollbar.rs
@@ -46,13 +46,14 @@ pub trait ScrollControlState: Data + PartialEq {
     /// for pull or bounce calculations on the container
     fn set_scroll_position(&mut self, val: f64);
 
-    fn set_scroll_pos_from_delta(&mut self, delta: f64) {
+    fn set_scroll_pos_from_delta(&mut self, delta: f64) -> f64 {
         let scroll_position = self.scroll_position() + delta;
+        let clamped_scroll_position = self
+            .min_scroll_position()
+            .max(scroll_position.min(self.max_scroll_position()));
+        self.set_scroll_position(clamped_scroll_position);
 
-        self.set_scroll_position(
-            self.min_scroll_position()
-                .max(scroll_position.min(self.max_scroll_position())),
-        );
+        clamped_scroll_position
     }
 }
 
@@ -245,7 +246,6 @@ impl<T: ScrollControlState> Widget<T> for Scrollbar {
                 if !self.animation_state.done {
                     self.opacity = self.animation_state.tick(*interval);
                     event_ctx.request_anim_frame();
-                    event_ctx.invalidate();
                 } else if !self.is_hot && self.opacity > 0. {
                     self.hide(env);
                     event_ctx.request_anim_frame();
@@ -263,7 +263,6 @@ impl<T: ScrollControlState> Widget<T> for Scrollbar {
                 };
                 self.show();
                 data.set_scroll_pos_from_delta(delta);
-                event_ctx.invalidate();
                 event_ctx.request_anim_frame();
             }
 
--- a/druid/src/widget/scrollbar.rs
+++ b/druid/src/widget/scrollbar.rs
@@ -1,8 +1,11 @@
 use std::mem;
 
-use crate::{BoxConstraints, Data, Env, Event, EventCtx, LayoutCtx, PaintCtx, RenderContext, Selector, UpdateCtx, Widget};
 use crate::kurbo::{Rect, RoundedRect, Size};
 use crate::theme;
+use crate::{
+    BoxConstraints, Data, Env, Event, EventCtx, LayoutCtx, PaintCtx, RenderContext, Selector,
+    UpdateCtx, Widget,
+};
 
 /// Trait used for shared state between
 /// the scrollbar and the scroll container.
@@ -46,9 +49,10 @@ pub trait ScrollControlState: Data + PartialEq {
     fn set_scroll_pos_from_delta(&mut self, delta: f64) {
         let scroll_position = self.scroll_position() + delta;
 
-        self.set_scroll_position(self.min_scroll_position()
-            .max(scroll_position
-                .min(self.max_scroll_position())));
+        self.set_scroll_position(
+            self.min_scroll_position()
+                .max(scroll_position.min(self.max_scroll_position())),
+        );
     }
 }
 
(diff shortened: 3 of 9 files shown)


Zarenor:
Hello!

Thanks for the PR!
This is exciting; virtualizing lists is something we've discussed but haven't prioritized implementing yet. This sounds like a big step ahead. I'll defer to others a bit more expert on the decoupled scroll and the Event changes.

I'm going to try to answer your questions a little quickly, and see if that brings about more questions.

1. Yes, that's correct, currently we clone data pretty aggressively. Right now, we don't have a way to change that behavior. We may change this implementation in the future, but the intended design isn't far from this. Instead, the idea is that 'large' portions of data would be cheaply clone-able and share-able by putting them behind an `Arc`. This makes the clone cheap, and lets us parcel out the pieces we need to check for same-ness.

2. Yes, the `Data` requirement is by-design. The choice over `PartialEq` is related to the aggressively-cloning behavior; `Data` boils down to a `ptr_eq` check for `Arc`, so if you structure your data around relevant 'chunks' behind `Arc`s, you get cheap comparison, and can easily propagate changes back up the tree by `Arc::make_mut`-ing and changing that one chunk. 

3. Unfortunately, I can't speak very well to the design of `MouseMoved`. I'll have to let someone else answer this a little better.

One more thing: We're pretty active in a [Zulip instance](https://xi.zulipchat.com/). You're welcome to come join us, if you like.

justinwilaby:
@Zarenor - Thank you for taking the time to answer!

Ah, it makes sense to place large data chunks behind an `Arc` for cheap cloning. I see that the `List` widget does this already. On the subject of data, what is the preferred recipe for widget configuration values? I seems reasonable to include these as `impl Data` when a shared mutable state is required or pass them to setters at the time of widget creation otherwise. I see a bit of both in there.

Zarenor:
Widget configuration data is something of a frontier, here. There are a couple of cases: 

- We're anticipating moving things you might find in styles (font sizes, border sizes, colors, maybe margins) into the `Env` - much of it is there already. The `Env` is primarily a key-based value store.
- Configuration which doesn't depend on state is done at widget instantiation. Usually, that's window creation, but as you can infer, `List` creates and destroys widgets as the collection changes. and your virtualizing list will necessarily need to manage widget lifetimes. So some of this more static config is typically in the widget struct.
 - Configuration which *is* dependent on state is going to be some sort of `Data`, and passed through the data tree. That most likely means it will be some part of your AppState, but through lenses you can focus down to the relevant parts. It is totally reasonable to implement `Widget` non-generically in some cases, where your needed data is only one type (`String`, for a `TextBox`, `f64` for a `Slider`). This idea isn't restricted to primitive or built-in types - we fully expect widgets built to business-domain types, or their UI-side mirrors.

Hopefully that's a little clarifying.

druid/src/core.rs
```diff
@@ -266,18 +266,14 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
...
         // Note: could also represent this as `Option<Event>`.
         let mut recurse = true;
         let mut hot_changed = None;
         let child_event = match event {
             Event::LifeCycle(event) => Event::LifeCycle(*event),
-            Event::Size(size) => {
-                recurse = ctx.is_root;
-                Event::Size(*size)
-            }
+            Event::Size(size) => Event::Size(*size),
```
justinwilaby: Size events propagate to children. In the case of the VirtualList, the max_scroll_position is calculated prior to the layout run so the scrollbar can layout the thumb in relation to the virtualized viewport size. 

This could alternately be accomplished if the `Data` were mutable in `layout` as well but I understand the purpose of the layout run is to commit positioning for `paint`. Mutating data in layout could cause side effects.

druid/src/theme.rs
```diff
@@ -50,6 +50,8 @@ pub const SCROLL_BAR_PAD: Key<f64> = Key::new("scroll_bar_pad");
 pub const SCROLL_BAR_RADIUS: Key<f64> = Key::new("scroll_bar_radius");
 pub const SCROLL_BAR_EDGE_WIDTH: Key<f64> = Key::new("scroll_bar_edge_width");
 
+pub const DEFAULT_ANIMATION_DURATION: Key<u64> = Key::new("default_animation_duration");
```
justinwilaby: Not sure about this one. It might make sense to have reasonable defaults for animation duration. Until this PR, I did not notice any time-based animation implementations.

druid/src/widget/scrollbar.rs
```diff
@@ -0,0 +1,375 @@
...
+use crate::kurbo::{Rect, RoundedRect, Size};
+use crate::theme;
+use crate::{
+    BoxConstraints, Data, Env, Event, EventCtx, LayoutCtx, PaintCtx, RenderContext, Selector,
+    UpdateCtx, Widget,
+};
+
+/// Trait used for shared state between
+/// the scrollbar and the scroll container.
+pub trait ScrollControlState: Data + PartialEq {
```
justinwilaby: This trait is needed to implement a shared state between the scrollbar and the widget being "scrolled". Normally, this information is derived from the container widget when the scrollbar is implemented as a child. This decoupled scrollbar acts more like a reskinned slider widget, taking in some values to scroll and interpolating between them.

druid/src/widget/scrollbar.rs
```diff
@@ -0,0 +1,375 @@
...
+        let scroll_position = self.scroll_position() + delta;
+
+        self.set_scroll_position(
+            self.min_scroll_position()
+                .max(scroll_position.min(self.max_scroll_position())),
+        );
+    }
+}
+
+struct AnimationState {
```
justinwilaby: A time-based animation struct. This should be replaced with something more robust later but seems to perform well for the time.

druid/src/widget/scrollbar.rs
justinwilaby: I should mention that the `scroll_position` can be influenced from the outside making it possible to scroll to a specific element in the list in response to an action originating anywhere in the app.

justinwilaby:
@cmyr - 

This is exactly the feedback I was hoping for! Thank you for taking the time to review it and for the thorough and articulate response. 

A lot of what you cited makes perfect sense and I agree with your suggestions.

## Rationale 
This was an experimental alternative to wrapping the `VirtualList` and `Scrollbar` in a parent `Widget` whose sole purpose would be to maintain and shuttle state and configuration between the two siblings. If implemented this way, the parent widget may need to duplicate and/or pass through the APIs for both the VirtualList as well as the Scrollbar. e.g. `child_factory`, `measurement_function`, `Axis`, etc.

Instead of sharing a parent, the two share state and can mutate it independently. Originally, I planned on providing a default implementation for the `ScrollControlState` to prevent the user form having to do this manually. The flexibility of this composition model would allow for the inclusion of smooth scrolling, drag and throw behaviors as well as easily implementing `scroll_to_index` or fixed content layout similar to that found in Pivot tables and the like. A scroll container with a fixed implementation is great for encapsulation and a low cognitive overhead but could mean some degree of duplication when more advanced control becomes necessary.

With that said, I get your point and can just as easily maintain the scrollbar as a direct child. 

## Generalization of the`VirtualList`
This is really just a virtualization container with some scrolling capabilities. It's used here as a "list" but I have used this type of control as the basis for `ViewStack, `DataGrid`, `PivotTable`, `Dropdown` and `TextDocument` widgets that I've created before. It's inherent ability to handle anywhere from a few to many millions of rows with fixed or variable sizes without blinking makes it a good general purpose layout tool. Packaging it with a fixed scrolling implementation may not suite all these use cases.

## Redefining the existing Scrollbar
The existing Scrollbar is very good and well architected. Besides the limitation you describe, there is no reason not to use it here. The scrollbar I build serves more as a POC for a decoupled Scrollbar and not necessarily a replacement. 

## Thoughts
The`ScrollControlState` can be removed or the interior mutability can. It was put in place as a precaution in case mutability was necessary outside that which is provided by the lifecylce and to make cloning cheap. If the Scrollbar becomes a child of the VirtualList, there are better alternatives. If the VirtualList is utilized for other purposes later, it may be necessary to revise or abstract the scrolling behavior. 

Your direction is invaluable and this project is very promising! Please let me know how I can best help.

druid/src/widget/scrollbar.rs
justinwilaby: I'll have to review this pattern and learn how it can be utilized here. Sounds exciting.

druid/src/widget/scrollbar.rs
```diff
@@ -0,0 +1,411 @@
...
+            equation: |t: f64, b: f64, c: f64, d: f64| -> f64 {
+                // cubic ease out
+                let t = t / d - 1.;
+                c * (t * t * t + 1.) + b
+            },
+        }
+    }
+}
+
+pub struct Scrollbar<S: ScrollControlState> {
```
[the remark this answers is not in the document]
justinwilaby: I wouldn't expect anything less!

druid/src/widget/scrollbar.rs
cmyr: the best place to start would probably be `examples/identity.rs`. There's a bunch of improvements landing in the next week or so too, 🤞.

cmyr:
Thank you for taking the time to do the initial work, and for being receptive, to feedback. :)

At the heart of things, right now, we're still figuring out what patterns make sense. Exploring that space is definitely valuable. In terms of the things that are being considered for inclusion in the core crate, however, my inclination is to prefer simplicity to conceptual elegance.

My basic concern at this point is maybe just that this PR is slightly.. premature? In the sense that it's really pushing things beyond what we've really tried yet, and so necessarily is raising questions that I don't have godo answers for, because I haven't really tried this stuff out myself. That makes it tricky to review, because I don't have the immediate experience of running into the constraints that motivate your solutions. My gut feeling is that we should probably sit on this until we've had a chance to get some of the more foundational stuff settled.

> The existing Scrollbar is very good and well architected. Besides the limitation you describe, there is no reason not to use it here. The scrollbar I build serves more as a POC for a decoupled Scrollbar and not necessarily a replacement.

I'm not 100% sure that the current scrollbar is quite right, although I'm sure it can get there. Some features I would like to see would include zooming, explicit content size, external control, and better behaviour when the inner widget is small. I'm sure there are other things as well. I've bumped into all of these in runebender, and my solutions have been pretty hacky.
 
> Your direction is invaluable and this project is very promising! Please let me know how I can best help.

This is an excellent question. Taking your offer in the broadest sense, there is one major todo on my radar right now that would be well suited to someone with experience doing UI and layout programming, and that is working to refine the layout primatives. I suspect there are a lot of layout helper widgets that we will want and that just aren't implemented yet, and some of our implementations are buggy or incomplete. I imagine that we will eventually require most of the [layout widgets] that are provided by flutter; our layout systems are very similar, and if they've decided they need a layout primitive I suspect we will need it at all.

If you're curious, a way I would start exploring this problem is to simply sketch out various increasingly complex layouts, and then try to build them in druid; if you can't get them to work, then we have some work to do.

Another component of this project might be improving layout debugging; the simplest version of this is just trying to catch and log common layout warnings (zero width or height widgets, for instance) but I'm open to fancier ideas. It would, for instance, be nice if we could run a druid app in a view debugger that would draw the view hierarchy, but... well one thing at a time. :)

This is a big and important part of the project that is not currently getting a ton of attention, and if you were interested I think there's a lot of fun stuff here.

[layout widgets]: https://flutter.dev/docs/development/ui/widgets/layout

PR #466, opened 2020-01-19
Z-order painting

scholtzan:
This addresses https://github.com/xi-editor/druid/issues/430

This requires https://github.com/linebender/piet/pull/117 to have access to `current_transform()` which is necessary to keep track of applied transformations.

One thing that is missing here (also because I'm not 100% sure) is keeping track of clipping. Right now previous clipping is not applied to z-order painting. It would be a bit trickier since `PaintCtx` would need to keep track of clipping.

One other thing that I was thinking about was thinking about is whether it would make sense to allow negative `z_index`es to allow painting below content?

### DIFF
--- a/druid/examples/anim.rs
+++ b/druid/examples/anim.rs
@@ -16,7 +16,8 @@
 
 use std::f64::consts::PI;
 
-use druid::kurbo::Line;
+use druid::kurbo::{Circle, Line, Affine};
+use druid::widget::{Align, Button, Checkbox, Flex, Label, Padding, ProgressBar, Slider};
 use druid::{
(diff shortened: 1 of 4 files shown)


cmyr:
cool, let's get that piet stuff merged and then this should be straightforward. Thanks for taking this on!

cmyr:
okay #472 brings your piet changes into druid, so this should be unblocked shortly. 👌

scholtzan:
I created a new PR https://github.com/xi-editor/druid/pull/504 since it was easier to just copy all the changes rather than fixing all merge conflicts

Issue #455, opened 2020-01-10
Schedule new event/update pass after paint if commands are queued


cmyr:
breaking out of #448. There's currently an issue where commands issued during or after `update` won't be handled until after the next normal event arrives. We shouldn't have to wait; if there are commands to handle, we should immediately schedule a new update, and handle them.

@edwin0cheng you're welcome to take this if you're interested!

> The best way to handle this right now would be to schedule an idle call with the runloop if there are commands to process after painting is finished, that would trigger a new event/update pass.
> 
> This is something we will need to do anyway, because `lifecycle` will work like `update` and you will be able to submit commands there, but we won't run them until after this paint cycle.
> 
> If you're interested in implementing this, you would be welcome to; it would work like `timer` does, where you request a callback from the `WindowHandle` or `WinCtx`, and then I would probably add a `fn idle(&mut self)` method to `druid-shell::WinHandler`, with an implementation in `druid::win_handler` that just processes commands?
> 
> Also the mechanism for scheduling idle tasks exists, but is a bit confusing: basically you call `WindowHandle::get_idle_handle()`, which returns an `Option<IdleHandle>`; then you call `add_idle` on that, passing a closure that will be called with an `&dyn Any` that is actually your `WindowHandler` type; you can downcast that to the correct concrete type and do stuff with it.
> 
> I'm not totally sure what the rationale for this approach is; the basic idea is that you can do arbitrary things with your `WindowHandler` in that closure, like create and dispatch events, but it's definitely an awkward API.
> 

_Originally posted by @cmyr in https://github.com/xi-editor/druid/pull/448#issuecomment-573147934_

edwin0cheng:
> @edwin0cheng you're welcome to take this if you're interested!

Yeah, I would like to take this task, would try to ask more question in Zulip too :)

edwin0cheng:
Although I didn't test it , but I think  it can be closed by #467.

PR #467, opened 2020-01-19
Issue Idle event to process command after paint

edwin0cheng:
This PR adds code for scheduling an idle call with the runloop if there are commands to process after painting is finished, that would trigger a new event/update pass.

Related #455

### DIFF
--- a/druid/src/win_handler.rs
+++ b/druid/src/win_handler.rs
@@ -25,7 +25,7 @@ use log::{error, info, warn};
 use crate::kurbo::{Rect, Size, Vec2};
 use crate::piet::{Piet, RenderContext};
 use crate::shell::{
-    Application, Cursor, FileDialogOptions, MouseEvent, WinCtx, WinHandler, WindowHandle,
+    Application, Cursor, FileDialogOptions, IdleToken, MouseEvent, WinCtx, WinHandler, WindowHandle,
 };
 
 use crate::app_delegate::{AppDelegate, DelegateCtx};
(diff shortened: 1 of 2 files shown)


druid/src/win_handler.rs
```diff
@@ -40,6 +40,8 @@ use crate::{
 
 use crate::command::sys as sys_cmd;
 
+const RUN_COMMANDS_TOKEN: IdleToken = IdleToken::new(1);
```
cmyr: ugh I so wish there was a way to have a `const` function that incremented a global counter at compile time, so we could have `const fn next() -> IdleToken` and it would ensure that each token was new. I guess something like this is possible with a proc macro, but that feels like overkill.

druid/src/win_handler.rs
```diff
@@ -171,6 +173,14 @@ impl<'a, T: Data> SingleWindowState<'a, T> {
         self.do_layout(piet);
         piet.clear(self.env.get(theme::WINDOW_BACKGROUND_COLOR));
         self.do_paint(piet);
+
+        // schedule an idle call with the runloop if there are commands to process after painting is finished,
+        // that would trigger a new event/update pass.
+        if !self.command_queue.is_empty() {
+            let mut handle = self.state.handle.get_idle_handle().unwrap();
```
cmyr: I'm not sure under what conditions this can be `None`, but I would probably be defensive? Although we'd also prefer not to fail silently, so I might also log the failure :(

druid/src/win_handler.rs
edwin0cheng: We can (ab)use rust macro : 
```rust
macro_rules! idle_id!() {
    {} => { uniq_hash(file!(), column!()) }
}
```
But I think it is not the best option either ;)

druid/src/win_handler.rs
edwin0cheng: Fixed 👍

PR #470, opened 2020-01-22
Unify atomically incrementing counter impls & id types

cmyr:
We had four or five different impls of atomically incrementing counters
for generating ids. This unifies those into a single impl, and also
standardizes on `u64` as our representation for identifiers.

### DIFF
--- a/druid-shell/src/common_util.rs
+++ b/druid-shell/src/common_util.rs
@@ -15,6 +15,8 @@
 //! Common functions used by the backends
 
 use std::any::Any;
+use std::num::NonZeroU64;
+use std::sync::atomic::{AtomicU64, Ordering};
 
 /// Strip the access keys from the menu string.
 ///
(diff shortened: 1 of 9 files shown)


druid-shell/src/platform/gtk/window.rs
```diff
@@ -624,23 +624,26 @@ impl<'a> WinCtx<'a> for WinCtxImpl<'a> {
     }
 
     fn request_timer(&mut self, deadline: std::time::Instant) -> TimerToken {
-        let interval = time_interval_from_deadline(deadline);
-        let token = next_timer_id();
+        let interval = deadline
+            .checked_duration_since(std::time::Instant::now())
+            .unwrap_or_default()
+            .as_millis() as u32;
```
raphlinus: This will overflow when the deadline is more than 48 days into the future or whatnot. I don't think this is important to solve right now, but maybe a note.

druid-shell/src/common_util.rs
```diff
@@ -48,3 +50,33 @@ impl<F: FnOnce(&dyn Any) + Send> IdleCallback for F {
         (*self)(a)
     }
 }
+
+/// An incrementing counter for generating unique ids.
```
raphlinus: Probably should write down the invariant that it can't be 0, though it's pretty clear from the safety requirement.

druid-shell/src/platform/gtk/window.rs
cmyr: fair, I can be more defensive here.

PR #472, opened 2020-01-23
Druid 0.5

cmyr:
This updates our crate versions, and also updates us to the latest
major versions of our dependencies, fixing any breakage.

### DIFF
--- a/druid-shell/examples/perftest.rs
+++ b/druid-shell/examples/perftest.rs
@@ -14,7 +14,7 @@
 
 use std::any::Any;
 
-use time::get_time;
+use time::Time;
 
 use piet_common::kurbo::{Line, Rect};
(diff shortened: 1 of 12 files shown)


druid-shell/examples/perftest.rs
```diff
@@ -14,7 +14,7 @@
 
 use std::any::Any;
 
-use time::get_time;
+use time::Time;
```
raphlinus: So this is certainly a valid crate update, but I now think this is the wrong time type. It should probably be std::time::Instant, followed by `elapsed` to get the duration, then formatted through `Duration::as_secs_f64` (new as of Rust 1.38). The time crate has these same types and methods, but I don't think they add a lot of value over std.

Reworking this is probably out of scope for this PR, could be considered part of #476.

cmyr: sounds good.

Issue #479, opened 2020-01-25
Touchpad scrolling broken under wayland


Malmz:
https://sourceforge.net/p/scintilla/bugs/1901/ seem to have the same problem but tldr; Gtk under wayland only produce smooth scroll events for devices like touchpads

fishrockz:
I have reproduced the bug and #481 fixes it for me.

PR #481, opened 2020-01-26
very basic smooth scroll implementation

Malmz:
Very basic handling of smooth scroll events from GTK since it is sometimes the only scroll event sent

### DIFF
--- a/druid-shell/src/platform/gtk/window.rs
+++ b/druid-shell/src/platform/gtk/window.rs
@@ -199,7 +199,8 @@ impl WindowBuilder {
                 | EventMask::KEY_PRESS_MASK
                 | EventMask::ENTER_NOTIFY_MASK
                 | EventMask::KEY_RELEASE_MASK
-                | EventMask::SCROLL_MASK,
+                | EventMask::SCROLL_MASK
+                | EventMask::SMOOTH_SCROLL_MASK,
         );
 
         drawing_area.set_can_focus(true);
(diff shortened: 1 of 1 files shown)


Malmz:
fixes #479

druid-shell/src/platform/gtk/window.rs
```diff
@@ -332,10 +333,11 @@ impl WindowBuilder {
                         handler.wheel(Vec2::from((120.0, 0.0)), modifiers, &mut ctx);
                     }
                     ScrollDirection::Smooth => {
-                        // TODO: support smooth scrolling via scroll.get_delta and get_is_stop
-                        eprintln!(
-                            "Warning: somehow the Druid widget got a smooth scroll event"
-                        );
+                        //TODO: Look at how gtk's scroll containers implements it
+                        let (mut delta_x, mut delta_y) = scroll.get_delta();
+                        delta_x *= 120.;
```
cmyr: I appreciate this is sort of unrelated to this patch, but it might be a good time to break out this `120.0` into some const somewhere with an informative name? Do we mention anywhere how we chose this value?

Also you can multiply a `Vec2` by a float, so you should be able to just write,

```rust
let scroll_delta = scroll.get_delta() * SCROLL_SCALE_FACTOR;
```
or something similar.

fishrockz:
Tried this out on my Linux laptop and it seems to improve behaviour on i3 desktop and fix the completely broken scrolling in gnome.

druid-shell/src/platform/gtk/window.rs
Malmz: There is a comment about it a few lines up
"The magic "120"s are from Microsoft's documentation for WM_MOUSEWHEEL. They claim that one "tick" on a scroll wheel should be 120 units."

It should probably be refactored into a const, I'll get on it when i have the time.

I am also interested in looking at how gnome handles scrolling in their widgets and mabye try to implement some form of kinetic scroll, but that will require some rework of the scroll event sent to druid.

cmyr: sounds good! I have some minor concerns about gtk borrowing this from a windows implementation detail but it's probably a reasonable default. That said, it might be worth digging into that, as well...

PR #482, opened 2020-01-26
Focus widgets

cmyr:
This adds the ability for a focused widget to give focus to the next or previous focusable widget. Widgets must register to participate in this "focus chain".

This also moves to representing focus state as an `Option<WidgetId>`, uses an internal event `RouteFocusChange` to notify only the old and new focus widgets.

I quite like this pattern of routing events to specific widgets, and I think it may be useful in other places as well.

This implements this behaviour for the textbox, using tab/shift+tab for next/previous. It can be tested in the textbox example.

Currently there's no way to request a timer from the `LifeCycleCtx`, so we don't get cursor blink in textbox when it is focused in this way.

### DIFF
--- a/druid/src/core.rs
+++ b/druid/src/core.rs
@@ -51,7 +51,6 @@ pub struct WidgetPod<T: Data, W: Widget<T>> {
     old_data: Option<T>,
     env: Option<Env>,
     inner: W,
-    id: WidgetId,
 }
 
 /// Generic state for all widgets in the hierarchy.
@@ -69,8 +68,8 @@ pub struct WidgetPod<T: Data, W: Widget<T>> {
 ///
 /// [`paint`]: trait.Widget.html#tymethod.paint
 /// [`WidgetPod`]: struct.WidgetPod.html
-#[derive(Default)]
 pub(crate) struct BaseState {
+    id: WidgetId,
     layout_rect: Rect,
 
     // TODO: consider using bitflags for the booleans.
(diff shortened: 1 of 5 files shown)


druid/src/core.rs
```diff
@@ -683,17 +681,20 @@ pub struct EventCtx<'a, 'b> {
...
 /// [`lifecycle`]: widget/trait.Widget.html#tymethod.lifecycle
 /// [`register_child`]: #method.register_child
 /// [`LifeCycleCtx::register_child`]: #method.register_child
-/// [`LifeCycle::RegisterChildren`]: enum.LifeCycle.html#variant.RegisterChildren
+/// [`LifeCycle::Register`]: enum.LifeCycle.html#variant.Register
 pub struct LifeCycleCtx<'a> {
     pub(crate) command_queue: &'a mut CommandQueue,
     /// the registry for the current widgets children;
-    /// only really meaninful during a `LifeCyle::RegisterChildren` call.
+    /// only really meaninful during a `LifeCyle::Register` call.
```
raphlinus: typo: meaningful

druid/src/core.rs
```diff
@@ -944,12 +975,21 @@ impl<'a> LifeCycleCtx<'a> {
         self.widget_id
     }
 
-    /// Registers a child wildget. This should only be called
-    /// in response to a `LifeCycle::RegisterChildren` event.
+    /// Registers a child wildget.
```
raphlinus: typo: widget

druid/src/window.rs
```diff
@@ -85,6 +90,11 @@ impl<T: Data> Window<T> {
         self.root.lifecycle(ctx, event, data, env);
         self.needs_inval |= ctx.needs_inval;
         self.children_changed |= ctx.children_changed;
+
+        if let LifeCycle::Register = event {
+            self.focus_widgets = std::mem::take(&mut ctx.focus_widgets);
+            dbg!(&self.focus_widgets);
```
cmyr: ahem

druid/src/window.rs
```diff
@@ -87,6 +90,11 @@ impl<T: Data> Window<T> {
...
```
raphlinus: remove

PR #483, opened 2020-01-27
Fix either initial state

SergejJurecko:
Layout/paint gets called before update and it means false branch was visible no matter what the actual lense value was initially.

### DIFF
--- a/druid/src/widget/either.rs
+++ b/druid/src/widget/either.rs
@@ -57,11 +57,8 @@ impl<T: Data> Widget<T> for Either<T> {
     }
 
     fn lifecycle(&mut self, ctx: &mut LifeCycleCtx, event: &LifeCycle, data: &T, env: &Env) {
-        match event {
-            LifeCycle::WidgetAdded => {
-                self.current = (self.closure)(data, env);
-            }
-            _ => {}
+        if let LifeCycle::WidgetAdded = event {
+            self.current = (self.closure)(data, env);
         }
         self.true_branch.lifecycle(ctx, event, data, env);
         self.false_branch.lifecycle(ctx, event, data, env);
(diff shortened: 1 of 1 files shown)


cmyr:
Hmm, this might be related to the recent lifecycle change.

I think the correct fix is to handle `LifeCycle::WidgetAdded`, and set our `current` value there; we should always receive this event before anything else happens.

SergejJurecko:
It is related to the recent lifecycle merge. It did not happen before. The widget as it is seems to me unnecessarily brittle, as it relies on unspecified behavior to work correctly.

cmyr:
Yep, things are definitely in flux, and widgets that dynamically change their children are particularly incomplete. This however feels like simply a bug introduced in the lifecycle change, and is specific to this widget.

There are likely other problems with this widget; the idea of widgets appearing and hiding dynamically is something that isn't fully supported yet.

In any case, if you call the closure in `LifeCycle::WidgetAdded` this should continue to work like before, without needing to make the larger changes.

SergejJurecko:
Yep that works.

cmyr:
Great, thanks. Just out of curiosity, are you using this anywhere?

Something that remains to be done with lifecycle is figuring out how to interact with widgets as they are dynamically added or removed from the graph. For instance,  we only  deliver events and updates to the currently active widget, in `Either`. Should we send a `WidgetRemoved` event to the hiding widget when `Either` is changing state?  Should we send `WidgetAdded` each time? Should these be different notifications?

A more philosophical question: should the inactive widget in an `Either` be able to receive commands?

All this is to say, I'm curious to see how people are using this widget. :)

SergejJurecko:
I'm using Either as a kind of wizard. I only need two steps so it fits easily. Multiple steps I assume would be relatively easily doable with Either's containing other Either's. The only shortcoming is I can't change the window size between switches. So pages with fewer controls look empty.

I can't think of a use case for WidgetRemoved for this widget. Maybe it's just my lack of imagination.

PR #484, opened 2020-01-28
raw_d2d and druid shell integration

hwchen:
piet windows backend is being updated to use raw_d2d instead of depending on the direct2d and directwrite crates.

druid-shell also needs to be updated in order to use the raw_d2d api instead of the direct2d and directwrite crates.

There's still a few last steps to clean up, but the functionality is working. I'm opening the PR now to allow discussion over the last few steps/cleanup

TODO:
- fix warnings
- clean comments
- change `QueryInterface` calls to `cast` method.
- before release, remove patch dependency on hwchen/piet:rawd2d

(NOTE): I've rebased using `git rebase -X theirs master', which I haven't used before, so this PR may be deleted if I mucked up the rebase

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,3 +4,7 @@ members = [
     "druid-shell",
     "druid-derive",
 ]
+
+[patch.crates-io]
+piet-common = { git = "https://github.com/hwchen/piet", branch = "raw_d2d" }
+
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,3 @@ members = [
     "druid-derive",
 ]
 
-[patch.crates-io]
-piet-common = { git = "https://github.com/hwchen/piet", branch = "raw_d2d" }
-
--- a/druid-shell/src/platform/windows/dcomp.rs
+++ b/druid-shell/src/platform/windows/dcomp.rs
@@ -38,8 +38,9 @@ use winapi::um::winnt::HRESULT;
 use winapi::Interface;
 use wio::com::ComPtr;
 
-use direct2d::math::Matrix3x2F;
-use direct2d::{self, RenderTarget};
+//old
+//use direct2d::math::Matrix3x2F;
+//use direct2d::{self, RenderTarget};
 
 use log::error;
 
--- a/druid-shell/src/platform/windows/dcomp.rs
+++ b/druid-shell/src/platform/windows/dcomp.rs
@@ -38,10 +38,6 @@ use winapi::um::winnt::HRESULT;
 use winapi::Interface;
 use wio::com::ComPtr;
 
-//old
-//use direct2d::math::Matrix3x2F;
-//use direct2d::{self, RenderTarget};
-
 use log::error;
 
 use super::util::OPTIONAL_FUNCTIONS;
--- a/druid-shell/src/platform/windows/mod.rs
+++ b/druid-shell/src/platform/windows/mod.rs
@@ -46,6 +46,170 @@ pub mod window;
 //
 // Basically, go from HwndRenderTarget or DxgiSurfaceRenderTarget (2d or 3d) to a Device Context.
 // Go back up for particular needs. Move up and down using query_interface.
+use piet_common::d2d::{D2DFactory, DeviceContext};
+use std::fmt::{Debug, Display, Formatter};
+use winapi::Interface;
+use winapi::shared::windef::HWND;
+use winapi::shared::winerror::{HRESULT, SUCCEEDED};
+use winapi::um::d2d1::{D2D1_HWND_RENDER_TARGET_PROPERTIES, D2D1_RENDER_TARGET_PROPERTIES,
+                       D2D1_SIZE_U, ID2D1HwndRenderTarget, ID2D1RenderTarget};
+use winapi::um::d2d1_1::ID2D1DeviceContext;
+use winapi::um::dcommon::D2D1_PIXEL_FORMAT;
 use wio::com::ComPtr;
-pub type HwndRenderTarget = ComPtr<winapi::um::d2d1::ID2D1HwndRenderTarget>;
-pub type DeviceContext = ComPtr<winapi::um::d2d1_1::ID2D1DeviceContext>;
+
+// TODO make newtypes
+// Can probably follow the pattern of direct2d crate
+// e.g. https://github.com/Connicpu/direct2d-rs/blob/v0.1.2/src/render_target/hwnd.rs
+//
+// error and wrap, copy from d2d.rs
+//
+// for creating rendertarget, use
+// https://github.com/hwchen/piet/blob/raw_d2d/piet-direct2d/src/d2d.rs#L165 as example, but fill
+// in properties etc.
+
+#[derive(Clone)]
+pub struct HwndRenderTarget {
+    ptr: ComPtr<ID2D1HwndRenderTarget>
+}
+
+impl HwndRenderTarget {
+    pub fn create<'a>(factory: &'a D2DFactory, hwnd: HWND, width: u32, height: u32) -> Result<Self, Error> {
+        // hardcode
+        // - RenderTargetType::Default
+        // - AlphaMode::Unknown
+        let rt_props = DEFAULT_PROPS;
+        let mut hwnd_props = DEFAULT_HWND_PROPS;
+
+        hwnd_props.pixelSize.width = width;
+        hwnd_props.pixelSize.height = height;
+
+        // now build
+        unsafe {
+            let mut ptr = std::ptr::null_mut();
+            let hr = (*factory.get_raw()).CreateHwndRenderTarget(
+                &rt_props,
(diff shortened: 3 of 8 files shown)


Cargo.toml
```diff
@@ -4,3 +4,7 @@ members = [
     "druid-shell",
     "druid-derive",
 ]
+
+[patch.crates-io]
+piet-common = { git = "https://github.com/hwchen/piet", branch = "raw_d2d" }
```
raphlinus: I know I already mentioned it, but just to make sure it's in the record, this will change to piet 0.0.9 as soon as that release happens. We'll coordinate the releases so they happen in succession.

druid-shell/src/platform/windows/dcomp.rs
```diff
@@ -38,8 +38,9 @@ use winapi::um::winnt::HRESULT;
 use winapi::Interface;
 use wio::com::ComPtr;
 
-use direct2d::math::Matrix3x2F;
-use direct2d::{self, RenderTarget};
+//old
```
raphlinus: Remove

druid-shell/src/platform/windows/mod.rs
```diff
@@ -26,3 +26,194 @@ pub mod runloop;
...
+//
+// A Render Target creates resources for drawing and performs drawing operations.
+//
+// - ID2D1HwndRenderTarget objects render content to a window.
+// - ID2D1DCRenderTarget objects render to a GDI device context.
+// - bitmap render target objects render to off-screen bitmap.
+// - DXGI render target objects render to  a DXGI surface for use with Direct3D.
+//
+// https://docs.microsoft.com/en-us/windows/win32/direct2d/devices-and-device-contexts
+// A Device Context, ID2D1DeviceContext, is used for windows 8 and higher. Render Target
```
raphlinus: This might be a misunderstanding. DeviceContext is available as of Windows 7  with platform update, which the minimum compatibility target for druid. We are not making an effort do to RenderTarget only (ie if the downcast to DC fails).

druid-shell/src/platform/windows/mod.rs
```diff
@@ -26,3 +26,194 @@ pub mod runloop;
...
+    ptr: ComPtr<ID2D1HwndRenderTarget>
+}
+
+impl HwndRenderTarget {
+    pub fn create<'a>(factory: &'a D2DFactory, hwnd: HWND, width: u32, height: u32) -> Result<Self, Error> {
+        // hardcode
+        // - RenderTargetType::Default
+        // - AlphaMode::Unknown
+        let rt_props = DEFAULT_PROPS;
+        let mut hwnd_props = DEFAULT_HWND_PROPS;
```
raphlinus: Pretty sure you need to set the hwnd here. I think it might make sense to write out the `D2D1_HWND_RENDER_TARGET_PROPERTIES { }` struct rather than setting fields, it means there's conscious thought to set all the fields.

Seeing the same pattern below, I think I know what's going on, a builder pattern is nice when you're exposing an API, but when the struct being built is in the same module as the use, it's less code (and imho simpler) just to write it as a struct.

druid-shell/src/platform/windows/mod.rs
```diff
@@ -26,3 +26,194 @@ pub mod runloop;
...
+        Self { ptr }
+    }
+
+    pub unsafe fn from_raw(raw: *mut ID2D1HwndRenderTarget) -> Self {
+        HwndRenderTarget {
+            ptr: ComPtr::from_raw(raw),
+        }
+    }
+
+    pub unsafe fn get_raw(&self) -> *mut ID2D1HwndRenderTarget {
```
raphlinus: Doesn't need to be unsafe.

druid-shell/src/platform/windows/mod.rs
```diff
@@ -26,3 +26,194 @@ pub mod runloop;
...
+                Err(hr.into())
+            }
+        }
+    }
+
+    pub unsafe fn from_ptr(ptr: ComPtr<ID2D1HwndRenderTarget>) -> Self {
+        Self { ptr }
+    }
+
+    pub unsafe fn from_raw(raw: *mut ID2D1HwndRenderTarget) -> Self {
```
raphlinus: Might be cleaner to use `from_ptr()` here.

druid-shell/src/platform/windows/mod.rs
```diff
@@ -26,3 +26,194 @@ pub mod runloop;
...
+        DxgiSurfaceRenderTarget {
+            ptr: ComPtr::from_raw(raw),
+        }
+    }
+
+    pub unsafe fn get_raw(&self) -> *mut ID2D1RenderTarget {
+        self.ptr.as_raw()
+    }
+
+    // TODO use https://docs.rs/wio/0.2.2/i686-pc-windows-msvc/wio/com/struct.ComPtr.html#method.cast
```
raphlinus: Sounds like this TODO is resolved?

druid-shell/src/platform/windows/paint.rs
```diff
@@ -110,16 +146,19 @@ pub(crate) unsafe fn create_render_target_dxgi(
...
-        self.d2d_factory
-    }
 
-    /// Return the raw Direct2D RenderTarget for this painting context. Note: it's possible
-    /// this will be wrapped to make it easier to port.
-    pub fn render_target(&mut self) -> &mut GenericRenderTarget {
-        self.render_target
+/// Casts hwnd variant to DeviceTarget
+///
+/// TODO: investigate whether there's a better way to do this.
```
raphlinus: Right, here I think you want a ComPtr cast, but this logic is *probably* correct.

druid-shell/src/platform/windows/util.rs
```diff
@@ -39,7 +39,9 @@ use winapi::um::winbase::{FILE_TYPE_UNKNOWN, STD_ERROR_HANDLE, STD_OUTPUT_HANDLE
 use winapi::um::wincon::{AttachConsole, ATTACH_PARENT_PROCESS};
 use winapi::um::winnt::{FILE_SHARE_WRITE, GENERIC_READ, GENERIC_WRITE};
 
-use direct2d::enums::DrawTextOptions;
+// old
```
raphlinus: Remove

druid-shell/src/platform/windows/window.rs
```diff
@@ -39,9 +39,20 @@ use winapi::um::winnt::*;
 use winapi::um::winuser::*;
 use winapi::Interface;
 
-use direct2d;
-use direct2d::math::SizeU;
-use direct2d::render_target::{GenericRenderTarget, HwndRenderTarget, RenderTarget};
+// old
```
raphlinus: Remove these comments

druid-shell/src/platform/windows/paint.rs
```diff
@@ -35,48 +35,84 @@ use winapi::um::dcommon::*;
...
+
+// end new
 use super::error::Error;
 use super::util::as_result;
 
 /// Context for painting by app into window.
 pub struct PaintCtx<'a> {
-    pub(crate) d2d_factory: &'a direct2d::Factory,
-    pub(crate) render_target: &'a mut GenericRenderTarget,
+    pub(crate) d2d_factory: &'a D2DFactory,
```
raphlinus: It seems really strange to me these fields aren't getting used. I don't know what's going on.

raphlinus:
In testing on an actual device, it looks like the dxgi render target is working, but hwnd not so much, here's the error:

```
2020-01-27 17:28:16,469 ERROR [druid_shell::platform::windows::paint] Creating hwnd render target failed: hresult 80070578
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src\libcore\option.rs:378:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
error: process didn't exit successfully: `target\debug\examples\calc.exe` (exit code: 0xc0000409, STATUS_STACK_BUFFER_OVERRUN)
```

This is probably related to not setting hwnd.

hwchen:
how do you test hwnd vs. dxgi targets?

raphlinus:
> how do you test hwnd vs. dxgi targets?

One is to resize the window. But to do it more systematically, you can make the dxgi creation fail.

hwchen:
Thanks, tried resizing first and that worked perfectly.

druid-shell/src/platform/windows/paint.rs
hwchen: From what I can tell, `PaintCtx` isn't used anywhere. There's another `PaintCtx` in `druid`, but in `druid-shell` there's no `PaintCtx` impls, and no direct construction.

The `create_render_target` and `create_render_target_dxgi` fns use those fields as args, but are not methods on `PaintCtx`.

Not sure if this clarifies anything?

hwchen: Left a TODO and commented out `PaintCtx` for now.

PR #486, opened 2020-01-28
Add a basic external event API

### LINKED ISSUE #471: External events
How would one react to some event not triggered by user input, like a background thread sending a message over a channel.

cmyr:
This adds API that allows the user to get a handle before
application launch that can be safely cloned and moved between
threads, and with which commands can be submitted back to
the running application.

closes #471


futurepaul:
Is there any limitation in this implementation that would make asynchronous round trips difficult? Like, send a url from the UI to a thread that makes a web request, receive the response back as an external event at some later time?

cmyr:
@futurepaul Currently this is only implemented in one direction, but I _think_ you should be able to create another channel in your thread and send that over to druid, and let your code use it to send messages back. You would have to write some kind of event loop or something to handle the details, though. So: bidirectional messaging isn't built-in, but should be possible?

PR #491, opened 2020-01-29
Refactoring

adamnemecek:


nilsmartel:
> Thanks for the PR! I'm not personally sure I necessarily prefer using Self in all cases? I think it's often more obvious to see the return type explicitly, instead of needing to figure out what type I'm in an impl block of.

Also the available language servers for rust (rust-analyser/rls) can handle explicitly typed out cases better. Often you will want to see. a functions definition. Since they will show you only a functions signature you can deduce more information, if the return type is written out explicitly.

Just a minor consideration

cmyr:
Thanks for this, but we're going to decline for the time being; if this convention becomes standard in the rust ecosystem we are happy to revisit, but for the time being the advantages do not clearly outweigh the disadvantages.

PR #501, opened 2020-01-31
Fixup WidgetId reuse logic

cmyr:
This fixes an issue revealed in the process of implementing a
testing harness.

The issue is basically that I had gotten a bit clever with the
widget id stuff. Specifically, certain widgets that manage a
single child using a WidgetPod (Padding and Align, specifically)
had attempted to be transparent for the purposes of widget id;
they would reuse the id of their inner widget, if one existed.
The problem with this was that you could end up with two
WidgetPods that had the same id, and so there would be no way
to address the inner pod.

So this reverts that fanciness.


futurepaul:
This is working for me, and I'll wait on this to merge #496.

PR #504, opened 2020-02-02
Z-order painting support

scholtzan:
This addresses https://github.com/xi-editor/druid/issues/430 and replaces https://github.com/xi-editor/druid/pull/466

This requires https://github.com/linebender/piet/pull/117 to have access to `current_transform()` which is necessary to keep track of applied transformations.

One thing that is missing here (also because I'm not 100% sure) is keeping track of clipping. Right now previous clipping is not applied to z-order painting. It would be a bit trickier since `PaintCtx` would need to keep track of clipping.

One other thing that I was thinking about was thinking about is whether it would make sense to allow negative `z_index`es to allow painting below content?

### DIFF
--- a/druid/src/contexts.rs
+++ b/druid/src/contexts.rs
@@ -96,7 +96,7 @@ pub struct LayoutCtx<'a, 'b: 'a> {
 }
 
 /// Z-order paint operations with transformations.
-pub struct ZOrderPaintOp {
+pub(crate) struct ZOrderPaintOp {
     pub z_index: u32,
     pub paint_func: Box<dyn FnOnce(&mut PaintCtx) + 'static>,
     pub transform: Affine,
@@ -114,7 +114,7 @@ pub struct PaintCtx<'a, 'b: 'a> {
     pub render_ctx: &'a mut Piet<'b>,
     pub window_id: WindowId,
     /// The z-order paint operations.
-    pub z_ops: Vec<ZOrderPaintOp>,
+    pub(crate) z_ops: Vec<ZOrderPaintOp>,
     /// The currently visible region.
     pub(crate) region: Region,
     pub(crate) base_state: &'a BaseState,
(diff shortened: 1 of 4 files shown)


druid/src/contexts.rs
```diff
@@ -103,6 +105,13 @@ pub struct LayoutCtx<'a, 'b: 'a> {
     pub(crate) window_id: WindowId,
 }
 
+/// Z-order paint operations with transformations.
+pub struct ZOrderPaintOp {
```
cmyr: Does this need to be public? It feels like an implementation detail. Maybe just `pub(crate)`?

druid/src/contexts.rs
```diff
@@ -528,11 +539,29 @@ impl<'a, 'b: 'a> PaintCtx<'a, 'b> {
...
             region: region.into(),
         };
-        f(&mut child_ctx)
+        f(&mut child_ctx);
+        self.z_ops.append(&mut child_ctx.z_ops);
+    }
+
+    /// Allows to specify order for paint operations.
+    ///
+    /// Larger `z_idx` indicate that an operation will be executed later.
```
cmyr: It feels like this is documenting the implementation more than the interface, but I've had trouble coming up with anything better, so I'm happy enough with this; at least it is clear.

edit: also the comment says `z_idx` but the code says `z_index` 🤷‍♀

druid/src/window.rs
```diff
@@ -139,6 +141,25 @@ impl<T: Data> Window<T> {
     pub fn paint(&mut self, paint_ctx: &mut PaintCtx, data: &T, env: &Env) {
         let visible = Rect::from_origin_size(Point::ZERO, self.size);
         paint_ctx.with_child_ctx(visible, |ctx| self.root.paint(ctx, data, env));
+
+        paint_ctx.z_ops.sort_by_key(|k| k.z_index);
```
cmyr: sort of silly but I might do the sorting after we take the items out of the context? It just feels very slightly easier to read.

druid/src/window.rs
```diff
@@ -139,6 +141,25 @@ impl<T: Data> Window<T> {
...
+
+        let z_ops = mem::replace(&mut paint_ctx.z_ops, Vec::new());
```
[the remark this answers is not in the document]
scholtzan: oh sweet!

cmyr:
Oh: about the negative z-index, we've decided against that for now, because it makes everything more complicated; all painting would have to be done in closures, and it isn't clear that it's worth the hassle.

PR #506, opened 2020-02-03
Fix FocusChanged routing

cmyr:
~This is based off of #499, which should go in first.~

This fixes *two* issues in focus routing; the first is that we weren't correctly checking children when deciding to recurse, and the second is that we were converting from `RouteFocusChanged` to `FocusChanged` without correctly checking if the current widget was actually involved in the change. 😬

### DIFF
--- a/druid/src/core.rs
+++ b/druid/src/core.rs
@@ -423,15 +423,23 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
             LifeCycle::HotChanged(_) => false,
             LifeCycle::RouteFocusChanged { old, new } => {
                 self.state.request_focus = None;
-                let this_changed = old.map(|_| false).or_else(|| new.map(|_| true));
+
+                let this_changed = if *old == Some(self.state.id) {
+                    Some(false)
+                } else if *new == Some(self.state.id) {
+                    Some(true)
+                } else {
+                    None
+                };
+
                 if let Some(change) = this_changed {
                     let event = LifeCycle::FocusChanged(change);
                     self.inner.lifecycle(ctx, &event, data, env);
                     false
                 } else {
-                    //FIXME: I think this is a bug? we should be checking `self.state.children`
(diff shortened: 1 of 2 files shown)


druid/src/core.rs
```diff
@@ -423,15 +423,25 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
             LifeCycle::HotChanged(_) => false,
             LifeCycle::RouteFocusChanged { old, new } => {
                 self.state.request_focus = None;
-                let this_changed = old.map(|_| false).or_else(|| new.map(|_| true));
+
+                let this_changed = if *old == Some(self.state.id) {
```
Zarenor: I think this respelling is much better. It took me several minutes tease apart the old version the first time.

druid/src/core.rs
```diff
@@ -423,15 +423,25 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
...
                     let event = LifeCycle::FocusChanged(change);
                     self.inner.lifecycle(ctx, &event, data, env);
                     false
                 } else {
-                    //FIXME: I think this is a bug? we should be checking `self.state.children`
-                    old.map(|id| ctx.children.contains(&id)).unwrap_or(false)
-                        || new.map(|id| ctx.children.contains(&id)).unwrap_or(false)
+                    old.map(|id| self.state.children.contains(&id))
+                        .unwrap_or(false)
+                        || new
```
Zarenor: Minor style nit: The or operator here tripped me up ("Why is there a closure here?!")
Can we spell this
`[...]
.unwrap_or(false)
||
new.map(...)
[...]`
Or will rustfmt move it back this way?

cmyr: ah interesting, I hadn't thought to see that as a closure. This formatting is rustfmt's doing, but I can rewrite this using `or_else`, which should be clearer.

Zarenor: Thanks. I only just now realized that GH discarded my newlines in what I'd written. Edited above to what I intended.

PR #511, opened 2020-02-05
Make WindowConnected be an Event, unify WidgetAdded and Register

cmyr:
Based on #510, which should go in first.

WindowConnected:

The distinction between Event events and LifeCycle events
is best expressed in terms of their source. Event events originate
in the platform; they are things like mouse movement, timers,
and key events.

LifeCycle events are things that are generated by the druid framework,
generally in the process of handling some other event or state change.

Under this framework, WindowConnected is more like an Event event than
like a LifeCycle event, and moving it there lets us delete some code,
which is generally a good sign.

WidgetAdded and Register:

In addition to that change, this also unifies the Register and WidgetAdded
events. These were pretty much always sent at the same time, and unifying
them should reduce confusion: from now on there is only WidgetAdded,
and that is the event that widgets need to handle to do their initial setup.

### DIFF
--- a/druid/src/event.rs
+++ b/druid/src/event.rs
@@ -46,6 +46,16 @@ use crate::{Command, Target, WidgetId};
 /// [`WidgetPod`]: struct.WidgetPod.html
 #[derive(Debug, Clone)]
 pub enum Event {
+    /// Sent to all widgets in a given window when that window is first instantiated.
+    ///
+    /// This should always be the first `Event` received, although widgets will
+    /// receive [`LifeCycle::WidgetAdded`] first.
+    ///
+    /// Widgets should handle this event if they need to do some addition setup
+    /// when a window is first created.
+    ///
+    /// [`LifeCycle::WidgetAdded`]: enum.LifeCycle.html#variant.WidgetAdded
+    WindowConnected,
     /// Called on the root widget when the window size changes.
     ///
     /// Discussion: it's not obvious this should be propagated to user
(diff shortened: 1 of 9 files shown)


druid/src/widget/textbox.rs
```diff
@@ -336,8 +336,10 @@ impl Widget<String> for TextBox {
 
     fn lifecycle(&mut self, ctx: &mut LifeCycleCtx, event: &LifeCycle, _data: &String, _env: &Env) {
         match event {
-            LifeCycle::WidgetAdded => ctx.invalidate(),
-            LifeCycle::Register => ctx.register_for_focus(),
+            LifeCycle::WidgetAdded => {
+                ctx.invalidate();
```
raphlinus: I don't think we need this. If removing it causes behavior changes, we should probably understand why. That said, it's probably not doing any harm.

druid/src/event.rs
```diff
@@ -129,18 +139,10 @@ pub enum LifeCycle {
...
+    /// Widgets should handle this event in order to do any initial setup.
     ///
-    /// This is sent after `WidgetAdded`. Widgets should handle this event if
-    /// they need to do some addition setup when a window is first created.
-    WindowConnected,
-    /// Sent to a widget when it or any of its children have changed.
-    ///
-    /// This event is used to register widgets with the framework, as well
-    /// as to register widgets to participate in certain framework features.
+    /// In addition, to setup, this event is also used by the framework to
```
raphlinus: "In addition to setup," (too much comma)

PR #513, opened 2020-02-06
Separate paint rects from layout rects

### LINKED ISSUE #401: Paint bounds
At the moment, there's no enforcement of where widgets can paint, so there's nothing stopping them from painting outside their layout bounds. However, container widgets might skip painting a widget when the paint region doesn't overlap the widget's layout bounds, leading to potential visible artifacts (the most likely is a widget "popping" into visibility on scrolling).

A much more systematic approach would be to track the actual paint bounds of a widget. I propose a new `set_paint_bounds` method on `LayoutCtx`. If this function is not called during a layout, then the paint bounds are set to the layout bounds (thus, this function is opt-in). Further, during layout we should keep track of the union of the paint bounds of all the children, and make that available. A simple container (that does no painting of its own) will call `set_paint_bounds` to this value.

An interesting question is whether an `update` should also be able to update the paint bounds (without also changing the layout bounds). However, it is difficult to figure out the logic of how this should propagate up, so for simplicity it's probably best to request layout (#400), and then in the future we can be finer grained about minimizing the work when the layout bounds are not actually different.

cmyr:
This is based off of #512, which should be merged first.

This introduces the idea of a 'paint rect', which is the region
that a widget wishes to paint inside. In the general case this is
equivalent to the widget's layout rect; the exceptions are cases
where widgets explicitly wish to paint outside of their bounds.

the paint region is specified as Insets relative to the layout rect.
Widgets that need an explicit paint region specify that region when
handling the Widget::layout method.

closes #401

### DIFF
--- a/druid/src/core.rs
+++ b/druid/src/core.rs
@@ -19,7 +19,7 @@ use std::collections::VecDeque;
 use log;
 
 use crate::bloom::Bloom;
-use crate::kurbo::{Affine, Point, Rect, Shape, Size};
+use crate::kurbo::{Affine, Insets, Point, Rect, Shape, Size};
 use crate::piet::RenderContext;
 use crate::{
     BoxConstraints, Command, Data, Env, Event, EventCtx, LayoutCtx, LifeCycle, LifeCycleCtx,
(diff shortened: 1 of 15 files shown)


druid/src/core.rs
```diff
@@ -166,13 +170,65 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
...
+    /// Given a parents layout size, determine the appropriate paint `Insets`
+    /// for the parent.
+    ///
+    /// This is a convenience method to be used from the [`layout`] method
+    /// of a `Widget` that manages a child; it allows the parent to correctly
+    /// propogate a child's desired paint rect, if it extends beyond the bounds
+    /// of the parent's layout rect.
+    ///
+    /// [`layout`]: widget/trait.Widget.html#tymethod.layout
+    pub fn compute_parent_paint_rect(&self, parent_size: Size) -> Insets {
```
cmyr: this should be `compute_parent_paint_insets`, and we can link to `Insets`.

PR #515, opened 2020-02-06
Fix incorrectly propogating is_hot to parents

### LINKED ISSUE #493: Scroll widget gets hot status when scrollbar fades
In the list example, if you hover over one of the lists scrollbar and then move over to the other one before it fades, both lists will have hot status and compete for scroll events.

Screencast from 2020-01-29 12_56_52

cmyr:
This was a simple bug; we were treating `is_hot` as something
that should be propogated from children to parents, where it is
actually calculated directly for each widget.

fixes #493

### DIFF
--- a/druid/src/core.rs
+++ b/druid/src/core.rs
@@ -554,7 +554,6 @@ impl BaseState {
         self.needs_inval |= child_state.needs_inval;
         self.request_anim |= child_state.request_anim;
         self.request_timer |= child_state.request_timer;
-        self.is_hot |= child_state.is_hot;
         self.has_active |= child_state.has_active;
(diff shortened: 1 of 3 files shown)


cmyr:
Oddly not, this is long standing; I assumed I'd introduced it when I added `merge_up`, but I actually seem to have just copy-pasted it.

cmyr:
actually now I'm suspicious of this patch, lemme write a test 🤓

cmyr:
Okay, updated this to do a bunch of testing and I'm feeling more confident about it; I had to include #517 in order to write the tests I wanted. 🤷‍♂

PR #516, opened 2020-02-06
Don't invalidate fade on scrollbar hover

### LINKED ISSUE #492: Sometimes scrollbars do not fade
When leaving the widget from the side of the scrollbar it does not fade. reset_scrollbar_fade can only called from MouseMoved events. Fade should also be triggered when hot status changes to false.

An example of this can be seen in the list example.

cmyr:
Previously we attempted to always have the scrollbars be visible if
the mouse is over them, which would cause them to never fade if
the mouse moved directly from a scrollbar out of the window.

This fixes that by allowing the scrollbars to fade even if hovered,
which feels more natural to me in any case.

fixes #492

### DIFF
--- a/druid/src/widget/scroll.rs
+++ b/druid/src/widget/scroll.rs
@@ -322,9 +322,7 @@ impl<T: Data, W: Widget<T>> Widget<T> for Scroll<T, W> {
                     } else {
                         self.scroll_bars.hovered = BarHoveredState::Horizontal;
                     }
-
-                    self.scroll_bars.opacity = env.get(theme::SCROLL_BAR_MAX_OPACITY);
-                    self.scroll_bars.timer_id = TimerToken::INVALID; // Cancel any fade out in progress
+                    self.reset_scrollbar_fade(ctx, &env);
                     ctx.invalidate();
                 }
                 Event::MouseDown(event) => {


ForLoveOfCats:
I would argue that this impedes usability. However that heads into the "should scrollbars be overlays at all" territory.

At the very least this should be modified show the scrollbars on mouse down before merging. With just these changes if the mouse is over the bar and starts dragging when they have faded they remain transparent.

cmyr:
mm, good call, it definitely shouldn't fade while dragging.

I don't mind if it fades when the mouse is overtop but unmoving; it will reappear immediately on the next mouse movement.

ForLoveOfCats:
It also needs to become visible again at the beginning of a drag.

I also feel like this is a hacky solution to the issue, wouldn't it beneficial to also start the fade once the scroll widget is no longer focused as suggested in the original issue? I don't know if that would get triggered on mouse out of window but that falls under tracking the mouse outside the border which is platform specific.

cmyr:
Yes, that could make sense, it's mostly just that there is currently no way to start a timer from a lifecycle event, and I was feeling too lazy to implement that, but you're right that might be the best approach.

cmyr:
Okay I think I've got a real solution to this, in #539. @ForLoveOfCats if you'd like to take a peek at that and make sure I'm not misunderstanding the rationale behind the original, that would be great.

ForLoveOfCats:
I was just about to comment on that new PR :smile:

PR #521, opened 2020-02-06
Swap horizontal and vertical descriptors. Fix up documentation to match.

Zarenor:
As discussed on Zulip - it appears that convention when naming splits is the opposite of what I had - which way the splitter runs, rather than which dimension it divides. This adjusts the `Split`'s functionality to match that.

Users will need to swap their invocations to get the same layouts.


cmyr:
okay, I think we can go ahead and get this in.

PR #528, opened 2020-02-08
Simple layout invalidation

### LINKED ISSUE #400: Medium grained layout invalidation
Currently we're calling layout every paint cycle, which is potentially quite wasteful. As a longer term project, we'd like truly fine grained invalidation of layout, so that we're careful to only do a layout on a subtree, and we're also careful not to propagate a relayout if the specific widget hasn't changed size. However, all that logic is fairly complicated.

What I'm proposing to do now is add a `request_layout()` method call on both `event` and `update` contexts, that ends up setting a single request bit for the entire window. Then, we skip the `layout` phase if no widgets have requested layout, and if there is any request, we compute layout from scratch starting at the root. This is a very simple change that should get rid of most of the needless layout recomputation.

cmyr:
This adds minimal layout invalidation, where we will not call
layout() unless the widget tree has changed, the  window has
been resized, or a widget has explicitly called ctx.request_layout().

closes #400

### DIFF
--- a/druid/src/window.rs
+++ b/druid/src/window.rs
@@ -261,8 +261,13 @@ impl<T: Data> Window<T> {
         data: &T,
         env: &Env,
     ) {
+        // FIXME: only do AnimFrame if root has requested_anim?
         self.lifecycle(queue, &LifeCycle::AnimFrame(0), data, env);
-        self.layout(piet, data, env);
+
+        if self.root.state().needs_layout {
+            self.layout(piet, data, env);
+        }
+
         piet.clear(env.get(crate::theme::WINDOW_BACKGROUND_COLOR));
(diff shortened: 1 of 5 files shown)


druid/src/window.rs
```diff
@@ -261,8 +261,14 @@ impl<T: Data> Window<T> {
         data: &T,
         env: &Env,
     ) {
+        // FIXME: only do AnimFrame if root has requested_anim?
         self.lifecycle(queue, &LifeCycle::AnimFrame(0), data, env);
-        self.layout(piet, data, env);
+
+        dbg!(self.root.state().needs_layout);
```
cmyr: oops!

raphlinus: Debug spam.

PR #533, opened 2020-02-08
[Windows]: Initialize on window creation event

msiglreith:
Do all initialize on WM_CREATE instead of the context where the window gets created.
Avoids issues in case of more than WM_CREATE gets issued on window creation (e.g WM_SIZE) with other window styles. This results in a panic as not everything is initialized at that time.

Noticed while trying to implement https://github.com/xi-editor/druid/issues/427.

### DIFF
--- a/druid-shell/src/platform/windows/window.rs
+++ b/druid-shell/src/platform/windows/window.rs
@@ -155,6 +155,7 @@ struct MyWndProc {
     d2d_factory: D2DFactory,
     dwrite_factory: DwriteFactory,
     state: RefCell<Option<WndState>>,
+    present_strategy: PresentStrategy,
 }
 
 /// The mutable state of the window.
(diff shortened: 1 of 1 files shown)


raphlinus:
Quick response: I see the point of this and want to incorporate the logic, but there's an internal layering of the low level window creation and the actual wndproc handling, and this breaks it. Is there any possibility this could be done in WM_CREATE in MyWndProc instead of the C window procedure?

The internal layering is not enforced, but the [win_win](https://github.com/xi-editor/druid/tree/win_win) branch would move this out to a separate crate. We don't have consensus to make that an explicit dependency, but I would like to preserve the layering (and also fix the soundness issue, for which I've been meaning to do a separate PR, which would conflict with this one).

Thanks for digging into this, I know it's tricky.

msiglreith:
sure, new try (:

cmyr:
@msiglreith this collided with #419, want to rebase or merge and then we'll get it in?

msiglreith:
Seems to build \o/

PR #539, opened 2020-02-11
Fix scrollbar hit-testing

### LINKED ISSUE #492: Sometimes scrollbars do not fade

cmyr:
So this should *actually* fix #492? The bug was subtle: when testing
if the mouse was hovered over the scrollbar, we were not checking
that it did not extend past the scrollbar to be outside the widget,
presuming (I assume) that we would not receive such events; but this
led to us incorrectly supressing the last mouse move event (the event
that moved the mouse outside of our bounds) and so not correctly
fading the scrollbars or updating our children (so they could mark
themselves as non-hot).

### DIFF
--- a/druid/src/widget/scroll.rs
+++ b/druid/src/widget/scroll.rs
@@ -256,19 +256,19 @@ impl<T, W: Widget<T>> Scroll<T, W> {
     fn point_hits_vertical_bar(&self, viewport: Rect, pos: Point, env: &Env) -> bool {
         if viewport.height() < self.child_size.height {
             let bounds = self.calc_vertical_bar_bounds(viewport, &env);
-            return pos.y > bounds.y0 && pos.y < bounds.y1 && pos.x > bounds.x0;
+            bounds.contains(pos)
+        } else {
+            false
         }
-
-        false
     }
 
     fn point_hits_horizontal_bar(&self, viewport: Rect, pos: Point, env: &Env) -> bool {
         if viewport.width() < self.child_size.width {
             let bounds = self.calc_horizontal_bar_bounds(viewport, &env);
-            return pos.x > bounds.x0 && pos.x < bounds.x1 && pos.y > bounds.y0;
+            bounds.contains(pos)
+        } else {
+            false
         }
-
-        false
     }
 }
 


ForLoveOfCats:
I don't think this is the cause of the original issue, however the current behavior in this path *is* incorrect which is exacerbating the issue.

The actual issue is still that when the mouse leaves the window border the scrollbar is not starting the fade if the bars are in the hovered or held state. Whether or not this is also present on mouse out of the widget bounds but still within the window bounds I do not know.

When I originally wrote this code in question here I was making sure that if the mouse was on the edge side of the scrollbar's padding that it would still trigger which is *very* important. However looking at it now, I can see that it is extending the check all the way to the window border instead of the widget border which is the incorrect behavior.

Merging this PR would simply hide the issue in one specific case (which is still not foolproof) by relying on the mouse hitting the empty area between the bar and the edge before exiting thus removing the hovered state. Not only is that fragile (what if the mouse is moving with speed?) that would also not help when dragging as the scrollbars stay visible the entire time when dragging (as they should) and there is still not yet logic to get mouse events when the pointer is outside the window borders.

If it is okay with you I will update this PR's singular commit to have the original behavior but with respect to the widget border (which is important for UX) and the mouse event's issue be pursued separately as there is no way that the scroll widget is the only one effected by this issue specifically with regards to the window border.

ForLoveOfCats:
On second thought the behavior fix mentioned would work best as a new PR. I'll do that now

cmyr:
I'm not sure that issue remains; if the mouse moves from inside to outside of a widget, even if it is outside of the window, we still deliver that move event, although we suppress subsequent ones.

It's possible that there is a separate bug in gtk where we are not receiving mouse events outside of the window, but I've played around just now and this is working on mac.

cmyr:
in any case feel free to try out any alternatives. The modified `list` example in #537 might be useful for experimentation, because it lets you visually see when widgets are hot.

cmyr:
Okay yea I can verify that the missing mouse events outside the window is a GTK issue.

ForLoveOfCats:
yeah I've been fighting with Google Drive for the past few minutes to get a decent video of reproduction to put on the original issue

cmyr:
Opened #541 for the "mouse outside the window" issue. I'm now pretty sure that this current patch, or some variation of it, is correct, but we will still have not-fading problems until we fix #541.

ForLoveOfCats:
Incorrect behavior fixed in https://github.com/xi-editor/druid/pull/542

PR #542, opened 2020-02-11
Keep scrollbar point hitbox within widget bounds

ForLoveOfCats:
I could have just added `SCROLL_BAR_PAD` to the values but recalculating them felt less hacky. Input on a better way to do it is welcome


cmyr:
hmm, just out of curiousity, what would happen if `calc_xxx_bar_bounds` went up to the edge of the widget?

ForLoveOfCats:
The scrollbar paint uses the bar bounds calculation to construct the rounded rect for the paint operation

PR #544, opened 2020-02-12
Remove WinCtx type

cmyr:
This removes `WinCtx` and moves all of its functionality over to `WindowHandle`.

yolo?

### DIFF
--- a/druid-shell/src/platform/windows/window.rs
+++ b/druid-shell/src/platform/windows/window.rs
@@ -59,7 +59,7 @@ use crate::dialog::{FileDialogOptions, FileDialogType, FileInfo};
 use crate::keyboard::{KeyEvent, KeyModifiers};
 use crate::keycodes::KeyCode;
 use crate::mouse::{Cursor, MouseButton, MouseEvent};
-use crate::window::{IdleToken, Text, TimerToken, WinCtx, WinHandler};
+use crate::window::{IdleToken, Text, TimerToken, WinHandler};
 
 extern "system" {
     pub fn DwmFlush();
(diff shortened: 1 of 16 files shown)


druid/src/contexts.rs
```diff
@@ -71,8 +69,8 @@ pub struct LifeCycleCtx<'a> {
 /// in the widget's appearance, to schedule a repaint.
 ///
 /// [`request_paint`]: #method.request_paint
-pub struct UpdateCtx<'a, 'b: 'a> {
-    pub(crate) text_factory: &'a mut Text<'b>,
+pub struct UpdateCtx<'a> {
+    //pub(crate) text_factory: &'a mut Text<'b>,
```
raphlinus: Probably this should be removed, it sounds like all access is now through the `WinHandle`.

druid/src/core.rs
```diff
@@ -569,7 +568,7 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
 
         let mut child_ctx = UpdateCtx {
             window: ctx.window,
-            text_factory: ctx.text_factory,
+            //text_factory: ctx.text_factory,
```
raphlinus: Same.

druid-shell/src/platform/windows/window.rs
```diff
@@ -103,9 +102,8 @@ pub enum PresentStrategy {
     FlipRedirect,
 }
 
-#[derive(Default)]
 pub struct WindowHandle {
-    // Note: this clone of the dwrite factory might move into WinCtxImpl.
+    //FIXME: why is this optional?
```
raphlinus: Looks like it's not needed. A long time ago, there was probably a two-stage initialization, where window state structures were created before window creation proper, then filled in after.

Note that when https://github.com/linebender/piet/pull/139 lands, we should be able to just derive `Clone` for `WindowHandle` here.

PR #549, opened 2020-02-12
[win] Move connect() & connected() into WM_CREATE

### LINKED ISSUE #548: Cannot set a timer when handling WinHandler::connect (windows)
At the point where the windows backend calls `WinHandler::connect`, the window has not been fully instantiated(?). Timers requested during this event (`Event::WindowConnected` in druid) never fire.

Noticed this because the `identity` example doesn't work correctly on windows.

cmyr:
This ensures our handle isn't null when we send it to the WinHandler.

fixes #548

### DIFF
--- a/druid-shell/src/platform/windows/window.rs
+++ b/druid-shell/src/platform/windows/window.rs
@@ -269,13 +269,9 @@ impl MyWndProc {
 }
 
 impl WndProc for MyWndProc {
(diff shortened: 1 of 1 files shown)


cmyr:
@msiglreith I think this makes sense? I'm not especially familiar with winapi.

PR #550, opened 2020-02-12
[gtk] Fix idle handler not working across threads

cmyr:
This is an issue in the generated code for gdk. Simple enough
work around.

upstream issue: https://github.com/gtk-rs/gdk/issues/304


sdroege:
See https://github.com/gtk-rs/gdk/issues/304#issuecomment-585553824

PR #552, opened 2020-02-14
Propagate GTK mouse leave notification as MouseEvent

ForLoveOfCats:
Fixes https://github.com/xi-editor/druid/issues/541

GTK doesn't seem to want to give mouse motion event outside the window (at the very least I couldn't find anything online on how to coerce it to) so this just grabs the mouse leave notification and forwards it as a MouseEvent


ForLoveOfCats:
Here are some before and afters of this PR (on master as of 4cd1ea0fa4b3611a83d3af9c16e5a105ad08721b)

before_leave
after_leave

PR #553, opened 2020-02-14
Handle events that do not originate with a window

### LINKED ISSUE #114: Add some kind of top-level application type

cmyr:
This adds a simple `AppHandler` trait to `druid-shell` and a corresponding implementation to `druid`, which together let the application handle events (right now  this is only menu commands) that do not originate in a window.

This is very specifically intended to address the case, on macOS, where no windows are open but there is still an application menu.

This involved quite a bit of refactoring, because we had baked in the assumption that all events originated with a window at a very low level.

This has exposed a few new issues; in particular the mechanism for opening a file is currently implemented on `WindowHandle`, so there's no way to open a file from an empty window, but it is  at least now possible to quit an application on mac after closing its last window.

closes #114

### DIFF
--- a/druid/src/win_handler.rs
+++ b/druid/src/win_handler.rs
@@ -76,6 +76,9 @@ struct Inner<T> {
     command_queue: CommandQueue,
     ext_event_host: ExtEventHost,
     windows: Windows<T>,
+    /// the application-level menu, only set on macos and only if there
+    /// are no open windows.
+    root_menu: Option<MenuDesc<T>>,
     pub(crate) env: Env,
     pub(crate) data: T,
 }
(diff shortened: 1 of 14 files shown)


druid/src/win_handler.rs
```diff
@@ -339,32 +389,70 @@ impl<T: Data> AppState<T> {
...
 
     /// Send an event to the widget hierarchy.
     ///
     /// Returns `true` if the event produced an action.
     ///
     /// This is principally because in certain cases (such as keydown on Windows)
     /// the OS needs to know if an event was handled.
-    fn do_event(&mut self, event: Event) -> bool {
-        let result = self.app_state.borrow_mut().do_event(self.window_id, event);
+    fn do_event(&mut self, event: Event, window_id: WindowId) -> bool {
```
raphlinus: Rename: do_window_event

PR #559, opened 2020-02-15
Correctly dispatch open/save commands

cmyr:
Missed this in #553.


cmyr:
yep!

cmyr:
you're only as clever as your knowledge of how dumb you are, or something?

PR #561, opened 2020-02-17
Add PaintCtx::widget_id()

cmyr:
At the very least this is useful during debugging.


cmyr:
It's mostly just for debugging; the reason that `LayoutCtx` doesn't have it (yet) is that `LayoutCtx` doesn't currently have a copy of `BaseState`, so adding access to widget id there would be a more invasive change. 🤷‍♂

PR #562, opened 2020-02-17
Don't always clear needs_inval before forwarding an event

cmyr:
This flag can be cleared during painting, but should persist between
events; specifically it should not be cleared when a Command is
handled after a non-command event, which could cause us to
ignore an invalidation that occured during the handling of the first
event.


raphlinus:
Agree it shouldn't be cleared every event. Is there some other place it's getting cleared?

cmyr:
https://github.com/xi-editor/druid/blob/master/druid/src/core.rs#L274

PR #564, opened 2020-02-18
Invalidate label layout when text changes

cmyr:
This generally improves our dynamic label so that it will recompute
less often, and behaves more like the localized string version.


futurepaul:
This seems to solve the problem in the hello app where it shows "hello-counter" until you do the first increment. Or is that just me?

cmyr:
I didn't even know we had that bug, but yes I think I understand why we do (lifecycle changes, not getting update when there is no data) and why this fixes it.

cmyr:
I'm going too self-approve this, it's a concrete bug that is frustrating me regularly 😅

PR #565, opened 2020-02-18
Use common code for svg and image filling

fishrockz:
This lets the svg widget use all of of the fill options in image.

It also takes some good ideas from @cmyr as to how to tidy up the code.

### DIFF
--- a/druid/src/widget/common.rs
+++ b/druid/src/widget/common.rs
@@ -1,0 +1,75 @@
+// Copyright 2020 The xi-editor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::{Affine, Size};
+
+// These are based on https://api.flutter.dev/flutter/painting/BoxFit-class.html
+#[derive(Clone, Copy, PartialEq)]
+pub enum FillStrat {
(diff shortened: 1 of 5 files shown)


druid/src/widget/common.rs
```diff
@@ -0,0 +1,85 @@
...
+        FillStrat::Contain
+    }
+}
+
+impl FillStrat {
+    /// Calculate an origin and scale for an image with a given `FillStrat`.
+    ///
+    /// This takes some properties of a widget and a fill strategy and returns an affine matrix
+    /// used to position and scale the image in the widget.
+    pub fn affine_to_fill(&self, parent: Size, fit_box: Size) -> Affine {
```
cmyr: this lint is saying that we should just use `self` instead of `&self` because the type is smaller than a pointer.

druid/src/widget/common.rs
```diff
@@ -0,0 +1,85 @@
...
+        let scalex = parent.width / fit_box.width;
+        let scaley = parent.height / fit_box.height;
+
+        let scale: Point = match self {
```
cmyr: this isn't really related to this patch, but more something that was lost in the shuffle when we eventually moved to `Affine`; it's now sort of weird that we are constructing `Point`s at all, being as we never use them as such?

I would probably write this `match` as,

```rust
let (scalex, scaley) = match self {
    FillStrat::Contain => {
        let scale = scalex.min(scaley);
        (scale, scale)
    }
    FillStrat::Fill => (scalex, scaley),
    // ...
}
```

druid/src/widget/common.rs
```diff
@@ -0,0 +1,85 @@
...
+            FillStrat::ScaleDown => {
+                let scale = scalex.min(scaley).min(1.0);
+                Point { x: scale, y: scale }
+            }
+            FillStrat::None => Point { x: 1.0, y: 1.0 },
+        };
+
+        let origin_x = (parent.width - (fit_box.width * scale.x)) / 2.0;
+        let origin_y = (parent.height - (fit_box.height * scale.y)) / 2.0;
+        let origin = Point::new(origin_x, origin_y);
```
cmyr: it feels like we can also just delete this line and use `origin_x` and `origin_y` directly.

PR #567, opened 2020-02-18
Fixup message when image example is run without image feature

### LINKED ISSUE #566: Error running image example
Was interested to try the new image widget, but when I try to run the example I get the error:

```
$ cargo run --example image --features "images"
error: --features is not allowed in the root of a virtual workspace
```

Am I missing something, or is it a cargo or druid issue?

> stable-x86_64-pc-windows-msvc 
> rustc 1.41.0

cmyr:
closes #566


fishrockz:
The current example works just fine in bash on linux, if this is a real issue on other platforms then should we also change the svg example as it is consistent with this?

PR #569, opened 2020-02-18
Move debug_paint_layout to WidgetExt

cmyr:
This lets us choose to only debug a subset of the tree,
and fixes an issue where we would overwrite an existing `modify_env`
closure in the AppLauncher.

### DIFF
--- a/druid/src/widget/widget_ext.rs
+++ b/druid/src/widget/widget_ext.rs
@@ -123,6 +123,13 @@ pub trait WidgetExt<T: Data>: Widget<T> + Sized + 'static {
         EnvScope::new(f, self)
     }
 
+    /// Draw the [`layout`] `Rect`s of  this widget and its children.
+    ///
+    /// [`layout`]: trait.Widget.html#tymethod.layout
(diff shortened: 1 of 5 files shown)


druid/src/widget/widget_ext.rs
```diff
@@ -175,6 +182,25 @@ impl<T: Data> SizedBox<T> {
     }
 }
 
+// if two things are modifying an env one after another, just combine the modifications
+impl<T: Data, W> EnvScope<T, W> {
+    pub fn env_scope(self, f2: impl Fn(&mut Env, &T) + 'static) -> EnvScope<T, W> {
+        let EnvScope { f, child } = self;
```
futurepaul: for some reason I've never used this style of named struct destructuring outside of a match statement, seems handy!

futurepaul:
This feels like a great change to me, always amazing how useful WidgetExt is.

It's not obvious to me how the `env` was being overwritten in the original version, so maybe I misunderstood when this closure was being run.

cmyr:
What was happening in the original version is that the `configure_env` method took a closure that it saved and ran later, so calling this method twice would cause the closure to be replaced. :)

PR #570, opened 2020-02-18
Add Env::DEBUG_WIDGET helper

cmyr:
Something I've wanted for a while, and just needed while doing some runebender stuff:

The intention is that, during debugging, you can set this key to
`true` for some widget and then check it in the Widget methods
to limit linespam when doing println debugging.

### DIFF
--- a/druid/src/env.rs
+++ b/druid/src/env.rs
@@ -102,6 +102,37 @@ pub struct ValueTypeError {
 }
 
 impl Env {
+    /// State for whether or not to paint colorful rectangles for layout
+    /// debugging.
+    ///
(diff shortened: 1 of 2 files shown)


druid/src/env.rs
```diff
@@ -102,6 +102,37 @@ pub struct ValueTypeError {
 }
 
 impl Env {
+    /// State for whether or not to paint colorful rectangles for layout
+    /// debugging.
+    ///
+    /// Set by the `debug_paint_layout()` method on [`AppLauncher`]'.
```
futurepaul: I believe the location of this method will be incorrect soon :)

PR #572, opened 2020-02-23
Remove clear from custom_widget example

futurepaul:
When I wrote this widget I thought `clear` was the right thing to do, and kept forgetting to go back and fix it. Would be happy to hear of any other improvements we could make to this example. If we want I can port my color picker tutorial to this example, because there's a lot in common.

### DIFF
--- a/druid/examples/custom_widget.rs
+++ b/druid/examples/custom_widget.rs
@@ -60,17 +60,22 @@ impl Widget<String> for CustomWidget {
     // Basically, anything that changes the appearance of a widget causes a paint.
     fn paint(&mut self, paint_ctx: &mut PaintCtx, data: &String, _env: &Env) {
         // Let's draw a picture with Piet!
-        // Clear the whole context with the color of your choice
-        paint_ctx.clear(Color::WHITE);
 
-        // Create an arbitrary bezier path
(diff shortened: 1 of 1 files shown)


druid/examples/custom_widget.rs
```diff
@@ -60,17 +60,22 @@ impl Widget<String> for CustomWidget {
...
     fn paint(&mut self, paint_ctx: &mut PaintCtx, data: &String, _env: &Env) {
         // Let's draw a picture with Piet!
-        // Clear the whole context with the color of your choice
-        paint_ctx.clear(Color::WHITE);
 
-        // Create an arbitrary bezier path
+        // Clear the whole widget with the color of your choice
         // (paint_ctx.size() returns the size of the layout rect we're painting in)
         let size = paint_ctx.size();
+        let rect = Rect::from_origin_size(Point::ORIGIN, size);
```
raphlinus: I guess https://github.com/linebender/kurbo/pull/86 hasn't been released yet?

cmyr: It is not, will have to wait for the next piet release.

futurepaul: ah that will be great to have

PR #578, opened 2020-02-25
Add alpha support for images

fishrockz:
This seems better but it would be even nicer if it determined if it was worth loading the alpha channel and then it could save some byte but I don't think it really hurts much. 

Also I don't like committing binary files but that png is smaller than some of the .rs files in that dir.


cmyr:
is this supposed to look like,

Screen Shot 2020-02-26 at 11 53 00 AM
?

fishrockz:
Yep, I was thinking of setting the background to something more distinctive so it's more obvious. My druid windows backgrounds aren't quite black so it's more obvious

fishrockz:
Screenshot from 2020-02-26 18-35-47
Screenshot from 2020-02-26 18-34-32
Please see before and after.

fishrockz:
I also acured to me that this should be a option for `new` and the binlinier interpolation should be user selectable in the same way that the FillStrat is. shall i add them to this MR?

PR #579, opened 2020-02-26
Add widget controller

cmyr:
This is an experiment for allowing easier overriding of event/lifecycle/update.


futurepaul:
This is def a step forward in ergonomics imo. I guess this question is if this should be three different widgets, one for each method, or if there's something inherently inter-related in event / lifecycle / update that makes more sense to package them together. I'm not proposing that we separate them, just checking on the reasoning.

cmyr:
I'm thinking of this as a distinction between behaviour and presentation. These are the  'behaviour' methods.

The inverse of this would be a `Painter` widget, that gets paint and _maybe_ layout. The reason I think maybe not layout is because I like the idea of using this `Paint` widget as basically another kind of `PaintBrush`, i.e. something you can set as the background of another widget, for instance.

There are other ways to do this; for instance `Controller` could be a widget and each of the methods could be a closure, but I tried that and it felt a bit more cumbersome. :shrug:

cmyr:
Okay, my inclination now is to merge this, and if it ends up not being useful we can pull it out again.

tbillington:
For a data point, I'm using the controller branch in [kondo](https://github.com/tbillington/kondo/pull/19) and it was useful. It may have been less useful if I had a better knowledge of how to implement a full event controller. Regardless, it got the job done :)

PR #580, opened 2020-02-26
Set Scroll active to true when dragging scrollbars

ForLoveOfCats:
Fixes https://github.com/xi-editor/druid/issues/576

### DIFF
--- a/druid/src/widget/scroll.rs
+++ b/druid/src/widget/scroll.rs
@@ -313,14 +313,11 @@ impl<T: Data, W: Widget<T>> Widget<T> for Scroll<T, W> {
                     }
                     ctx.request_paint();
                 }
-                Event::MouseUp(event) => {
+                Event::MouseUp(_) => {
                     self.scroll_bars.held = BarHeldState::None;
                     ctx.set_active(false);
 
-                    let offset_pos = event.pos + self.scroll_offset;
-                    if !self.point_hits_vertical_bar(viewport, offset_pos, &env)
-                        && !self.point_hits_horizontal_bar(viewport, offset_pos, &env)
-                    {
+                    if !scroll_bar_is_hovered {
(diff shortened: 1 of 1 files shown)


druid/src/widget/scroll.rs
```diff
@@ -313,7 +313,18 @@ impl<T: Data, W: Widget<T>> Widget<T> for Scroll<T, W> {
                     }
                     ctx.request_paint();
                 }
-                Event::MouseUp(_) => self.scroll_bars.held = BarHeldState::None,
+                Event::MouseUp(event) => {
+                    self.scroll_bars.held = BarHeldState::None;
+                    ctx.set_active(false);
+
+                    let offset_pos = event.pos + self.scroll_offset;
```
cmyr: If this is the same as the `scroll_bar_is_hovered` check that we do at the top of this method, we could just reuse that here?

Also: did you run into this as an issue, or are you being defensive? Basically I _think_ that `MouseUp` should always be preceded by `MouseMoved`, that is we should be handling hover changes in `MouseMoved` and not have to worry about this, but that invariant isn't written down or checked, so 🤷‍♂

ForLoveOfCats: Are you referring to calculating `offset_pos` at the top of the function as opposed to doing it at usage site?

Yes I ran into an issue, if the mouse point is outside the Scroll's borders then it is not notified of the new position any other way. So releasing within bounds would work without the scrollbar hit checks here but not if released when outside the individual Scroll's bounds.

cmyr: I mean there is a bool called `scroll_bar_is_hovered` that we calculate at the top of this method.

If we're active we should always receive mouse events, regardless of where the mouse is? Maybe we have some other bug?

Issue #582, opened 2020-02-28
control widgets and container alignment


cmyr:
We have a bunch of nice control widgets, but they don't  play nicely together. The major issue is vertical alignment; widgets all have  their own heights, and this makes  them look really weird when used together.

As an illustration, here is a row  of our common control widgets, laid out naively:

<img width="547" alt="Screen Shot 2020-02-28 at 3 10 15 PM" src="https://user-images.githubusercontent.com/3330916/75583841-6f4bd600-5a3c-11ea-9596-c85334c6520c.png">

There are various things going on here:
- the widgets have different heights and are top-aligned, which looks jagged
- some widgets include internal padding and others don't, which makes it hard to use them interchangeably
- different text widgets have different internal padding, causing misaligned baselines

This stuff is a bit tricky. The easy one is the interior h-padding; I think we can just say that widgets shouldn't do that.

The alignment stuff is tricky, given our layout model. The basic problem is that in the current implementation of `Flex`, we do a single pass, measuring each widget, and then laying it out immediately. This means that we do not know our own final size when laying out widgets, so we can't do things like center widgets vertically in the container.

One option we would have is to have a canonical 'control height', which would be the declared height of all control widgets, and they could handle centering themselves.

Another option would be to have a `ControlGroup` widget that would know the default control height, would size itself to that height, and then center or otherwise align all children in that space; this feels hacky.

A third option is to allow the user to specify an alignment for a row/column. This would make layout more expensive; we would need to do two passes at layout, one to get sizes and then one to resolve position.

This maybe needs more research. SwiftUI has [alignment guides](https://swiftui-lab.com/alignment-guides/) which seem interesting, but I don't know anything about the implementation.

Baseline alignment is going to be an additional challenge, because we will need to find some way of getting that information from widgets.

futurepaul:
<img width="683" alt="sized_box" src="https://user-images.githubusercontent.com/543668/75616406-7c40f600-5b1e-11ea-9dcd-1b2b09540baf.png">

Fwiw, the widgets were designed to center vertically, but can only center vertically if given a height to center themselves within:

```rust
    Flex::row()
        .with_child(textbox, 0.0)
        .with_child(checkbox, 0.0)
        .with_child(slider, 0.0)
        .with_child(progress_bar, 0.0)
        .with_child(stepper, 0.0)
        .with_child(switch, 0.0)
        .with_child(button, 0.0)
        .with_child(label, 0.0)
        .fix_height(25.0)
        .center()
[...]

PR #583, opened 2020-02-29
Make flex child alignment configurable

cmyr:
This was much easier than expected; the flex widget was already
doing layout in multiple passes, so it is simply a matter of
calculating the correct position on the minor axis for a given
alignment.

Screen Shot 2020-02-28 at 7 20 36 PM

This is related to/inspired by #582

### DIFF
--- a/druid/src/widget/flex.rs
+++ b/druid/src/widget/flex.rs
@@ -24,6 +24,7 @@ use crate::{
 /// A container with either horizontal or vertical layout.
 pub struct Flex<T> {
     direction: Axis,
+    alignment: Alignment,
     children: Vec<ChildWidget<T>>,
 }
 
@@ -35,6 +36,26 @@ struct ChildWidget<T> {
 pub(crate) enum Axis {
     Horizontal,
     Vertical,
+}
+
+/// The alignment of the widgets inside of the container.
+///
+/// If a widget is smaller than the container on the minor axis, this determines
+/// where it is positioned.
+#[derive(Debug, Clone, Copy)]
+pub enum Alignment {
+    /// Top or leading.
+    ///
+    /// In a vertical container, widgets are top aligned. In a horiziontal
+    /// container, their leading edges are aligned.
+    Start,
+    /// Widgets are centered in the container.
+    Center,
+    /// Bottom  or trailing.
+    ///
+    /// In a vertical container, widgets are bottom aligned. In a horiziontal
+    /// container, their trailing edges are aligned.
+    End,
 }
 
 #[derive(Copy, Clone, Default)]
@@ -73,6 +94,7 @@ impl<T> Flex<T> {
         Flex {
             direction: Axis::Horizontal,
             children: Vec::new(),
+            alignment: Alignment::Start,
         }
     }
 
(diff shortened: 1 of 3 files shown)


druid/src/widget/flex.rs
```diff
@@ -37,6 +38,26 @@ pub(crate) enum Axis {
...
+///
+/// If a widget is smaller than the container on the minor axis, this determines
+/// where it is positioned.
+#[derive(Debug, Clone, Copy)]
+pub enum Alignment {
+    /// Top or leading.
+    ///
+    /// In a vertical container, widgets are top aligned. In a horiziontal
+    /// container, their leading edges are aligned.
+    First,
```
ForLoveOfCats: This feels weird. Would something like this be possible or even desirable?
```rust
pub enum VerticalAlignment {
    Top,
    Center,
    End,
}

pub enum HorizontalAlignment {
    Leading,
    Center,
    Trailing,
}
```
Then modify `Axis`
```rust
pub(crate) enum Axis {
    Horizontal(HorizontalAlignment),
    Vertical(VerticalAlignment),
}
```
This could be even worse UX wise, just tossing around an idea :man_shrugging:

cmyr: heh I actually had almost literally that code, except with still having my `Alignment` type internally, and using `From<VerticalAlignment>` and `From<HorizontalAlignment>` for it. The problem I ran into is that the vertical/horizontal axis is not really enforced in the type system; that is to say, I can't at compile time ensure that a vertical `Flex` can accept a `HorizontalAlignment` but not a `VerticalAlignment`.

The one way I could think of would be to *require* the `row()` or `column()` constructor to take an explicit vertical or horizontal alignment argument, but this breaks a very common api in a situation where there is probably a reasonable default.

So I ended up with this.

In a funny way, making this change does start to argue for having separate `Row`/`Column` or `HStack`/`VStack` types, because then we could tie the allowable alignment to the type system. As is, though, I really couldn't find a satisfying solution.

ForLoveOfCats: Yeah I was trying to avoid suggesting separate vertical and horizontal versions of the struct :thinking: Thankfully `Flex` doesn't seem to have more than ~120 LOC will eventually lessen the pain of duplicating most of it....

cmyr: Oh, we wouldn't need to duplicate it I don't think, we could just use it as private API, with types like:


```rust
pub struct Row<T> {
    inner: Flex<T>,
    alignment: HorizontalAlignment,
}

pub struct Column<T> {
    inner: Flex<T>,
    alignment: VerticalAlignment,
}
```

cmyr: but again I don't love this. Another option would be to have `Leading`, `Top`, `Trailing`, `Bottom`, be essentially aliases for each other. 🤷‍♂

ForLoveOfCats: Oh nice! That makes me feel better about that lol

cmyr: In any case my inclination is to leave this as-is until at least we implement baseline alignment, and then we can consider switching the API again?

cmyr:
yea, I think if that isn't true we've violated an invariant and crashing is the right thing to do.

Issue #427, opened 2019-12-29
Windows without decorations


crsaracco:
*(Priority: super low)*

I'd like to be able to create windows without the titlebar or borders, in order to make my own custom thing.

For example, Bitwig made its own titlebar so it could add more stuff to it:

![bitwig-no-titlebar](https://i.imgur.com/kK8LVoj.png)

In GTK land, this is done with [`gtk_window_set_decorated`](https://developer.gimp.org/api/2.0/gtk/GtkWindow.html#gtk-window-set-decorated).

Zarenor:
For Windows, we'd need a way to adjust the window style being provided to CreateWindow - currently it's always `WS_OVERLAPPED` - we'd need `WS_POPUP` for borderless, I think. Then we will need to make it clear how to create the minimize, maximize/restore, and close buttons. And any resize border the user wants. 

So, maybe a little complex, but definitely something I'd like to see.

In general, if you have the time, I think we're open to seeing something created for one platform and letting the others trail until we can get a good design story and make a workable API. I'm not sure if we're okay with that approach for something this large

cmyr:
So this is totally a goal, and is highly platform dependent.

I see there being two ways of handling this. The first is better API but less flexible: a `WindowDescExt` traits exposed on a per-platform basis, that allow doing platform-specific customization. This is inspired by the various traits in stdlib like [`OsStrExt`], [`DirBuilderExt`], etcetera. For platform specific behaviour (like setting the [`NSWindowStyleMask`] on macOS) you would do that via a method on an extension trait.

The other option is to allow (per-platform) for the `WindowDesc` to take a closure that will be given a pointer to a window object on that platform, where you can do whatever you want; we would call this right after instantiating the  window. This would let you do anything you can do with access to the window object, although it wouldn't work for things that you do during a window creation routine.

It may be but this particular case is general enough that we can just add it to `WindowDesc` and `WindowBuilder` directly, in which case you can ignore my comment above, although I still think something like this is something we will want eventually.

futurepaul:
Here are the related options in Electron for reference: https://electronjs.org/docs/api/frameless-window

iulianR:
[...]

PR #587, opened 2020-03-01
'resizable' and 'show_titlebar' options

Psykopear:
This PR solves #427 for linux only. I created the needed functions on windows and mac too, but they are not used in the 'build' function, so for now they do nothing.

### DIFF
--- a/druid-shell/src/platform/gtk/window.rs
+++ b/druid-shell/src/platform/gtk/window.rs
@@ -86,6 +86,8 @@ pub struct WindowBuilder {
     title: String,
     menu: Option<Menu>,
     size: Size,
+    resizable: bool,
+    decorated: bool,
 }
 
 #[derive(Clone)]
@@ -114,6 +116,8 @@ impl WindowBuilder {
             title: String::new(),
             menu: None,
             size: Size::new(500.0, 400.0),
+            resizable: true,
+            decorated: true,
         }
     }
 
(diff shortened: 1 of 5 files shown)


druid-shell/src/platform/gtk/window.rs
```diff
@@ -415,6 +429,18 @@ impl WindowHandle {
         }
     }
 
+    pub fn set_resizable(&self, resizable: bool) {
```
cmyr: In general we expect the various platform `WindowHandle` and other types to be used through the shared `WindowHandle` struct in `druid-shell/src/window.rs`, and so I would add these methods there, wrapping the method on each platform type; this will require you to add stubs of these methods to the other platforms.

druid-shell/src/window.rs
```diff
@@ -213,6 +213,16 @@ impl WindowBuilder {
         self.0.set_size(size)
     }
 
+    /// Set wheter the window should be resizable
+    pub fn resizable(&mut self, resizable: bool) {
+        self.0.resizable(resizable)
+    }
+
+    /// Set wheter the window should have a titlebar and decorations
```
cmyr:
```suggestion
    /// Set whether the window should have a titlebar and decorations
```

druid-shell/src/window.rs
```diff
@@ -213,6 +213,16 @@ impl WindowBuilder {
         self.0.set_size(size)
     }
 
+    /// Set wheter the window should be resizable
```
cmyr:
```suggestion
    /// Set whether the window should be resizable
```

druid-shell/src/platform/gtk/window.rs
Psykopear: Uh, I already added the stubs for other platforms, so it should be enough to add the two methods to the WindowHandle right?

cmyr: I saw you had the methods for builder, but I didn't see them for WindowHandle? I might have missed it. :)

Psykopear: Yep you're right, I missed the WindowHandle, adding it now

Psykopear:
CI looks good, anything else you want me to look at?

PR #591, opened 2020-03-02
Process commands and do update after external event

cmyr:
Previously we skipped update and just went directly to
painting in this case, but that doesn't make any sense.

This also now processes any commands submitted while processing 
external events before returning.


cmyr:
All other events come through another code path (`do_window_event`) and that handles update as normal; this is the only current special case.

Issue #526, opened 2020-02-07
add a PaintBrush type


cmyr:
I'm reminded by #523 of something we'd talked about a little while ago, and might as well have an issue for. There is currently a `PaintBrush` type in piet, which is a way of representing disparate different things that act like colors; the idea is that we can treat colors or gradients equivalently for the sake of our API.

@futurepaul brought up a good point, which is that it would be nice if we could treat env/theme keys as colors, where appropriate, so that you could write,

```rust
let my_widget = TextBox::raw().padding(5.0).background(theme::BACKGROUND_LIGHT);
```

And then have your color be determined by the value of that key at runtime.

I think we can do this, and I think it's pretty simple. I propose that we add a type to druid, also called `PaintBrush`, which should look like this:

```rust
pub enum PaintBrush {
    Concrete(piet::PaintBrush),
    Key(Key<Color>),
    // other variants possible
}
```

And then we should be able to take one of these types, pass it an `Env`, and get something we can actually paint, like:

```rust
impl PaintBrush {
    fn resolve(&self, env: &Env) -> piet::PaintBrush { ... }
}
```

The idea is that a method like `Container::background` would take as its argument `impl Into<PaintBrush>`, and so you could then pass it *either* an actual color/gradient or a key that could be resolved to a color/gradient at runtime.

This isn't too tricky, but it does require some understanding of the type system and of how `Env` works, but not much more than that.

futurepaul:
I have this working for `Key<Color>`, having a lifetime issue with `Key<LinearGradient>`

cmyr:
push a branch and we can take a look?

PoignardAzur:
With #592 merged, is this still relevant? KeyOrValue is ubiquitous in the codebase now.

PR #592, opened 2020-03-02
Add KeyOrValue for styling purposes

futurepaul:
Here's a proof of concept of `KeyOrValue`, which is sort of the logical conclusion of #526 .

When providing style values to a widget, the user can use a named value from the `Env`, which will be resolved later by the widget, or they can just use a "concrete" value. I've only implemented this for `Container`, as a proof of concept, but I believe this should work for all widgets that accept an `impl ValueType` in their constructor.

I'm not sure of the performance implications, especially with how much cloning I'm doing. Also, this is with `Key` still taking a `'static str`, lifetimes get really hairy when `Key` isn't static.

The other limitation is that this currently doesn't work for gradients.


druid/src/widget/container.rs
```diff
@@ -16,18 +16,18 @@
...
-    width: f64,
-    brush: PaintBrush,
+    width: KeyOrValue<f64>,
+    brush: KeyOrValue<Color>,
 }
 
 /// A widget that provides simple visual styling options to a child.
 pub struct Container<T> {
-    background: Option<PaintBrush>,
+    background: Option<KeyOrValue<Color>>,
```
cmyr: I think this isn't quite right, but is useful as an illustration.

We'll still likely want a `PaintBrush` enum in druid proper that includes `KeyOrValue<Color>` but perhaps some other things as well.

druid/examples/panels.rs
```diff
@@ -15,25 +15,23 @@
...
 
 fn build_app() -> impl Widget<()> {
     let solid = Color::rgb8(0x3a, 0x3a, 0x3a);
-    let gradient = LinearGradient::new(
-        UnitPoint::TOP_LEFT,
-        UnitPoint::BOTTOM_RIGHT,
-        (Color::rgb8(0x11, 0x11, 0x11), Color::rgb8(0xbb, 0xbb, 0xbb)),
-    );
+
+    // let gradient = LinearGradient::new(
```
cmyr: we can get rid of this file too I think, then looks good to me!

druid/src/env.rs
```diff
@@ -76,6 +78,18 @@ pub enum Value {
     String(String),
 }
 
+/// Either a concrete `T` or a [`Key<T>`] that can be resolved in the [`Env`].
+///
+/// This is a way to allow widgets to interchangeably use either a specific
+/// value or a value from the environment for some purpose.
+///
+/// [`Key<T>`]: struct.Key.html
+/// [`Env<T>`]: struct.Env.html
```
cmyr: oops my mistake, this is wrong, should just be `Env`.

cmyr:
```suggestion
/// [`Env`]: struct.Env.html
```

futurepaul: not sure what you're referring to

PR #596, opened 2020-03-03
game of life

kindlychung:
Suggestions are welcome.

### DIFF
--- a/druid/examples/game_of_life.rs
+++ b/druid/examples/game_of_life.rs
@@ -276,8 +276,7 @@ impl Widget<AppData> for GameOfLifeWidget {
         _event: &LifeCycle,
         _data: &AppData,
         _env: &Env,
-    ) {
-    }
+    ) {}
 
     fn update(&mut self, ctx: &mut UpdateCtx, _old_data: &AppData, _data: &AppData, _env: &Env) {
         ctx.request_paint();
@@ -384,11 +383,11 @@ fn make_widget() -> impl Widget<AppData> {
                                     ctx.request_paint();
                                 },
                             )
-                            .lens(AppData::paused)
-                            .center()
-                            .fix_width(80.)
-                            .fix_height(40.)
-                            .padding((2.0, 2.0)),
+                                .lens(AppData::paused)
+                                .center()
+                                .fix_width(80.)
+                                .fix_height(40.)
+                                .padding((2.0, 2.0)),
                             1.0,
                         )
                         .with_child(
@@ -397,10 +396,10 @@ fn make_widget() -> impl Widget<AppData> {
                                 data.clear();
                                 ctx.request_paint();
                             })
-                            .lens(AppData::grid)
-                            .center()
-                            .fix_height(30.)
-                            .padding((2.0, 2.0)),
+                                .lens(AppData::grid)
+                                .center()
+                                .fix_height(30.)
+                                .padding((2.0, 2.0)),
                             1.0,
                         )
                         .fix_height(35.)
@@ -414,7 +413,7 @@ fn make_widget() -> impl Widget<AppData> {
                                 .padding(3.0),
                             0.,
                         )
-                        .with_child(Slider::new().lens(AppData::speed).padding(3.0), 1.)
+                        .with_child(Slider::new().lens(AppData::speed).padding(3.0).expand(), 1.)
                         .alignment(Alignment::Center)
                         .padding(4.0),
                     0.,
--- a/druid/examples/game_of_life.rs
+++ b/druid/examples/game_of_life.rs
@@ -47,8 +47,8 @@ struct GridPos {
 }
 
 #[derive(Clone)]
-struct ColorScheme<'a> {
-    colors: Vec<&'a Color>,
+struct ColorScheme {
+    colors: Vec<Color>,
     current: usize,
 }
 
@@ -207,13 +207,13 @@ impl Grid {
     }
 }
 
-struct GameOfLifeWidget<'a> {
+struct GameOfLifeWidget {
     timer_id: TimerToken,
     cell_size: Size,
-    color_scheme: ColorScheme<'a>,
+    color_scheme: ColorScheme,
 }
 
-impl GameOfLifeWidget<'_> {
+impl GameOfLifeWidget {
     fn grid_pos(&self, p: Point) -> Option<GridPos> {
         let w0 = self.cell_size.width;
         let h0 = self.cell_size.height;
@@ -229,7 +229,7 @@ impl GameOfLifeWidget<'_> {
     }
 }
 
-impl Widget<AppData> for GameOfLifeWidget<'_> {
+impl Widget<AppData> for GameOfLifeWidget {
     fn event(&mut self, ctx: &mut EventCtx, event: &Event, data: &mut AppData, _env: &Env) {
         match event {
             Event::WindowConnected => {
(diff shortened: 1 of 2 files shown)


druid/examples/game_of_life.rs
```diff
@@ -0,0 +1,171 @@
...
+    ) -> Size {
+        bc.max()
+    }
+
+    fn paint(&mut self, paint_ctx: &mut PaintCtx, data: &AppData, _env: &Env) {
+        let size: Size = paint_ctx.size();
+        let w0 = size.width/40.0;
+        let h0 = size.height/40.0;
+        let cell_size = Size{ width: w0, height: h0 };
+        for i in 0..1600 {
```
cmyr: and a general rust thought: I would probably (in a "real" program) write this as using an iterator returned from my `Grid` type; there would be a function like,

```rust
iter_live_cells(&self) -> impl Iterator<Item=(usize, usize)> { .. }
```
and then here it would just be like,

```rust
for (row, col) in grid.iter_live_cells() {
// make rect
// paint
}
```

druid/examples/game_of_life.rs
```diff
@@ -0,0 +1,171 @@
...
+
+use druid::{Affine, AppLauncher, BoxConstraints, Color, Env, Event, EventCtx, LayoutCtx, LifeCycle, LifeCycleCtx, LocalizedString, PaintCtx, Point, Rect, RenderContext, Size, UpdateCtx, Widget, WindowDesc, Data};
+
+
+const POOL_SIZE: i32 = 1600;
+const CELL_COLOR: Color = Color::rgb8(0xf3 as u8, 0xf4 as u8, 8 as u8);
+
+#[derive(Clone)]
+struct AppData {
+    grid: [i32; 1600],
```
cmyr: does the game have more state than just on/off? is it ternary? In the former case I would use a `bool` here (or a custom enum with `Alive` and `Dead` variants), and in the latter case I would use an enum as well; this lets our types better reflect the program we're trying to write.

druid/examples/game_of_life.rs
```diff
@@ -0,0 +1,171 @@
...
+}
+
+impl AppData {
+    pub fn evolve(&mut self) {
+        let mut indices_to_mutate: Vec<usize> = vec![];
+        for i in 0..POOL_SIZE {
+            let neighbors = AppData::neighbors(i);
+            let mut sum = 0;
+            for nb_i in &neighbors {
```
cmyr: I would also be tempted to make this an iterator that can filter out out-of-bounds neighbours; or I might at least write,

```rust
for nb_i in neighbors.iter().filter(|n| n < POOL_SIZE) { ... }
```

druid/examples/game_of_life.rs
```diff
@@ -0,0 +1,171 @@
...
+//! Game of life
+
+use druid::kurbo::BezPath;
+use druid::piet::{FontBuilder, ImageFormat, InterpolationMode, Text, TextLayoutBuilder};
+
+
+use druid::{Affine, AppLauncher, BoxConstraints, Color, Env, Event, EventCtx, LayoutCtx, LifeCycle, LifeCycleCtx, LocalizedString, PaintCtx, Point, Rect, RenderContext, Size, UpdateCtx, Widget, WindowDesc, Data};
+
+
+const POOL_SIZE: i32 = 1600;
```
cmyr: I think this should be a `usize`, which is the type used for most stdlib indexes?

druid/examples/game_of_life.rs
cmyr: I would also probably (caveat: all of this is imagined for some "real program") call `AppData` `Grid`, and I would maybe even completely abstract away the underlying storage, coming up with something like this:

```rust
struct Grid {
    storage: [bool; POOL_SIZE],
}

#[derive(Clone, Copy)]
struct GridPos {
    row: usize,
    col: usize,
}

impl Index<GridPos> for Grid {
    type Output = bool;

    fn index(&self, pos: GridPos) -> &Self::Output {
        let idx = pos.row * GRID_SIZE + pos.col;
        self.storage.index(idx)
    }
}

impl IndexMut<GridPos> for Grid {
    type Output = bool;

    fn index_mut(&mut self, pos: GridPos) -> &Self::Output {
        let idx = pos.row * GRID_SIZE + pos.col;
        self.storage.index_mut(idx)
    }
}
```

And then I would implement everything in terms of that?

Again, this is overkill for this problem but would be very reasonable for something of larger scope.

druid/examples/game_of_life.rs
Zarenor: Also: I think the const can be used in the array declaration. Since I think you mean them to match, best to use the const.
Secondly, if you intend a square, and want to make your life 'easier', I would define a 'SIDE_LENGTH' const, and make this 'SIDE_LENGTH * SIDE_LENGTH'. (SIDE_LENGTH = 40, here)

druid/examples/game_of_life.rs
```diff
@@ -0,0 +1,508 @@
...
+
+#[derive(Clone, Copy, PartialEq)]
+struct GridPos {
+    row: usize,
+    col: usize,
+}
+
+#[derive(Clone)]
+struct ColorScheme<'a> {
+    colors: Vec<&'a Color>,
```
cmyr: I don't think there's any need for this to be borrowed data, `Vec<Color>` should be fine. Was this about the iterator?

druid/examples/game_of_life.rs
```diff
@@ -0,0 +1,508 @@
...
+//! Game of life
+
+use std::ops::{Index, IndexMut};
+use std::time::{Duration, Instant};
+
+use druid::widget::{Alignment, Button, Flex, Label, Slider, WidgetExt};
+use druid::{
+    AppLauncher, BoxConstraints, Color, Data, Env, Event, EventCtx, LayoutCtx, Lens, LifeCycle,
+    LifeCycleCtx, LocalizedString, MouseButton, PaintCtx, Point, Rect, RenderContext, Size,
+    TimerToken, UnitPoint, UpdateCtx, Widget, WindowDesc,
```
cmyr: lets clean up the unused import

druid/examples/game_of_life.rs
```diff
@@ -0,0 +1,508 @@
...
+        let min_side = max_size.height.min(max_size.width);
+        Size {
+            width: min_side,
+            height: min_side,
+        }
+    }
+
+    fn paint(&mut self, paint_ctx: &mut PaintCtx, data: &AppData, _env: &Env) {
+        let size: Size = paint_ctx.size();
+        let w0 = size.width / 40.0;
```
cmyr: could this be `GRID_SIZE`?

cmyr:
@kindlychung there are a couple little outstanding things to poke at here; additionally #620 is going to change the default behaviour of the slider, so if you want it to be wide you're going to have to give it an explicit width.

druid/examples/game_of_life.rs
kindlychung: yeah, i used an iterator before.

kindlychung: woudn't that result in the cloning of colors every time paint is called?

druid/examples/game_of_life.rs
kindlychung: yeah, it's done.

druid/examples/game_of_life.rs
kindlychung: never mind, i was wrong, a simple `&self.color_scheme` solves the problem.

kindlychung:
> @kindlychung there are a couple little outstanding things to poke at here; additionally #620 is going to change the default behaviour of the slider, so if you want it to be wide you're going to have to give it an explicit width.

Not sure what I need do about this. I saw this in #620  : 

```rust
        let height = env.get(theme::BASIC_WIDGET_HEIGHT);
        let default_width = 100.0;
        let width = env.get(theme::WIDE_WIDGET_WIDTH);
        bc.constrain((width, height))
```

So there is already a default WIDE_WIDGET_WIDTH in the theme, right?

cmyr:
Thanks, nice to have some more useful examples!

Issue #547, opened 2020-02-12
Remove WinHandler::connected method


cmyr:
This was added because `WinHandler::connect` previously didn't have access to a `WinCtx`, which meant that we couldn't actually send an event into druid from it. Now that `WinCtx` is gone, we can remove `connected`.

PR #597, opened 2020-03-03
Remove WinHandler::connected method

pyroxymat:
Solves issue #547 

I am not sure how to verify if my code still runs on Mac and Windows since I use Linux for development.

### DIFF
--- a/druid/src/win_handler.rs
+++ b/druid/src/win_handler.rs
@@ -602,8 +602,6 @@ impl<T: Data> crate::shell::AppHandler for AppHandler<T> {
 
 impl<T: Data> WinHandler for DruidHandler<T> {
     fn connect(&mut self, handle: &WindowHandle) {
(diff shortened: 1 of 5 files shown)


druid/src/win_handler.rs
```diff
@@ -606,9 +606,7 @@ impl<T: Data> WinHandler for DruidHandler<T> {
         //it's possible that we should move those calls to occur in connected?
```
cmyr: we can also delete this comment while we're in here

pyroxymat: Whoops that comment is indeed outdated. I removed it.

cmyr:
huh, CI is in a weird state, I'm not really sure what's going on 😕

cmyr:
I'm going to go ahead and merge this, I think ci is just borked.

Issue #536, opened 2020-02-10
unify WindowDesc and PendingWindow


cmyr:
A `WindowDesc` is a description of a window that does not exist. A `PendingWindow` is a collection of state for a window that does not exist. A `PendingWindow` is created out of a `WindowDesc`, and exists only for as long as it takes for the platform to create and return an actual window handle.

Basically: we could get rid of `PendingWindow`, and just let `WindowDesc` serve both purposes, if we want?

This is basically just clean up, and is very non-urgent, but if someone is interested in doing some refactoring I think it makes sense.

PR #598, opened 2020-03-03
Unify WindowDesc and PendingWindow

pyroxymat:
Solves issue #536

### DIFF
--- a/druid/src/win_handler.rs
+++ b/druid/src/win_handler.rs
@@ -93,7 +93,6 @@ impl<T> Windows<T> {
     fn connect(&mut self, id: WindowId, handle: WindowHandle) {
         if let Some(pending) = self.pending.remove(&id) {
             let win = Window::new(id, handle, pending);
-            // let win = pending.into_window(id, handle);
             assert!(self.windows.insert(id, win).is_none(), "duplicate window");
         } else {
             log::error!("no window for connecting handle {:?}", id);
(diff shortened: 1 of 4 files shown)


druid/src/win_handler.rs
```diff
@@ -85,21 +85,22 @@ struct Inner<T> {
...
+    pending: HashMap<WindowId, WindowDesc<T>>,
     windows: HashMap<WindowId, Window<T>>,
 }
 
 impl<T> Windows<T> {
     fn connect(&mut self, id: WindowId, handle: WindowHandle) {
         if let Some(pending) = self.pending.remove(&id) {
-            let win = pending.into_window(id, handle);
+            let win = Window::new(id, handle, pending);
+            // let win = pending.into_window(id, handle);
```
cmyr: :grin:

druid/src/window.rs
```diff
@@ -53,27 +46,14 @@ pub struct Window<T> {
...
+impl<T> Window<T> {
+    pub(crate) fn new(id: WindowId, handle: WindowHandle, desc: WindowDesc<T>) -> Window<T> {
         Window {
             id,
-            root,
-            size: Size::ZERO,
-            title,
-            menu,
+            root: WidgetPod::new(desc.root),
+            size: desc.size.unwrap_or(Size::ZERO),
```
cmyr: I think we want to keep this as `Size::ZERO`; the size in the window description is passed to the window builder, but is not the actual size of the window itself; we will get that back from the platform once the window connects.

pyroxymat: I changed it back to Size::ZERO

cmyr:
oh one other thing: does the multiwin example still work after this change? I'll check in a few minutes but feel free to check first if you're around.

cmyr:
thanks!

Issue #475, opened 2020-01-24
X11 backend tracking issue


crsaracco:
This is a tracking issue for implementing an X11 `druid-shell` backend.

I expect that I'll be doing a lot of the implementation, but others are free to jump in as well!

### Background / Motivation

[VST](https://en.wikipedia.org/wiki/Virtual_Studio_Technology) plugins are always given a platform-native window to attach to (see #468).

Druid has platform-native backends for [Mac OS](https://github.com/xi-editor/druid/tree/master/druid-shell/src/platform/mac) (cocoa) and [Windows](https://github.com/xi-editor/druid/blob/master/druid-shell/src/platform/windows/window.rs) (winapi), but uses [GTK](https://github.com/xi-editor/druid/tree/master/druid-shell/src/platform/gtk) as the backend for Linux, which is non-"native".

In order to use Druid for VST plugin GUIs on X11 platforms, we need to implement an X11 `druid-shell` backend.

~~X11 windows are also compatible inside Wayland, so I *think* this should work for VSTs in Wayland environments, although I haven't tested this myself yet.~~ (maybe not, see comments)

See #468 for some more VST-related background/motivation. Others may also find having a native X11 backend useful for other use cases.

### Implementation details

I intend to keep GTK as the main backend for Linux, and only enable X11 if users specifically request it through a feature flag.

I'm planning to use the [`rust-xcb`](https://github.com/rtbo/rust-xcb) crate for interfacing with X11/XCB.

### Progress

Each check mark corresponds roughly to one PR. This list might be incomplete; I will add to it as I discover more requirements.

 - [x] (PR: #599) Bare-minimum X11 backend implemented
    - backend is behind a feature flag, basic window handling implemented, painting to the window implemented, and mouse + keyboard events are handled. Some hacks were involved in completing this, which will be handled in future bullet points / PRs.
 - [x] (Issue: #829) Fix program not terminating when the window closes ([users have to `^C` to get it to quit](https://github.com/xi-editor/druid/pull/599#issuecomment-594309564))
 - [ ] Fix window syncing hack
    - Currently figure out the screen's refresh rate and **sleep** (!) for however long based on the refresh rate, which is probably not what we want to do. Better way to sync draws?
    - This might also be what's causing the window to [blink a lot during re-rendering](https://github.com/xi-editor/druid/pull/599#issuecomment-609376693); if fixing this hack doesn't solve that issue, make a new task in this list.
 - [ ] Menu support
[...]

Issue #468, opened 2020-01-20
Support for attaching to a parent window


crsaracco:
This is a tracking issue for supporting passing in a `parent` pointer when creating a window. This functionality will enable users to use Druid as a child window for platform-native windows *not* created by Druid.

### Background / Motivation

This functionality has been desired by many in the [RustAudio](https://github.com/RustAudio) group for a while now: it is a requirement for creating GUIs in the [VST](https://en.wikipedia.org/wiki/Virtual_Studio_Technology) plugin format, one of the most popular plugin formats in modern Digital Audio Workstations (DAWs). Currently, no GUI crate in the Rust ecosystem fully meets all of the requirements to create GUIs for VST plugins, so this will be a step towards making Druid the first one.

*(A "plugin", here, is basically a third-party musical "device" that the user can load into the DAW to synthesize or alter sounds. Plugins typically have a GUI that enables the user to change parameters for this device so that they can tweak the sounds/effects as they see fit, among other features.)*

In DAWs following the VST specification, the DAW is responsible for handling the lifecycle of all plugin GUI windows: creation, opening, closing, etc. When the user loads a plugin into the DAW, the DAW probes the plugin to figure out if it supports having a GUI window. If so, the DAW creates the window for the plugin, then passes the window's "parent" pointer to the plugin so that it can connect to that window.

This parent pointer is always a pointer to a platform-native window (`HWND` on Windows, `NSView` on Mac OS, or an X11 window ID on X11 systems), which can be used in the platform-native window creation function to "connect" to the parent window. (examples: [Windows](https://github.com/crsaracco/vst2-gui-prototypes/blob/master/windows-opengl-vst/src/editor/window/win32_window.rs#L23-L36), [Mac OS](https://github.com/rust-dsp/rtb-rs/blob/master/src/platform/cocoa/mod.rs#L48-L79), [Linux/X11](https://github.com/crsaracco/vst2-gui-prototypes/blob/master/linux-opengl-vst/src/editor/window/mod.rs#L112-L114))

Therefore, the ideal workflow from a VST developer's perspective would be:

 1. DAW loads the plugin, opens a window, and passes a pointer to that window to the plugin.
 2. The plugin code takes that parent pointer and passes it to Druid via Druid's app creation API; Druid then handles the rest of the window creation/attaching process.
 3. The plugin code now has access to a window that it can draw on, using standard Druid API calls. The GUI portion of the plugin now functions as a typical Druid application.

[...]

PR #599, opened 2020-03-04
X11 implementation of druid-shell

crsaracco:
This PR implements the first bullet point of #475: "bare-minimum X11 backend".

As stated in #475, I am opting to keep GTK as the main Linux backend since it is much more feature-complete; you have to explicitly ask for this new X11 backend by using `--features use_x11`.

I had to make some architectural hacks in the `druid-shell` code to make this work, so I *don't* expect this to plug right in to `druid` itself quite yet. I hope to either hash this problem out in this PR, or identify a plan of attack going forward to align all of the backends together so that they all work in `druid`. More details on this below.

@cmyr laid out a good goal for this first PR in Zulip, so I've tried to follow it as closely as possible here.

> I think a good goal for an initial PR would be that all of the interfaces are present (e.g. the crate can compile) and that the most essential functionality is in place; this would mean at least window handling, painting, and keyboard and mouse event handling.

Some other stuff isn't implemented quite yet -- but again, more details below.

## What is implemented

 - All interfaces are present, and the crate compiles! ... at least `druid-shell` does. I don't think it's compiling quite yet in `druid`, because of the hacks I had to write (see below). This is okay-ish for now though, since this code is all behind a feature flag. You can test this with `cargo run --example perftest --features use_x11` in the `druid-shell` directory on an X11 system.
 - Most of the window handling functionality is in place: `WindowBuilder`, `set_title`, `show`, `close`, `bring_to_front_and_focus`. I go over stuff that is explicitly *not* implemented yet in the next section.
 - Painting the window works (including animations), with two exceptions:
    - The draw surface doesn't resize when the window does; it will always stay the same as the initial window dimensions. This should be a relatively easy fix (I need to get the resize dimensions somehow, and then I either need to pass the dimensions to `render()` or to `XWindow`; haven't decided the better API yet).
    - Requesting the next redraw during animations is currently a terrible hack, but I'll go over that below.
 - Keyboard events are working, thank you @perlindgren! There was some discussion about the best way to implement this in the future, so this feature is currently in more of a "proof-of-concept, hey-look-we-can-handle-keyboard-events" state, rather than a "100% finished product" state. We expect that the actual implementation might change in the future, but some more discussion is needed I think.
 - Mouse events are roughly in the same state as keyboard events, AFAIK (but thanks again to @perlindgren for implementing this too!)

## What is explicitly *not* implemented yet

 - menus: I *think* we're going to have to draw these ourselves for X11 (???). There may be some other way to do it by requesting the window manager to do it for you, but I didn't dig too deep into it yet. Suggestions welcome here.
 - errors
 - timers
 - the "idle handle"
 - custom cursors
 - file dialogs
 - DPI scaling
 - the clipboard
 - locales (I think this is still a TODO for the rest of `druid`?)

## Ugly, ugly hacks

 - X11 uses a "global" XCB connection to the X server (see `application.rs`). We handle events by fetching them from this global connection, parsing the event, and figuring out which window to route the event to for event handling purposes. Because of this, I need to store a map from window ID to `XWindow` inside the RunLoop object. Because of *this*, I needed to pass the `RunLoop` instance into `WindowBuilder`, so that we can connect the new window in that map, so that we can handle events. This breaks the API that `druid` was using previously, because `WindowBuilder::build` now has a new argument. This causes us to not be able to use the X11 backend in `druid` itself, which I would like to resolve as part of this PR. What do?
    - Grep for `TODO(x11/architecture)` to see areas of code that are affected by this.
 - AFAIK, there is no good way to sync window redrawing to the screen's refresh rate. AFAIK, there's also no system timer to use in X11 to schedule stuff to happen asynchronously in the future. To **(temporarily)** remedy this, I just manually figure out the screen's refresh rate and **sleep** (!) for however long that is. I think I'm going to need to write some sort of redraw scheduler to handle this for a multi-window setup? Suggestions definitely welcome here.

## Next steps

This feature is partly an exploration into whether or not we can use Druid for VST plugins; we plan on working on #468 next, and make a proof-of-concept VST that works in a Digital Audio Workstation (DAW). We are hoping that this works out pretty smoothly, but there may be opportunities to identify some issues with Druid architecture (for certain use-cases) if not.


crsaracco:
One additional thing I found: Closing the window via the window manager's `X` button doesn't terminate the program. You have to manually `^C` it to get it to quit.

Is this something I missed in `druid-shell`, or is this something I need to go implement in XCB code?

druid-shell/Cargo.toml
```diff
@@ -11,6 +11,7 @@ edition = "2018"
 
 [features]
 use_gtk = ["gtk", "gtk-sys", "gio", "gdk", "gdk-sys", "glib", "glib-sys", "cairo-rs"]
+use_x11 = ["xcb", "cairo-sys-rs"]
```
cmyr: I would like to just call this feature `x11`. `use_gtk` should also be called `gtk`, I didn't understand at the time that an optional crate was automatically a feature.

(this is not to say you should change `use_gtk`, just not to treat it as an example).

druid-shell/Cargo.toml
```diff
@@ -43,8 +46,9 @@ cocoa = "0.20.0"
 objc = "0.2.5"
 cairo-rs = { version = "0.8.0", default_features = false }
 
+# TODO(x11/dependencies): only use feature "xcb" if using XCB
 [target.'cfg(target_os="linux")'.dependencies]
-cairo-rs = {  version = "0.8.0", default_features = false }
+cairo-rs = {  version = "0.8.0", default_features = false, features = ["xcb"] }
```
cmyr: it's annoying that we need to always have these on for linux; it means that our x11 backend depends on gtk. I'm not sure what the best solution is.

druid-shell/examples/perftest.rs
```diff
@@ -118,7 +118,16 @@ fn main() {
     builder.set_handler(Box::new(perf_test));
     builder.set_title("Performance tester");
 
-    let window = builder.build().unwrap();
-    window.show();
+    // TODO(x11/architecture): super hacky way to connect the XWindow to the RunLoop. Better way to do it?
+    cfg_if::cfg_if! {
+        if #[cfg(all(target_os = "linux", feature = "use_x11"))] {
```
cmyr: I'm confused that this is still here, I thought I removed it?? In any case we want all backends to have the same public api. (Other backends currently use some global state to handle things like this; it isn't perfect but if we're going to rethink that we should do it all at once.)

druid-shell/examples/shello.rs
```diff
@@ -64,9 +64,17 @@ impl WinHandler for HelloState {
     }
 
     fn key_down(&mut self, event: KeyEvent) -> bool {
-        let deadline = std::time::Instant::now() + std::time::Duration::from_millis(500);
-        let id = self.handle.request_timer(deadline);
-        println!("keydown: {:?}, timer id = {:?}", event, id);
+        // TODO(x11/timers): Timers not implemented yet on X11, so this won't work there.
+        cfg_if::cfg_if! {
```
cmyr: okay this is some merge mess? I have no idea what happened. 😕

druid-shell/examples/shello.rs
```diff
@@ -133,8 +141,17 @@ fn main() {
     builder.set_handler(Box::new(HelloState::default()));
     builder.set_title("Hello example");
     builder.set_menu(menubar);
-    let window = builder.build().unwrap();
 
-    window.show();
+    // TODO(x11/architecture): super hacky way to connect the XWindow to the RunLoop. Better way to do it?
```
cmyr: ditto this, I could've sworn I deleted this?

druid-shell/src/platform/x11/application.rs
```diff
@@ -0,0 +1,152 @@
...
+        let conn = XCB_CONNECTION.connection_cloned();
+        loop {
+            if let Some(ev) = conn.wait_for_event() {
+                let ev_type = ev.response_type() & !0x80;
+                //NOTE: I don't think we should be doing this here, but I'm trying to keep
+                //the code mostly unchanged. My personal feeling is that the best approach
+                //is to dispatch events to the window as early as possible, that is to say
+                //I would send the *raw* events to the window and then let the window figure
+                //out how to handle them. - @cmyr
+                match ev_type {
```
cmyr: so to expand on this comment: I think we should make this runloop as minimal as possible, and basically dispatch raw events to our internal window object with as little handling as possible, and do all this logic there.

In particular I feel like this will make it easier to eventually support the VST case; the window object can basically stand alone.

druid-shell/src/platform/x11/application.rs
```diff
@@ -0,0 +1,152 @@
...
+        }
+    }
+
+    pub fn quit() {
+        // No-op.
+    }
+
+    pub fn clipboard() -> Clipboard {
+        // TODO(x11/clipboard): implement Application::clipboard
+        unimplemented!();
```
cmyr: for things like this I think it makes more sense to do `log::warn` with a message about what is unimplemented instead of using the unimplemented macro, which will just crash us? 🤷‍♂

druid-shell/src/platform/x11/error.rs
```diff
@@ -0,0 +1,30 @@
...
+#[derive(Debug, Clone)]
+pub enum Error {
+    // TODO(x11/errors): enumerate `Error`s for X11
+    NoError,
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, _f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+        // TODO(x11/errors): implement Error::fmt
+        unimplemented!();
```
cmyr: I would just do like `write!(f, "X11 Error")` here; this is going to panic in a situation where we are most likely already panicking, which isn't too helpful.

druid-shell/src/platform/x11/keycodes.rs
```diff
@@ -0,0 +1,335 @@
...
+use crate::keyboard::StrOrChar;
+use crate::keycodes::KeyCode;
+
+pub type RawKeyCode = u32;
+
+impl From<RawKeyCode> for KeyCode {
+    #[allow(clippy::just_underscores_and_digits, non_upper_case_globals)]
+    fn from(raw: u32) -> KeyCode {
+        match raw {
+            9 => KeyCode::Escape,
```
cmyr: for these kinds of magic numbers I find it helpful to point to some documentation source that supports the mapping used.

druid-shell/src/platform/x11/keycodes.rs
```diff
@@ -0,0 +1,335 @@
...
+
+            _ => {
+                log::warn!("Warning: unknown keyval {}", raw);
+                KeyCode::Unknown(raw)
+            }
+        }
+    }
+}
+
+impl Into<StrOrChar> for KeyCode {
```
cmyr: a few things going on here: `StrOrChar` is not intended to be used directly; it is a way to allow the `KeyEvent` constructor to take either a `str` or a `char`.

More importantly, `KeyCode` cannot be used on its own to determine the string value of a key press. The `KeyCode` refers to a specific physical key; to interpret this we need to map it through the user's current locale and keyboard layout. (I think we talked about this? It's on zulip somewhere)

druid-shell/src/platform/x11/keycodes.rs
```diff
@@ -0,0 +1,335 @@
...
+            KeyCode::NumpadMultiply => KP_Multiply,
+            KeyCode::NumpadDivide => KP_Divide,
+            KeyCode::NumLock => Num_Lock,
+            KeyCode::NumpadEnter => KP_Enter,
+
+            KeyCode::ArrowUp => Up,
+            KeyCode::ArrowDown => Down,
+            KeyCode::ArrowLeft => Left,
+            KeyCode::ArrowRight => Right,
+            KeyCode::Unknown(_) => unreachable!(
```
cmyr: The intention here is that the value of the variant *is* the raw key code for this platform, and you should just return it directly.

druid-shell/src/platform/x11/runloop.rs
```diff
@@ -0,0 +1,94 @@
+// Copyright 2018 The xi-editor Authors.
```
cmyr: I'm very confused, this should have been deleted 😖

druid-shell/src/platform/x11/window.rs
```diff
@@ -0,0 +1,407 @@
...
+
+        let xwindow = XWindow::new(window_id, self.handler.unwrap(), self.size);
+        Application::add_xwindow(window_id, xwindow);
+
+        Ok(WindowHandle::new(window_id))
+    }
+}
+
+// X11-specific event handling and window drawing (etc.)
+pub struct XWindow {
```
cmyr: this shouldn't need to be public; at most it should be `pub(crate)`.

druid-shell/src/platform/x11/window.rs
```diff
@@ -0,0 +1,407 @@
...
+#[derive(Clone)]
+pub struct IdleHandle;
+
+impl IdleHandle {
+    pub fn add_idle_callback<F>(&self, _callback: F)
+    where
+        F: FnOnce(&dyn Any) + Send + 'static,
+    {
+        // TODO(x11/idle_handles): implement IdleHandle::add_idle_callback
+        unimplemented!();
```
cmyr: again I'd just log here instead of panicking

druid-shell/Cargo.toml
crsaracco: Done; changed all `use_x11` to just `x11`

druid-shell/examples/perftest.rs
crsaracco: Fixed in my branch :)

druid-shell/examples/shello.rs
crsaracco: Fixed in my branch :)

druid-shell/src/platform/x11/runloop.rs
crsaracco: Lol -- updated all files in `druid-shell/src/platform/x11` to 2020. Unless you wanted all of those headers deleted, in which case I could do that too.

druid-shell/Cargo.toml
crsaracco: Hmm, I guess we could technically just split that and make more-specific `cfg` targets?

druid-shell/src/platform/x11/runloop.rs
crsaracco: Oh, you meant the entire `runloop.rs` file. Hmmm

crsaracco: Ok, yep, I was able to just delete it!

druid-shell/examples/shello.rs
crsaracco: Nope, this code currently breaks because we don't have timers implemented in X11 yet, so I `cfg_if`'d it away for X11.

druid-shell/src/platform/x11/application.rs
crsaracco: Well -- to get which window we need to send the event to, we actually need to parse the event within this loop and fetch the window ID out of it (`expose.window()`, `key_press.event()`, `button_press.event()` etc)

We *could* send the raw event and let the window parse the event again, though, in the spirit of handling everything in the window -- it's probably not too much overhead to parse twice.

druid-shell/src/platform/x11/application.rs
crsaracco: Changed *all* `unimplemented!()`s to `log::warn!()`s.

Might be overboard? Let me know if I should delete any.

druid-shell/src/platform/x11/error.rs
crsaracco: Done.

druid-shell/src/platform/x11/keycodes.rs
crsaracco: Gonna have to defer to @perlindgren here

druid-shell/src/platform/x11/keycodes.rs
crsaracco: (I'm unsure what to do here too -- paging @perlindgren for help)

druid-shell/src/platform/x11/keycodes.rs
crsaracco: This one I was able to handle myself ;)

druid-shell/src/platform/x11/window.rs
crsaracco: Done.

druid-shell/src/platform/x11/window.rs
crsaracco: Done.

druid-shell/Cargo.toml
cmyr: unfortunately you can't use feature targets in this way. :/

druid-shell/examples/shello.rs
cmyr: what I would do is implement the API, returning a timer token, but also log a warning that timers don't work?

druid-shell/src/common_util.rs
```diff
@@ -21,7 +21,8 @@ use std::sync::atomic::{AtomicU64, Ordering};
 /// Strip the access keys from the menu string.
 ///
 /// Changes "E&xit" to "Exit". Actual ampersands are escaped as "&&".
-#[cfg(any(target_os = "macos", target_os = "linux"))]
+// TODO(x11/menus): Re-evaluate this #[cfg]; used to be just "macos OR linux" before.
+#[cfg(any(target_os = "macos", all(target_os = "linux", not(feature = "x11"))))]
```
cmyr: at this point it might just be easier to have a `#[allow(dead_code)]`. 😒

druid-shell/src/dialog.rs
```diff
@@ -23,6 +23,8 @@ pub struct FileInfo {
 }
 
 /// Type of file dialog.
+// TODO(x11/file_dialogs): Remove this #[cfg]
+#[cfg(not(all(target_os = "linux", feature = "x11")))]
```
cmyr: does this matter? do things complain if this exists but is unused?

druid-shell/src/platform/x11/application.rs
cmyr: Hmm, okay. Another option I might consider is to have our own `XcbEvent` enum; we would parse these raw events into that type here, extract the window id, and then forward the event to the window?

I'm just spitballing, really, I don't have super-strong opinions.

druid-shell/src/dialog.rs
crsaracco: Clippy does, yep. Could make it an `#[allow(dead_code)]` like the above instead?

druid-shell/src/common_util.rs
crsaracco: haha, fair. Can do.

druid-shell/examples/shello.rs
crsaracco: Oh right, I *did* just do that :P I'll look into it.

druid-shell/Cargo.toml
crsaracco: Ah, dang. Sounds like it should be a feature. Wonder if there's a Rust / Cargo feature request laying around somewhere...

druid-shell/src/dialog.rs
cmyr: I'm confused, if it's a public type clippy shouldn't complain? is it not exposed anywhere? Is there some place where it's used on other platforms but which isn't implemented here?

druid-shell/src/platform/x11/application.rs
crsaracco: I'm thinking it's very possible that I'll have to work in this area when I'm exploring VST compatibility, so I'm leaning towards re-thinking this then? Agree that it's not ideal, though.

druid-shell/src/dialog.rs
crsaracco: Seems to work now without it, not sure what the problem was before. Removed.

druid-shell/examples/shello.rs
crsaracco: Reverted.

druid-shell/src/common_util.rs
crsaracco: Done.

druid-shell/src/dialog.rs
crsaracco: Oops, I lied. Failing clippy as of the last commit

crsaracco: Fixed.

Finnerale:
I've been playing around with the examples using the X11 back end *on Wayland*
Will try it on X11 next.

Issues:
- Window flashes white when resizing quickly
- Resizing the window on the left site makes it blink temporarily towards the direction I drag on the right. This goes completely crazy when sizing it < 0 width, the window just flies away to the right :joy: 
    Does not happen when resizing on the right side
- Application does not exit when closing the window
- Some background flashing when scrolling (dragging the scroll bar) or when hovering over buttons  in the list demo

Maybe issues, maybe not indented yet:
- No keyboard modifiers? At least `shift` does not work
- No scrolling

Finnerale:
So on X11 could confirm the same issues but:
- They are all much more subtle, especially the flashing
- The issue with the window flying exists here as well but behaves slightly different:
    On Wayland the window gets smashed to the right end of my monitor, then stays there
    On X11 the window keeps cycling through my monitors until I stop
    In both cases I can make it large again by dragging left

druid-shell/src/platform/x11/window.rs
```diff
@@ -0,0 +1,444 @@
...
+}
+
+impl XWindow {
+    pub fn new(window_id: u32, window_handler: Box<dyn WinHandler>, size: Size) -> XWindow {
+        let conn = Application::get_connection();
+        let setup = conn.get_setup();
+        let screen_num = Application::get_screen_num();
+        // TODO(x11/errors): Don't unwrap for screen or visual_type?
+        let screen = setup.roots().nth(screen_num as usize).unwrap();
+        let mut visual_type = get_visual_from_screen(&screen).unwrap();
```
Finnerale, line 167: This is not used anywhere and the compiler complains about it, I suppose this `visual_type` can go away?

druid-shell/src/platform/x11/window.rs
```diff
@@ -0,0 +1,444 @@
...
+            )
+        };
+        let cairo_surface = cairo::XCBSurface::create(
+            &cairo_xcb_connection,
+            &cairo_drawable,
+            &cairo_visual_type,
+            size.width as i32,
+            size.height as i32,
+        )
+            .expect("couldn't create a cairo surface");
```
Finnerale, line 217: This and one line in `application.rs` needs a `cargo fmt`

druid-shell/src/common_util.rs
```diff
@@ -21,6 +21,7 @@ use std::sync::atomic::{AtomicU64, Ordering};
 /// Strip the access keys from the menu string.
 ///
 /// Changes "E&xit" to "Exit". Actual ampersands are escaped as "&&".
+#[allow(dead_code)]
 #[cfg(any(target_os = "macos", target_os = "linux"))]
 pub fn strip_access_key(raw_menu_text: &str) -> String {
```
Finnerale, line 26: Seems like you can actually do
```rust
#[cfg(not(feature = "x11"))]
#[cfg(any(target_os = "macos", target_os = "linux"))]
pub fn strip_access_key(raw_menu_text: &str) -> String {
``` 
Which seems better to me than using `dead_code`

crsaracco: Oh yeah, that's a lot nicer than the `#[cfg]` mess I made in a previous commit, haha. Done.

druid-shell/src/platform/x11/window.rs
crsaracco: Done

druid-shell/src/platform/x11/window.rs
crsaracco: Whoops, missed it when I moved this chunk of code to a different function. Deleted (and a few other variables too).

druid-shell/src/platform/x11/application.rs
```diff
@@ -0,0 +1,210 @@
...
+    screen_num: i32,
+}
+
+lazy_static! {
+    static ref XCB_CONNECTION: XcbConnection = XcbConnection::new();
+}
+
+thread_local! {
+    static WINDOW_MAP: RefCell<HashMap<u32, XWindow>> = RefCell::new(HashMap::new());
+}
```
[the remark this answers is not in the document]
crsaracco: We need to call `get_connection`, `get_screen_num`, and `add_xwindow` from other modules, and (a reference to) `Application` doesn't seem to be passed around at this time -- so those methods need to be static.

Ideally these can just move to member fields in `Application` in the future, though.

crsaracco:
Okay, I think a recap would be useful here:

 - First, the PR description/comment/thing is a bit out of date now. I can update it if anyone thinks it would be worth it (for referencing in the future, for example)
 - Closing the window doesn't actually terminate the program; I'm not sure if I missed some `druid`/`druid-shell` code to do this, or if I missed some `xcb` code to do this. Any insight? If it's not something obviously druid-related, it's probably XCB-related.
 - Keyboard and mouse event support are in a "proof-of-concept" phase, and will likely need some work in the future to make it production-grade. There are some lingering and unresolved conversations buried in the Zulip history about it, so we'll probably have to dig those up. I think we can leave these as-is for this first PR though -- it should be good enough to start POCing VSTs (and I don't think anyone else currently has an X11 backend usecase yet?).
   - In particular: nope, no keyboard modifiers yet. Not sure how complicated that is -- might be easy, might be difficult.
   - Scrolling support in particular isn't implemented yet, but that actually should be a fairly quick fix -- you *should* just need to follow the pattern of the other mouse/keyboard events and enable the scroll-related mask in the window event mask (I'd have to look up which one it is). Could fix it now or leave it for now, indifferent.
 - Window redraw flashes: I noticed this on my system too. It might be related to the terrible way I'm refreshing the screen for animations (sleeping), or it may be that I just need to figure out double-buffering the draws or something. Good enough to POC VSTs IMO, but definitely need to be fixed in the future.
 - The window flying error sounds hilarious, haha. I couldn't repro on my DE (Cinnamon), but that should probably be fixed too. VSTs aren't typically resized, so we probably won't notice that there.

I think that's everything -- does anyone consider any of these blocking? I can see what I can do to implement them; otherwise, I'll make sure they're noted in #475.

Finnerale:
@cmyr I think we could merge this now.

Might be interesting to create issues for the things that still need to be done and maybe create a project for the X11 back end, that could make further contributions to this a lot easier (although that is quite some organizational work :sweat_smile: )

crsaracco:
I was thinking the same thing, and would be happy to help if possible!

cmyr:
Sounds good, and would be very happy to have any issue opened for outstanding stuff.

crsaracco:
I've updated the "Progress" list in #475 to reflect all of the findings here.

As of right now there are 15 unchecked items in that list -- I'm not sure if you want me to flood the issue tracker writing up stuff for all 15, but I'd be happy to if that's what's desired.

For now, I figure what I'll do is "trickle" the issues into the tracker and write up the N things I find most important to fix next.

PR #601, opened 2020-03-04
Allow setting child spacing in Flex widget

cmyr:
This lets you specify the space betweeeen children. There are a number
of ways we could do this; I've chosen the simplest, but also the
least flexible.

If you need to customize the padding, you can manually add SizedBox
widgets where needed.

Screen Shot 2020-03-04 at 3 29 58 PM


cmyr:
I'm not happy with this, going to do something more... flexible 😷.

PR #602, opened 2020-03-04
Use KeyOrValue to set Label size and color

futurepaul:
First stop on the `KeyOrValue` tour!

This is a direct followup to #592, which implemented but did not use `KeyOrValue`.

Had a little fun with the demo. Is this our first example of dynamic styling? Stepper seemed more appropriate than a slider here, but I would like to improve the numeric textbox behavior.

### DIFF
--- a/druid/examples/styled_text.rs
+++ b/druid/examples/styled_text.rs
@@ -12,11 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//! Example of dynamic text styling
+
 use druid::widget::{Alignment, Flex, Label, Parse, Stepper, TextBox, WidgetExt};
 use druid::{
-    theme, AppLauncher, Data, Lens, LensExt, LensWrap, LocalizedString, PlatformError, Widget,
+    theme, AppLauncher, Data, Key, Lens, LensExt, LensWrap, LocalizedString, PlatformError, Widget,
     WindowDesc,
 };
+
+// This is a custom key we'll use with Env to set and get our text size.
+const MY_CUSTOM_TEXT_SIZE: Key<f64> = Key::new("styled_text.custom_text_size");
 
 #[derive(Clone, Lens, Data)]
 struct AppData {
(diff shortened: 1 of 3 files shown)


druid/examples/styled_text.rs
```diff
@@ -0,0 +1,83 @@
...
+        Label::new(|data: &String, _env: &_| format!("Default: {}", data)).lens(AppData::text);
+
+    // The text_color and text_size builder methods can override the defaults
+    // with either a theme key (like we're doing here), or a concrete value
+    // (Color and f64, respectively).
+    let styled_label =
+        Label::new(|data: &AppData, _env: &_| format!("Size {:.1}: {}", data.size, data.text))
+            .text_color(theme::PRIMARY_LIGHT)
+            .text_size(theme::TEXT_SIZE_LARGE)
+            .env_scope(|env: &mut druid::Env, data: &AppData| {
```
cmyr: this is interesting: I see why it's necessary, but it also feels kind of bad.

My basic feeling is that we should ultimately not be overriding the text sizes in the environment except in very specific circumstances; what I would suggest here is defining our own `Key` ('MY_CUSTOM_TEXT_SIZE')? and using that here. This will also help illustrate a bit better how the env works?

druid/src/widget/label.rs
```diff
@@ -80,9 +84,21 @@ impl<T: Data> Label<T> {
         self
     }
 
+    /// Set text color.
```
cmyr: I would indicate that these are 'builder-style' methods, and I would also consider having mutating versions of these, called `set_text_color` and `set_text_size`, for when we want to modify things directly at runtime? Although it's possible we want that to only be done through the environment; I'm not totally sure what patterns we'll want?

My *hunch* is that it makes sense to allow a parent to modify these on a child directly when needed, but I'm not 100% confident..

futurepaul: I went with adding the setters. I figure if it causes problems or we never use it in practice we can remove?

druid/examples/styled_text.rs
futurepaul: I'm following you. I think this did help illustrate Env better!

druid/src/widget/label.rs
```diff
@@ -80,9 +84,47 @@ impl<T: Data> Label<T> {
...
+    pub fn set_text_color(&mut self, color: impl Into<KeyOrValue<Color>>) {
+        self.color = color.into();
+    }
+
+    /// Builder-style method for setting the text size.
+    ///
+    /// The argument can be either an `f64` or a [`Key<f64>`].
+    ///
+    /// [`Key<f64>`]: struct.Key.html
+    pub fn text_size(mut self, size: impl Into<KeyOrValue<f64>>) -> Self {
```
cmyr: okay one last nit, but I like to have all the builder methods at the top so that they appear together in the documentation.

futurepaul: haha so very nit

Issue #529, opened 2020-02-08
Add LifeCycle::RouteWidgetAdded


cmyr:
`WidgetAdded` is dispatched from the root of the tree, which means that it is handled by every widget on the way down. This isn't bad, but it could be confusing for container widgets, which will require some other way of checking if this means that _they_ were added, or just some child.

It would be nice if we could make it an invariant that `WidgetAdded` will be both the first thing a widget sees, as well as something that will only ever be seen once.

Fortunately, this has a pretty simple solution: instead of dispatching `WidgetAdded` directly, we should instead dispatch `RouteWidgetAdded`. In `WidgetPod`, when we receive this event, we should check if `self.old_data.is_none()`. If the check is `true`, it means that our widget (and by definition all of our widget's children) are brand new, and we can replace `RouteWidgetAdded` with normal `WidgetAdded` for that section of the tree.

If the check is `false`, then we check if we have `children_changed` set, and if so we forward the `RouteWidgetAdded` event down to our widget.

Normal widgets should never need to know that `RouteWidgetAdded` exists; they just need to be sure that they forward events to any children they might have.

If someone is interested in this, it will be a  good way to  learn about how event flow works, and it shouldn't be too hard; otherwise I'll get around to it in the next little while.

PR #603, opened 2020-03-05
Add RouteWidgetAdded dispatch

pyroxymat:
Solves #529

### DIFF
--- a/druid/src/core.rs
+++ b/druid/src/core.rs
@@ -350,7 +350,7 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
         if self.old_data.is_none() {
             log::error!(
                 "widget {:?} is receiving an event without having first \
-                 recieved WidgetAdded.",
+                 received WidgetAdded.",
                 ctx.widget_id()
             );
         }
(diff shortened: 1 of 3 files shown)


druid/src/event.rs
```diff
@@ -137,29 +137,30 @@ pub enum Event {
 #[derive(Debug, Clone)]
 pub enum LifeCycle {
     /// Sent to a `Widget` when it is added to the widget tree. This should be
-    /// the first message that each widget receives.
+    /// the first message that each widget receives. This should only be sent once.
+    /// If any new children are added a RouteWidgedAdded event will be sent instead.
```
cmyr: I appreciate the effort to try and keep the docs up to date; that's an important thing to do.

That said, I'm not 100% sure what the right docs are, here. My basic feeling is that these docs might be slightly outdated, since registering children is handled entirely in the `WidgetPod`.

So I think the easiest thing to do is to leave these docs as-is, and we can open an issue about them being slightly out of date. For `RouteWidgetAdded`, I would include minimal docs, similar to `RouteFocusChanged`; basically I would say something like, "Internal: used by the framework to route the `WidgetAdded` event."

We might also want to consider adding `#[doc(hidden)]` to the various 'route' events, but that's also a separate concern.

druid/src/core.rs
```diff
@@ -473,17 +473,29 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
                 r
             }
             LifeCycle::WidgetAdded => {
-                // if this is called, it means widgets were added. That might be
-                // be us; if we don't have old_data we set it now.
+                assert!(self.old_data.is_none());
```
cmyr: happy to see this, I'm  not 100% confident about some of the logic either so it's good to keep ourselves honest.

druid/src/core.rs
```diff
@@ -473,17 +473,29 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
...
+                // we just pass this event down
                 if self.old_data.is_none() {
-                    self.old_data = Some(data.clone());
-                    self.env = Some(env.clone());
-                }
-                if self.state.children_changed {
-                    self.state.children.clear();
-                    self.state.focus_chain.clear();
+                    self.lifecycle(ctx, &LifeCycle::WidgetAdded, data, env);
+                    return;
```
cmyr: oh, this is confusing? If we return here then we aren't going to end up registering ourselves, which normally happens at the end of this fn (where we call `ctx.register_child(self.id())`).

It's possible that we don't have a test case to catch this?

let me take a look and see, it's worth having if we don't.

cmyr: okay, I've opened #606 and I'm going to merge that shortly; when that's merged you can rebase this and then if that test is passing I'll be happy. :)

pyroxymat: I rebased on top of master!

Since we enter the function again recursively with a different argument (WidgetAdded instead of RouteWidgetAdded) we still end up at the end of the function for 'self'. So we end up registering ourselves I think.

cmyr: ah okay I didn't catch that, I read this as a call to the child's lifecycle fn.

PR #605, opened 2020-03-05
No inner align in control widgets

cmyr:
This removes inner alignment wrappers from control widgets; alignment should be handled explicitly by the user.

cc @futurepaul: can you think of anything (examples etc) that this might break, that we should update alongside?

### DIFF
--- a/druid/src/widget/progress_bar.rs
+++ b/druid/src/widget/progress_bar.rs
@@ -16,19 +16,21 @@
 
 use crate::kurbo::{Point, RoundedRect, Size};
 use crate::theme;
-use crate::widget::Align;
 use crate::{
     BoxConstraints, Env, Event, EventCtx, LayoutCtx, LifeCycle, LifeCycleCtx, LinearGradient,
     PaintCtx, RenderContext, UnitPoint, UpdateCtx, Widget,
 };
 
 /// A progress bar, displaying a numeric progress value.
+///
+/// This type impls `Widget<f64>`, expecting a float in the range `0.0..1.0`.
 #[derive(Debug, Clone, Default)]
-pub struct ProgressBar {}
+pub struct ProgressBar;
 
 impl ProgressBar {
-    pub fn new() -> impl Widget<f64> {
-        Align::vertical(UnitPoint::CENTER, Self::default())
+    /// Return a new `ProgressBar`.
+    pub fn new() -> ProgressBar {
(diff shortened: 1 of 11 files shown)


futurepaul:
Looks like you missed progress bar? Also, label is being centered vertically but not by an `Align` widget, just through the good old fashioned way: layout and paint methods.

Overall it looks like you're still returning the same size from all these, and therefore nothing should break. You're just by default top-aligned now. If these widgets returned only the size of their paint area (didn't honor bc.min()) that's when I'm not sure what would happen.

I ran all the widgets and it mostly looks good. The `slider.rs` example is a little busted because the checkbox is now top-aligned but the label it's next to is vertically centered. `switch.rs` has the opposite problem: the stepper widget never had the vertical align stuff, so it's been centered in the example.

druid/src/widget/progress_bar.rs
```diff
@@ -16,19 +16,21 @@
...
 #[derive(Debug, Clone, Default)]
-pub struct ProgressBar {}
+pub struct ProgressBar;
 
 impl ProgressBar {
-    pub fn new() -> impl Widget<f64> {
-        Align::vertical(UnitPoint::CENTER, Self::default())
+    /// Return a new `ProgressBar`.
+    pub fn new() -> ProgressBar {
+        Self::default()
```
futurepaul: Is there a reason this is Self::default() and checkbox is Default::default()?

druid/src/widget/textbox.rs
```diff
@@ -74,6 +60,18 @@ impl TextBox {
         }
     }
 
+    /// Builder-style method to set the `TextBox`'s placeholder text.
+    pub fn with_placeholder(mut self, placeholder: impl Into<String>) -> Self {
+        self.placeholder = placeholder.into();
+        self
+    }
+
+    #[deprecated(since = "0.5.0", note = "Use TxtBox::new instead")]
```
futurepaul: TextBox::new

futurepaul:
I'm not super familiar with all the spacing of Switch, so I'm just assuming that works.

druid/src/widget/progress_bar.rs
cmyr: zero reason!

cmyr:
I think the switch thing was a bug; basically there  was always extra padding on the right.

Issue #505, opened 2020-02-03
QUIT_APP command panics


xliiv:
Running
```rust
use druid::widget::{Button, Flex};
use druid::{AppLauncher, Widget, WindowDesc};

fn main () {
    let window = WindowDesc::new(build_widget);
    AppLauncher::with_window(window)
        .use_simple_logger()
        .launch(0_u32)
        .expect("launch failed");
}

fn build_widget() -> impl Widget<u32> {
    Flex::column()
        .with_child(
            Button::new(
                "Quit",
                |evt_ctx , data: _, _env| {
                    dbg!(&data);
                    let command = druid::Command::new(druid::commands::QUIT_APP, evt_ctx.window_id());
                    evt_ctx.submit_command(command, None);
                },
            ),
            1.0,
        )
}
```
gives
```
> env RUST_BACKTRACE=1 cargo run --example quit
    Finished dev [unoptimized + debuginfo] target(s) in 0.12s
     Running `target/debug/examples/quit`
DEBUG [druid::localization] available locales [], current en-US
DEBUG [druid::localization] resolved: [en-US]
Activated application
[druid/examples/quit.rs:18] &data = 0
thread 'main' panicked at 'Attempted to quit a GTK main loop when none is running.', /home/xliiv/.cargo/registry/src/github.com-1ecc6299db9ec823/gtk-0.8.0/src/rt.rs:127:13
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:77
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:61
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1028
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1412
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:65
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:50
[...]

PR #609, opened 2020-03-07
fix gtk quit panic #505

Dmitry-Borodin:
#505 is a druid issue, 

[related gtk issue](https://github.com/gtk-rs/gtk/issues/964)

I've tested, I cannot reproduce the crash after this change and app is closing now.


cmyr:
Thanks!

PR #610, opened 2020-03-07
removed gtk::init() because application will initialize it

Dmitry-Borodin:
https://developer.gnome.org/gtk3/stable/GtkApplication.html#gtk-application-new

> When using GtkApplication, it is not necessary to call gtk_init() manually. It is called as soon as the application gets registered as the primary instance.

We are registering application in the same call frame, so no need to init() manually, which leads to confusion because we shouldn't tear down gtk manually - it will crash #505

We can use application as the main api point here.


cmyr:
Thanks!

PR #611, opened 2020-03-07
Remove use GTK optional feature

### LINKED ISSUE #604: Remove use-gtk feature
The intent of this feature was to allow selectively using gtk on mac, which was specifically so that *I* could test the gtk backend without needing to boot linux.

Ultimately it's way easier for me to just run a VM than to maintain this feature, and it's an annoying burden, maintenance wise.

Removing this feature will involve\
- removing it from the `Cargo.toml` (druid & druid-shell)
- removing it anywhere it is used; we should use the gtk backend by default on linux, only swapping to the x11 backend when it is specified by feature; 
- and updating CI to no longer use this flag anywhere, if it is currently.

This should wait until #599 lands, otherwise it will be a painful merge.

pyroxymat:
Closes #604

### DIFF
--- a/druid-shell/Cargo.toml
+++ b/druid-shell/Cargo.toml
@@ -10,7 +10,6 @@ categories = ["os::macos-apis", "os::windows-apis", "gui"]
 edition = "2018"
 
 [features]
-use_gtk = ["gtk", "gtk-sys", "gio", "gdk", "gdk-sys", "glib", "glib-sys", "cairo-rs"]
 x11 = ["xcb", "cairo-sys-rs"]
 
 [package.metadata.docs.rs]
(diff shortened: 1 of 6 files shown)


cmyr:
Thanks! I'd like to hold this off until we get #599 merged, to try and keep that as uncomplicated as possible, but we can revisit then. :)

cmyr:
@pyroxymat with #599 merged, we can look at this again?

cmyr:
That does look useful! Ideally things will simplify a bit with us dropping GTK as a feature, but we can definitely keep that in mind if it still feels hairy.

totsteps:
@pyroxymat can we have the merge conflicts fixed? No rush though.
Thanks.

pyroxymat:
Fixed the PR

Issue #563, opened 2020-02-18
Lens derive fails if method exists with same name as member


cmyr:
e.g.

```rust
#[derive(Lens)]
struct Boots {
    leather: bool,
}

impl Boots {
    fn leather(&self) -> bool {
        self.leather
    }
}
```

will produce the cryptic error,

```
error[E0592]: duplicate definitions with name `leather`
  --> src/your_file.rs:N:N+M
   |
20 |   #[derive(Lens)]
   |            ---- other definition for `leather`
...
26 | /     fn leather(&self) -> bool {
27 | |         self.leather
28 | |     }
   | |_____^ duplicate definitions for `leather`
```

I think the _best_ solution here is to have something like a `#[druid(lens_name = "leather_lens")]` annotation that can be added to the field, and then to print a message suggesting that.

I think the _easier_ solution would be to just keep appending underscores until we find a valid identifier 😬.

emigr2k1:
Shouldn't this be closed?

cmyr:
Thanks!

Issue #431, opened 2019-12-31
Update lens module documentation to also document deriving Lens


cmyr:
The best place to look for examples of how to use lens is in `lens.rs`, but these examples don't mention `derive(Lens)`. In particular, reading these docs might suggest someone should use the `lens!` macro when `Type::field` syntax would be easier.

It might also be worth mentioning the `WidgetTrait::lens` method in here, although that is less important? That can be communicated in examples.

PoignardAzur:
The Lens documentation is missing on several points. I'm adding it to my list.

PR #612, opened 2020-03-08
Implement overriding of lens name

pyroxymat:
Closes issue #563 

I am not sure how to check if there is a function with a given name already. So this allows you to set a different lens_name but I don't know yet how to print a nice error message if the name for the lens is already taken by a function.

### DIFF
--- a/druid-derive/src/attr.rs
+++ b/druid-derive/src/attr.rs
@@ -22,6 +22,7 @@ use syn::{Error, ExprPath, Meta, NestedMeta};
 const BASE_ATTR_PATH: &str = "druid";
 const IGNORE_ATTR_PATH: &str = "ignore";
 const SAME_FN_ATTR_PATH: &str = "same_fn";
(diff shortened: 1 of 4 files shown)


pyroxymat:
After checking I don't think it is possible to search for impls in a proc-derive macro. So the PR is merge-able.

The docs are not adjustable properly yet but I guess that can be updated as well in issue #431

druid-derive/src/attr.rs
```diff
@@ -22,6 +22,7 @@ use syn::{Error, ExprPath, Meta, NestedMeta};
 const BASE_ATTR_PATH: &str = "druid";
 const IGNORE_ATTR_PATH: &str = "ignore";
 const SAME_FN_ATTR_PATH: &str = "same_fn";
+const LENS_NAME_OVERRIDE_ATTR_PATH: &str = "lens_name";
```
cmyr: I feel like in the future we might want to replace the `#[druid]` attribute with separate `#[lens]` and `#[data]` attributes, but I think this is reasonable for now.

Issue #540, opened 2020-02-11
Improvements to Split widget


cmyr:
There are a number of features that are not yet implemented for the `Split` widget, but which we will want eventually.

- explicit size: currently the split widget spit point is represented as a ratio, but in many situations I think it makes sense to provide an explicit value. This should probably mean supporting both possible representations?

- minimum sizes/maximum sizes? it should be possible to provide a minimum and maximum size for one or both sides of the panel. This is slightly annoying because it creates the possibility that the constraints will not be satisfiable, and we will need to figure out how to handle that. The fewer constraints we allow the simpler a problem this is; a good place to start would just be to have a `min_size` that would apply to both sides.

PoignardAzur:
Is this still relevant?

This is the current declaration of Split:

```rust
/// A container containing two other widgets, splitting the area either horizontally or vertically.
pub struct Split<T> {
    split_axis: Axis,
    split_point_chosen: f64,
    split_point_effective: f64,
    min_size: (f64, f64), // Integers only
    bar_size: f64,        // Integers only
    min_bar_area: f64,    // Integers only
    solid: bool,
    draggable: bool,
    /// The split bar is hovered by the mouse. This state is locked to `true` if the
    /// widget is active (the bar is being dragged) to avoid cursor and painting jitter
    /// if the mouse moves faster than the layout and temporarily gets outside of the
    /// bar area while still being dragged.
    is_bar_hover: bool,
    /// Offset from the split point (bar center) to the actual mouse position when the
    /// bar was clicked. This is used to ensure a click without mouse move is a no-op,
    /// instead of re-centering the bar on the mouse.
    click_offset: f64,
    child1: WidgetPod<T, Box<dyn Widget<T>>>,
    child2: WidgetPod<T, Box<dyn Widget<T>>>,
}
```

PR #613, opened 2020-03-08
Implement min size constraint for splitter

pyroxymat:
I am not sure about the API still for the split constraints. But this is a draft of how to implement them. You can check it out in the "split_demo" example.

Closes issue #540

### DIFF
--- a/druid/src/widget/split.rs
+++ b/druid/src/widget/split.rs
@@ -21,28 +21,11 @@ use crate::{
     PaintCtx, RenderContext, UpdateCtx, Widget, WidgetPod,
 };
 
-pub struct SplitConstraints {
-    min_size: f64,
-}
-
-impl Default for SplitConstraints {
-    fn default() -> Self {
-        Self { min_size: 0.0 }
-    }
-}
-
-impl SplitConstraints {
-    pub fn with_min_size(min_size: f64) -> Self {
-        Self { min_size }
-    }
-}
-
 ///A container containing two other widgets, splitting the area either horizontally or vertically.
 pub struct Split<T> {
     split_direction: Axis,
     draggable: bool,
-    constraints_left: SplitConstraints,
-    constraints_right: SplitConstraints,
+    min_size: f64,
     split_point: f64,
     splitter_size: f64,
     child1: WidgetPod<T, Box<dyn Widget<T>>>,
@@ -58,8 +41,7 @@ impl<T> Split<T> {
     ) -> Self {
         Split {
             split_direction,
-            constraints_left: SplitConstraints::default(),
-            constraints_right: SplitConstraints::default(),
+            min_size: 0.0,
             split_point: 0.5,
             splitter_size: 10.0,
             draggable: false,
@@ -85,10 +67,11 @@ impl<T> Split<T> {
         self.split_point = split_point;
         self
     }
-    pub fn split_constraints(mut self, left: SplitConstraints, right: SplitConstraints) -> Self {
-        // TODO add checks
-        self.constraints_left = left;
-        self.constraints_right = right;
+    /// Set the minimum size for both sides of the split
+    /// The value must be atleast 0.0
+    pub fn min_size(mut self, min_size: f64) -> Self {
+        assert!(min_size >= 0.0);
+        self.min_size = min_size;
 
         self
     }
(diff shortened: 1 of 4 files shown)


pyroxymat:
Okay. I simplified the API additionally, I change the splitter demo to only have a single splitter now. Since it doesn't make a lot of sense to have a nested splitter in case the min-sizes are really off (the min_size for the right 2 is together the same as the minsize for the left split).

I guess a good solution at some point might be to move to a splitter where you can have multiple splits? This is how QT does it as far as I know.

Zarenor:
I think we should index the panes left-to-right and top-to-bottom. So to min-size them independently, we have `first_min`(`min_first`?) and `min_second`. I could imagine max-sizes too, but I'm not sure how we handle the possibility of being asked to split a container too large for our maximum sizes if they're both/all set.

I had gone for 'simple, fast, and composable' with the single-splitter-only design. It handles, I think, the 80% case and has much simpler code than a multiway split. I also think filling out the constraints (mins and maxes) makes having to split several times more tolerable. I think being able to set the split point in units (pixels?) gets us the rest of the way to workable.

However, I wouldn't be opposed to creating a separate multisplitter, or altering this one with a vec of parameters, and so an arbitrary number of splits. Whichever we think is more appropriate. I think in that case agreeing to indexing in a direction would matter.

druid/src/widget/split.rs
```diff
@@ -65,6 +67,14 @@ impl<T> Split<T> {
         self.split_point = split_point;
         self
     }
+    /// Set the minimum size for both sides of the split
+    /// The value must be atleast 0.0
```
cmyr:
```suggestion
    /// The value must be greater than or equal to `0.0`.
```

druid/src/widget/split.rs
```diff
@@ -92,33 +102,33 @@ impl<T> Split<T> {
             }
         }
     }
+    fn calculate_limits(&self, size: Size) -> (f64, f64) {
+        let size_in_split_direction = match self.split_direction {
```
cmyr: Just fyi I believe you can do, `self.split_direction.major(size)` to get the relevant 'main axis length' for this direction.

druid/src/widget/split.rs
```diff
@@ -316,3 +355,15 @@ impl<T: Data> Widget<T> for Split<T> {
         self.child2.paint_with_offset(paint_ctx, &data, env);
     }
 }
+
+// Move to std lib clamp as soon as https://github.com/rust-lang/rust/issues/44095 lands
+fn clamp(mut x: f64, min: f64, max: f64) -> f64 {
+    assert!(min <= max);
+    if x < min {
```
cmyr: Not a blocker, but I would probably just write this as `x.max(min).min(max)`?

pyroxymat: I just did it how Rust (nightly) did it, to keep compatibility in edge-cases. There is quite some [discussion](https://github.com/rust-lang/rfcs/pull/1961#issuecomment-289893089) going on what is the proper way to write clamp. Seems like at edge cases weird things happens with these small (seemingly) similar implementations.

druid/src/widget/split.rs
pyroxymat: Nice that is cleaner! FYI: I had to make the function public (for the crate) though.

druid/src/widget/split.rs
pyroxymat: Seems a bit clearer what happens in case of 0.0 indeed :)

druid/src/widget/split.rs
pyroxymat: I just noticed I actually had them the wrong way around as well...

pyroxymat: For anyone wondering what the answer is, as seen in a [previous discussion](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/split.20vertical.2Fhorizontal.20confusion):

The axis tells you the axis of the splitter itself.

druid/src/widget/split.rs
```diff
@@ -65,6 +67,14 @@ impl<T> Split<T> {
         self.split_point = split_point;
         self
     }
+    /// Set the minimum size for both sides of the split
```
cmyr:
```suggestion
    /// Builder-style method to set the minimum size for both sides of the split.
    ///
```

Issue #615, opened 2020-03-08
Add ctrl support in textbox


pyroxymat:
Currently textbox allows selection of text using the shift key and arrow keys to add/remove individual letters to the selection. Ctrl + arrows keys causes you to jump to the front or end.

A common use-case is to use ctrl + shift to select entire words in their entirely. Ctrl + arrow keys to jump through words. Ctrl + backspace to delete word before you (or selection). Ctrl + Delete to delete word behind you (or selection).

cmyr:
So this is slightly complicated: basically, in the long term, we don't want to have all of the actual event handling in textbox; we want to have something like the [`Edit` enum in xi](https://github.com/xi-editor/xi-editor/blob/master/rust/core-lib/src/rpc.rs#L383),  and then we want to have some modular component that maps keyboard events into edit actions in a way that can easily vary per-platform, and be reused by different editable components.

That said, if you wanted to start working on something in this direction you would be welcome to. A good start would be to remove all the direct handling of keyboard events from the textbox (at least those related to text editing) and to move to something that consumes key events and produces edit actions.

I think realistically this should be a trait, something like,

```rust
trait TextInput {
    fn handle_event(&self, event: SomeKeyEvent) -> Option<EditAction>;
}
```

or something. This would let us be very flexible in terms of how we handle keyboard input.

Ciantic:
This is pretty good already, only one thing that didn't work:

- [x] Ctrl+Left jumps word left
- [x] Ctrl+Right jumps word left
- [x] Ctrl+Shift+Left selects word on left
- [x] Ctrl+Shift+Right selects word on right
- [ ] Ctrl+Backspace deletes word on left
- [x] Ctrl+Del deletes word on right

Ciantic:
Now

- [x] <kbd>Ctrl</kbd>+<kbd>Backspace</kbd> deletes word on left

PR #617, opened 2020-03-08
Implement EditAction for keyboard gestures

pyroxymat:
This moves our textbox actions into an `EditAction` enum with the goal of eventually moving to a xi-based implementation for our textbox.

This is a step in solving issue #615

### DIFF
--- a/druid/src/text/text_input.rs
+++ b/druid/src/text/text_input.rs
@@ -1,0 +1,91 @@
+use crate::{HotKey, KeyCode, SysMods};
+use druid_shell::KeyEvent;
+
+// This enumeration is heavily inspired by xi-editors edit notification
+// https://github.com/xi-editor/xi-editor/blob/066523b2a57f719cd93cc9ac0dda7687194badd4/rust/core-lib/src/rpc.rs#L374
+// This is done with the goal of eventually being able to easily switch
+// to a xi-based implementation of our EditActions.
+#[derive(Clone, Debug)]
+pub enum EditAction {
+    Insert { chars: String },
+    DeleteForward,
+    DeleteBackward,
+
+    MoveLeft,
+    MoveLeftAndModifySelection,
+    MoveRight,
+    MoveRightAndModifySelection,
+
+    MoveToLeftEndOfLine,
+    MoveToRightEndOfLine,
+
+    SelectAll,
+}
+
+pub trait TextInput {
+    /// Handle a key event and return an edit action to be executed
+    /// for the key event
+    fn handle_event(&self, event: &KeyEvent) -> Option<EditAction>;
+}
+
+/// Handles key events and returns actions that are applicable to
+/// single line textboxes
+pub struct SingleLineTextInput {}
+
+impl SingleLineTextInput {
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+
+impl TextInput for SingleLineTextInput {
+    fn handle_event(&self, event: &KeyEvent) -> Option<EditAction> {
+        let action = match event {
+            // Select all (Ctrl+A || Cmd+A)
+            k_e if (HotKey::new(SysMods::Cmd, "a")).matches(k_e) => EditAction::SelectAll,
+            // Jump left (Ctrl+ArrowLeft || Cmd+ArrowLeft)
+            k_e if (HotKey::new(SysMods::Cmd, KeyCode::ArrowLeft)).matches(k_e)
+                || HotKey::new(None, KeyCode::Home).matches(k_e) =>
+            {
+                EditAction::MoveToLeftEndOfLine
+            }
+            // Jump right (Ctrl+ArrowRight || Cmd+ArrowRight)
+            k_e if (HotKey::new(SysMods::Cmd, KeyCode::ArrowRight)).matches(k_e)
+                || HotKey::new(None, KeyCode::End).matches(k_e) =>
+            {
+                EditAction::MoveToRightEndOfLine
+            }
+            // Select left (Shift+ArrowLeft)
+            k_e if (HotKey::new(SysMods::Shift, KeyCode::ArrowLeft)).matches(k_e) => {
+                EditAction::MoveLeftAndModifySelection
+            }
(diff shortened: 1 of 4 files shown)


futurepaul:
This already feels like a big improvement to me! I don't know exactly where this is headed, so the `SingleLineTextInput` / `TextInput` abstraction feels a little premature to me, but I'll defer to @cmyr on this.

And a potential point of clarification: the way I've been thinking of text's future in druid is making it so that xi's `Rope` data structure would be easy to swap in to replace `String` if necessary. The other thinking is that the text editing facilities could be useful to other UI toolkits, so should stay generic if possible. I've ignored the broader abstractions of xi editor, because afaik it's not obvious that a simple text editor in druid should use RPC, for instance.

futurepaul:
It was annoying but we were resetting the cursor blink everywhere it made sense. I believe this is a regression in that regard. Is there a slick way to get that back without being messy?

futurepaul:
Movements don't seem to be updating the hscroll. When I jump left or jump right on a line that's longer than the textbox, the cursor goes but the hscroll doesn't. Same when I use the arrow keys to move through a piece of text longer than textbox.

pyroxymat:
> It was annoying but we were resetting the cursor blink everywhere it made sense. I believe this is a regression in that regard. Is there a slick way to get that back without being messy?

I updated the logic that no matter which "EditAction" we currently execute, we always reset the Blinking. Notice that an edit action will always cause a change in either selection or text. If this changes we would need to re-evaluate

> Movements don't seem to be updating the hscroll. When I jump left or jump right on a line that's longer than the textbox, the cursor goes but the hscroll doesn't. Same when I use the arrow keys to move through a piece of text longer than textbox.

I changed it that any movement (except for select all) causes the hscroll to be updated. This adds the really nice feature (that wasn't in before) that if drag select more than currently fits in the textbox it will automatically scroll to track your cursor.

druid/src/text/text_input.rs
```diff
@@ -0,0 +1,96 @@
+use super::Movement;
+use crate::{HotKey, KeyCode, SysMods};
+use druid_shell::KeyEvent;
+
+// This following enumerations are heavily inspired by xi-editors enumerations found at
+// https://github.com/xi-editor/xi-editor/blob/e2589974fc4050beb33af82481aa71b258358e48/rust/core-lib/src/edit_types.rs
+// This is done with the goal of eventually being able to easily switch
+// to a xi-based implementation of our Events.
+
+/// Events that only modify view state
```
cmyr: This comment is now out of date

druid/src/text/text_input.rs
```diff
@@ -0,0 +1,96 @@
+use super::Movement;
```
cmyr: can we get a license header and a `//! short module comment` here?

druid/src/text/text_input.rs
```diff
@@ -0,0 +1,96 @@
...
+    Click(MouseAction),
+    Drag(MouseAction),
+    Delete, // { movement: Movement, kill: bool },
+    Backspace,
+    Insert(String),
+    Paste(String),
+}
+
+#[derive(PartialEq, Eq, Debug, Clone)]
+pub struct MouseAction {
```
cmyr: I'd like doc comments on all public items please. :)

Also: who is doing click detection?

druid/src/text/text_input.rs
pyroxymat: Done

druid/src/text/text_input.rs
pyroxymat: Changed!

druid/src/text/text_input.rs
pyroxymat: The textbox itself

druid/src/text/text_input.rs
```diff
@@ -0,0 +1,113 @@
...
+use super::Movement;
+use crate::{HotKey, KeyCode, SysMods};
+use druid_shell::KeyEvent;
+
+// This following enumerations are heavily inspired by xi-editors enumerations found at
+// https://github.com/xi-editor/xi-editor/blob/e2589974fc4050beb33af82481aa71b258358e48/rust/core-lib/src/edit_types.rs
+// This is done with the goal of eventually being able to easily switch
+// to a xi-based implementation of our Events.
+
+/// Defines actions that edit the state of a textbox widget
```
cmyr:
```suggestion
/// An enum that represents actions in a text buffer.
```

druid/src/text/text_input.rs
```diff
@@ -0,0 +1,113 @@
...
+    Backspace,
+    Insert(String),
+    Paste(String),
+}
+
+/// Extra information related to mouse actions
+#[derive(PartialEq, Eq, Debug, Clone)]
+pub struct MouseAction {
+    pub column: usize,
+    pub shift: bool,
```
cmyr: I think this should be row, column, and `KeyModifiers`. We do care about ctrl/cmd/etc.

druid/src/text/text_input.rs
```diff
@@ -0,0 +1,113 @@
...
+
+pub trait TextInput {
+    /// Handle a key event and return an edit action to be executed
+    /// for the key event
+    fn handle_event(&self, event: &KeyEvent) -> Option<EditAction>;
+}
+
+/// Handles key events and returns actions that are applicable to
+/// single line textboxes
+pub struct SingleLineTextInput {}
```
cmyr: We don't need to overthink this; single line input is a subset of normal input, and doesn't really require special logic; we just don't handle certain events later on. I would just call this `BasicTextInput` or something. (We will want some platform-specific additions, as well, like `CocoaTextInput`; we can hopefully chain these, like): 

```rust
let edit = basic_input.handle_event(&event).or(platform_input.handle_event(&event));
```

... or something.

cmyr:
@pyroxymat ping on this? very interested in merging.

druid/src/text/text_input.rs
pyroxymat: Seems like a smart idea. Fixed.

PR #619, opened 2020-03-09
Add tests to SizedBox

cmyr:
I was slightly confused by some behaviour I saw, and wanted to
confirm that this was doing what I thought it was.

### DIFF
--- a/druid/src/widget/sized_box.rs
+++ b/druid/src/widget/sized_box.rs
@@ -76,6 +76,31 @@ impl<T> SizedBox<T> {
         self
     }
 
+    fn child_constraints(&self, bc: &BoxConstraints) -> BoxConstraints {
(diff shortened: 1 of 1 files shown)


druid/src/widget/sized_box.rs
```diff
@@ -76,6 +76,33 @@ impl<T> SizedBox<T> {
         self
     }
 
+    fn child_constraints(&self, bc: &BoxConstraints) -> BoxConstraints {
+        // if we don't have a width/height, we don't change that axis.
+        // if we have an infinite width/height, we tighten that axis
```
futurepaul: I'm not following how this behaves differently for infinite and non-infinite? Just that infinite will always be the max?

cmyr: fair point, that was confusing, I reworded this comment.

PR #620, opened 2020-03-09
Give 'wide' widgets a shared default width

cmyr:
This also removes the tendency for these widgets to always
explicitly fill their provided constraints when possible.

The rationale here is that if the parent wants the child to fill
the bounds, it can do this by providing tight constraints.

cc @futurepaul thoughts? I'm going through a bunch of layout stuff, and will try and write up guidelines for writing a `layout` fn, but I don't think in the general case we should just be deferring to `bc.max()`; if we want that we can wrap ourselves in a `SizedBox` with infinite bounds.

### DIFF
--- a/druid/src/theme.rs
+++ b/druid/src/theme.rs
@@ -42,6 +42,9 @@ pub const CURSOR_COLOR: Key<Color> = Key::new("cursor_color");
 pub const FONT_NAME: Key<&str> = Key::new("font_name");
 pub const TEXT_SIZE_NORMAL: Key<f64> = Key::new("text_size_normal");
 pub const BASIC_WIDGET_HEIGHT: Key<f64> = Key::new("basic_widget_height");
+
+/// The default minimum width for a 'wide' widget; a textbox, slider, progress bar, etc.
+pub const WIDE_WIDGET_WIDTH: Key<f64> = Key::new("druid.widgets.long-widget-width");
 pub const BORDERED_WIDGET_HEIGHT: Key<f64> = Key::new("bordered_widget_height");
 
 pub const TEXTBOX_BORDER_RADIUS: Key<f64> = Key::new("textbox_radius");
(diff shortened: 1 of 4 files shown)


futurepaul:
So the way I'm reading this is: widgets have a default width and height set in the theme. If you want to override that, you can pass a "tight" box constraint in one or both dimensions.

Should the widget complain if it's given a bc height or width that's smaller than the corresponding theme value?

Also, what happens if there's a tight bc that's min infinite in a dimension?

Overall I agree with the idea of making this default width more intuitive and less surprising.

cmyr:
Correct.

I think it's fine to intentionally shrink the widget by passing it small max constraints.

I think that having any constraint be min infinite is a bug? I would have to think about this.

futurepaul:
I believe we'll need some more logic to handle the case of shrinking / growing a widget with tight constraints, at least in the height dimension. Right now afaik the paint methods assume they'll get their height from the theme::BORDERED_WIDGET_HEIGHT, not from the result of layout.

cmyr:
This is one open question for me: how should a widget handle constraints that it does not expect?

In flutter I see three different behaviours: widgets that have strong opinions about their appearance (such as a slider) will draw themselves centered in the available space; widgets where there is a sensible way to fill the space (such as a progress indicator) will draw themselves stretched; and label/textbox widgets will draw their contents at the top left. 🤷‍♂

cmyr:
Although also: if a widget just gets totally wonky constraints, we don't need to worry too much about that?

futurepaul:
We are guarding against truly wild stuff with bc.debug_check. The only problem I'm expecting is that widgets will be painting outside their boundary (if they're not clipped), which isn't the end of the world. (Slider is already an offender in this case, btw). Seems like documenting "how will this widget fill its given space" in the non-obvious cases will get us pretty far.

cmyr:
yea, that sounds good. And I've noticed through this that all of the stroked widgets are drawing outside of their bounds; this is something we'll need to address.

futurepaul:
Yeah I'm thinking about it and I think the only way to get a minimum infinity right now is to put a SizedBox inside a Scroll and then expand it (or explicitly set its width and height to infinity).

cmyr:
infinite min can also happen if you expand in other places that want you to freely choose a size, such as if you're a non-flex child of a flex widget. 😕 

I'm... pretty sure it's nonsensical, and we should check for it?

futurepaul:
Would be pretty easy to add it to `BoxConstraints` `debug_check` method

cmyr:
I think that makes sense. I think it also might make sense at some point in the future for widgets to be able to pass their intrinsic sizes to that method (or an alternate method) to catch places where widgets are being used incorrectly? but: future work.

PR #624, opened 2020-03-09
Flex main axis alignment

cmyr:
This implements two additional options from the flutter Flex widget: the ability to set the [main axis alignment](https://api.flutter.dev/flutter/rendering/MainAxisAlignment-class.html) (which I'm calling `Justification`; not sure I like this name) as well as the ability to force the container to use all available space on the main axis, which flutter calls [`MainAxisSize`](https://api.flutter.dev/flutter/rendering/MainAxisSize-class.html).

These are implemented together because without the latter you don't have any extra space with which to use the former.

Screen Shot 2020-03-09 at 7 49 24 PM
Screen Shot 2020-03-09 at 7 49 33 PM

### DIFF
--- a/druid/src/widget/flex.rs
+++ b/druid/src/widget/flex.rs
@@ -25,6 +25,7 @@ use crate::{
 pub struct Flex<T> {
     direction: Axis,
     cross_alignment: CrossAxisAlignment,
+    main_alignment: MainAxisAlignment,
     children: Vec<ChildWidget<T>>,
 }
 
@@ -42,7 +43,7 @@ pub(crate) enum Axis {
 ///
 /// If a widget is smaller than the container on the minor axis, this determines
 /// where it is positioned.
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq)]
 pub enum CrossAxisAlignment {
     /// Top or leading.
     ///
@@ -56,6 +57,33 @@ pub enum CrossAxisAlignment {
     /// In a vertical container, widgets are bottom aligned. In a horiziontal
     /// container, their trailing edges are aligned.
     End,
+}
+
+/// Arrangement of children on the main axis.
+///
+/// If there is surplus space on the main axis after laying out children, this
+/// enum represents how children are laid out in this space.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum MainAxisAlignment {
+    /// Top or leading.
+    ///
+    /// Children are aligned with the top or leading edge, without padding.
+    Start,
+    /// Children are centered, without padding.
+    Center,
+    /// Bottom or trailing.
+    ///
+    /// Children are aligned with the bottom or trailing edge, without padding.
+    End,
+    /// Extra space is divided evenly between each child.
+    SpaceBetween,
+    /// Extra space is divided evenly between each child, as well as at the ends.
+    SpaceEvenly,
+    /// Space between each child, with less at the start and end.
+    ///
+    /// This divides space such that each child is separated by `n` units,
+    /// and the start and end have `n/2` units of padding.
+    SpaceAround,
 }
 
 #[derive(Copy, Clone, Default)]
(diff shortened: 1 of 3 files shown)


druid/src/widget/flex.rs
```diff
@@ -130,6 +171,23 @@ impl<T> Flex<T> {
...
 
+    /// Builder-style method for specifying the childrens' [`Justification`].
+    ///
+    /// [`Justification`]: enum.Justification.html
+    pub fn justification(mut self, justification: Justification) -> Self {
+        self.justification = justification;
+        self
+    }
+
+    /// Builder-style method for setting whether the container must expand
```
futurepaul: For some reason I'm having a hard time understanding this. I guess because I assumed this is what a Flex widget was doing already?

cmyr: yea, I can understand the confusion.

Let's say a flex row is passed a width constraint `(0., 400.)`. None if its children are flex. Is it still obliged to have a final width of `400.0`, or can it be smaller? This controls that behaviour.

I'll think about how to make this clearer.

futurepaul: Gotcha.

If any children are flex this Flex container will expand to fill all available space on its main axis. But if no children are flex, `fill_main_axis` determines whether or not Flex should shrink to fit, or still fill its main axis (TKTK something about what it does with the extra space?)

cmyr:
Another big thing to figure out here is what we should call `Alignment` and `Justification`. Without better namespacing, these feel sort of ambiguous. (I wish we could declare types in types like in swift, so that this could be `Flex::Alignment` and `Flex::Justification`). In any case, the main alternative I can think of is `MainAxisAlignment` and `CrossAxisAlignment` (or `Major`/`Minor`) which is more verbose, but also less ambiguous.

futurepaul:
I like MainAxisAlignment and CrossAxisAlignment. It's verbose, but Alignment and Justification are overloaded and confusing. I have to look them up every time I use them in CSS.

cmyr:
I've removed the second commit; in light of earlier changes, this is now useable without needing the `fit_main_axis` setting.

It's possible that setting will come back, but I'm not 100% sure it's necessary; I think that the equivalent might be achievable with a tight constraint.

futurepaul:
Hmm playing around with this and I feel like I'd expect the default behavior of something called "flex" to actually fill the main axis and that its children would stretch unless told otherwise. That might just be me?

cmyr:
It is not the defualt behaviour in flutter, which is approximately the extent of my analysis. 😁

druid/src/widget/flex.rs
```diff
@@ -225,15 +271,21 @@ impl<T: Data> Widget<T> for Flex<T> {
...
         }
 
+        // figure out if we have extra space on major axis, and if so how to use it
+        let major_used = total_non_flex + flex_used;
+        let extra = (self.direction.major(bc.min()) - major_used).max(0.0);
+        let spacing = self.main_alignment.spacing(extra, self.children.len());
+
         // Finalize layout, assigning positions to each child.
-        let mut major = 0.0;
+        let mut major = spacing.pre;
```
futurepaul: I don't know if there should be a warning here or really anything different at all, I just know I wouldn't feel comfortable modifying this section. My brain compiler is tracking at like 85%.

druid/src/widget/flex.rs
```diff
@@ -281,3 +336,62 @@ impl CrossAxisAlignment {
         }
     }
 }
+
+impl MainAxisAlignment {
```
futurepaul: This looks good to my brain compiler, and I really like this style of spinning out straight logic.

druid/src/widget/flex.rs
cmyr: so 'major' here represents our 'current' position in the major axis; that is, where we're going to place the next item. This means that we should place the first item after any initial padding.

cmyr:
mm, hadn't noticed panels. There is a reasonable solution, which is to expand each panel.

About Row/Column; I was thinking about this earlier in the week, so that we could have better type names for the alignment options (not having 'start' and 'end') but I'm happy enough with the current options, and I'm not sure there's a huge gain now in going back to row/column?

cmyr:
conclusion: let's get this in now and clean up afterwards.

Issue #607, opened 2020-03-07
Feature parity with flutter Flex widget


cmyr:
I've been playing around a bunch with layout, and have been digging back into flutter a bit more. The flutter `Row` & `Column` widgets have a bunch of additional flexibility that we are currently missing, and I'd like to try and get us more or less to parity; row & column really are the foundation of most layouts.

For an excellent overview of what we would like, you can play around with https://flutter.dev/docs/codelabs/layout-basics.

A brief rundown of what I would like:

- [ ] [`mainAxisSize`](https://flutter.dev/docs/codelabs/layout-basics#mainaxissize-property) (or equivalent): basically determines whether a row or column is sized to contain its children, or expands to fill its constraints.
- [ ] [`mainAxisAlignment`](https://flutter.dev/docs/codelabs/layout-basics#mainaxisalignment-property) (or equivalent): options for how children are positioned along the main axis.
- [ ] per-child overrides: it should be possible to override certain properties on a per-child basis. For instance, if widgets are centered in the container by default it should be possible to override that alignment for one child.
- [ ] API for adding empty widgets with and without flex.
- [ ] support 'loose' flex widgets; currently a flex item is given 'tight' constraints, i.e. it is forced to occupy the full flex space. We should allow 'loose' flex constraints, where a widget can only optionally fill its flex space.
- [ ] by default, items in a flex should have a flex factor of 1.0 and 'loose' constraints.

What does this look like?

We will need to expose an explicit `FlexItem` type (currently `ChildWidget`). It will expose `flex_factor`, `flex_fit` (tight/loose), optional `cross_alignment` (currently just called `alignment`).

The API for adding a child maybe to look like,

```rust
fn with_child(mut self, child: impl Into<FlexItem<T>>) -> Self { .. }
```

where there is a blanket `impl <W: Widget<T>> From<W> for FlexItem<T>`; so you can just pass in a widget and we will use sane defaults, and if you want to get fancy you can create the `FlexItem` manually.

As a convenience, `Flex` can also get two methods for adding spacers:

```rust
impl<T> Flex<T> {
    // creates a SizedBox with the given dimension
    fn with_spacer(mut self, size: f64) -> FlexItem<T> { .. }
   // creates an empty SizedBox with the given flex:
    fn with_flex_spacer(mut factor: f64) -> FlexItem<T> { .. }
}
```

I hope that with these additions, we will significantly expand the range of layouts possible in druid.

PR #625, opened 2020-03-10
Loosen incoming constraints in Flex

cmyr:
The flex widget should not force its children to match its
own minimum size, particularly on the minor axis; the container
is allowed to be larger than the sum of its children, and many
of the configurable options on the flex widget expect this to be
the case.

If the user wants a specific widget to fill its available space,
this can be achieved with an expanded SizedBox or possibly with a
(not yet implemented) 'stretch' alignment.

before:
Screen Shot 2020-03-09 at 10 02 02 PM
after:
Screen Shot 2020-03-10 at 11 03 50 AM

progress on #607

### DIFF
--- a/druid/examples/calc.rs
+++ b/druid/examples/calc.rs
@@ -16,7 +16,7 @@
 
 use druid::{AppLauncher, Data, Lens, LensWrap, LocalizedString, Widget, WindowDesc};
 
-use druid::widget::{Button, Flex, Label, Padding};
+use druid::widget::{Button, Flex, Label, WidgetExt};
 
 #[derive(Clone, Data, Lens)]
 struct CalcState {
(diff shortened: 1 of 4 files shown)


futurepaul:
Now that #605 is in this should work better yeah?

cmyr:
yep, and this should also make that stretch alignment meaningful.

futurepaul:
I'm not getting your second screenshot. Not sure what I'm doing wrong. I checked out this branch and rebased it to master.

Screen Shot 2020-03-10 at 11 17 38 AM

cmyr:
oh sorry I updated the screenshot and then found a bug and didn't push the branch!

cmyr:
also this patch breaks calculator pretty bad, I haven't checked everything else; I'm okay with some example breakage, I'd like to really cull the examples  soon

futurepaul:
Another way to fix calc would be to "fill main axis" (#624) so the buttons have more room vertically, but I believe they wouldn't stretch in that case.

cmyr:
that would make the container fill the main axis, but would not oblige the children to fill the available flex (this is another option that is coming soon)

cmyr:
great, happy to get this in so that the justify thing makes sense.

PR #627, opened 2020-03-10
Add the ability to make the Split handle a solid rectangle

s3thi:
This PR adds a new method to `Split` that lets it draw its draggable handle as a solid rectangle instead of two lines.

Screenshots for comparison: [before](https://imgur.com/a/XuTKYRQ), [after](https://imgur.com/a/xRHzAOV)

I'd love some help with this PR. Everything renders correctly when I use this functionality in my own app, but the splitter handle is 1px off in the split demo from the examples in the repo. I can't figure out why this behaves differently in different scenarios.

### DIFF
--- a/druid/src/widget/split.rs
+++ b/druid/src/widget/split.rs
@@ -96,7 +96,7 @@ impl<T> Split<T> {
     /// Builder-style method to set whether the splitter handle is drawn as a solid rectangle.
     ///
     /// If this is `false` (the default), it will be drawn as two parallel lines.
-    pub fn solid(mut self, solid: bool) -> Self {
+    pub fn fill_splitter_handle(mut self, solid: bool) -> Self {
         self.solid = solid;
         self
     }
--- a/druid/src/widget/split.rs
+++ b/druid/src/widget/split.rs
@@ -93,6 +93,7 @@ impl<T> Split<T> {
         self.draggable = draggable;
         self
     }
+    /// Set whether the splitter handle is a solid rectangle or two parallel lines.
     pub fn solid(mut self, solid: bool) -> Self {
         self.solid = solid;
         self
--- a/druid/src/widget/split.rs
+++ b/druid/src/widget/split.rs
@@ -17,13 +17,14 @@
 use crate::kurbo::{Line, Point, Rect, Size};
 use crate::widget::flex::Axis;
 use crate::{
-    theme, BoxConstraints, Cursor, Data, Env, Event, EventCtx, LayoutCtx, LifeCycle, LifeCycleCtx,
-    PaintCtx, RenderContext, UpdateCtx, Widget, WidgetPod,
+    theme, BoxConstraints, Color, Cursor, Data, Env, Event, EventCtx, LayoutCtx, LifeCycle,
+    LifeCycleCtx, PaintCtx, RenderContext, UpdateCtx, Widget, WidgetPod,
 };
 
 ///A container containing two other widgets, splitting the area either horizontally or vertically.
 pub struct Split<T> {
     split_direction: Axis,
+    solid: bool,
     draggable: bool,
     min_size: f64,
     split_point: f64,
@@ -42,6 +43,7 @@ impl<T> Split<T> {
         Split {
             split_direction,
             min_size: 0.0,
+            solid: false,
             split_point: 0.5,
             splitter_size: 10.0,
             draggable: false,
@@ -91,6 +93,10 @@ impl<T> Split<T> {
         self.draggable = draggable;
         self
     }
+    pub fn solid(mut self, solid: bool) -> Self {
+        self.solid = solid;
+        self
+    }
     fn splitter_hit_test(&self, size: Size, mouse_pos: Point) -> bool {
         match self.split_direction {
             Axis::Vertical => {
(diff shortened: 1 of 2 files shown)


cmyr:
it sounds like maybe stroke behaviour? stroke is centered on the line, so can paint outside of the bounds if the line is a boundary.

druid/src/widget/split.rs
```diff
@@ -91,6 +93,11 @@ impl<T> Split<T> {
         self.draggable = draggable;
         self
     }
+    /// Set whether the splitter handle is a solid rectangle or two parallel lines.
+    pub fn solid(mut self, solid: bool) -> Self {
```
cmyr: The over-the-top thing to do here would be to have a  `SplitterStyle` enum but I don think that's necessary, however I might just give this a slightly more descriptive name; something like `draw_solid_splitter` or `solid_splitter`  or ??

druid/src/widget/split.rs
```diff
@@ -91,6 +93,11 @@ impl<T> Split<T> {
         self.draggable = draggable;
         self
     }
+    /// Set whether the splitter handle is a solid rectangle or two parallel lines.
```
cmyr:
```suggestion
    /// Builder-style method to set whether the splitter handle is drawn as a solid rectangle.
    ///
    /// If this is `false` (the default), it will be drawn as two parallel lines.
```

druid/src/widget/split.rs
Zarenor: I've been stewing for a while on wanting to offer more appearance options here and more generally. I think an enum adds a small amount of flexibility, but is kinda constraining.. But it's the obvious choice if we go beyond two appearance styles.
Something I was kicking around was a Fn that gets the paint ctx, start and end coordinates (of the bar centerline) and width as set. Then the expectation is that you draw within the rectangle this provides, however you like.
It's possible we could use a hidden inner widget for this, if we want to leverage the built-in clipping ability and coordinate space?

Zarenor: To be clear: I still wouldn't be opposed to having one or several built-in options for this.

Zarenor:
I've been stewing for a while on wanting to offer more appearance options here and more generally. I think an enum adds a small amount of flexibility, but is kinda constraining.. But it's the obvious choice if we go beyond two appearance styles.
Something I was kicking around was a Fn that gets the paint ctx, start and end coordinates (of the bar centerline) and width as set. Then the expectation is that you draw within the rectangle this provides, however you like.
It's possible we could use a hidden inner widget for this, if we want to leverage the built-in clipping ability and coordinate space?

To be clear: I still wouldn't be opposed to having one or several built-in options for this.

druid/src/widget/split.rs
s3thi: I started out with exactly that design, but it seemed a bit overkill. Besides the solid and stroked styles we have, is there another type of splitter that is in common use at the moment? Windows 10, macOS, and GNOME 3 all use a solid filled splitter.

KDE and some versions of GNOME 3 have sort of 3D embossed splitters, but IMO something like that would look out of place in Druid at the moment. We could reconsider this if/when Druid has more ways of styling all the other widgets.

(Now that I think of it, an ideal way of handling the styling of the splitter handle could be to set some values on the enclosing `Env`. We draw a solid rectangle as the splitter in all cases, and let the user set custom values on the `Env` for its stroke and fill properties. But I do feel our current method is enough for now.)

As for the name, how does `fill_splitter_handle` sound? Nice and verbose, and fits nicely into this snippet from my app:

```rust
Split::vertical(make_sidebar(), make_todo_list())
        .draggable(true)
        .split_point(0.3)
        .splitter_size(4.0)
        .min_size(160.0)
        .fill_splitter_handle(true)
```

s3thi:
@Zarenor I was thinking perhaps a more declarative solution would be easier to work with. I mentioned it in my comment [here](https://github.com/xi-editor/druid/pull/627#discussion_r390776827). What do you think?

I have one more major issue with this widget: the user can only interact with it if they place their cursor _exactly_ within the bounds of the splitter handle. I don't know how splitters work on Windows/Linux, but on macOS you can place your cursor a tiny bit outside the bounds of the splitter and still be able to interact with it.

In general on iOS (and maybe also macOS?), the interactive area of a widget is not limited to its painted area. You can often interact with a widget by performing a gesture slightly outside of its painted area. E.g, you can tap a few pixels below a button on iOS and still manage to send it a tap event. I think Android works the same way.

It would be useful to have something like this for the split widget. I would like to make my splitter barely visible -- say only 4px wide -- but still let the user interact with it if they place their cursor 2px to the left or right of the visible area. So, effectively, the splitter would be 8px wide but the painted area would only be 4px wide. Other widgets would be allowed to draw inside that 2px extra space on either side, but any events that occur there would be sent to the splitter.

Hope this makes some sort of sense?

This would be a huge architectural change, of course, but potentially very valuable from a usability PoV. It would also help with the transition to mobile.

s3thi:
> it sounds like maybe stroke behaviour? stroke is centered on the line, so can paint outside of the bounds if the line is a boundary.

I'm sorry, I'm not sure I understand this. Do you mean to say that stroke can paint extra pixels in certain cases, whereas fill can't?

The issue I'm seeing is that the filled splitter handle is moved about 1px upward in the `split_demo`, whereas it behaves correctly in my own app. The stroked handle works perfectly fine in both places. Adjusting the coordinates of the filled handle fixes it in the `split_demo`, but breaks it in my app.

Zarenor:
@s3thi I think I see what you're asking for with the splitter. I'm not sure of other apps, but in what I have to hand VS2019 and Notepad++ on Windows, the splitter has no fuzziness - meaning some of them are quite narrow! I don't think it's a particularly large refactor to add fuzziness, or even to make it configurable. I'm not sure I love the idea of capturing input over area other widgets draw on, but it is certainly possible and fairly easy in the `event` code.

As for more appearance options: I'm open to plenty of ideas, and I don't have strong opinions as to what the default should be. I picked what I did (one or two lines, depending on width) just to have something. In VS2019, the splitters are the color of the window chrome, and the controls all have some border to demarcate their boundary. In N++, the splitter is the color of the chrome, but with 'dimples' - shaded dots - to indicate the boundary of the draggable area.

The more I think on it, the more I think a few basic defaults, and the option to draw your own, is what I'd prefer.

druid/src/widget/split.rs
cmyr: this sounds good to me!

cmyr:
Also: we have an idea to allow customization of these sorts of drawing things, by allowing certain elements of UI (like a button background) take a closure like `Fn(&mut PaintCtx, &mut T, &mut Env)` and can just paint whatever they want.

cmyr:
Also: I do think that there should be a minimum hit target size, and a 2x splitter should have a larger-than-2px hit target, but I'm happy to treat that as a separate project.

s3thi:
Created #641 for the minimum hit area feature.

PR #630, opened 2020-03-10
Add spacer widget helpers to Flex

cmyr:
This makes it easy to add fixed and flex-size padding between
widgets.

progress on #607

### DIFF
--- a/druid/examples/flex.rs
+++ b/druid/examples/flex.rs
@@ -19,10 +19,12 @@ use druid::widget::{
     SizedBox, Slider, Stepper, Switch, TextBox, WidgetExt,
 };
 use druid::{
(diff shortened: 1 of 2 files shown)


cmyr:
good Q. it would be limited; you could maybe turn them on or off, but you couldn't, for instance, place them.

cmyr:
okay i've updated the example to let you add spacers.

futurepaul:
Thanks for adding to the example! This is feeling SO expressive for layout now. And the example pretty much solves the "discoverability" aspect for these features.

cmyr:
yea I think so too. Also  I think that the spacer should take a `KeyOrValue`, but that requires changing `SizedBox`.

cmyr:
okay I couldn't help myself: you can now set what the spacer value is, when it is fixed.
Screen Shot 2020-03-11 at 3 18 21 PM

Screen Shot 2020-03-11 at 3 18 37 PM

PR #637, opened 2020-03-11
Make Flex widget sizing on main axis be configurable

cmyr:
With this patch, the container can can be optionally set to always
use the available space on the main axis.

Basically: a flex container that has no flex items will happily just lay out its children, returning the union of their bounds as its own. If you want to force a container to occupy the full space on the major axis, you call `set_must_fill_main_axis(true)`.


cmyr:
An expanded sizedbox expands both directions (you can get around this manually, but wouldn't continue to work between row & column), and I think i had another case, but I can't think of it right now :/

PR #638, opened 2020-03-11
Add label to Checkbox widget and fix examples

### LINKED ISSUE #635: checkbox widget should have a label
I don't think it's meaningful to have a checkbox without text, although we can expose a `RawCheckbox` if we wish. 

The checkbox should have the same general layout properties as the radio widget, and clicking on the label should toggle the check.

futurepaul:
closes #635 

Forgoing `RawCheckbox` for simplicity, but someone can feel free to add it if they need it!

### DIFF
--- a/druid/src/widget/checkbox.rs
+++ b/druid/src/widget/checkbox.rs
@@ -14,22 +14,26 @@
 
 //! A checkbox widget.
 
-use crate::kurbo::{BezPath, Point, RoundedRect, Size};
+use crate::kurbo::{BezPath, Point, Rect, RoundedRect, Size};
 use crate::piet::{LineCap, LineJoin, LinearGradient, RenderContext, StrokeStyle, UnitPoint};
 use crate::theme;
+use crate::widget::{Label, LabelText, WidgetExt};
 use crate::{
     BoxConstraints, Env, Event, EventCtx, LayoutCtx, LifeCycle, LifeCycleCtx, PaintCtx, UpdateCtx,
-    Widget,
+    Widget, WidgetPod,
 };
 
 /// A checkbox that toggles a `bool`.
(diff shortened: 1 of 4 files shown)


druid/src/widget/checkbox.rs
```diff
@@ -69,16 +73,29 @@ impl Widget<bool> for Checkbox {
...
-        _layout_ctx: &mut LayoutCtx,
+        layout_ctx: &mut LayoutCtx,
         bc: &BoxConstraints,
-        _data: &bool,
+        data: &bool,
         env: &Env,
     ) -> Size {
         bc.debug_check("Checkbox");
 
+        // bc.constrain(Size::new(
```
cmyr: ✂️

druid/src/widget/checkbox.rs
```diff
@@ -69,16 +73,29 @@ impl Widget<bool> for Checkbox {
...
+        //     env.get(theme::BASIC_WIDGET_HEIGHT),
+        //     env.get(theme::BASIC_WIDGET_HEIGHT),
+        // ))
+
+        let label_size = self.child_label.layout(layout_ctx, &bc, data, env);
+        let padding = 8.0;
```
cmyr: it feels like this should maybe be a constant somewhere, and at least shared between the radio button and the checkbox? That can be a separate PR, though.

`CONTROL_WIDGET_HORIZ_PADDING`?

futurepaul: Yeah I'll do this along with fixing the padding of radio.

druid/src/widget/checkbox.rs
futurepaul: Oops thank you!

PR #639, opened 2020-03-11
Add FlexChild type

### LINKED ISSUE #607: Feature parity with flutter Flex widget

cmyr:
This is the last major bit of work on flex? maybe? But it enables
an important property of flex  children, which is that they can
be either 'loose' or 'tight'; that is, they can be *allowed* or
*required* to use the space they are provided during layout.

This also includes a big overhaul of flex-related docs

my intention is that at some point I will change `add_child` to not take a flex factor, but this will break everyones code so I want to hold off til at least I have the rest of the API nailed down.

This also currently has a bizarre test failure, in a case that does not, as far as  I can see,  touch anything that I've done here. I'll go for a walk and talk a closer look. 

closes #607

### DIFF
--- a/druid/src/widget/flex.rs
+++ b/druid/src/widget/flex.rs
@@ -23,6 +23,78 @@ use crate::{
 };
 
 /// A container with either horizontal or vertical layout.
+///
+/// This widget is the foundation of most layouts, and is highly configurable.
+///
+/// # Flex layout algorithm
+///
+/// Children of a `Flex` container can have an optional `flex` parameter.
+/// Layout occurs in several passes. First we measure (calling their [`layout`]
+/// method) our non-flex children, providing them with unbounded space on the
+/// main axis. Next, the remaining space is divided between the flex children
+/// according to their flex factor, and they are measured, with constraints
+/// determined by the available flex space and whether they are [`tight`] or
+/// [`loose`].
+///
+/// The difference between a widget with *no* flex and a widget with [`loose`]
+/// flex is that the non-flex widget is laid out first, and can take as much
+/// space as it needs; the [`loose`] widget is laid out afterwards, and is
+/// constrained based on how much space is available.
+///
+///
+/// # Options
+///
+/// To experiment with these options, see the `flex` example in `druid/examples`.
+///
+/// - [`CrossAxisAlignment`] determines how children are positioned on the
+/// cross or 'minor' axis. The default is `CrossAxisAlignment::Center`.
+///
+/// - [`MainAxisAlignment`] determines how children are positioned on the main
+/// axis; this is only meaningful if the container has more space on the main
+/// axis than is taken up by its children.
+///
+/// - [`must_fill_main_axis`] determines whether the container is obliged to
+/// be maximally large on the major axis, as determined by its own constraints.
+/// If this is `true`, then the container must fill the available space on that
+/// axis; otherwise it may be smaller if its children are smaller.
+///
+/// Additional options can be set (or overridden) in the [`FlexParams`].
+///
+/// # Examples
+///
+/// Construction with builder methods
+///
+/// ```
+/// use druid::widget::{Flex, FlexParams, Label, Slider, CrossAxisAlignment};
+///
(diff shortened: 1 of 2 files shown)


futurepaul:
Couple thoughts on the wording of things: it's a little confusing to me that `tight` and `loose` methods add a child, but `flex` mutates the flex factor. I think `add_tight` and `add_loose` would be a little clearer.

I'm also thinking that the word "flex" is a little overloaded. Would it be possible to explicitly use the term `flex_factor` in some places where that makes sense? That way `flex` can mostly mean "is this flex or not?" and `flex_factor` can mean "how much?"

cmyr:
thanks for the notes, I think you're right.

`tight` and `loose` are constructors, but `flex` is a builder method. I agree this is a bit confusing.

I actually think I might have a better way; maybe we should forget about `FlexChild` and instead do something like,

```rust
fn add_flex_child(&mut self, child: impl Widget<T> + 'static, params: impl Into<FlexParams>);

// with like,

impl From<f64> for FlexParams {
    fn from(self: f64) -> FlexParams {
        FlexParams {
            flex: self,
            tight: false,
            alignment: CrossAxisAlignment::Center
        }
     }
}
```

futurepaul:
Nice I like this. And then tight and loose can just mutate these params like flex does?

cmyr:
updated this, take a look.

cmyr:
didn't do the `impl Into<>` business, i'm not sure if it's a huge win

futurepaul:
Yeah this makes a lot more sense to me! And it seems a simpler change overall, too. Is there any limitation with this where you wouldn't be able to update dynamically, like in the flex.rs example?

cmyr:
No, I think everything should be possible still? 🤔 I'll play around a bit to make sure.

cmyr:
closing in favour of #691

Issue #420, opened 2019-12-27
Expose a larger number of design tokens in druid::theme


s3thi:
The current `EventScope` API allows me to override the variables listed in `druid::theme` to customize how my application looks. However, there are a large number of variables used implicitly in the default widgets that I can't override. Instead of being picked up from the `Env`, these variables appear as literals in the source code.

For example, the border radius for a `Button` is 4.0. It appears in `druid/src/widget/button.rs`, but I can't retrieve or override it because it's just a literal value.

Finding such variables and making them explicit in `druid::theme` will make styling a Druid application easier. There is precedent for this in the Design Systems communities, where these kinds of atomic base units are called design tokens. [This article](https://medium.com/eightshapes-llc/tokens-in-design-systems-25dd82d58421) is a decent explainer, and [this is a list of tokens](https://www.lightningdesignsystem.com/design-tokens/) used in the Salesforce design system.

Adopting this approach would reduce inconsistencies across the toolkit and makes design decisions more explicit. E.g, `Button` has a border radius of 4.0, but a `TextBox` has a border radius of 2.0. Both values appear as literals in the source code for those widgets. Exposing border radius values in `druid::theme` would've either prevented this inconsistency, or documented it as an explicit design decision.

This would also make it easy for custom widgets to fit in with the default set widget set.

It might not be an explicit goal of Druid, but this approach would also allow radical re-skinning of applications, with variable changes affecting both default, custom and community-built widgets.

I'm sure there are more ideas that Druid could pick up from the Design Systems community, but I can't think of anything at the moment. For now, setting up explicit design tokens would be a great start.

cmyr:
Cool, I think there's two separate related things here but they're both basically a matter of "haven't gotten to it yet":

- there are definitely a bunch of fields that should be customizable that have not been; totally happy for PRs, here.
- There has not been a lot of thought put into what is involved in making the theme system coherent: What keys should there be, what should their relationships be, what concepts should we express, etcetera. Raph and I were talking about this in relation to doing GSoC again this year; we were thinking that really diving deep into this would be a good project for somebody. GSoC or not, that is totally the case; this is really just an area that needs more thought and focus, and I'm happy to hear any thoughts or ideas. :)

Thanks for the issue and the links, it looks like a good starting place. 👍

s3thi:
[...]

PR #642, opened 2020-03-12
Add a color hierarchy

s3thi:
This PR is one step towards the design-token-based approach to basic UI units and measures as discussed in #420.

I'm going to break the work in #420 up into two or three smaller PRs. In this first PR, I'm looking for feedback on what the community thinks of this design. I've used this successfully on the Web, but does it make sense at all in the context of Druid?

In this PR, I took all the different colors used in `theme.rs` and put them in a color hierarchy, from lightest to darkest. You can visually see the colors I used in [this CodePen](https://codepen.io/s3thi/pen/KKpQaVo). After creating the hierarchy, I replaced all uses of color across the codebase with colors from the hierarchy. This resulted in minor changes to the appearances of some widgets because I had to eliminate the use of some colors, but the differences are not striking.

As new widgets are written, they should try to work with the colors within this hierarchy before defining their own colors. This will prevent an overcrowding of the theme with widget-specific variables such as `SCROLL_BAR_BORDER_COLOR` or `BUTTON_LIGHT`. It will also allow easy customization, because now you only have to change a couple of variables to affect the appearance of your entire app.

My next step is to do this same exercise for border radiuses, text sizes, opacities, margins & paddings, and widget sizes. Before that, though, I'd love to know if this is even worthwhile.

### DIFF
--- a/druid/examples/styled_text.rs
+++ b/druid/examples/styled_text.rs
@@ -57,7 +57,7 @@ fn ui_builder() -> impl Widget<AppData> {
     // text_size gets a custom key which we set with the env_scope wrapper.
     let styled_label =
         Label::new(|data: &AppData, _env: &_| format!("Size {:.1}: {}", data.size, data.text))
-            .text_color(theme::PRIMARY_LIGHT)
+            .text_color(theme::COLOR_PRIMARY_LIGHT)
             .text_size(MY_CUSTOM_TEXT_SIZE)
             .env_scope(|env: &mut druid::Env, data: &AppData| {
                 env.set(MY_CUSTOM_TEXT_SIZE, data.size)
--- a/druid/examples/styled_text.rs
+++ b/druid/examples/styled_text.rs
@@ -46,7 +46,7 @@ fn main() -> Result<(), PlatformError> {
 
 fn ui_builder() -> impl Widget<AppData> {
     // This is druid's default text style.
-    // It's set by theme::LABEL_COLOR and theme::TEXT_SIZE_NORMAL
+    // It's set by theme::COLOR_BASE_LIGHTEST and theme::TEXT_SIZE_NORMAL
     let label =
         Label::new(|data: &String, _env: &_| format!("Default: {}", data)).lens(AppData::text);
 
--- a/druid/examples/textbox.rs
+++ b/druid/examples/textbox.rs
@@ -27,7 +27,7 @@ fn main() {
             env.set(theme::COLOR_BASE_DARKEST, Color::WHITE);
             env.set(theme::COLOR_BASE_LIGHTEST, Color::BLACK);
             env.set(theme::CURSOR_COLOR, Color::BLACK);
-            env.set(theme::BACKGROUND_LIGHT, Color::rgb8(230, 230, 230));
+            env.set(theme::COLOR_BASE_DARK, Color::rgb8(230, 230, 230));
         })
         .use_simple_logger()
         .launch("typing is fun!".to_string())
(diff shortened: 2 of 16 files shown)


cmyr:
Interesting.

I haven't used this particular pattern myself, but I can see its appeal. My immediate gut reaction isn't 100% supportive, though, and I think I have two primary criticisms.

Before going too far, I want to say that I think this *would* be an improvement on the current set of keys, which we've sort of added pretty haphazardly, and which feel difficult to use correctly. That said, I think that desktop UI is different from web UI; there is less expectation of customization on the desktop.

Additionally: my main experience doing user-facing application development has been on iOS and the mac, so I am heavily inspired by AppKit/UIKit, and for druid I am paying a lot of attention to SwiftUI.

So: I have two related concerns with this approach:

- *semantics*: I think there is tremendous value in giving *semantic* rather than *descriptive* names to colors. Take a look at the list of [ui colors exposed by AppKit](https://developer.apple.com/documentation/appkit/nscolor/ui_element_colors?language=objc), and for a less extreme version, the [material design color system](https://material.io/design/color/#color-theme-creation). These names don't tell you what a color looks like, they tell you *how it is used*. By using these sorts of established names, we make it easier to match the appearance of other parts of the OS, *even between releases*.

- *accessibility (and dark mode)*: another important quality of semantic color names is that they provide a lot of flexibility to implement accessibility features and dark mode. For accessibility this might mean increasing color contrast, inverting colors, or otherwise avoiding certain problematic color combinations (red/green color blindness being the best example). And for dark mode this is even more stark; If someone has asked for `COLOR_BASE_LIGHTEST` for their text, it seems incorrect for us to simply choose to replace this with a dark color based on dark-mode settings. If, however they have chosen `TEXT_COLOR`, however, it is only reasonable that this color changes based on other factors.

In any case, I'm happy to hear an alternative perspective; I haven't used a system like what is described here, so I don't feel I can really evaluate its strengths. Let me know what you think!

Zarenor:
I think I agree with Colin here, in both thinking a more systematic, well-defined system is a much better thing, and in thinking that naming them in a way which describes their usage rather than their color is important. 

As I've discussed before, the UI system I've done the most work in is WPF. [Here](https://docs.microsoft.com/en-us/dotnet/api/system.windows.systemcolors?view=netframework-4.8) is the list of color keys provided by WPF. You can shadow these to override colors locally, or whole-app, and you can shadow them with dynamic options to allow runtime customization, if you want. I'm not sure if we want to reach for such a large hammer yet, but I think this rename and the KeyOrValue change gets us close to this idea (when also combined with EnvScope).

These are named in a mostly-semantic manner. The exceptions are the 'light' and 'dark' colors (which are intended for 3d effect, i;e shadowing, and might not be inverted in a dark or high-contrast theme), and the 'gray' keys, which are intended for disabled objects (I don't think we currently have a 'disabled' concept in druid).

While I think we should come up with our own list, one much like the AppKit list or this one feels to me more suitable for what we're doing here.

s3thi:
@cmyr @Zarenor fair points. This was my first stab at fixing an issue that had been annoying me for a while. I'm going spend a bit of time looking at how AppKit, Material, WPF, and Flutter handle this stuff and update this PR.

I want to mention that this is not the only way design systems are put together on the Web, it just happens to be what I'm comfortable with. I've seen folks build systems that are very similar to what AppKit does, and they work out well too.

That said, the advantage of the color hierarchy in this PR over what desktop toolkits do is that this system makes it easy to build new widgets that visually fit well with the existing set of widgets. You just have to use colors (or margins, or paddings, or ...) from the palette defined here, and things will look coherent. With semantic keys, this becomes harder.

Perhaps a middle ground would work better, where you first define a color palette and then define semantic keys that are derived from that palette. So you could define a `COLOR_BASE_LIGHT`, and then a `COLOR_LABEL_SECONDARY` that is set to the value of `COLOR_BASE_LIGHT` in the dark theme. If there is a light theme, `COLOR_LABEL_SECONDARY` could probably be set to `COLOR_BASE_DARKER`. This would ensure we have semantic keys for the most important bits of the UI, while also giving users a fixed color palette that they can use when they build custom widgets.

Seems like this is what Material is doing, but I'll have to dig deeper.

cmyr:
Yep I 100% agree that this PR is an improvement over what we currently have; I just also think that if we're going to dig into this problem (which we should do!) we should consider all our options and try to figure out what makes the most sense.

I 100% agree that whatever scheme we come up with should make it easy for controls to work together.

PR #644, opened 2020-03-12
Set parse widget state in WidgetAdded

cmyr:
This was an oversight from the lifecycle change; the parse state
would not be set until the first update was received.


futurepaul:
Another helpful illustration of lifecycle!

PR #653, opened 2020-03-12
Sunflower

kindlychung:


cmyr:
This is neat but I think it's going to go at the end of the review queue, I have a bunch of other stuff I need to look at :(

cmyr:
ultimately I think this is very cool but don't think it really illustrates enough currently un-illustrated features to warrant being in `examples`. Thank you though!

PR #657, opened 2020-03-14
Fix: update not propagated to the children

chris-zen:
`ViewSwitcher` does not propagate updates to the children when the data has changed.
The fix will propagate the update to the active child whenever the data changes.

### DIFF
--- a/druid/src/widget/view_switcher.rs
+++ b/druid/src/widget/view_switcher.rs
@@ -70,7 +70,7 @@ impl<T: Data, U: PartialEq> Widget<T> for ViewSwitcher<T, U> {
         }
     }
 
-    fn update(&mut self, ctx: &mut UpdateCtx, old_data: &T, data: &T, env: &Env) {
+    fn update(&mut self, ctx: &mut UpdateCtx, _old_data: &T, data: &T, env: &Env) {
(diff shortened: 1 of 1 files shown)


druid/src/widget/view_switcher.rs
```diff
@@ -79,6 +79,9 @@ impl<T: Data, U: PartialEq> Widget<T> for ViewSwitcher<T, U> {
         }
 
         if !old_data.same(data) {
+            if let Some(child) = self.active_child.as_mut() {
```
cmyr: cool, yea, this was definitely a bug.

I don't actually think we should be doing the old_data.same(data) check; we should just entirely delegate this to the child, which can do what it needs. We also shouldn't call `request_paint`; that's also up to the child.

chris-zen: @cmyr Thanks for the feedback.
I addressed your comment, and checked that it still works with my code.

chris-zen:
@cmyr sorry, I messed up while trying to amend a commit and deleted the remote branch by mistake, so Github closed the PR. 🤦‍♂ 

I've recreated this same PR here: https://github.com/xi-editor/druid/pull/662

PR #656, opened 2020-03-14
Update main crate docs

cmyr:

PR #658, opened 2020-03-14
Update examples/hello

cmyr:
I wrote this for #656, and I think it's an improvement, but it is no longer a counter, which is a nice reactive example. thoughts?

image


fishrockz:
Verified to work nicely on Linux

The UI is getting nicer and nicer every time I look, This shows off what can be achieved  very easily and is nicer that the old example.

+1

PR #660, opened 2020-03-14
fix #648

### LINKED ISSUE #648: request_layout should imply request_paint
I think that if you call `request_layout`  on a context, it should also mark the widget as needing painting; I don't think it makes a lot of sense to do layout without painting, and this will save always needing to call both methods.

If someone makes this PR, they should also look for anywhere in our code where we do call these together, and remove the `request_paint` call.

kindlychung:


druid/src/contexts.rs
```diff
@@ -150,6 +150,7 @@ impl<'a> EventCtx<'a> {
     /// [`layout`]: widget/trait.Widget.html#tymethod.layout
     pub fn request_layout(&mut self) {
         self.base_state.needs_layout = true;
+        self.request_paint();
```
cmyr: Maybe let's just do `self.base_state.needs_inval = true` directly?

kindlychung: done.

cmyr:
@raphlinus can I get a nod from you that this makes sense, and I haven't overlooked any future plans?

PR #662, opened 2020-03-14
Fix: update not propagated to the children

chris-zen:
ViewSwitcher does not propagate updates to the children when the data has changed.
The fix will propagate the update to the active child whenever the data changes.


s3thi:
Looks good, thanks :)

PR #663, opened 2020-03-15
Interpolation options and better Alpha

fishrockz:
Complete tec debt identified in #578 

This MR allows the user of the image widget to determine the interpolation used for plotting.

And only stores the alpha channel if it is used.

### DIFF
--- a/druid/src/widget/image.rs
+++ b/druid/src/widget/image.rs
@@ -118,6 +118,7 @@ pub struct ImageData {
     pixels: Vec<u8>,
     x_pixels: u32,
     y_pixels: u32,
+    format: ImageFormat,
 }
 
 impl ImageData {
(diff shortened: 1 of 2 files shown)


druid/src/widget/image.rs
```diff
@@ -58,6 +60,17 @@ impl<T: Data> Image<T> {
...
     }
+
+    /// A builder-style method for specifying the interpolation strategy.
+    pub fn interpolation_mode(mut self, interpolation: InterpolationMode) -> Self {
+        self.interpolation = interpolation;
+        self
+    }
+
+    /// Modify the widget's `InterpolationMode`.
+    pub fn set_interpolation(&mut self, interpolation: InterpolationMode) {
```
cmyr: I'd like to give these consistent names, even if it's long, so this would be `set_interpolation_mode`.

druid/src/widget/image.rs
```diff
@@ -113,17 +128,36 @@ impl ImageData {
...
 
     /// Load an image from a DynamicImage from the image crate
     pub fn from_dynamic_image(image_data: image::DynamicImage) -> ImageData {
-        let rgb_image = image_data.to_rgba();
-        let sizeofimage = rgb_image.dimensions();
-        ImageData {
-            pixels: rgb_image.to_vec(),
-            x_pixels: sizeofimage.0,
-            y_pixels: sizeofimage.1,
+        match image_data.color() {
```
cmyr: this method is a bit cluttered but I don't have a great suggestion. I might add a constructor method to `ImageData`? But this isn't a blocker or anything.

druid/src/widget/image.rs
fishrockz: The fill has mode for the builder and set for the updater. I think that's what you wanted so i just did the same but changed fill to interpolation.

I can change both..

druid/src/widget/image.rs
fishrockz: I have broken this up in to the match and then separate functions for with and without alpha.

This seems like a good idea as it means users can pick as well as having it workout what seems best.

druid/src/widget/image.rs
cmyr: what I mean is that I think the name should be `interpolation_mode` and `set_interpolation_mode`; they should match, just one has  `set`.

fishrockz: I mean that we have
```
    pub fn fill_mode(mut self, mode: FillStrat) -> Self
    pub fn set_fill(&mut self, newfil: FillStrat)
```
so to be consistent we would have
```
    pub fn interpolation_mode(mut self, interpolation: InterpolationMode) -> Self 
    pub fn set_interpolation(&mut self, interpolation: InterpolationMode) 
```
But do we want `set_fill` to be `set_fill_mode` so we have

```
    pub fn fill_mode(mut self, mode: FillStrat) -> Self
    pub fn set_fill_mode(&mut self, newfil: FillStrat)
    pub fn interpolation_mode(mut self, interpolation: InterpolationMode) -> Self 
    pub fn set_interpolation_mode(&mut self, interpolation: InterpolationMode) 
```

cmyr: ah gotcha. yea, that's an oversight, we can fix that in another patch? Let's do these 'right', and then we can patch up the other one. :)

fishrockz:
@cmyr Bump

I have crafted the history to keep all the changes in different commits but as we don't have merge commits I don't really see the point of braking this in to more PRs.

PR #664, opened 2020-03-15
[Do not merge] prototype Container::background as a KeyOrValue

chris-zen:
This is a prototype of how `Container::background` could work with a `KeyOrValue<Brush>` rather than a `PaintBrush`.

The PR is mainly created for making discussions easier at this stage.

I only implemented one case of Gradient (Linear) to prototype the idea, the rest of them (Radix and Fixed) could be implemented later on following the same idea.


cmyr:
see #666 for my take on this

chris-zen:
Thanks for the work on #666 !
That supersedes this work ;-P

PR #665, opened 2020-03-15
Add set_text method to Label.

xStrom:
This PR adds the `set_text` method to `Label` which allows changing the text. For changing static strings, previously the only solution was to recreate `Label` and for changing dynamic strings there's only `update`. (A bit more context in [Zulip](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/Dynamic.20label))

Unfortunately, in order to support `LabelText::Dynamic` the `set_text` method must also take in `Data` and `Env`.

As it stands right now, just recreating `Label` might be reasonable. However as `Label` might grow with all sorts of style properties, the option to just change the text and not have to recreate the whole style can be a significant win.

### DIFF
--- a/druid/src/widget/label.rs
+++ b/druid/src/widget/label.rs
@@ -81,6 +81,19 @@ impl<T: Data> Label<T> {
         }
     }
 
+    /// Set a new text.
+    ///
+    /// Takes an already resolved string as input.
+    ///
+    /// If you're looking for full [`LabelText`] support,
+    /// then you need to create a new [`Label`].
+    ///
+    /// [`Label`]: #method.new
+    /// [`LabelText`]: enum.LabelText.html
+    pub fn set_text(&mut self, text: impl Into<String>) {
+        self.text = LabelText::Specific(text.into());
+    }
+
     /// Set text alignment.
     #[deprecated(since = "0.5.0", note = "Use an Align widget instead")]
     pub fn text_align(self, _align: UnitPoint) -> Self {


cmyr:
hmm, I hadn't considered that issue with data/env.

I think we should say that if you need to do this with a dynamic label, you should just resolve it yourself first before passing it in?

xStrom:
It isn't great, I agree. Also I think the `Data` and `Env` are actually also needed for `LocalizedString`.

I think it would be a stumbling block to require the `LabelText` to already be resolved, because you don't need to resolve it yourself when creating the `Label`. Also it would require more typing, because it couldn't be a one liner anymore.

So in that sense, I think having `Data` and `Env` required by `set_text` makes things easier for 2/3 cases, because you would need access to those references anyway.

cmyr:
Yea, I see there isn't a good clear answer. My feeling is that changing the text in `AnimFrame` in this way (the only way that doesn't work) is a pretty major edge case, and I don't mind if the user has to do more typing in that particular case; it's more important that the common case be clear and smooth.

I am coming back then to requiring the user to do the resolution. It will only be required in this very specific case.

I think for convenience we add a constructor to `LabelText`.

```rust
impl<T> LabelText<T> {
    fn new(text: impl Into<LabelText<T>>) -> Self {
        text.into()
    }
}

// while you are handling anim frame:
let mut new_text = LabelText::new(|d: &MyData, _: &Env| format!("data: {}", d));
new_text.resolve(data, env);
self.label.set_text(new_text);
```

this isn't that bad I don't think. And again, you will only _ever_ need to resolve in this specific case.

cmyr:
Also: in the world where you're doing it this way you have other options; for instance you could just manually compute the new string (holding onto the `LocalizedString` and calling `resolve` yourself) and then call `set_text` with just a `String`.

xStrom:
In my own specific case I'm already just using `String` via `format!` so I don't need to resolve anything.

Now if we require pre-resolved text, do you think the method documentation will be enough or is it maybe also worth it to call the new method something like `set_resolved_text`?

cmyr:
Again, you only need resolved text in the very specific case that you are changing this during `AnimFrame`.

I think it's reasonable to mention in the docs that changing the text during `AnimFrame` won't trigger resolution, or something.

It's also quite likely that we should add more detail to the docs for `AnimFrame`. Part of me kind of thinks that it should be a separate function in `Widget`, but 🤷‍♂.

xStrom:
You keep focusing on `AnimFrame` so I just want to make sure we're on the same page here.

The `Label` widget does text resolving in only two cases:
1. While handling `LifeCycle::WidgetAdded`
2. In `update` when `old_data` doesn't match `data`

In all other cases changing of text would require manual pre-resolving, e.g. if you want to call `set_text` in the parent's `event`, `lifecycle`, `layout`, `paint`, or even if you want to update the text without changing data in `update`.

I don't have a long history of using druid so my only real-life experience is indeed `AnimFrame` for which I don't need resolving in any case, because I just use `format!`.

So I guess my point is that *all* calls to `set_text` with `LabelText::Localized` / `LabelText::Dynamic` would require pre-resolving, unless you immediately follow up with an `update` that actually changes the `Data`.

That's why I included `Data` and `Env` in the `set_text` signature, because it's practically always required for 2/3 of the `LabelText` types.

If I am misunderstanding something, let me know. If we're in agreement about the ramifications here and you still think that it makes sense to optimize `set_text` for the `LabelText::Specific` use case and require pre-resolving for other types, then I'm happy to make that change.

cmyr:
ah good point. Label expects to get `WidgetAdded`, which means we would need to call `children_changed` after `set_text`. The more I think about it the more it all feels like a footgun.

My new thinking: maybe `set_text` *only* takes a `String` (or `impl into<String>`). If you want to do some fancy non-data based thing, you need to make a String on your own. This would also let you just hold onto your own `LocalizedString` and handle resolving it on your own. This will make it hard to misuse; if you see you can only pass a `String`, you can read the docs, which will explain the situation.

xStrom:
I think minimizing footguns is good, especially because right now the only real world demand (by me) is satisfied by the `String`-only method.

PR #666, opened 2020-03-15
Painter & BackgroundBrush sketch

cmyr:
how to have more fully custom painting without having to do full
custom widgets.


druid/src/paintbrush.rs
```diff
@@ -0,0 +1,97 @@
...
+///
+/// [`paint`]: widget/trait.Widget.html#tymethod.paint
+/// [`Data`]: trait.Data.html
+/// [`Env`]: struct.Env.html
+pub enum BackgroundBrush<T> {
+    Color(Color),
+    ColorKey(Key<Color>),
+    Linear(LinearGradient),
+    Radial(RadialGradient),
+    Fixed(FixedGradient),
```
chris-zen: I guess that we might have `LinearKey`, `RadialKey` and `FixedKey` in the future without breaking changes.

cmyr: possibly; I'm not 100% sure those types belong in the env. But good catch, I will mark this as `non_exhaustive` to ensure adding an item is non-breaking.

PR #668, opened 2020-03-15
Stack layout with an example

kindlychung:


cmyr:
sorry to leave this hanging, getting this widget right is subtle, and will require doing some thinking about what the correct behaviour should be, but I'll try and block out some time for this in the next few days!

druid/src/widget/stack.rs
```diff
@@ -0,0 +1,112 @@
...
+    ///
+    /// See also `with_child`.
+    pub fn add_child(&mut self, child: impl Widget<T> + 'static) {
+        self.children.push(WidgetPod::new(child).boxed());
+    }
+}
+
+impl<T: Data> Widget<T> for Stack<T> {
+    fn event(&mut self, ctx: &mut EventCtx, event: &Event, data: &mut T, env: &Env) {
+        for child in &mut self.children.iter_mut().rev() {
```
cmyr: this looks generally good, but one thing I would want is to treat mouse events specially; they should only be delivered to the topmost widget that they intersect with; this should be as easy as going (for the different mouse events):

```rust
if let Some(active_child) = self.children.iter_mut().rev().find(|child| child.layout_rect().contains(mouse_pos) {
    active_child.event(..);
}
```

And then we can also get rid of the `is_handled` check.

druid/src/widget/stack.rs
```diff
@@ -0,0 +1,112 @@
...
+        bc: &BoxConstraints,
+        data: &T,
+        env: &Env,
+    ) -> Size {
+        bc.debug_check("Stack");
+        let loosened_bc = bc.loosen();
+        let mut max_width: f64 = 0.;
+        let mut max_height: f64 = 0.;
+        for child in &mut self.children {
+            let child_size: Size = child.layout(layout_ctx, &loosened_bc, data, env);
```
cmyr: This type annotation is unnecessary, not that it's a big deal

druid/src/widget/stack.rs
```diff
@@ -0,0 +1,112 @@
...
+            max_width = max_width.max(child_size.width);
+            max_height = max_height.max(child_size.height);
+            // Stash size.
+            let rect = Rect::from_origin_size(Point::ORIGIN, child_size);
```
cmyr: we can now just do `child_size.to_rect()`.

druid/src/widget/stack.rs
```diff
@@ -0,0 +1,112 @@
...
+            child.set_layout_rect(rect);
+        }
+        Size {
```
cmyr: I would prefer using `Size::new` for this; it is more concise.

druid/src/widget/stack.rs
```diff
@@ -0,0 +1,112 @@
...
+    PaintCtx, UpdateCtx, Widget, WidgetPod,
+};
+
+/// A container that lays out its children along the z-axis, first child at bottom, last child on top.
+#[derive(Default)]
+pub struct Stack<T> {
+    children: Vec<BoxedWidget<T>>,
+}
+
+// struct ChildWidget {
```
cmyr: cleanup :)

druid/src/widget/stack.rs
```diff
@@ -0,0 +1,112 @@
...
+use crate::kurbo::{Point, Rect, Size};
+use crate::{
+    BoxConstraints, BoxedWidget, Data, Env, Event, EventCtx, LayoutCtx, LifeCycle, LifeCycleCtx,
+    PaintCtx, UpdateCtx, Widget, WidgetPod,
+};
+
+/// A container that lays out its children along the z-axis, first child at bottom, last child on top.
+#[derive(Default)]
+pub struct Stack<T> {
+    children: Vec<BoxedWidget<T>>,
```
cmyr: I find that we are using this inconsistently, and I'm inclined to remove it; I find the name unclear. I would prefer the more annoying `WidgetPod<T, Box<dyn Widget<T>>` for now, thanks.

kindlychung:
@cmyr Please have a look at the changes.

druid/src/widget/stack.rs
```diff
@@ -0,0 +1,118 @@
+// Copyright 2018 The xi-editor Authors.
```
cmyr: 2020

druid/src/widget/stack.rs
```diff
@@ -0,0 +1,118 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! A widget that arranges its children on top of one another.
+
+use crate::kurbo::Size;
```
cmyr: you can actually import this from `crate` below.

druid/src/widget/stack.rs
```diff
@@ -0,0 +1,118 @@
...
+impl<T: Data> Widget<T> for Stack<T> {
+    fn event(&mut self, ctx: &mut EventCtx, event: &Event, data: &mut T, env: &Env) {
+        match event {
+            Event::MouseDown(e) | Event::MouseUp(e) | Event::MouseMoved(e) => {
+                if let Some(active_child) = self
+                    .children
+                    .iter_mut()
+                    .rev()
+                    .find(|child| child.layout_rect().contains(e.pos))
+                {
```
cmyr: thinking harder this doesn't actually work; it doesn't take account of whether or not a widget is active, and it won't correctly deliver the mouse moved event when the mouse first exits a widget.

I think getting this correct is going to take some fairly tricky logic. One option would be to introduce the idea of a 'hit zone' or a 'hit mask' that lets us represent the obscured portion of a widget. Another option would be to always deliver events to all widgets; and a last option would be to only deliver events to the top widget. 😒

kindlychung: One last move move event is sent when mouse is about to leave, right?

cmyr: yes, if the mouse has been over a widget, and moves outside that widget, we deliver the first event that arrives after it has moved.

PR #669, opened 2020-03-15
Align Container and debug borders to pixels.

xStrom:
### How stroke works
The coordinates provided to stroke are fully respected and not rounded to an integer. Furthermore the stroke is positioned exactly on top of the provided coordinates. This means that if you draw a 1px wide line at pixel boundaries, 0.5px of the line width will be on one side pixel and  the other 0.5px on the other side. Of course the pixel is usually the smallest unit a display has, so the 1px line ends up looking like a 2px line, although usually alpha blended. The solution is to specify the midpoint of a pixel instead, i.e. if you want the line to appear on the 2nd pixel row you specify 1.5 as the point.

stroke-point

### Steps towards improving druid
This PR changes the way the `Container` widget border and the debug borders are drawn. In both cases they are drawn inwards, i.e. within the widget size, with pixel accuracy.

This PR also moves the `Container` border drawing after the background drawing so that you can have both active at the same time.

### Real life example (400% zoom)
real-changes

### Future work
* There is still more stroke usage in druid that isn't pixel perfect. These should all be addressed.
* The widget layouts themselves [should be quantized to integers](https://github.com/xi-editor/druid/pull/148#issuecomment-530441059). Otherwise even if the stroke usage is pixel-perfect, it will have a non-integer origin. This is currently affecting `Label` and probably everything that uses `Label`'s layout in calculations of its own layout, e.g. `Button`.

### DIFF
--- a/druid/src/core.rs
+++ b/druid/src/core.rs
@@ -19,7 +19,7 @@ use std::collections::VecDeque;
 use log;
 
 use crate::bloom::Bloom;
-use crate::kurbo::{Affine, Insets, Point, Rect, Shape, Size};
+use crate::kurbo::{Affine, Insets, Rect, Shape, Size};
 use crate::piet::RenderContext;
 use crate::{
     BoxConstraints, Command, Data, Env, Event, EventCtx, LayoutCtx, LifeCycle, LifeCycleCtx,
@@ -268,10 +268,11 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
         paint_ctx.z_ops.append(&mut ctx.z_ops);
 
         if env.get(Env::DEBUG_PAINT) {
-            let rect = Rect::from_origin_size(Point::ORIGIN, ctx.size());
+            const BORDER_WIDTH: f64 = 1.0;
+            let rect = ctx.size().to_rect().inset(BORDER_WIDTH / -2.0);
             let id = self.id().to_raw();
             let color = env.get_debug_color(id);
-            ctx.stroke(rect, &color, 1.0);
+            ctx.stroke(rect, &color, BORDER_WIDTH);
         }
 
         self.state.needs_inval = false;
--- a/druid/src/widget/container.rs
+++ b/druid/src/widget/container.rs
@@ -15,7 +15,7 @@
 //! A widget that provides simple visual styling options to a child.
 
 use super::BackgroundBrush;
-use crate::shell::kurbo::{Point, Rect, RoundedRect, Size};
+use crate::shell::kurbo::{Point, Rect, Size};
 use crate::{
     BoxConstraints, Color, Data, Env, Event, EventCtx, KeyOrValue, LayoutCtx, LifeCycle,
     LifeCycleCtx, PaintCtx, RenderContext, UpdateCtx, Widget, WidgetPod,
@@ -114,8 +114,8 @@ impl<T: Data> Widget<T> for Container<T> {
         bc.debug_check("Container");
 
         // Shrink constraints by border offset
-        let border_width = match self.border {
-            Some(ref border) => border.width.resolve(env),
+        let border_width = match &self.border {
+            Some(border) => border.width.resolve(env),
             None => 0.0,
         };
         let child_bc = bc.shrink((2.0 * border_width, 2.0 * border_width));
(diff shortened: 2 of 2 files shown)


cmyr:
Awesome, this is 100% welcome and important. There are a few places where I've done this manually (mostly in runebender I guess?) and I haven't written it up anywhere. Would be good to have this be in the docs somewhere...

cmyr:
and also 100% layouts should be on the grid. Maybe the simplest option would be to `floor` all sizes that are used as layouts? this could cause some weird issues though. Maybe also `BoxConstraints::constrain` should enforce that sizes are integers; this happens earlier, so we'd be less likely to do things like accumulate non-rounded numbers when doing layout math.

xStrom:
I think `ceil` is better than `floor` because otherwise there will be drawing overflow. I was thinking of sending a PR to kurbo to add both methods to `Size`.

Solving this in some more generic way like `BoxConstraints::constrain` could be worth it, as it will help people who aren't that familiar with this whole mid-point drawing business. That said, I haven't gone over all the widgets yet to properly comment on how to best do it.

cmyr:
Noted about ceil. That pr would be welcome in kurbo.

xStrom:
Okay so the kurbo changes have been merged and I'll try to tackle some of the future work (i.e. fix this issue more broadly and document it better) later this week. I'd prefer to do that in separate PRs.

So I'd like to move forward with merging these changes here. Let me know if anything needs adjusting.

cmyr:
looking now.

druid/src/core.rs
```diff
@@ -268,10 +268,16 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
...
 
         if env.get(Env::DEBUG_PAINT) {
-            let rect = Rect::from_origin_size(Point::ORIGIN, ctx.size());
+            const BORDER_WIDTH: f64 = 1.0;
+            const BORDER_MIDPOINT: f64 = BORDER_WIDTH / 2.0;
+            let size = ctx.size();
+            let rect = Rect::from_origin_size(
+                (BORDER_MIDPOINT, BORDER_MIDPOINT),
+                (size.width - BORDER_WIDTH, size.height - BORDER_WIDTH),
+            );
```
cmyr: another way to think about this  is in terms of insets:

```rust
const BORDER_WIDTH: f64 = 1.0;
let bounds = ctx.size().to_rect();
let stroke_inset = Insets::uniform(BORDER_WIDTH / 2.0);
let stroke_rect = rect - stroke_inset;
```

druid/src/widget/container.rs
```diff
@@ -135,29 +135,35 @@ impl<T: Data> Widget<T> for Container<T> {
...
-        paint_ctx.clip(panel);
-
         if let Some(background) = self.background.as_mut() {
+            let bg_rect = RoundedRect::from_origin_size(
+                Point::ORIGIN,
+                paint_ctx.size().to_vec2(),
+                self.corner_radius,
+            );
+
+            if let Err(e) = paint_ctx.save() {
```
cmyr: I really get annoyed by this api, will look into a helper method (this is unrelated to the current  PR)

druid/src/widget/container.rs
```diff
@@ -135,29 +135,35 @@ impl<T: Data> Widget<T> for Container<T> {
...
-
-        if let Err(e) = paint_ctx.save() {
-            log::error!("{}", e);
-            return;
-        }
-
-        paint_ctx.clip(panel);
-
         if let Some(background) = self.background.as_mut() {
+            let bg_rect = RoundedRect::from_origin_size(
```
cmyr: my current preferred way to do this (until https://github.com/linebender/kurbo/pull/92 merges) is `paint_ctx.to_rect().to_rounded_rect(self.corner_radius);` 😒

druid/src/core.rs
xStrom: The inset helper methods are definitely helpful and made the code more concise. In this case the insets do have to be negative though, to shrink the rect.

druid/src/widget/container.rs
xStrom: Yeah the `with_save` method will reduce human failure chance.

druid/src/widget/container.rs
xStrom: Using the helper methods now!

xStrom:
I did some additional testing and found that `Container` had a bug where it was using `paint` instead of `paint_with_offset` even though it sets an offset in its `layout` method. Fixed that as well.

PR #671, opened 2020-03-15
example of using DynamicImage

kindlychung:
changing pixels on the fly


fishrockz:
What are you trying to achieve here?

If you just want to create a image at run time I think we can do that without using unsafe.

kindlychung:
> What are you trying to achieve here?
> 
> If you just want to create a image at run time I think we can do that without using unsafe.
> 
> I'll try and make a example

Sure, would love see it.

PR #673, opened 2020-03-16
Adding a reference to the data to the child_builder of a ViewSwitcher

chris-zen:
My model contains some fields that are only necessary while building the view. The current `child_builder` signature does not include a reference to the data, which forces me to clone the whole data and transfer the ownership to the closure.

For an example of what I mean you can have a look to this code:

- The clone happens [here](https://github.com/chris-zen/kiro-synth/blob/druid-tabs/kiro-synth-host/src/ui/view.rs#L103)
- The building happens [here](https://github.com/chris-zen/kiro-synth/blob/druid-tabs/kiro-synth-host/src/ui/view.rs#L108)

In that example I store in [the data model](https://github.com/chris-zen/kiro-synth/blob/druid-tabs/kiro-synth-host/src/ui/model.rs#L39) two kind of things (for convenience):

- data that never changes from the user interaction like origin, min, max, step.
- data that changes on user interaction like the value and the modulation.

So I need to pass the data model across the view building functions, as well as the regular Widget data (which can be adapted using lenses).

### DIFF
--- a/druid/src/widget/view_switcher.rs
+++ b/druid/src/widget/view_switcher.rs
@@ -22,9 +22,12 @@ use crate::{
 /// A widget that can switch dynamically between one of many views depending
 /// on application state.
 
+type ChildPicker<T, U> = dyn Fn(&T, &Env) -> U;
+type ChildBuilder<T, U> = dyn Fn(&U, &T, &Env) -> Box<dyn Widget<T>>;
+
 pub struct ViewSwitcher<T, U> {
-    child_picker: Box<dyn Fn(&T, &Env) -> U>,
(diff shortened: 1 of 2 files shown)


s3thi:
This looks good to me. Could you fix that Clippy error that's showing up on the CI?

chris-zen:
@s3thi all checks passed ;-)

druid/src/widget/view_switcher.rs
```diff
@@ -22,9 +22,12 @@ use crate::{
 /// A widget that can switch dynamically between one of many views depending
 /// on application state.
 
+type ChildPicker<T, U> = dyn Fn(&T, &Env) -> U;
```
cmyr: I approve of this refactor 👍

PR #678, opened 2020-03-16
Encapsulation

fishrockz:
I have written a couple of widgets using encapsulation and it works quite nicely, but I only realised I could do this after @kinnison told me about it when talking about rust.

Would it be helpful to have a example using this patten or would we rather discourage this patten?

There are a couple of issues with this code i would like to get some advice on so even if we dont merge this the review would be helpful.

### DIFF
--- a/druid/examples/image_encapsulate.rs
+++ b/druid/examples/image_encapsulate.rs
@@ -1,0 +1,109 @@
+// Copyright 2020 The xi-editor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! This example shows how to draw an png image.
+//!
+//! Requires the non-default "image" feature to be enabled:
+//! `cargo run --example image_encapsulate --features "images"`
+//!
+
+#[cfg(not(feature = "image"))]
+fn main() {
+    eprintln!("This examples requires the \"image\" feature to be enabled:");
+    eprintln!("cargo run --example image_encapsulate --features \"image\"");
+}
+
+#[cfg(feature = "image")]
+fn main() {
+    use druid::{
+        widget::{Flex, Image, ImageData, Slider, WidgetExt},
+        AppLauncher, BoxConstraints, Env, Event, EventCtx, LayoutCtx, LifeCycle, LifeCycleCtx,
+        PaintCtx, Size, UpdateCtx, Widget, WindowDesc,
+    };
+
+    pub struct ImageSwitcher {
+        image_index: u32,
+        images: Vec<Box<dyn Widget<u32>>>,
+    }
+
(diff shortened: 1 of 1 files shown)


druid/examples/image_encapsulate.rs
```diff
@@ -0,0 +1,108 @@
...
+        ));
+        let png_3 = Box::new(Image::new(
+            ImageData::from_file("examples/PicWithAlpha.png").unwrap(),
+        ));
+
+        let mut col = Flex::column();
+
+        col.add_child(ImageSwitcher::new(vec![png_1, png_2, png_3]), 1.);
+        let mut row = Flex::row();
+        row.add_child(Slider::new(), 1.0);
```
fishrockz: I could not work out how to do this nicely, were should i look for docs? maybe @futurepaul

druid/examples/image_encapsulate.rs
```diff
@@ -0,0 +1,108 @@
...
+                images: newimages,
+            }
+        }
+    }
+
+    impl Widget<f64> for ImageSwitcher {
+        fn event(&mut self, ctx: &mut EventCtx, _event: &Event, data: &mut f64, _env: &Env) {
+            let index_float = *data * (self.images.len() as f64);
+            self.image_index = (self.images.len() as u32 - 1).min(index_float as u32);
+            ctx.request_paint();
```
fishrockz: Im pretty sure this is wrong and i should be doing something in update

druid/examples/image_encapsulate.rs
futurepaul: This behavior has recently changed, check out the flex.rs example for all the options.

druid/examples/image_encapsulate.rs
futurepaul: Yeah, this should be in update. You want all three of these lines in update and it should work. Update is for dealing with changes to data, and that's what you have driving changes to your view. Event is for things that are literally Events (like mouse and keyboard stuff) or Commands.

fishrockz: Thanks paul

druid/examples/image_encapsulate.rs
```diff
@@ -0,0 +1,109 @@
...
+#[cfg(feature = "image")]
+fn main() {
+    use druid::{
+        widget::{Flex, Image, ImageData, Slider, WidgetExt},
+        AppLauncher, BoxConstraints, Env, Event, EventCtx, LayoutCtx, LifeCycle, LifeCycleCtx,
+        PaintCtx, Size, UpdateCtx, Widget, WindowDesc,
+    };
+
+    pub struct ImageSwitcher {
+        image_index: u32,
```
cmyr: I would always use `usize` for any type that is representing an index.

druid/examples/image_encapsulate.rs
```diff
@@ -0,0 +1,109 @@
...
+    impl ImageSwitcher {
+        pub fn new(newimages: Vec<Box<dyn Widget<u32>>>) -> Self {
+            Self {
+                image_index: 0,
+                images: newimages,
+            }
+        }
+    }
+
+    impl Widget<f64> for ImageSwitcher {
```
cmyr: semantically, it feels like the data type for this widget should also be `usize`, assuming it is supposed to represent an index. If you want to modify that data with a slider, you could use a lens to map the float to an int.

cmyr:
(these comments are just for informational purposes, no need to make changes)

cmyr:
@fishrockz I think this probably doesn't make sense as a baked-in example right now, so I'm going to close. Thanks for the illustration, though. If you're motivated, the `Either` example could definitely use some work...

PR #683, opened 2020-03-17
Clickable controller widget

futurepaul:
A possible "clickable" widget design.


druid/src/widget/click.rs
```diff
@@ -0,0 +1,84 @@
...
+// limitations under the License.
+
+//! A clickable [`Controller`] widget.
+//!
+//! [`Controller`]: struct.Controller.html
+
+use crate::widget::Controller;
+use crate::{Data, Env, Event, EventCtx, LifeCycle, LifeCycleCtx, Widget};
+
+/// A clickable [`Controller`] widget. Pass this and a child widget to a
```
cmyr: I would maybe mention the relationship with button; that this is intended for places where you want something to be clickable, but don't want to use the standard button widget?

PR #686, opened 2020-03-17
Add PaintCtx::with_save

### LINKED ISSUE #646: Add PaintCtx::paint_with_offset
I think there should be an easy API for painting something at some offset. This should look a lot like `PaintCtx::with_child_ctx` or `PaintCtx::paint_with_offset`.

Under the covers, to be safe, I think this actually _should_ just call `with_child_ctx`. I think this might be  as simple as taking the `paint` fn  from the `Scroll` widget, and pasting that into a new method on `PaintCtx`?

(and I think that `with_child_ctx` should probably not be public; it's too tricky to use, and `paint_with_offset` should serve the same purpose? Unless I'm missing something? 🤔

cmyr:
This reduces boilerplate when you need to clip or transform briefly
during painting.

- closes #646

### DIFF
--- a/druid/src/contexts.rs
+++ b/druid/src/contexts.rs
@@ -585,6 +585,40 @@ impl<'a, 'b: 'a> PaintCtx<'a, 'b> {
         self.z_ops.append(&mut child_ctx.z_ops);
     }
 
+    /// Saves the current context, executes the closures, and restores the context.
+    ///
+    /// This is useful if you would like to transform or clip or otherwise
+    /// modify the drawing context but do not want that modification to
+    /// effect other widgets.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use druid::{Env, PaintCtx, RenderContext, theme};
+    /// # struct T;
+    /// # impl T {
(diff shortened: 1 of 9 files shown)


druid/src/widget/container.rs
```diff
@@ -135,21 +135,14 @@ impl<T: Data> Widget<T> for Container<T> {
     }
 
     fn paint(&mut self, ctx: &mut PaintCtx, data: &T, env: &Env) {
+        let panel = ctx.size().to_rounded_rect(self.corner_radius);
```
xStrom: This shouldn't be moved out of the `if let Some(background) = self.background.as_mut() {}` block.

druid/src/contexts.rs
```diff
@@ -585,6 +585,40 @@ impl<'a, 'b: 'a> PaintCtx<'a, 'b> {
         self.z_ops.append(&mut child_ctx.z_ops);
     }
 
+    /// Saves the current context, executes the closures, and restores the context.
+    ///
+    /// This is useful if you would like to transform or clip or otherwise
+    /// modify the  drawing context but do not want that modification to
```
xStrom: Nit: double space in 'the  drawing'.

PR #687, opened 2020-03-17
Add expand_width & expand_height

cmyr:
This comes out of my remaining lack of satisfaction with #639. Basically I find the  `FlexParams` business still overkill in the general case, and I  want to get rid of the `loose` and  `tight` distinction. My thinking here is that `tight` is strictly representable in terms of  a child that is set to `expand`, but very often in a container if you use `expand` you will end up with an infinite size on one of the axes. This makes it easier to use `expand` on a specific axis, which will let me delete the loose/tight distinction in `FlexParams`.

### DIFF
--- a/druid/src/widget/sized_box.rs
+++ b/druid/src/widget/sized_box.rs
@@ -69,9 +69,31 @@ impl<T> SizedBox<T> {
     }
 
     /// Expand container to fit the parent.
-    /// It is equivalent to setting width and height to Infinity.
+    ///
+    /// Only call this method if you want your widget to occupy all available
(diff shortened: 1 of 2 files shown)


futurepaul:
Just to make sure I'm following this: the plan is that a child widget of Flex will be given loose constraints by default, and then calling `expand_appropriateaxis` on that widget will make it behave as if it was "tight"?

I like that this is handing behavior to individual widgets where applicable, instead of their container. Makes this knowledge useful in other layout situations.

cmyr:
yep, that is exactly my thinking.

Issue #659, opened 2020-03-14
Add missing layout widgets


cmyr:
There are a number of basic layout widgets in flutter that we don't have.

- [ ] [`LimitedBox`]: A widget that checks if incoming constraints are infinite, and constraints them.
- [ ] [`FittedBox`]: A widget that fits a child within itself, in a number of ways (fill, fit, fillHeight, etc).
- [ ] [`FractionallySizedBox`]: use a fraction of available space
- [ ] [`AspectRatio`]: maintain an aspect ratio (this should report if it is passed tight constraints)
- [ ] [`OverflowBox`]: allow a child to overflow
- [ ] [`ConstrainedBox`]: impose arbitrary(?) constraints on child? I think this should probably be represented as `Fn(BoxConstraints) -> BoxConstraints` or something.

[`LimitedBox`]: https://api.flutter.dev/flutter/widgets/LimitedBox-class.html
[`FractionallySizedBox`]: https://api.flutter.dev/flutter/widgets/FractionallySizedBox-class.html
[`AspectRatio`]: https://api.flutter.dev/flutter/widgets/AspectRatio-class.html
[`OverflowBox`]: https://api.flutter.dev/flutter/widgets/OverflowBox-class.html
[`ConstrainedBox`]: https://api.flutter.dev/flutter/widgets/ConstrainedBox-class.html

In any case: none of these are super-duper-urgent, but I think we would like to have all of them eventually.

kindlychung:
[] Rotate: allows rotation of its child.

JAicewizard:
just a question, but how is drawing outside the window border being handled? would the OverflowBox clip the content on the edge of the window? or would it signal the window to grow somehow. If the later is chosen, will other widgets resize too? will the background be painted on the part of the window that wasn't there before.
If the OverflowBox would grow the window without resizing or painting the background on the new area, it might be nice to use it for dropdowns, since they might not fit withing the window.

cmyr:
Outside of the window or outside of the widget?

I do not expect us to support drawing outside of the window; for cases like drop-down menus, we will be creating new modal windows for the menu.

PR #689, opened 2020-03-17
Add ConstrainedBox widget

futurepaul:
This is progress on #659 

I could see this being more like SizedBox (have an `Option<f64>` for each of min_width, min_height, max_width, max_height instead of this closure), but I wanted to give the closure a shot because it will obviously offer more flexibility. It's very possible that some convenience methods on BoxConstraints could make this easier to use?

The other question is if there's any reason to pass in LayoutCtx and Env?


druid/src/widget/constrained_box.rs
```diff
@@ -0,0 +1,73 @@
...
+
+//! A widget which modifies the BoxConstraints passed to its child
+
+use crate::shell::kurbo::Size;
+use crate::{
+    BoxConstraints, Data, Env, Event, EventCtx, LayoutCtx, LifeCycle, LifeCycleCtx, PaintCtx,
+    UpdateCtx, Widget, WidgetId,
+};
+
+/// A widget with custom BoxConstraints.
```
cmyr: let's link to [`BoxConstraints`].

druid/src/widget/constrained_box.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    fn update(&mut self, ctx: &mut UpdateCtx, old_data: &T, data: &T, env: &Env) {
+        self.inner.update(ctx, old_data, data, env);
+    }
+
+    fn layout(&mut self, ctx: &mut LayoutCtx, bc: &BoxConstraints, data: &T, env: &Env) -> Size {
+        bc.debug_check("ConstrainedBox");
+
+        let child_bc = (self.constrain)(bc);
```
cmyr: we _could_ subsequently call `constrain` ourselves, to make sure that the returned constraints are valid? or we could otherwise validate them here?

druid/src/widget/widget_ext.rs
```diff
@@ -93,6 +95,43 @@ pub trait WidgetExt<T: Data>: Widget<T> + Sized + 'static {
...
+                (bc.max().width, (bc.max().height).max(height)).into(),
+            )
+        })
+    }
+
+    /// Wrap this widget in a [`ConstrainedBox`]. This mutates the widget's
+    /// minimum BoxConstraints width.
+    ///
+    /// [`ConstrainedBox`]: struct.ConstrainedBox.html
+    fn min_width(self, width: f64) -> ConstrainedBox<T> {
```
cmyr: these make me feel like we should *some* validation. It would be very easy to pass these some value that is larger than max width, and I think we should at least log that?

futurepaul: I'm actually guarding against that with: `(bc.max().height).max(height)` but I could log a warning instead? (bc.debug_check also complains about it)

cmyr: that sounds good. Maybe there should be a method on `BoxConstraints` that constraints _other_ box constraints? 🤔

futurepaul: Yeah that makes sense to me.

cmyr:
What's the status on this?

futurepaul:
I need a good motivating example where "min_height" / "min_width" are actually required to get a sense if this is working as intended.

cmyr:
I would check the flutter docs; in particular see if they have one of their little info videos for this widget? they often come up with a good motivating case.

PR #691, opened 2020-03-18
2 flex 2 params

### LINKED ISSUE #607: Feature parity with flutter Flex widget

cmyr:
This is a different way of doing flex params, described in #687. This includes changing all examples to the new api.

The tldr:
- add_child/with_child no longer takes the flex factor
- add_flex_child & with_flex_child takes `FlexParams`
- `f64` has `impl Into<FlexParams>`, so you can still just pass value
- flex is loose by default, and you need to explicitly expand to fill the space
- strictly more layouts are possible!
- there are *gasp* docs!

I'm fairly confident that this is the  correct direction, but welcome all input.

This  includes a commit from #687, supersedes #639  and closes #607.

cc @futurepaul

### DIFF
--- a/druid/src/widget/flex.rs
+++ b/druid/src/widget/flex.rs
@@ -71,7 +71,7 @@ use crate::{
 /// let my_row = Flex::row()
 ///     .cross_axis_alignment(CrossAxisAlignment::Center)
 ///     .must_fill_main_axis(true)
-///     .with_child(Label::new("hello"), 0.0)
+///     .with_child(Label::new("hello"))
 ///     .with_spacer(8.0)
 ///     .with_flex_child(Slider::new(), 1.0);
 /// ```
@@ -84,7 +84,7 @@ use crate::{
 /// let mut my_row = Flex::row();
 /// my_row.set_must_fill_main_axis(true);
 /// my_row.set_cross_axis_alignment(CrossAxisAlignment::Center);
-/// my_row.add_child(Label::new("hello"), 0.0);
+/// my_row.add_child(Label::new("hello"));
 /// my_row.add_spacer(8.0);
 /// my_row.add_flex_child(Slider::new(), 1.0);
 /// ```
@@ -338,8 +338,8 @@ impl<T: Data> Flex<T> {
     /// Builder-style variant of `add_child`.
     ///
     /// Convenient for assembling a group of widgets in a single expression.
-    pub fn with_child(mut self, child: impl Widget<T> + 'static, flex: f64) -> Self {
-        self.add_flex_child(child, flex);
+    pub fn with_child(mut self, child: impl Widget<T> + 'static) -> Self {
+        self.add_flex_child(child, 0.0);
         self
     }
 
@@ -378,10 +378,6 @@ impl<T: Data> Flex<T> {
     }
 
     /// Builder-style method for adding a `flex` spacer to the container.
-    ///
-    /// See [`add_child`] for an overview of `flex`.
-    ///
-    /// [`add_child`]: #method.add_child
     pub fn with_flex_spacer(mut self, flex: f64) -> Self {
         self.add_flex_spacer(flex);
         self
@@ -407,20 +403,13 @@ impl<T: Data> Flex<T> {
         self.fill_major_axis = fill;
     }
 
-    /// Add a child widget.
+    /// Add a non-flex child widget.
     ///
-    /// If `flex` is zero, then the child is non-flex. It is given the same
-    /// constraints on the "minor axis" as its parent, but unconstrained on the
-    /// "major axis".
+    /// See also [`with_child`].
     ///
-    /// If `flex` is non-zero, then all the space left over after layout of
-    /// the non-flex children is divided up, in proportion to the `flex` value,
-    /// among the flex children.
-    ///
-    /// See also `with_child`.
-    //TODO: make this fn not take `flex`; I just don't want to break api yet.
-    pub fn add_child(&mut self, child: impl Widget<T> + 'static, flex: f64) {
-        self.add_flex_child(child, flex);
+    /// [`with_child`]: #method.with_child
+    pub fn add_child(&mut self, child: impl Widget<T> + 'static) {
+        self.add_flex_child(child, 0.0);
     }
 
     /// Add a flexible child widget.
@@ -461,12 +450,12 @@ impl<T: Data> Flex<T> {
     }
 
     /// Add an empty spacer widget with a specific `flex` factor.
-    ///
-    /// See [`add_child`] for an overview of `flex`.
-    ///
-    /// [`add_child`]: #method.add_child
     pub fn add_flex_spacer(&mut self, flex: f64) {
-        self.add_flex_child(SizedBox::empty(), flex);
+        let child = match self.direction {
+            Axis::Vertical => SizedBox::empty().expand_height(),
+            Axis::Horizontal => SizedBox::empty().expand_width(),
+        };
+        self.add_flex_child(child, flex);
     }
 }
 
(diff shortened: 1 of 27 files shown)


s3thi:
This looks great!

Not sure what I feel about the `FlexParams` struct, though. From what I understand, you only need to construct it manually if a child of the `Flex` container has a different cross-axis alignment than the container. Is that correct? Could this be set with a method instead?

I don't have a very strong opinion on this, but I find it a little overwhelming when an API exposes too many small types like this. Reading through a bunch of supported method calls is easier than looking at the definitions for a bunch of tiny enums or structs, each one of which encapsulates just a tiny bit of functionality.

Bikeshedding aside, these changes are great and the docs are very welcome :)

druid/src/widget/flex.rs
```diff
@@ -338,8 +515,8 @@ impl<T: Data> Widget<T> for Flex<T> {
         for child in &mut self.children {
             if child.params.flex != 0.0 {
                 let major = remaining * child.params.flex / flex_sum;
+                let min_major = 0.0;
 
-                let min_major = if major.is_infinite() { 0.0 } else { major };
```
futurepaul: This feels like a behavior change, but I'm not quite tracking.

cmyr: it is; we used to always pass tight constraints to flex children, and now we always pass loose constraints.

cmyr:
@s3thi I agree that `FlexParams` now feels a bit iffy. Previously there were three possible arguments that a child could have, and this felt more necessary.

I do agree that adding lots of little types is annoying. My feeling here is that this type only needs to be used rarely, though, so it isn't as much of a paper cut.

I could imagine one alternative; the `Flex` itself could have an API that lets you override the `CrossAxisAlignment` of a child. It's hard to know what this would look like, though; does it take an index, or does it apply only to the last-added child, or what? I'm really not sure.

s3thi:
@cmyr I can't come up with an API that feels right, either. Using an index to modify the alignment would require the user to somehow keep track of the index they want to modify. Applying the new alignment to the last-added child feels stateful in a weird way.

As I said, I don't have strong opinions here. It's just something I wanted to mention, because I've been looking at Flutter docs and running into a ton of small types. I often end up opening 10 browser tabs for documentation and forgetting what I was even going to look up.

I suppose these types only become a problem if you're required to use them in the normal course of building your UI. `FlexParams` is something that's you're absolutely required to know of and use, so it's probably okay.

cmyr:
well `FlexParams` is only *kind* of something you need to know? You can certainly build lots of applications without ever using it, and only using `f64`. We'll just need to make sure that the documentation is clear.

cmyr:
well `FlexParams` is only *kind* of something you need to know? You can certainly build lots of applications without ever using it, and only using `f64`. We'll just need to make sure that the documentation is clear.

cmyr:
Okay, I think I'm convinced this is the right direction. I've added some more docs, and will get this in shortly.

druid/src/widget/flex.rs
```diff
@@ -23,6 +23,121 @@ use crate::{
...
+/// Imagine, for instance, we have a row that is 30 px wide, and we have
+/// two children, both of which want to be 20 px wide. If child #1 is non-flex
+/// and child #2 is flex, the first widget will take up its 20px, and the second
+/// widget will be constrained to 10px.
+///
+/// If, instead, both widgets are flex, they will each be given equal space,
+/// and both will end up taking up 15px.
+///
+/// If both are non-flex, they will both take up 20px, and will overflow the
+/// container.
```
xStrom: Nit: The second comma seems excessive.

druid/src/widget/flex.rs
```diff
@@ -42,6 +157,43 @@ struct Spacer {
...
+/// Generally, when you would like to add a flexible child to a container,
+/// you can simply call [`with_flex_child`] or [`add_flex_child`], passing the
+/// child and the desired flex factor as a `f64`, which has an impl of
+/// `Into<FlexParams>`.
+///
+/// If you need to set additional paramaters, such as a custom [`CrossAxisAlignment`],
+/// you can construct `FlexParams`  directly. By default, the child has the
+/// same `CrossAxisAlignment` as the container.
+///
+/// For an overview  of the flex layout algorithm, see the [`Flex`] docs.
```
xStrom: Nit: Double space in 'overview  of'.

druid/src/widget/flex.rs
```diff
@@ -95,9 +247,34 @@ pub enum MainAxisAlignment {
...
-#[derive(Copy, Clone, Default)]
-struct Params {
-    flex: f64,
+impl FlexParams {
+    /// Create custom `FlexParams` with a specific `flex_factor` and an optional
+    /// [`CrossAxisAlignment`].
+    ///
+    /// You likely only need to create these manually if you need to specify
+    /// a custom alignment; if you only need to use a custom `flex_factor` you
+    /// can pass an `f64`  to any of the functions that take `FlexParams`.
```
xStrom: Nit: There's a double space after \`f64\`.

druid/src/widget/flex.rs
cmyr: good nit!

druid/src/widget/flex.rs
cmyr: ask my about my macbook keyboard 🤢

druid/src/widget/flex.rs
cmyr: ⌨️

Issue #632, opened 2020-03-11
allow setting minimum window size


cmyr:
This will be very similar in implementation to #587.

Finnerale:
I've implemented it but I'm (again) not sure about naming :sweat_smile: 

The function to set a windows initial size is called `WindowDesc::window_size` thus is would be consistent to call the function for minimal size `window_min_size` but it seems odd to me when another function is called `WindowDesc::title` and not `window_title`.

Also based on the discussion in #697 it would make most sense to call it `with_min_size` and `with_title`.

Thus I would call the function `WindowDesc::with_min_size` in my PR and then make another one to rename all other methods of `WindowDesc` to builder style naming

cmyr:
yea, naming is a disaster.

something like `WindowDesc` is _only_ a builder, so I don't feel too bad about not using `with` everywhere. That's mostly a problem with widgets. I would be happy with either `min_size` or  `with_min_size`.

cmyr:
This is now added in druid, but we still need druid-shell support for mac & windows.

xStrom:
This was implemented for Windows by #727. There's also a macOS specific issue for this at #714.

cmyr:
okay, lets close in favour of #714

PR #697, opened 2020-03-18
Rename Stepper builder methods

### LINKED ISSUE #684: replace Stepper::min/max with 'with_range'
I think these two methods are slightly too generally named, and I would like to be consistent with the new method on Slider.

Finnerale:
This fuses `Stepper::min` and `Stepper::max` to `Stepper::with_range` and renames `Stepper::step` as well as `Stepper::wrap` to `Stepper::with_step` and `Stepper::with_wrap` respectively.

Closes #684

### DIFF
--- a/druid/src/widget/stepper.rs
+++ b/druid/src/widget/stepper.rs
@@ -77,7 +77,7 @@ impl Stepper {
     /// Set whether the stepper should wrap around the minimum/maximum values.
     ///
     /// The default is no wrap
-    pub fn with_wrap(mut self, wrap: bool) -> Self {
+    pub fn with_wraparound(mut self, wrap: bool) -> Self {
         self.wrap = wrap;
         self
     }
--- a/druid/src/widget/stepper.rs
+++ b/druid/src/widget/stepper.rs
@@ -57,26 +57,27 @@ impl Stepper {
         }
     }
 
-    /// Set the stepper's maximum value.
-    pub fn max(mut self, max: f64) -> Self {
+    /// Set the range covered by this slider.
+    ///
+    /// The default range is `std::f64::MIN..std::f64::MAX`.
+    pub fn with_range(mut self, min: f64, max: f64) -> Self {
+        self.min = min;
         self.max = max;
         self
     }
 
-    /// Set the stepper's minimum value.
-    pub fn min(mut self, min: f64) -> Self {
-        self.min = min;
-        self
-    }
-
     /// Set the steppers amount by which the value increases or decreases.
-    pub fn step(mut self, step: f64) -> Self {
+    ///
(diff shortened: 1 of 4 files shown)


cmyr:
Looks good! `with_wrap` stands out as feeling a little weird, but 🤷‍♂

cmyr:
To bikeshed: how do we feel about `with_wrapping` or `with_wraparound`?

Finnerale:
`with_wraparound` seems good to me, should I change it to that or wait a little longer?

xStrom:
You can just change it to `with_wraparound` now.

Not related to your work per se, but I was looking at this `Stepper` code and the wraparound doesn't even work. With wraparound set the `Stepper` can only have its value as the `min` or `max`.

Looking at the PR (#308) it seems that initially the code was maybe more functional.
```rust
if (*data - old_data).abs() > EPSILON {
    // callback
    (self.value_changed)(ctx, data, env);
} else if self.wrap {
```
However in a commit titled [Adress stepper review feedback](https://github.com/xi-editor/druid/pull/308/commits/4f76e71eb2001a33bd82980252498c7f5b950b6f) that was removed and changed to the current non-working code. Also that change doesn't seem to be based on any actual reviewer feedback, so I guess some sort of mistake.

Feel free to fix the wraparound code here, or if you'd rather not that's fine too and we'll just merge the method renames only.

Finnerale:
For now I've just renamed `with_wrap` to `with_wraparound`

The functionality in itself is indeed completely broken, I think I will deal with that in a separate PR after I'm done with #632

druid/src/widget/stepper.rs
```diff
@@ -57,26 +57,27 @@ impl Stepper {
...
+    /// The default step is 1.0
+    pub fn with_step(mut self, step: f64) -> Self {
         self.step = step;
         self
     }
 
     /// Set whether the stepper should wrap around the minimum/maximum values.
-    pub fn wrap(mut self, wrap: bool) -> Self {
+    ///
+    /// The default is no wrap
```
cmyr:
```suggestion
    /// The default is `false`.
```

druid/src/widget/stepper.rs
```diff
@@ -57,26 +57,27 @@ impl Stepper {
...
+    pub fn with_range(mut self, min: f64, max: f64) -> Self {
         self.min = min;
+        self.max = max;
         self
     }
 
     /// Set the steppers amount by which the value increases or decreases.
-    pub fn step(mut self, step: f64) -> Self {
+    ///
+    /// The default step is 1.0
```
cmyr:
```suggestion
    /// The default step is `1.0`.
```

PR #698, opened 2020-03-18
Delete some redundant examples

cmyr:
All of these examples demonstrate widgets that are also present in
other examples.


cmyr:
@xStrom indeed. 🤒

PR #701, opened 2020-03-18
Allow setting minimum window size

Finnerale:
This implements setting the minimum window size (#632) **for Linux**.
The minimum size is also passed to the Mac and Windows WindowBuilder but they are not being utilized.

I've added this to the `calc` example and `TODO` tags to the implementation missing places as it was done in #587.

### DIFF
--- a/druid/examples/calc.rs
+++ b/druid/examples/calc.rs
@@ -200,9 +200,9 @@ fn build_calc() -> impl Widget<CalcState> {
 }
 
 fn main() {
-    let window = WindowDesc::new(build_calc)
-        .title(LocalizedString::new("calc-demo-window-title").with_placeholder("Simple Calculator"))
-        .with_min_size((250.0, 350.0));
+    let window = WindowDesc::new(build_calc).title(
+        LocalizedString::new("calc-demo-window-title").with_placeholder("Simple Calculator"),
+    );
     let calc_state = CalcState {
         value: "0".to_string(),
         operand: 0.0,
--- a/druid/examples/calc.rs
+++ b/druid/examples/calc.rs
@@ -200,9 +200,9 @@ fn build_calc() -> impl Widget<CalcState> {
 }
 
 fn main() {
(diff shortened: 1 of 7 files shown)


druid/examples/calc.rs
```diff
@@ -200,9 +200,9 @@ fn build_calc() -> impl Widget<CalcState> {
 }
 
 fn main() {
-    let window = WindowDesc::new(build_calc).title(
-        LocalizedString::new("calc-demo-window-title").with_placeholder("Simple Calculator"),
-    );
+    let window = WindowDesc::new(build_calc)
+        .title(LocalizedString::new("calc-demo-window-title").with_placeholder("Simple Calculator"))
+        .with_min_size((250.0, 350.0));
```
cmyr: can we move this to another example? I'd like to make calc be non-resizable (see #683)

Finnerale: Sure, maybe the flex example?

cmyr: I like that idea!

druid/examples/flex.rs
```diff
@@ -323,7 +323,8 @@ fn make_ui() -> impl Widget<AppState> {
 
 fn main() -> Result<(), PlatformError> {
     let main_window = WindowDesc::new(make_ui)
-        .window_size((600., 600.00))
+        .window_size((620., 265.00))
```
cmyr:
```suggestion
        .window_size((620., 600.00))
```
I think it's clearer when these are different, and we currently can't display the  column example  at our initial size.

druid/src/app.rs
```diff
@@ -179,6 +181,14 @@ impl<T: Data> WindowDesc<T> {
         self
     }
 
+    /// Set the minimal window size, similar to [`window_size`]
```
cmyr:
```suggestion
    /// Set the minimum window size.
    ///
    /// To  set the initial window size, see [`window_size`].
```

PR #702, opened 2020-03-19
Add UnitWrap for easier use of Widget<()>.

xStrom:
A discussion came up [in zulip](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/Lens.20to.20unit.20type) about what the best way to use a `Widget<()>` in a `Flex` layout is. Some [ad hoc solutions](https://gist.github.com/emigr2k1/a687b49e35c95079fb0df3395fa4f121) were tried.

Perhaps there's already an efficient and simple way to do this? If so, then this PR might not be useful but we would like to know how.

Otherwise I cooked up this PR here which adds a new `UnitWrap` type and a corresponding `WidgetExt::unit` method. This makes it easy to use these dataless `Widget<()>` widgets in a `Flex` layout.

### DIFF
--- a/druid/src/lens.rs
+++ b/druid/src/lens.rs
@@ -293,6 +293,75 @@ where
     }
 }
 
+/// A wrapper for widgets that don't use data, i.e. `Widget<()>`.
+///
+/// You can think of it like a [`LensWrap`] with a [`Lens`] that
+/// always focuses to `()` for every source type.
+///
+/// Wrapping such dataless widgets in `UnitWrap` will allow you
+/// to conveniently use them in [`Flex`] layouts for example.
+///
+/// Keep in mind that because the data will always be `()`,
+/// the [`update`] method of the wrapped widget will never be called.
+///
+/// [`LensWrap`]: struct.LensWrap.html
+/// [`Lens`]: trait.Lens.html
+/// [`Flex`]: ../widget/struct.Flex.html
+/// [`update`]: ../widget/trait.Widget.html#tymethod.update
+pub struct UnitWrap<W>
+where
+    W: Widget<()>,
+{
+    inner: W,
+}
+
+impl<W> UnitWrap<W>
+where
+    W: Widget<()>,
+{
+    /// Wrap a `Widget<()>` with a `UnitWrap` for use in cases
+    /// where `Widget<T: Data>` for any value of `T` can be used.
+    pub fn new(inner: W) -> UnitWrap<W> {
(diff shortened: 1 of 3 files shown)


druid/src/lens.rs
```diff
@@ -293,6 +293,75 @@ where
     }
 }
 
+/// A wrapper for widgets that don't use data, i.e. `Widget<()>`.
```
cmyr: I'm not sure that `lens` is quite the right place for this? I get that it's _kind_ of a lens, like a specialization of one particular lens, but still not 100% clear. 🤔

druid/src/widget/widget_ext.rs
```diff
@@ -178,12 +178,22 @@ pub trait WidgetExt<T: Data>: Widget<T> + Sized + 'static {
...
+    /// [`LensWrap`]: ../lens/struct.LensWrap.html
+    /// [`Lens`]: ../lens/trait.Lens.html
     fn lens<S: Data, L: Lens<S, T>>(self, lens: L) -> LensWrap<T, L, Self> {
         LensWrap::new(self, lens)
     }
 
+    /// Wrap a `Widget<()>` in a [`UnitWrap`] widget.
+    ///
+    /// [`UnitWrap`]: ../lens/struct.UnitWrap.html
+    fn unit(self) -> UnitWrap<Self>
```
cmyr: I'm starting to have some concerns about how general the names we're using in `WidgetExt` is. I'd be happy to hear some opinions.

I like `with_` for things that take an argument; I could imagine `lens` being `with_lens`. I'm not sure what makes sense here? `into_`?

One option: if we think of `UnitWrap` as an escape hatch, we could exclude it from `WidgetExt`.

xStrom:
After discussing this and alternative solutions some more in zulip, it seems best to not merge this right now.

The main benefit of `UnitWrap` right now would be performance in heavy use of `Widget<()>`. We have no actual cases of this yet, so this code might act as a burden instead. Once there are real life complex UIs that can benefit from optimizations we can return to this solution with benchmarks.

For people wanting to use dataless widgets (e.g. @emigr2k1), the current canonical way to do that is:
```rust
pub struct MyWidget {
    // Some fields
}

impl MyWidget {
    // Some methods etc
}

impl<T: Data> Widget<T> for MyWidget {
    fn update(&mut self, _ctx: &mut UpdateCtx, _old_data: &T, _data: &T, _env: &Env) {}
    // .. and other required Widget methods
}
```

PR #709, opened 2020-03-19
Redirect vertical scrolling as horizontal scrolling when holding shift

ForLoveOfCats:
Because macOS is apparently a magical place where this is done at the OS level I implemented this feature in `druid-shell` for both GTK and Windows. The Windows patch is very straightforward but GTK has two codepaths for scroll events. The "smooth" scroll path receives data for both axis at once so I wrote it such that when shift is held the two axis are capable of fighting each other as that made the most sense as well as is consistent with the other codepath and Windows patch.

My understanding is that the shift key does not affect trackpad scrolling on macOS but that was not feasible to emulate here. On GTK the "smooth" events are not only used for scrolling but can in some cases be used for mousewheel input (I observed this firsthand on my laptop) and with Windows there only seems to be a single source of scroll data regardless of if it originated from a mousewheel or trackpad.

### DIFF
--- a/druid-shell/src/platform/gtk/window.rs
+++ b/druid-shell/src/platform/gtk/window.rs
@@ -346,10 +346,18 @@ impl WindowBuilder {
                 let mut handler = state.handler.borrow_mut();
                 match scroll.get_direction() {
                     ScrollDirection::Up => {
-                        handler.wheel(Vec2::from((0.0, -120.0)), modifiers);
+                        if modifiers.shift {
+                            handler.wheel(Vec2::from((-120.0, 0.0)), modifiers);
+                        } else {
+                            handler.wheel(Vec2::from((0.0, -120.0)), modifiers);
+                        }
                     }
                     ScrollDirection::Down => {
-                        handler.wheel(Vec2::from((0.0, 120.0)), modifiers);
+                        if modifiers.shift {
+                            handler.wheel(Vec2::from((120.0, 0.0)), modifiers);
+                        } else {
+                            handler.wheel(Vec2::from((0.0, 120.0)), modifiers);
+                        }
                     }
                     ScrollDirection::Left => {
                         handler.wheel(Vec2::from((-120.0, 0.0)), modifiers);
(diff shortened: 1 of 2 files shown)


cmyr:
I will need some input on this; I'm really not sure what the conventions are on these platforms. In principal this sounds good to me.

ForLoveOfCats:
I'm currently wrangling some git stuff on this, turns out I made a silly mistake so hold off on reviewing just yet

xStrom:
This is not a platform behavior on Windows. Holding shift while scrolling has no effect.

That said some applications do implement this on Windows, including some big ones like Adobe Reader, Chrome, Firefox. Amusingly in Internet Explorer shift+scroll cycles through the back button history.

I guess I'm fine with this being a druid feature on Windows. It doesn't conflict with any platform behavior really and can be useful.

ForLoveOfCats:
That is really surprising to me, even when I used to use Windows this feature was so widespread that I never thought about it being a platform convention or not. I do remember FF having that history scrolling feature for the longest time which bugged me endlessly back in the day :laughing:

I may be biased but I think that this is a reasonable thing to expect to work regardless of the platform

cmyr:
Does this also apply to trackpad gestures? Is that even a relevant question? Basically: this makes sense to me for the scroll wheel, I think it's a reasonable default there, and if an application wants to get fancy they still have the opportunity. I think it's weirder for trackpad gestures.

ForLoveOfCats:
Yes these changes should affect scrolling with a trackpad when holding shift in the GTK shell and Windows. Neither platforms seem to offer a way to differentiate between trackpad and mousewheel scroll events. It would be *great* if a Windows user with a trackpad can verify that `WM_MOUSEWHEEL` does indeed get fired when scrolling with a trackpad which would mean that this change 100% does affect the trackpad as well on Windows.

cmyr:
maybe ask on zulip?

ForLoveOfCats:
That's a good idea, I forget that not everyone watches the repo and has a notification queue a mile long :sweat_smile:

ForLoveOfCats:
This is simple both conceptually and in diff size and I would like to have it hang out in master for a while before 0.5 drops which is rapidly approaching. Any changes before a final review?

xStrom:
I played around with this and can confirm that:
* This works on Windows.
* The `WM_MOUSEWHEEL` event gets fired with a trackpad.

ForLoveOfCats:
Awesome, thanks! Now I suppose someone on Linux also needs to confirm

cmyr:
@xStrom if you're happy to +1 this I won't block on testing with gtk.

totsteps:
I tested the `scroll_colors` example and can confirm that this works on Linux as well.

Issue #633, opened 2020-03-11
implement window resize option


cmyr:
This was implemented for gtk in #587, but still needs to be implemented for mac and  windows.

These can be done separately, that is if someone wants to do mac but not windows that's fine!

emigr2k1:
I think this should be closed

xStrom:
It's [not implemented](https://github.com/xi-editor/druid/blob/267e06870894bff78d9cd60ab5966f070b2007d3/druid-shell/src/platform/mac/window.rs#L692-L693) on macOS yet.

emigr2k1:
My bad. I thought that this issue was about the resizable option in `WindowDesc` that already works.

totsteps:
I think this can be closed as it was implemented in #773 at [code](https://github.com/xi-editor/druid/blob/3b3d7a00b44824fae98cf48390e948e5d73a6329/druid-shell/src/platform/mac/window.rs#L698).

cmyr:
yep thanks!

Issue #437, opened 2020-01-01
Compilation of druid-shell fails on windows


Zwo1in:
Hi,
I wanted to give druid a try, but I'm unable to compile druid for windows.
This is the error I get:
```
Compiling druid-shell v0.4.0
error[E0308]: mismatched types
 --> C:\Users\Zwolin\.cargo\registry\src\github.com-1ecc6299db9ec823\druid-shell-0.4.0\src\platform\windows\window.rs:1007:48
     |
1007 |         SetWindowLongPtrW(hwnd, GWLP_USERDATA, wndproc_ptr as LONG_PTR);
     |                                                ^^^^^^^^^^^^^^^^^^^^^^^ expected i32, found isize
```
And the toolchain I use:
```
stable-i686-pc-windows-msvc
rustc 1.40.0 (73528e339 2019-12-16)
```

raphlinus:
Any special reason you're using a 32 bit target? It might not be too hard to fix this, but it's not been a priority.

PR #712, opened 2020-03-19
Implement WindowBuilder::resizable and WindowHandle::resizable for windows

teddemunnik:
Implementation of #633 for windows. 

When `WindowBuilder::resizable` or `WindowHandle::resizable` is set to `false` the window can no longer be resized by dragging on the borders, and the window can no longer be maximized.
not resizable

### DIFF
--- a/druid-shell/src/platform/windows/window.rs
+++ b/druid-shell/src/platform/windows/window.rs
@@ -33,6 +33,7 @@ use winapi::shared::minwindef::*;
 use winapi::shared::windef::*;
 use winapi::shared::winerror::*;
 use winapi::um::d2d1::*;
+use winapi::um::errhandlingapi::GetLastError;
 use winapi::um::unknwnbase::*;
 use winapi::um::winnt::*;
 use winapi::um::winuser::*;
@@ -68,7 +69,6 @@ extern "system" {
 /// Builder abstraction for creating new windows.
 pub struct WindowBuilder {
     handler: Option<Box<dyn WinHandler>>,
-    dwStyle: DWORD,
     title: String,
     menu: Option<Menu>,
     present_strategy: PresentStrategy,
@@ -699,7 +699,6 @@ impl WindowBuilder {
     pub fn new() -> WindowBuilder {
         WindowBuilder {
             handler: None,
-            dwStyle: WS_OVERLAPPEDWINDOW,
             title: String::new(),
             menu: None,
             resizable: true,
@@ -719,7 +718,6 @@ impl WindowBuilder {
     }
 
     pub fn resizable(&mut self, resizable: bool) {
-        // TODO: Use this in `self.build`
         self.resizable = resizable;
     }
 
@@ -796,6 +794,12 @@ impl WindowBuilder {
                 }
                 None => (0 as HMENU, None),
             };
+
+            let mut dwStyle = WS_OVERLAPPEDWINDOW;
+            if !self.resizable {
+                dwStyle &= !(WS_THICKFRAME | WS_MAXIMIZEBOX);
+            }
+
             let mut dwExStyle = 0;
             if self.present_strategy == PresentStrategy::Flip {
                 dwExStyle |= WS_EX_NOREDIRECTIONBITMAP;
@@ -804,7 +808,7 @@ impl WindowBuilder {
                 dwExStyle,
                 class_name.as_ptr(),
                 self.title.to_wide().as_ptr(),
-                self.dwStyle,
+                dwStyle,
                 CW_USEDEFAULT,
                 CW_USEDEFAULT,
                 width,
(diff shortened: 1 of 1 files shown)


druid-shell/src/platform/windows/window.rs
```diff
@@ -1075,8 +1079,20 @@ impl WindowHandle {
     // TODO: Implement this
     pub fn show_titlebar(&self, _show_titlebar: bool) {}
 
-    // TODO: Implement this
-    pub fn resizable(&self, _resizable: bool) {}
+    pub fn resizable(&self, resizable: bool) {
+        if let Some(w) = self.state.upgrade() {
+            let hwnd = w.hwnd.get();
+            unsafe {
+                let mut style = GetWindowLongPtrW(hwnd, GWL_STYLE);
```
xStrom: Need to check for errors here. `GetWindowLongPtrW` returns zero in case of error.

druid-shell/src/platform/windows/window.rs
```diff
@@ -1075,8 +1079,20 @@ impl WindowHandle {
...
+                if resizable {
+                    style |= (WS_THICKFRAME | WS_MAXIMIZEBOX) as LONG_PTR;
+                } else {
+                    style &= !(WS_THICKFRAME | WS_MAXIMIZEBOX) as LONG_PTR;
```
xStrom: Need to use `WindowLongPtr` instead of `LONG_PTR` for 32-bit support.

druid-shell/src/platform/windows/window.rs
```diff
@@ -1075,8 +1079,20 @@ impl WindowHandle {
...
+                }
+                SetWindowLongPtrW(hwnd, GWL_STYLE, style);
```
xStrom: Need to check for errors here. `SetWindowLongPtrW` returns zero in case of error. As for what to do about it, should log the error I guess.

druid-shell/src/platform/windows/window.rs
teddemunnik: Is `WindowLongPtr` a type? I can't seem to find it in the winapi docs. From what I understand `GetWindowLongPtrW` and `SetWindowLongPtrW` returns and uses `LONG_PTR` which is 32 bit on 32 bit windows, and 64 bit on 64 bit windows.

teddemunnik:
Thanks for the review @xStrom! I completely neglected error handling, giving it another pass soon :)

druid-shell/src/platform/windows/window.rs
xStrom: It's a bit confusing for sure.

1. `WindowLongPtr` is a druid-shell type defined in `window.rs` aka this very same file. You can search for it to see its definition.
2. `GetWindowLongPtrW` only works on 64-bit but the Rust `winapi` crate we're using correctly aliases that function to `GetWindowLongW` in 32-bit mode.

teddemunnik: Ah, thanks! I didn't think to look in druid for `WindowLongPtr` type.. ;) I'm a bit confused though as to why it is neccesary: 

- According to MSDN `LONG_PTR` is defined to be the size of a pointer, that is 4 bytes on 32 bit windows, and 8 bytes on 64 bit windows. [LONG_PTR docs](https://docs.microsoft.com/en-us/windows/win32/winprog/windows-data-types)
- In the winapi crate `LONG_PTR` is defined to be `isize`, which also is 4 bytes on 32 bit windows, and 8 bytes on 64 bit windows: [isize docs](https://doc.rust-lang.org/std/primitive.isize.html), [winapi LONG_PTR docs](https://docs.rs/winapi/0.3.8/winapi/shared/basetsd/type.LONG_PTR.html) 

Am I missing something?

teddemunnik: Ah I found issue #437, I had expected this to be the same between using the windows API in C++ and rust.. Will change it :)

xStrom: Well the docs for [GetWindowLongPtrW](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getwindowlongptrw) state that for 32-bit support you should use `GetWindowLongPtr` instead which is aliased to `GetWindowLong` on 32-bit.

The rust `winapi` crate aliases `GetWindowLongPtrW` to `GetWindowLongW` on 32-bit. The `GetWindowLongW` expects `i32` while `GetWindowLongPtrW` expects `isize`.

teddemunnik: Alright, makes sense. [Seems like it's something that might be getting changed in winapi 0.4](https://github.com/retep998/winapi-rs/pull/735). Thanks for explaining!

teddemunnik:
Updated with error handling and `WindowLongPtr`. I re-used `Error::Hr`, because I am hoping that in a future change we can then use [FormatMessage](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-formatmessage) in the `Display` implementation to get a nice error description as well as just the error code. Could you give it another look @xStrom ?

xStrom:
This seems good, but we're having issues with the testing system here (#724 fixed it for the future, but doesn't apply here). Could you re-push so that it triggers the testing on a new commit? For example combine your two commits into one and force push to your branch.

PR #715, opened 2020-03-20
fix image path in image example

totsteps:


cmyr:
hmm I'm confused this seems to be working for me?

totsteps:
Oh, wait. You right.
I guess I cd'ed into `examples` dir which broke the path!!

Will close this now.

emigr2k1:
Could this use CARGO_MANIFEST_DIR to avoid that problem?

totsteps:
To be honest I have no idea how `CARGO_MANIFEST_DIR` works!

PR #718, opened 2020-03-20
Print HRESULT description when formatting Error::Hr

teddemunnik:
When implementing error handling for #712 I thought it'd be nice to add some details to the error `HRESULT` error messages instead of just error codes. 

This change adds a short description from the [FormatMessage](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-formatmessage) windows API when printing an `Error::Hr`, when one is available. 

As an example the following messages:
> failed to set window style. HRESULT 0x80070585
> failed to set window style. HRESULT 0x80070578

are now displayed as: 
> failed to set window style. HRESULT 0x80070585: Invalid index. 
> failed to set window style. HRESULT 0x80070578: Invalid window handle.

### DIFF
--- a/druid-shell/src/platform/windows/error.rs
+++ b/druid-shell/src/platform/windows/error.rs
@@ -40,6 +40,7 @@ pub enum Error {
     NullHwnd,
 }
 
+#[allow(clippy::crosspointer_transmute)]
 fn hresult_description(hr: HRESULT) -> Option<String> {
     unsafe {
         let mut message_buffer: LPWSTR = std::ptr::null_mut();
--- a/druid-shell/src/platform/windows/error.rs
+++ b/druid-shell/src/platform/windows/error.rs
@@ -16,7 +16,16 @@
 
 use std::fmt;
 
+use super::util::FromWide;
+use std::mem::transmute;
+use std::ptr::{null, null_mut};
+use winapi::shared::minwindef::HLOCAL;
+use winapi::shared::ntdef::LPWSTR;
 use winapi::shared::winerror::HRESULT;
+use winapi::um::winbase::{
+    FormatMessageW, LocalFree, FORMAT_MESSAGE_ALLOCATE_BUFFER, FORMAT_MESSAGE_FROM_SYSTEM,
+    FORMAT_MESSAGE_IGNORE_INSERTS, FORMAT_MESSAGE_MAX_WIDTH_MASK,
+};
 
 /// Error codes. At the moment, this is little more than HRESULT, but that
 /// might change.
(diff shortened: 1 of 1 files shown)


druid-shell/src/platform/windows/error.rs
```diff
@@ -31,10 +40,42 @@ pub enum Error {
...
+fn hresult_description(hr: HRESULT) -> Option<String> {
+    unsafe {
+        let mut message_buffer: LPWSTR = std::ptr::null_mut();
+        let format_result = FormatMessageW(
+            FORMAT_MESSAGE_FROM_SYSTEM
+                | FORMAT_MESSAGE_ALLOCATE_BUFFER
+                | FORMAT_MESSAGE_IGNORE_INSERTS
+                | FORMAT_MESSAGE_MAX_WIDTH_MASK,
+            null(),
+            hr as u32,
```
xStrom: Let's change this `u32` to `DWORD` (from `winapi::shared::minwindef`). I know functionally they're the same and the `u32` looks nicer but matching the types used in Microsoft docs when calling a winapi function will help whomever is reading this code in the future.

teddemunnik: Comepletely agree, changed it

druid-shell/src/platform/windows/error.rs
```diff
@@ -31,10 +40,42 @@ pub enum Error {
...
+            0,
+            transmute(&mut message_buffer as *const LPWSTR),
```
xStrom: I think we can remove the `as *const LPWSTR` part. Otherwise we might as well just write `transmute(&message_buffer)`.

Probably best to have this instead:
```rust
transmute(&mut message_buffer)
```

The `mut` part doesn't seem required for the code to work, but it communicates the fact that the function will modify the pointer so we can keep the `mut`.

xStrom: Sorry, I was a bit slow with this comment. I wasn't waiting until you push the `DWORD` change, I swear! :grimacing:

teddemunnik: It seems to work without the `as *const LPWSTR`, but then clippy complains:

> warning: transmute from a reference to a pointer
  --> druid-shell\src\platform\windows\error.rs:55:13
   |
55 |             transmute(&mut message_buffer),
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try: `&mut message_buffer as *mut *mut u16 as *mut u16`
   |
   = note: `#[warn(clippy::useless_transmute)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_transmute

Is that another false positive? I already turned off [clippy::crosspointer_transmute](https://rust-lang.github.io/rust-clippy/master/index.html#crosspointer_transmute) for this function in an earlier change.

xStrom: Clippy seems unhappy indeed. However ` as *mut LPWSTR` seems to satisfy clippy and is more to the truth, so maybe use that.

teddemunnik: Ah just listening to clippy apparantly this just works: `&mut message_buffer as *const LPWSTR as LPWSTR`, didn't know you could do that without transmute. Then I can remove the clippy ignore flags. Does that sound good?

xStrom: Great find! I was trying to get rid of `transmute` earlier but didn't know how. It looks like the following works as well:
```rust
&mut message_buffer as *mut LPWSTR as LPWSTR
```
Let's go with that if possible.

teddemunnik: Sounds good! I just pushed this change :)

raphlinus:
I was going to complain about the obviously wrong casting, but that's winapi being winapi, and I agree this is the best way to write that in Rust.

PR #720, opened 2020-03-20
use dejavu sans on linux

kindlychung:


cmyr:
I think we need to be careful to ensure that the default font will always be found; that's the reason we currently use a generic sans-serif. When we have better font discovery and fallback tools we can be more intelligent about choosing the best font available.

PR #725, opened 2020-03-20
Add Label::dynamic constructor

cmyr:
This is just a convenience; because of how type resolution works
with the `Into<LabelText>` paramater, you are forced to always
provide explicit types when creating a label with dynamic text.

This is a papercut, and is particularly annoying when trying to
write sample code.

### DIFF
--- a/druid/src/widget/label.rs
+++ b/druid/src/widget/label.rs
@@ -81,6 +81,31 @@ impl<T: Data> Label<T> {
         }
     }
 
+    /// Construct a new dynamic label.
+    ///
+    /// The contents of this label are generated from the data using a closure.
+    ///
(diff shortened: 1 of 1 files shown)


druid/src/widget/label.rs
```diff
@@ -81,6 +81,31 @@ impl<T: Data> Label<T> {
...
+    /// The following are equivalent.
+    ///
+    /// ```
+    /// use druid::Env;
+    /// use druid::widget::Label;
+    /// let button1: Label<u32> = Label::new(|data: &u32, _: &Env| format!("total is {}", data));
+    /// let button2: Label<u32> = Label::dynamic(|data, _| format!("total is {}", data));
+    /// ```
+    ///
+    /// [`new`] #method.new
```
xStrom: Missing colon.

Issue #714, opened 2020-03-19
implement  minimum window sizes on mac


cmyr:
when #701 is merged we will have  the API for this, but it will only be implemented in GTK.

mobile-bungalow:
Has there been any progress on this? This seems straight forward and I'd like to give it a try.

xStrom:
@mobile-bungalow you are welcome to send us a PR. I don't think anyone else is working on this right now.

PR #727, opened 2020-03-21
Implement WindowBuilder::set_min_size for windows.

teddemunnik:
Implements #714 for windows.


druid-shell/src/platform/windows/window.rs
```diff
@@ -161,6 +161,7 @@ struct WndState {
     render_target: Option<DeviceContext>,
     dcomp_state: Option<DCompState>,
     dpi: f32,
+    min_size: Option<kurbo::Size>,
```
xStrom: `Size` is already imported from `kurbo` so this can be changed to just `Option<Size>`.

teddemunnik: Good point, fixed

PR #729, opened 2020-03-21
Add with_ prefix to builder-like methods in Label

emigr2k1:


druid/src/widget/label.rs
```diff
@@ -106,22 +106,9 @@ impl<T: Data> Label<T> {
...
-    ///
-    /// [`Label`]: #method.new
-    /// [`LabelText`]: enum.LabelText.html
-    pub fn set_text(&mut self, text: impl Into<String>) {
-        self.text = LabelText::Specific(text.into());
-    }
-
     /// Set text alignment.
     #[deprecated(since = "0.5.0", note = "Use an Align widget instead")]
-    pub fn text_align(self, _align: UnitPoint) -> Self {
```
cmyr: being as this is deprecated, I wouldn't bother renaming it; the point of the deprecation is to avoid breaking existing code, while pointing to the new API.

(I'm not worried about breaking these other methods; they are relatively new and I don't think they are widely used. You could do this deprecation trick for fun if you wanted, but i'm not concerned)

emigr2k1:
Mmm I messed up my branch. I fixed it but I can't reopen the PR. I'll create another one. Sorry

xStrom:
Just for future info @emigr2k1 you can always reset your branch to a previous commit if you make any commits you don't actually want. Then later you can "force" push it to your github branch.

emigr2k1:
Thanks! I fixed it but I could not reopen the pull request. The button was greyed out. I shouldn't have closed the PR I guess. I closed it just so nobody bothered to check it while I was fixing it

xStrom:
No need to worry about people checking it while you're fixing it. Most likely no one would have even noticed. Also if you're planning on taking longer to fix it (like a full day or week) you can just post a comment that you'll fix it later. The PR can remain open. :+1:

PR #731, opened 2020-03-21
Allow druid_derive macros in druid crate

### LINKED ISSUE #650: cannot derive Data within druid module
This is one of those macro gotchas. There's some trick to make this work consistently, which I'm sure david tolnay has an example of somewhere. 😁

emigr2k1:
This makes use of feature `extern_crate_self` to refer to the current
crate with another name, in this case, druid. This allows druid_derive
macros to expand to something like `druid::Data` instead of `crate::Data`.

Closes #650

### DIFF
--- a/druid/src/widget/flex.rs
+++ b/druid/src/widget/flex.rs
@@ -204,7 +204,7 @@ pub(crate) enum Axis {
 ///
 /// If a widget is smaller than the container on the minor axis, this determines
 /// where it is positioned.
-#[derive(Debug, Clone, Copy, PartialEq)]
+#[derive(Debug, Clone, Copy, PartialEq, Data)]
 pub enum CrossAxisAlignment {
     /// Top or leading.
     ///
@@ -224,7 +224,7 @@ pub enum CrossAxisAlignment {
 ///
 /// If there is surplus space on the main axis after laying out children, this
 /// enum represents how children are laid out in this space.
-#[derive(Debug, Clone, Copy, PartialEq)]
+#[derive(Debug, Clone, Copy, PartialEq, Data)]
 pub enum MainAxisAlignment {
     /// Top or leading.
     ///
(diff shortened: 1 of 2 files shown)


emigr2k1:
If putting `druid` in the namespace is not a good solution, I could implement something like the solution provided in this issue[1].

[1] https://github.com/rust-lang/rust/issues/54647#issuecomment-429166393

cmyr:
I haven't seen this solution before, can you link to any information on it? Are there any major drawbacks?

emigr2k1:
This feature was primarily implemented to solve problems like this. Here[1] is the tracking issue for the feature and here is some discussion[2]. The only drawback is that you have `druid` in the global namespace (not exported).
Some other crates do this too: yew[3], RustPython[4]
[1] https://github.com/rust-lang/rust/issues/56409
[2] https://github.com/rust-lang/rust/issues/54647
[3] https://github.com/yewstack/yew/blob/c33fe73f6caad8701a40cd488a00ea9e373930d5/src/lib.rs#L82
[4] https://github.com/RustPython/RustPython/blob/098532b283a98e8b98b70fafbed4ec4187311e97/vm/src/lib.rs#L35

cmyr:
Okay, then I have one additional request for this PR; in `druid/src/widget/flex.rs` we currently manually implement `Data` for two types; can you change those to use `derive`?

druid/src/widget/flex.rs
```diff
@@ -700,20 +700,6 @@ impl Spacing {
     }
 }
 
-// we have these impls mostly for our 'flex' example, but I could imagine
-// them being broadly useful?
```
emigr2k1: I didn't know what to do with this comment

cmyr: you can just delete that.

PR #738, opened 2020-03-22
Improve Split accuracy.

### LINKED ISSUE #641: The handle of the Split widget should have a minimum hit area
You can only interact with the handle of the `Split` widget if you place your cursor exactly within its bounds. On some operating systems -- for example macOS -- you can place your cursor a tiny bit outside the bounds of the splitter handle and still be able to interact with it.

It would be useful to have this feature in Druid's `Split` widget as well. This would allow you to make your splitter barely visible -- say only 4px wide -- but still let users interact with it if they place their cursor 2px to the left or right of the visible area.

A good way to do this would be to specify a minimum hit area for the splitter handle. If you ask for a handle that has a width less than the minimum hit area, the widget will still capture events that occur within the minimum area.

xStrom:
This PR includes a slew of changes to the `Split` widget.

* All layout and drawing is now quantized to integers as a continuation of work in #669. This means that single pixel lines never bleed out to side pixels. This applies to both the inner children and the split bar.
* Adds `min_splitter_area(f64)` method which basically acts as padding around the split bar. This padding still reacts to mouse events. As a result it is possible to use extremely narrow split bars while still having a reasonable area for the user to click on. Fixes #641.
* No longer forgets the `split_point` chosen by the user. The widget still modifies the split point when the size constraints require it (e.g. when resizing the window to extremely small) however now the user chosen split point is restored as soon as the constraints allow it.
* The meaning of `splitter_size` has slightly changed, to a less surprising interpretation.
    * Previously a splitter size of 10 with a solid fill meant:
     4px wide solid fill with 3px padding on each side.
    With the fill width formula being: size - 2*((size/3).floor())
    Now a 10 with a solid fill means a 10px wide solid fill.
    The new `min_splitter_area` can be used to add padding.
    * Previously a splitter size of 10 with a line style meant:
    2px pad | ~2px line | 2px pad | ~2px line | 2px pad
    With 1px lines being centered (size/3).floor() pixels away from edges.
    Now a 10 with a line style means: 3px line | 4px pad | 3px line
    Padding at edges can be added optionally with `min_splitter_area`.
* Renamed a few internal functions and added a few lines of documentation for better readability.

Generally speaking the `Split` widget still has issues, which were there before and aren't addressed here, e.g. children still being drawn when they have zero sized layout. I'll probably tackle this at least partially in a future PR.

### DIFF
--- a/druid/src/widget/split.rs
+++ b/druid/src/widget/split.rs
@@ -37,6 +37,9 @@ pub struct Split<T> {
 
 impl<T> Split<T> {
     /// Create a new split panel, with the specified axis being split in two.
+    ///
+    /// Horizontal split axis means that the children are left and right.
+    /// Vertical split axis means that the children are up and down.
     fn new(
         split_axis: Axis,
         child1: impl Widget<T> + 'static,
@@ -58,13 +61,13 @@ impl<T> Split<T> {
 
     /// Create a new split panel, with the horizontal axis split in two by a vertical bar.
     /// The children are laid out left and right.
-    pub fn horizontal(child1: impl Widget<T> + 'static, child2: impl Widget<T> + 'static) -> Self {
+    pub fn columns(child1: impl Widget<T> + 'static, child2: impl Widget<T> + 'static) -> Self {
         Self::new(Axis::Horizontal, child1, child2)
     }
 
     /// Create a new split panel, with the vertical axis split in two by a horizontal bar.
     /// The children are laid out up and down.
-    pub fn vertical(child1: impl Widget<T> + 'static, child2: impl Widget<T> + 'static) -> Self {
+    pub fn rows(child1: impl Widget<T> + 'static, child2: impl Widget<T> + 'static) -> Self {
         Self::new(Axis::Vertical, child1, child2)
     }
 
--- a/druid/src/widget/split.rs
+++ b/druid/src/widget/split.rs
@@ -21,166 +21,166 @@ use crate::{
     LifeCycleCtx, PaintCtx, RenderContext, UpdateCtx, Widget, WidgetPod,
 };
 
-///A container containing two other widgets, splitting the area either horizontally or vertically.
+/// A container containing two other widgets, splitting the area either horizontally or vertically.
 pub struct Split<T> {
-    split_direction: Axis,
-    solid: bool,
-    draggable: bool,
-    min_size: f64,
+    split_axis: Axis,
     split_point_chosen: f64,
     split_point_effective: f64,
-    splitter_size: f64,
-    min_splitter_area: f64,
+    min_size: f64,     // Integers only
+    bar_size: f64,     // Integers only
+    min_bar_area: f64, // Integers only
+    solid: bool,
+    draggable: bool,
     child1: WidgetPod<T, Box<dyn Widget<T>>>,
     child2: WidgetPod<T, Box<dyn Widget<T>>>,
 }
 
 impl<T> Split<T> {
-    ///Create a new split panel.
+    /// Create a new split panel, with the specified axis being split in two.
     fn new(
-        split_direction: Axis,
+        split_axis: Axis,
         child1: impl Widget<T> + 'static,
         child2: impl Widget<T> + 'static,
     ) -> Self {
         Split {
-            split_direction,
-            min_size: 0.0,
-            solid: false,
+            split_axis,
             split_point_chosen: 0.5,
             split_point_effective: 0.5,
-            splitter_size: 6.0,
-            min_splitter_area: 6.0,
+            min_size: 0.0,
+            bar_size: 6.0,
+            min_bar_area: 6.0,
+            solid: false,
             draggable: false,
             child1: WidgetPod::new(child1).boxed(),
             child2: WidgetPod::new(child2).boxed(),
         }
     }
 
-    /// Create a new split panel, with a vertical splitter between two children.
-    pub fn vertical(child1: impl Widget<T> + 'static, child2: impl Widget<T> + 'static) -> Self {
-        Self::new(Axis::Vertical, child1, child2)
-    }
-
-    /// Create a new split panel, with a horizontal splitter between two children.
+    /// Create a new split panel, with the horizontal axis split in two by a vertical bar.
+    /// The children are laid out left and right.
     pub fn horizontal(child1: impl Widget<T> + 'static, child2: impl Widget<T> + 'static) -> Self {
         Self::new(Axis::Horizontal, child1, child2)
     }
 
-    /// Set container's split point as a fraction of the split dimension
-    /// The value must be between 0.0 and 1.0, exclusive
+    /// Create a new split panel, with the vertical axis split in two by a horizontal bar.
+    /// The children are laid out up and down.
+    pub fn vertical(child1: impl Widget<T> + 'static, child2: impl Widget<T> + 'static) -> Self {
+        Self::new(Axis::Vertical, child1, child2)
+    }
+
+    /// Builder-style method to set the split point as a fraction of the split axis.
+    ///
+    /// The value must be between `0.0` and `1.0`, inclusive.
+    /// The default split point is `0.5`.
     pub fn split_point(mut self, split_point: f64) -> Self {
         assert!(
-            split_point > 0.0 && split_point < 1.0,
-            "split_point must be between 0.0 and 1.0!"
+            split_point >= 0.0 && split_point <= 1.0,
+            "split_point must be in the range [0.0-1.0]!"
         );
         self.split_point_chosen = split_point;
         self
     }
 
-    /// Builder-style method to set the minimum size for both sides of the split.
+    /// Builder-style method to set the minimum size for both sides of the split axis.
     ///
     /// The value must be greater than or equal to `0.0`.
+    /// The value will be rounded up to the nearest integer.
     pub fn min_size(mut self, min_size: f64) -> Self {
         assert!(min_size >= 0.0);
-        self.min_size = min_size;
+        self.min_size = min_size.ceil();
         self
     }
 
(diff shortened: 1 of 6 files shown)


cmyr:
@Zarenor are you around? Any interest in taking a look at this?

druid/src/widget/split.rs
```diff
@@ -88,38 +98,89 @@ impl<T> Split<T> {
...
+    /// Set the minimum width of the splitter area.
+    ///
+    /// The minimum splitter area defines the minimum width of the area
+    /// where mouse hit detection is done for the splitter.
+    /// The final area is either this or the splitter size, whichever is greater.
+    ///
+    /// This can be useful when you want to use a very narrow visual splitter,
+    /// but don't want to sacrifice user experience.
+    ///
+    /// The default minimum splitter area is `6.0`.
```
cmyr, line 111: I would include a note that this panics if the size is less than 0.0.

druid/src/widget/split.rs
```diff
@@ -44,42 +46,50 @@ impl<T> Split<T> {
...
     pub fn min_size(mut self, min_size: f64) -> Self {
         assert!(min_size >= 0.0);
         self.min_size = min_size;
-
         self
     }
-    /// Set the width of the splitter bar, in pixels
-    /// The value must be positive or zero
+
+    /// Set the width of the splitter bar.
```
cmyr, line 89:
```suggestion
    /// Builder-style method to set the width of the splitter bar.
```

druid/src/widget/split.rs
```diff
@@ -88,38 +98,89 @@ impl<T> Split<T> {
         self.splitter_size = splitter_size;
         self
     }
+
+    /// Set the minimum width of the splitter area.
```
cmyr, line 102:
```suggestion
    /// Builder-style method to set the minimum width of the splitter area.
```

druid/src/widget/split.rs
```diff
@@ -88,38 +98,89 @@ impl<T> Split<T> {
...
     /// Builder-style method to set whether the splitter handle is drawn as a solid rectangle.
     ///
     /// If this is `false` (the default), it will be drawn as two parallel lines.
     pub fn fill_splitter_handle(mut self, solid: bool) -> Self {
         self.solid = solid;
         self
     }
-    fn splitter_hit_test(&self, size: Size, mouse_pos: Point) -> bool {
+
+    /// Returns the width of the area of mouse hit detection.
```
cmyr, line 135: 'width' is kind of confusing for a vertical splitter, but whatever 🤷‍♂

druid/src/widget/split.rs
```diff
@@ -27,8 +27,10 @@ pub struct Split<T> {
     solid: bool,
     draggable: bool,
     min_size: f64,
-    split_point: f64,
+    split_point_chosen: f64,
+    split_point_effective: f64,
     splitter_size: f64,
+    min_splitter_area: f64,
```
cmyr, line 33: Maybe this should be an `Option<f64>`, so that we can default to using the splitter size when this isn't passed explicitly? Maybe?

I'm not sure. I'm just thinking of in the example where you set the splitter size to 3, should we keep the same default min_splitter_area? Maybe...

druid/src/widget/split.rs
Zarenor: 'breadth', so we're not implying directionality (vs 'width' or 'height')?

Zarenor: I'd used 'size' before, because I hadn't been able to come up with 'breadth'

druid/src/widget/split.rs
Zarenor: Should we start adding methods that allow mutation of this kind of thing at runtime (so that this distinction is meaningful)?

I think the reason we haven't thus far is the type-erased nature of children contained by built-in container widgets? But it feels like we're moving in a direction where it's acceptable to work around that or downcast out of type-erasure. So the distinction could *become* meaningful.

druid/src/widget/split.rs
```diff
@@ -88,38 +98,89 @@ impl<T> Split<T> {
...
     }
-    fn splitter_hit_test(&self, size: Size, mouse_pos: Point) -> bool {
+
+    /// Returns the width of the area of mouse hit detection.
+    #[inline]
+    fn splitter_area(&self) -> f64 {
+        self.splitter_size.max(self.min_splitter_area)
+    }
+
+    /// Returns the padding width added to the splitter on both sides.
```
Zarenor, line 141:
```suggestion
    /// Returns the padding width added to the splitter on each side.
```
I think the 'each side' phrasing makes more clear that we're dividing the difference by two - we *should* add this to each side, not redivide.

druid/src/widget/split.rs
```diff
@@ -88,38 +98,89 @@ impl<T> Split<T> {
...
-        // we need the minor axis to get the size of the 'splitted' direction
-        let size_in_splitted_direction = self.split_direction.minor(size);
+        // we need the minor axis to get the size of the split direction
+        let size_in_split_direction = self.split_direction.minor(size);
 
-        let min_offset = (self.splitter_size * 0.5).min(5.0);
-        let mut min_limit = self.min_size.max(min_offset);
-        let mut max_limit = (size_in_splitted_direction - self.min_size.max(min_offset)).max(0.0);
+        let mut min_limit = self.min_size;
+        let mut max_limit = (size_in_split_direction - min_limit).max(0.0);
```
Zarenor, line 183: I see we've removed the absolute '5.0' minimum. This was originally added to 'fix' the possibility that the splitter would get 'lost' in the edge of it's container - or that nested splitters could collide.
I'm okay with removing it if the behavior is still acceptable in these extrema, but it's something worth checking for.

Zarenor:
I'm sorry I didn't see this earlier. I added a few more comments here.

I'm really glad to see the work on quantizing move forward, and I appreciate the other improvements made.

Looks great!

druid/src/widget/split.rs
cmyr: I'm just trying to make this a convention in the docs for any builder-style methods, since we have decided to forgo having actual builders.

For mutating methods, my feeling is "it depends". Some cases feel really obvious (changing label color) but others I'm less clear about. I don't think we need to add them everywhere; we can see what use dictates.

druid/src/widget/split.rs
cmyr: I think breadth is probably more confusing than advantageous. I think size is good?

druid/src/widget/split.rs
cmyr: agreed.

druid/src/widget/split.rs
xStrom: Yeah the usage of `width` is a bit confusing. Previously `Split` used both `size` and `width` to refer to this and I didn't spend much time thinking about it. I think it's best to use only one word for this concept and I'll do that update.

As for what word to use, `size` is decent. I can also think of `thickness` and `cross length`. I kind of like `thickness`.

cmyr: whichever you like; I still prefer size I think, thickness has z-axis implications for me?

xStrom: Well there's no z-axis involved with `Split` so that's an unlikely clash. Also [Flutter seems to use thickness](https://api.flutter.dev/flutter/material/Divider/thickness.html).

That said, I went with `size` for now due to its shortness and better fit with `area`.

druid/src/widget/split.rs
xStrom: I think having it set by default with some reasonable value is best. The main reason this exists is to allow extremely narrow splitter bars without sacrificing the user experience by requiring them to pixel hunt. So if someone wants a 3px bar, or even a 1px bar, they can easily do so but druid will default to a 6px bar area. If they want the area to also be 1px then they can set the minimum area to e.g. 0px.

druid/src/widget/split.rs
xStrom: This minimum isn't needed with the new logic, because the splitter bar is guaranteed to always be visible. This specific function calculates the min/max limits given the provided size and doesn't need to know about the splitter bar at all. The bar logic is elsewhere.

xStrom:
I addressed the feedback and made some other improvements.

* Improved documentation, including unifying language for builder methods.
* Renamed a bunch of variables and methods for consistency and reducing confusion, e.g.:
    * What was known previously as **splitter** is now the **splitter bar** to make it clear it's different from `Split` which is the whole widget. So **splitter bar** is how it's referenced in documentation, but actual methods are just `bar` for shortness.
    * The width/breadth/thickness of the splitter bar is now always called `size`.
    * `fill_splitter_handle` to `solid_bar` to use the new `bar` nomenclature and to make it more concise.
    * `split_direction` is now `split_axis` and this has some functional changes too, which I describe below.
* Swapped the meaning of the `Split::new(axis)`, `Split::horizontal()`, and `Split::vertical()` constructors. Previously it meant that the splitter bar is horizontal/vertical, but now it means that the horizontal/vertical axis is split in half. I think this reads better and this also matches what other frameworks are doing, e.g. [Qt](https://doc.qt.io/qt-5/qsplitter.html#orientation-prop) and [Metro Components](https://metroui.org.ua/splitter.html).
* The effective split point calculation now properly takes the bar area into account.
* The minimum split child size, bar size, and bar area is now rounded up to integers.
* Fixed the check for infinite height.
* Reduced code size a bit by some refactoring.

xStrom:
Based on the discussion in [zulip](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/split.20vertical.2Fhorizontal.20confusion) I renamed the constructor helpers from vertical/horizontal to row/cols.
```rust
Split::columns(child1, child2) // left and right
Split::rows(child1, child2)    // top and bottom
```

Issue #685, opened 2020-03-17
Add debug warnings for layout gotchas


futurepaul:
There are some layouts that are expressible in our system, but nonsensical, that we want to watch out for. When we can we'd like to log a warning, and if possible fix the API to make nonsensical layouts harder or impossible to layout. But warnings are a good first step.

Here are a few:
1. A "minimum" constraint of infinity makes no sense.
2. A child returning infinity as its size makes no sense (infinity should only be passed down to children, never bubbled back up). I believe this can currently happen with a flexible widget inside an align widget, for instance.
3. As far as I know, passing an infinite width to an align widget that's trying to center horizontally makes no sense, and same for height / vertical. However, an align widget could center something in one axis and still be fine with max infinity in the other — centering in a vertical scroll comes to mind.
4. If you try to "expand" a SizedBox inside of infinity that should be bad.

When possible, warnings can be added to the BoxConstraints method `debug_check`. `constrain` might also be a good spot. Not sure if that will work for all of these.

If you know of any other gotchas please add them!

PoignardAzur:
Are there still gotchas that aren't addressed?

PR #739, opened 2020-03-22
Infinite size warnings

futurepaul:
Seems like warning when a widget is attempting to return an infinite size fixes most of #685. This is not a systematic fix-up, but just what I've thought of so far.

These specific fixes are motivated by a layout issue I was having:

```rust
Flex::row().with_child(
  Label::new(|data: &ListItem, _env: &Env| data.file_name.clone())
      .expand_width()
      .on_click(|_, data, _| {
          println!("hey");
      }),
)
```

Flex gives SizedBox an infinite max constraint, SizedBox passes an infinite minimum constraint to Label (this is where we go wrong I believe), and Label measures itself and then "constrains" itself within the infinite BoxConstraints.

When I switched to `width_flex_child` I get my desired result.

### DIFF
--- a/druid/src/widget/mod.rs
+++ b/druid/src/widget/mod.rs
@@ -252,6 +252,11 @@ pub trait Widget<T> {
     fn id(&self) -> Option<WidgetId> {
         None
     }
+
+    /// Get the (verbose) type name of the widget for debugging purposes.
+    fn type_name(&self) -> &'static str {
+        std::any::type_name::<Self>()
+    }
 }
 
 impl WidgetId {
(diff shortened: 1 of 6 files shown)


druid/src/widget/mod.rs
```diff
@@ -252,6 +252,11 @@ pub trait Widget<T> {
     fn id(&self) -> Option<WidgetId> {
         None
     }
+
+    /// Get the (verbose) type name of the widget for debugging purposes.
+    fn type_name(&self) -> &'static str {
```
cmyr: on the `Box` impl I would call this method on the inner type; otherwise all boxed widgets will be `Box<dyn Widget<_>>`. Although maybe that's fine?

cmyr: oh one other request: can we do `#[doc(hidden)]` here, and can we say "you should not override this method"?

We'll come up with a better way to do this eventually.

PR #740, opened 2020-03-22
Alignment RadioGroup

HoNile:
Hi,

During one of the last updates the default alignment for `Flex::column` changed  wich changed the alignment from `RadioGroup` too.

Currently there is no way to set the alignment inside the `RadioGroup` as the return from `RadioGroup::new()` is `impl Widget<T>` .

So I update `RadioGroup::new()` to return a `Flex<T>` to be able to change the alignment if necessary and I reset the default alignment from RadioGroup like previously as I think it make more sence in the RadioGroup case.

Let me know if it's good to you or if it's too weird to return a `Flex<T>` for `RadioGroup::new()`.

### DIFF
--- a/druid/examples/flex.rs
+++ b/druid/examples/flex.rs
@@ -169,6 +169,7 @@ fn make_control_row() -> impl Widget<AppState> {
                 .with_child(Checkbox::new("Fix minor axis size").lens(Params::fix_minor_axis))
                 .with_spacer(10.)
                 .with_child(Checkbox::new("Fix major axis size").lens(Params::fix_major_axis))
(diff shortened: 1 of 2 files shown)


futurepaul:
I had been thinking of returning a Vec of Radio to solve this, but simply returning Flex feels like a way better solution!

HoNile:
Oh it look like I miss PR #717 part of it is the same as this one.

I have think a little more and I think we could rename `new` in `generate` if you find that returning a Flex is a good idea else of course you can close this PR.

cmyr:
yea sorry about this, will merge #717 shortly. I think a separate method on radio group that just returns the items as a vec makes sense.

PR #741, opened 2020-03-22
Improve multiwin

### LINKED ISSUE #728: ui inconsistency in multiwin example
afbeelding

When the menu count is reduced to 0, one of the windows is still showing `custom` menu.

kindlychung:
* fix #728 
* make use of the command handlers in the `Delegate`

### DIFF
--- a/druid/examples/multiwin.rs
+++ b/druid/examples/multiwin.rs
@@ -41,7 +41,7 @@ fn main() {
             LocalizedString::new("multiwin-demo-window-title").with_placeholder("Many windows!"),
         );
     AppLauncher::with_window(main_window)
-        .delegate(Delegate)
+        .delegate(Delegate{ windows: Vec::new()})
         .launch(State::default())
         .expect("launch failed");
 }
@@ -90,7 +90,9 @@ fn ui_builder() -> impl Widget<State> {
     col
 }
 
-struct Delegate;
+struct Delegate {
+    windows: Vec<WindowId>,
+}
 
 impl AppDelegate<State> for Delegate {
     fn event(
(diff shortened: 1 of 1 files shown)


cmyr:
@kindlychung could you break this into separate PRs? I would like to review the fix without it necessarily including the other changes, thanks!

kindlychung:
@cmyr The added buttons are part of the fix.

cmyr:
I keep on taking a look at this but I'm having trouble following along. Can you maybe explain what your understanding to the problem is, and why this fix makes sense to you? That would help me figure out how to evaluate this. :)

kindlychung:
> I keep on taking a look at this but I'm having trouble following along. Can you maybe explain what your understanding to the problem is, and why this fix makes sense to you? That would help me figure out how to evaluate this. :)

Sure.

The `Delegate` struct now keeps a list of window IDs, once a increment/decrement command arrives, it replays to all the windows, which makes sure all the windows consistently and correctly show the menu entries.

kindlychung:
The inconsistency is described in #728. You can compare the behavior after pulling this pr.

jneem:
Why add two new buttons instead of replacing the old buttons? It seems to me like the old buttons have the incorrect behavior and the new buttons have the correct behavior.

kindlychung:
> Why add two new buttons instead of replacing the old buttons? It seems to me like the old buttons have the incorrect behavior and the new buttons have the correct behavior.

Yeah, that's true. Will delete the old ones.

kindlychung:
Had a lot of trouble rebasing this. Close and open new pr.

PR #743, opened 2020-03-23
fancier calc

kindlychung:
i made some visual changes to the calc example, taking advantage of the .background(painter) method.

removed the C button, since it has exactly the same function as the CE button from what i read.

the C button now expands over 2 "columns", which imo is a nice demo of using flex to get a more complex layout.

### DIFF
--- a/druid/examples/calc.rs
+++ b/druid/examples/calc.rs
@@ -25,7 +25,7 @@ use druid::{
 };
 use druid::{
     widget::{CrossAxisAlignment, Flex, Label, MainAxisAlignment, Padding, Painter},
-    UnitPoint,
+    Insets, UnitPoint,
 };
 
 #[derive(Clone, Data, Lens)]
@@ -179,10 +179,11 @@ fn flex_row3<T: Data>(
     w2: impl Widget<T> + 'static,
     w3: impl Widget<T> + 'static,
 ) -> impl Widget<T> {
+    let insets = Insets::new(1., 1., 0., 0.);
     Flex::row()
-        .with_flex_child(Padding::new(1., w1), 2.0)
-        .with_flex_child(Padding::new(1., w2), 1.0)
-        .with_flex_child(Padding::new(1., w3), 1.0)
+        .with_flex_child(Padding::new(insets.clone(), w1), 2.0)
+        .with_flex_child(Padding::new(insets.clone(), w2), 1.0)
+        .with_flex_child(Padding::new(insets, w3), 1.0)
 }
 
 fn flex_row4<T: Data>(
@@ -191,11 +192,12 @@ fn flex_row4<T: Data>(
     w3: impl Widget<T> + 'static,
     w4: impl Widget<T> + 'static,
 ) -> impl Widget<T> {
+    let insets = Insets::new(1., 1., 0., 0.);
     Flex::row()
-        .with_flex_child(Padding::new(1., w1), 1.0)
-        .with_flex_child(Padding::new(1., w2), 1.0)
-        .with_flex_child(Padding::new(1., w3), 1.0)
-        .with_flex_child(Padding::new(1., w4), 1.0)
+        .with_flex_child(Padding::new(insets.clone(), w1), 1.0)
+        .with_flex_child(Padding::new(insets.clone(), w2), 1.0)
+        .with_flex_child(Padding::new(insets.clone(), w3), 1.0)
+        .with_flex_child(Padding::new(insets, w4), 1.0)
 }
 
 fn build_calc() -> impl Widget<CalcState> {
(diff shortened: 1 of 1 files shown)


kindlychung:
@futurepaul Could you review this? Thanks.

druid/examples/calc.rs
```diff
@@ -19,7 +19,14 @@ use druid::{
...
 };
 
-use druid::widget::{CrossAxisAlignment, Flex, Label, Painter};
+use druid::{
+    piet::{FixedLinearGradient, GradientStop},
+    Point,
+};
+use druid::{
+    widget::{CrossAxisAlignment, Flex, Label, MainAxisAlignment, Padding, Painter},
+    UnitPoint,
```
xStrom: Remove unused import `UnitPoint`.

druid/examples/calc.rs
```diff
@@ -165,44 +178,79 @@ fn digit_button(digit: u8) -> impl Widget<CalcState> {
...
-        .with_spacer(1.0)
-        .with_flex_child(w4, 1.0)
+        .with_flex_child(Padding::new(1., w1), 1.0)
+        .with_flex_child(Padding::new(1., w2), 1.0)
+        .with_flex_child(Padding::new(1., w3), 1.0)
+        .with_flex_child(Padding::new(1., w4), 1.0)
 }
 
 fn build_calc() -> impl Widget<CalcState> {
+    let display_painter = Painter::new(|ctx, _, env| {
```
xStrom: Rename unused variable `env` to `_env`.

druid/examples/calc.rs
```diff
@@ -165,44 +178,79 @@ fn digit_button(digit: u8) -> impl Widget<CalcState> {
...
+        let bounds = ctx.size().to_rect();
+        if let Ok(brush) = ctx.gradient(FixedLinearGradient {
```
futurepaul: Is there a reason this gradient can't be made with `LinearGradient::new`?

druid/examples/calc.rs
```diff
@@ -165,44 +178,79 @@ fn digit_button(digit: u8) -> impl Widget<CalcState> {
...
+            start: Point::new(0., 0.),
+            end: Point::new(100., 100.),
+            stops: vec![
+                GradientStop {
+                    pos: 0.0,
+                    color: Color::from_rgba32_u32(0x225553ff),
```
futurepaul: I feel like these colors don't match the rest of the calculator. What if this was just a subtle gray / dark gray gradient?

druid/examples/calc.rs
```diff
@@ -165,44 +178,79 @@ fn digit_button(digit: u8) -> impl Widget<CalcState> {
...
 }
 
-fn flex_row<T: Data>(
+fn flex_row3<T: Data>(
+    w1: impl Widget<T> + 'static,
+    w2: impl Widget<T> + 'static,
+    w3: impl Widget<T> + 'static,
+) -> impl Widget<T> {
+    Flex::row()
+        .with_flex_child(Padding::new(1., w1), 2.0)
```
futurepaul: Adding 1 pixel padding is different than what I had. What I'm going for is exactly 1 pixel between each button, but nothing around the outside edge. Here's the mac calculator for reference:
Screen Shot 2020-03-23 at 10 43 33 AM

druid/examples/calc.rs
```diff
@@ -145,15 +150,19 @@ fn op_button(op: char) -> impl Widget<CalcState> {
 fn digit_button(digit: u8) -> impl Widget<CalcState> {
     let painter = Painter::new(|ctx, _, env| {
         let bounds = ctx.size().to_rect();
+        let color = env.get(theme::BACKGROUND_LIGHT);
+        let color_hot: Color = Color::from_rgba32_u32(color.clone().as_rgba_u32() + 0x303030);
```
futurepaul: I like the idea of deriving a color from another color, but it feels a little verbose for an example. Is this something we should consider making easier in Piet, or at least refactoring as a function in here?

kindlychung: Totally for adding a method in piet.

druid/examples/calc.rs
kindlychung: yeah, i was trying to do that also, but after expanding the AC button to two columns, there is a 1-pixel mismatch with the button below AC.

druid/examples/calc.rs
kindlychung:
```
204 |         if let Ok(brush) = ctx.gradient(LinearGradient::new(
    |                                ^^^^^^^^ the trait `std::convert::From<piet::gradient::LinearGradient>` is not implemented for `piet::gradient::FixedGradient`
```

Maybe i missed something, but it didn't work.

druid/examples/calc.rs
kindlychung: Until then let's keep this as it is?

druid/examples/calc.rs
kindlychung: tried this, doesn't look so good. it's fighting with the rest of the ui, which is also mostly gray.

PR #744, opened 2020-03-23
delete word by pressing ctrl + backspace

kindlychung:


cmyr:
So this might be reasonable per-platform behaviour (on macOS option-backspace deletes a word, cmd-backspace deletes to the start of line) but It will need to be built on top of #617.

I'd also avoid the unrelated changes in `backspace`; that code is carefully written and probably never needs to be touched again. :)

kindlychung:
i certainly hope i didn't break anything. all the changes here is additive, existing behaviors shouldn't be affected at all.

cmyr:
Oops, I'd misread the patch slightly.

In any case, this is unfortunately not how we will implement this logic.

The first step of this will be to implement the concept of a `Movement`, which can be a word, a line, a paragraph, to the start of the document, etc; these will also largely borrow from implementations we've done in xi. Then this 'delete backwards' logic will be a subset of 'delete by movement'.

So I applaud you taking up this pretty tricky project, but this isn't the direction we're going for this. 😢

PR #747, opened 2020-03-23
Update piet to 0.0.11, bump other deps

cmyr:


druid/src/widget/image.rs
```diff
@@ -131,14 +131,10 @@ impl ImageData {
...
     pub fn from_dynamic_image(image_data: image::DynamicImage) -> ImageData {
-        match image_data.color() {
-            image::ColorType::RGBA(_) | image::ColorType::BGRA(_) | image::ColorType::GrayA(_) => {
-                Self::from_dynamic_image_with_alpha(image_data)
-            }
-            image::ColorType::RGB(_)
-            | image::ColorType::Gray(_)
-            | image::ColorType::Palette(_)
-            | image::ColorType::BGR(_) => Self::from_dynamic_image_without_alpha(image_data),
+        match image_data.color().channel_count() {
```
fishrockz: I sort of really like this.
I also really dont. 

If we bump the version of image in the future and they introduce a new type that dose not obay this rule it might be hard to debug with the new change but if we try to update image with the old code then rust will warn us that a new type has been added and we can think about what were are doing.
The idea would be for /channel_count/alpha_present/ I will take a action to go make a issue with the image crate for this.

But the commit message dose not at all match this change, I know druid is much less worried about our history than I normally am so I'm not sure that should block this.

But if it was in a separate PR then we could discus this while not blocking most of this PR

fishrockz: Actioned https://github.com/image-rs/image/issues/1167

druid-shell/Cargo.toml
```diff
@@ -16,19 +16,20 @@ use_gtk = ["gtk", "gtk-sys", "gio", "gdk", "gdk-sys", "glib", "glib-sys", "cairo
...
 
 [dependencies]
-piet-common = "0.0.9"
+piet-common = "0.0.10"
 log = "0.4.8"
 lazy_static = "1.0"
 time = "0.2.7"
 cfg-if = "0.1.10"
+# NOTE: if changing, ensure version is compatible with the version in piet
 kurbo = "0.5.11"
```
xStrom: So we're keeping the direct kurbo dependency and not going back to piet's re-export? I guess it can be useful again if we expect fast changes in kurbo.

druid-shell/Cargo.toml
```diff
@@ -16,19 +16,20 @@ use_gtk = ["gtk", "gtk-sys", "gio", "gdk", "gdk-sys", "glib", "glib-sys", "cairo
 default-target = "x86_64-pc-windows-msvc"
 
 [dependencies]
-piet-common = "0.0.9"
+piet-common = "0.0.10"
```
xStrom: Let's update piet to 0.0.11 instead after [piet#150](https://github.com/linebender/piet/pull/150) is merged.

druid-shell/Cargo.toml
cmyr: That was my thinking, yea.

druid/src/widget/image.rs
cmyr: yea, this change was a result of updating our `image` dependency; it broke its API to rename a bunch of things in `ColorType`, and when I went and looked at the API this was the solution that seemed simplest. I'll acknowledge it's a bit gross. 😒

xStrom: @fishrockz perhaps you can open a new issue for us here too to track this? Then we can get the version updates done and deal with the robustness of the `Image` widget separately.

fishrockz: Do we need to bump image now? I am happy to create a issue bump image once its fixed? but if we have a stonge reason to want a newer version of image for 0.0.5 then i dont think this should block.

cmyr: I'll do a more robust impl here, no worries

fishrockz: follow up issue when the image crate people get back to us, https://github.com/xi-editor/druid/issues/748

cmyr: okay, I updated this. You can no longer ensure you're matching all items, because the enum is non-exhaustive, but it should be more robust.

druid/src/widget/image.rs
```diff
@@ -203,6 +199,14 @@ impl ImageData {
     }
 }
 
+fn has_alpha_channel(image: &image::DynamicImage) -> bool {
+    use image::ColorType::*;
+    match image.color() {
+        La8 | Rgb8 | Rgba8 | La16 | Rgba16 | Bgra8 => true,
```
xStrom: `Rgb8` has alpha? Doesn't seem right.

cmyr: nope I just can't read 😄

druid/src/widget/image.rs
fishrockz: Apart from @xStrom's comment. I am very happy with this. in gitlab i would click resolve but i think think github has a concept of unresolved threads.

xStrom: Threads can be resolved, but merging isn't blocked here due to unresolved threads. Resolving hides the comments though, so I tend to only do it for trivial things.

fishrockz:
All the image stuff looks good to me!
I am not supper keen on the new way rs-image manges dynamic images but we have issues in place to improve things.

cmyr:
thanks all!

PR #749, opened 2020-03-23
Remove BoxedWidget

cmyr:
This was never used consistently and has a slightly confusing name, because it is not exactly what it says it is.


raphlinus:
I'm surprised this isn't used more. Isn't it one of the main ways for containers to own children?

cmyr:
We just don't use the type alias.

PR #750, opened 2020-03-23
Update readme

cmyr:
A few small things.


cmyr:
As we move more of this stuff to be in the book, we get rustfmt / CI for free, so I think that's the longer-term play.

cmyr:
(good catch though!)

PR #751, opened 2020-03-23
Immediate mode widget

Finnerale:
This will add a widget that allows writing parts of the ui in an immediate-mode style.

The reason I wrote this was for an event logger which should display events which look like this:
```rust
pub enum Event {
    WindowConnected,
    Size(Size),
    MouseDown(MouseEvent),
    MouseUp(MouseEvent),
    MouseMoved(MouseEvent),
    KeyDown(KeyEvent),
    KeyUp(KeyEvent),
    Paste(Clipboard),
    Wheel(WheelEvent),
    Zoom(f64),
    Timer(TimerToken),
    Command(Command),
    TargetedCommand(Target, Command),
}
```

There are two problems with representing this in druid:
- As of now, druid has no good way to represent enums
- The enum in question is not meant to be displayed

The latter means it does not derive `Data`, nor does any of its variants and even when druid had support for enums using some sort of proc-macro derive it still would be no use as such a derive would not be part of an `Event`.

The solution I came up with is this `Immediate` widget. Immediate mode fits Rust really well and makes this (mostly) a trivial, standard Rust `match` statement:
```rust
#[derive(Data)]
struct UiEvent {
    #[druid(same_fn = "PartialEq::eq")]
    inner: Event
}

Immediate::new(|event: &UiEvent| {
    match event.inner {
        Event::Size(size) => Label::new(format!("w: {}, h: {}", size.width, size.height)),
        _ => Label::new("Unhandled Event"),
    }
});
```
And such a wrapper is probably not necessary in most cases as the data in question would already be inside some other `Data` deriving structure of the ui state.

The `Immediate::new` function expects a function `&Data -> impl Widget<()>`.

I chose `Widget<()>` because it prevents one from using anything that depends on data, such as `Label::dynamic`. Using `Label::dynamic` does not make any sens in immediate-mode as the `Label` will be reconstructed anyway when the data changes.

A limitation of this design as of now is that there is no way to mutate the app state from the immediate-mode code, for example if a `Button` was in there. In order to do mutations such a `Button` would have to issue a command using `EventCtx` to do so.

I have not yet had the need to do state mutation inside `Immediate` but was wondering if there could be a nicer way to allow mutations. (One way to do so could be using `Widget<Data>` instead of `Widget<()>`, but that feels wrong as well, for the reason above)

### DIFF
--- a/druid/src/widget/immediate.rs
+++ b/druid/src/widget/immediate.rs
@@ -1,11 +1,39 @@
+// Copyright 2020 The xi-editor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 use crate::{widget::prelude::*, Data};
 
+/// A widget that allows writing parts of the ui in an immediate-mode style.
+///
+/// The primary usecase for this is displaying data that doesn't fit into druids model, such as
+/// enums you have no control over.
+///
+/// Whenever the state represented by the `Immediate` changes,
+/// it will reconstruct its content for the new data.
+///
+/// While this is not the most efficient thing to do, it is very simple and performs perfectly fine
+/// for small or rarely changed data.
+///
+/// Note that this is more like a last resort when the data to be represented is in a for druid particularly
+/// unfitting format.
 pub struct Immediate<D, W: Widget<()>> {
     constructor: Box<dyn Fn(&D) -> W>,
     content: Option<W>,
 }
 
 impl<D, W: Widget<()>> Immediate<D, W> {
+    /// Takes a constructor for a stateless widget
     pub fn new(constructor: impl Fn(&D) -> W + 'static) -> Self {
         Self {
             constructor: Box::new(constructor),
--- a/druid/src/widget/immediate.rs
+++ b/druid/src/widget/immediate.rs
@@ -1,0 +1,58 @@
+use crate::{widget::prelude::*, Data};
+
+pub struct Immediate<D, W: Widget<()>> {
+    constructor: Box<dyn Fn(&D) -> W>,
+    content: Option<W>,
+}
+
+impl<D, W: Widget<()>> Immediate<D, W> {
+    pub fn new(constructor: impl Fn(&D) -> W + 'static) -> Self {
+        Self {
+            constructor: Box::new(constructor),
+            content: None,
+        }
+    }
+}
+
+impl<D: Data, W: Widget<()>> Widget<D> for Immediate<D, W> {
+    fn event(&mut self, ctx: &mut EventCtx, event: &Event, _data: &mut D, env: &Env) {
+        if let Some(content) = &mut self.content {
+            content.event(ctx, event, &mut (), env);
+        }
+    }
+
+    fn lifecycle(&mut self, ctx: &mut LifeCycleCtx, event: &LifeCycle, data: &D, env: &Env) {
+        if let LifeCycle::WidgetAdded = event {
+            self.content = Some((self.constructor)(data));
+        }
+        if let Some(content) = &mut self.content {
+            content.lifecycle(ctx, event, &(), env);
+        }
+    }
+
+    fn update(&mut self, ctx: &mut UpdateCtx, old_data: &D, data: &D, env: &Env) {
+        if !old_data.same(data) {
+            self.content = Some((self.constructor)(data));
+            ctx.children_changed();
+        } else {
+            // This can happen when env changes, right?
+            if let Some(content) = &mut self.content {
+                content.update(ctx, &(), &(), env);
+            }
+        }
+    }
+
+    fn layout(&mut self, ctx: &mut LayoutCtx, bc: &BoxConstraints, _data: &D, env: &Env) -> Size {
+        if let Some(content) = &mut self.content {
+            content.layout(ctx, bc, &(), env)
+        } else {
+            Size::ZERO
+        }
+    }
+
+    fn paint(&mut self, ctx: &mut PaintCtx, _data: &D, env: &Env) {
+        if let Some(content) = &mut self.content {
+            content.paint(ctx, &(), env);
+        }
+    }
+}
--- a/druid/src/widget/immediate.rs
+++ b/druid/src/widget/immediate.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::{widget::prelude::*, Data};
+use crate::{widget::prelude::*, Data, WidgetPod};
 
 /// A widget that allows writing parts of the ui in an immediate-mode style.
 ///
@@ -27,13 +27,13 @@ use crate::{widget::prelude::*, Data};
 ///
 /// You should only use `Immediate` if your data format can't be reasonably used with other widgets.
 pub struct Immediate<D, W> {
-    constructor: Box<dyn Fn(&D) -> W>,
-    content: Option<W>,
+    constructor: Box<dyn Fn(&D) -> Option<W>>,
+    content: Option<WidgetPod<(), W>>,
 }
 
 impl<D, W: Widget<()>> Immediate<D, W> {
     /// Takes a constructor for a stateless widget
-    pub fn new(constructor: impl Fn(&D) -> W + 'static) -> Self {
+    pub fn new(constructor: impl Fn(&D) -> Option<W> + 'static) -> Self {
         Self {
             constructor: Box::new(constructor),
             content: None,
@@ -50,7 +50,7 @@ impl<D: Data, W: Widget<()>> Widget<D> for Immediate<D, W> {
 
     fn lifecycle(&mut self, ctx: &mut LifeCycleCtx, event: &LifeCycle, data: &D, env: &Env) {
         if let LifeCycle::WidgetAdded = event {
-            self.content = Some((self.constructor)(data));
+            self.content = (self.constructor)(data).map(WidgetPod::new);
         }
         if let Some(content) = &mut self.content {
             content.lifecycle(ctx, event, &(), env);
@@ -59,12 +59,12 @@ impl<D: Data, W: Widget<()>> Widget<D> for Immediate<D, W> {
 
     fn update(&mut self, ctx: &mut UpdateCtx, old_data: &D, data: &D, env: &Env) {
         if !old_data.same(data) {
-            self.content = Some((self.constructor)(data));
+            self.content = (self.constructor)(data).map(WidgetPod::new);
             ctx.children_changed();
         } else {
             // This can happen when env changes, right?
             if let Some(content) = &mut self.content {
-                content.update(ctx, &(), &(), env);
+                content.update(ctx, &(), env);
             }
         }
     }
--- a/druid/src/widget/immediate.rs
+++ b/druid/src/widget/immediate.rs
@@ -26,7 +26,7 @@ use crate::{widget::prelude::*, Data};
 /// for small or rarely changed data.
 ///
 /// You should only use `Immediate` if your data format can't be reasonably used with other widgets.
-pub struct Immediate<D, W: Widget<()>> {
+pub struct Immediate<D, W> {
     constructor: Box<dyn Fn(&D) -> W>,
     content: Option<W>,
 }
@@ -70,11 +70,10 @@ impl<D: Data, W: Widget<()>> Widget<D> for Immediate<D, W> {
     }
 
     fn layout(&mut self, ctx: &mut LayoutCtx, bc: &BoxConstraints, _data: &D, env: &Env) -> Size {
-        if let Some(content) = &mut self.content {
-            content.layout(ctx, bc, &(), env)
-        } else {
-            Size::ZERO
-        }
+        self.content
+            .as_mut()
+            .map(|c| c.layout(ctx, bc, &(), env))
+            .unwrap_or_default()
     }
 
     fn paint(&mut self, ctx: &mut PaintCtx, _data: &D, env: &Env) {
--- a/druid/src/widget/immediate.rs
+++ b/druid/src/widget/immediate.rs
@@ -16,7 +16,7 @@ use crate::{widget::prelude::*, Data};
 
 /// A widget that allows writing parts of the ui in an immediate-mode style.
 ///
-/// The primary usecase for this is displaying data that doesn't fit into druids model, such as
+/// The primary use case for this is displaying data that doesn't fit into druid's model, such as
 /// enums you have no control over.
 ///
 /// Whenever the state represented by the `Immediate` changes,
@@ -25,8 +25,7 @@ use crate::{widget::prelude::*, Data};
 /// While this is not the most efficient thing to do, it is very simple and performs perfectly fine
 /// for small or rarely changed data.
 ///
-/// Note that this is more like a last resort when the data to be represented is in a for druid particularly
-/// unfitting format.
+/// You should only use `Immediate` if your data format can't be reasonably used with other widgets.
 pub struct Immediate<D, W: Widget<()>> {
     constructor: Box<dyn Fn(&D) -> W>,
     content: Option<W>,
--- a/druid/src/widget/mod.rs
+++ b/druid/src/widget/mod.rs
@@ -28,6 +28,7 @@ mod identity_wrapper;
 #[cfg(feature = "image")]
 #[cfg_attr(docsrs, doc(cfg(feature = "image")))]
 mod image;
+mod immediate;
 mod label;
 mod list;
 mod padding;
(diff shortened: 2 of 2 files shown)


Finnerale:
I've pushed the event logger example branch (based on this PR) here:
https://github.com/Finnerale/druid/blob/event-logger/druid/examples/event_log.rs

druid/src/widget/immediate.rs
```diff
@@ -0,0 +1,86 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::{widget::prelude::*, Data};
+
+/// A widget that allows writing parts of the ui in an immediate-mode style.
+///
+/// The primary usecase for this is displaying data that doesn't fit into druids model, such as
```
xStrom: `usecase` -> `use case`
`druids` -> `druid's`

druid/src/widget/immediate.rs
```diff
@@ -0,0 +1,86 @@
...
+/// enums you have no control over.
+///
+/// Whenever the state represented by the `Immediate` changes,
+/// it will reconstruct its content for the new data.
+///
+/// While this is not the most efficient thing to do, it is very simple and performs perfectly fine
+/// for small or rarely changed data.
+///
+/// Note that this is more like a last resort when the data to be represented is in a for druid particularly
+/// unfitting format.
```
xStrom: This sentence seems a bit chaotic, especially the end. Maybe replace it with something like ```You should only use `Immediate` if your data format can't be reasonably used with other widgets.```

druid/src/widget/immediate.rs
```diff
@@ -0,0 +1,86 @@
...
+            self.content = Some((self.constructor)(data));
+        }
+        if let Some(content) = &mut self.content {
+            content.lifecycle(ctx, event, &(), env);
+        }
+    }
+
+    fn update(&mut self, ctx: &mut UpdateCtx, old_data: &D, data: &D, env: &Env) {
+        if !old_data.same(data) {
+            self.content = Some((self.constructor)(data));
```
xStrom: A `LifeCycle::WidgetAdded` event should be sent to the newly constructed widget.

druid/src/widget/immediate.rs
```diff
@@ -0,0 +1,86 @@
...
+/// Whenever the state represented by the `Immediate` changes,
+/// it will reconstruct its content for the new data.
+///
+/// While this is not the most efficient thing to do, it is very simple and performs perfectly fine
+/// for small or rarely changed data.
+///
+/// Note that this is more like a last resort when the data to be represented is in a for druid particularly
+/// unfitting format.
+pub struct Immediate<D, W: Widget<()>> {
+    constructor: Box<dyn Fn(&D) -> W>,
```
xStrom: Perhaps it makes sense to have `Box<dyn Fn(&D) -> Option<W>>` for cases where we don't want to display any widget for events we don't care about?

druid/src/widget/immediate.rs
Finnerale: Doesn't this happen when I call `ctx.children_changed()`?

druid/src/widget/immediate.rs
Finnerale: Thanks, that sounds a lot better :sweat_smile:

druid/src/widget/immediate.rs
Finnerale: Hmm, I think I would prefer some kind of `Empty` widget that could be returned instead.
That could be useful in other places and allows keeping the api a bit simpler.

druid/src/widget/immediate.rs
xStrom: I think `ctx.children_changed()` will trigger a `LifeCycle::RouteWidgetAdded` but not `LifeCycle::WidgetAdded`. Having a `WidgetPod` encapsulate `self.content` might also fix it but seems heavyweight for `Immediate`.

I haven't researched the whole `WidgetAdded` system too much, so perhaps @cmyr can chime in here. However I think that we could even remove `ctx.children_changed()` and just send the `WidgetAdded` event after creating the inner widget.

druid/src/widget/immediate.rs
xStrom: Well as a developer using `Immediate` you could still return an `Empty` widget if you wish. I think the `Option` would enable a significant performance boost in certain cases. For example let's say you're only interested in displaying mouse click event info. Then every other event, especially events like size, will still cause an allocation of `Empty`. With `None` there would be no allocation.

druid/src/widget/immediate.rs
cmyr: yes, all you need to do after adding a widget is call `ctx.children_changed()`. `RouteWidgetAdded` is an implementation detail; when we see that event, we check to see if the current widget is new, and if it is we replace it with `WidgetAdded`.

druid/src/widget/immediate.rs
```diff
@@ -0,0 +1,85 @@
...
+        } else {
+            // This can happen when env changes, right?
+            if let Some(content) = &mut self.content {
+                content.update(ctx, &(), &(), env);
+            }
+        }
+    }
+
+    fn layout(&mut self, ctx: &mut LayoutCtx, bc: &BoxConstraints, _data: &D, env: &Env) -> Size {
+        if let Some(content) = &mut self.content {
```
cmyr: this looks like a good opportunity for a `map`:

```rust
self.content.as_mut().map(|c| c.layout(ctx, bc, &(), env)).unwrap_or_default()
```

druid/src/widget/immediate.rs
```diff
@@ -0,0 +1,85 @@
...
+/// enums you have no control over.
+///
+/// Whenever the state represented by the `Immediate` changes,
+/// it will reconstruct its content for the new data.
+///
+/// While this is not the most efficient thing to do, it is very simple and performs perfectly fine
+/// for small or rarely changed data.
+///
+/// You should only use `Immediate` if your data format can't be reasonably used with other widgets.
+pub struct Immediate<D, W: Widget<()>> {
```
cmyr: As a general guideline, you should have as few bounds as possible on a type itself. In this case, we don't need to know that `W` is a widget yet; we can add that bound in the impl.

druid/src/widget/immediate.rs
```diff
@@ -0,0 +1,85 @@
...
+    /// Takes a constructor for a stateless widget
+    pub fn new(constructor: impl Fn(&D) -> W + 'static) -> Self {
+        Self {
+            constructor: Box::new(constructor),
+            content: None,
+        }
+    }
+}
+
+impl<D: Data, W: Widget<()>> Widget<D> for Immediate<D, W> {
```
cmyr: I think the inner widget should have the same `Data` as self, and should use a lens to determine what its inner data type is? I appreciate this is a sort of academic point since you're explicitly trying to not have the inner type use data? But it seems like an unnecessary limitation.

druid/src/widget/immediate.rs
Finnerale: But if `Empty` is defined as
```rust
struct Empty;
```
It should not allocate anything, thus I kind of doubt that it would make a performance difference except having one more method call for each widget method :thinking:

druid/src/widget/immediate.rs
Finnerale: So this would work as expected? Would there be any benefit in sending the `WidgetAdded` event directly?

cmyr: This is correct. The only time it would be okay to send `WidgetAdded` directly is when we are already handling a `WidgetAdded` event; otherwise state will not correctly propagate back up the tree.

druid/src/widget/immediate.rs
Finnerale: I chose `()` explicitly because it prevents one from using anything that depends on data, such as Label::dynamic. Using Label::dynamic does not make any sens in immediate-mode as the Label will be reconstructed anyway when the data changes. Using this widget means giving up the advantages druid offers, this should be clear.

Finnerale:
I agree that this is not the best way to solve a data model problem.
What I would prefer is first class support for representing enums in the ui.
However I could not come up with any solution that would allow me to represent enums that I have no control over (except copying the enum declaration, derive the things i need and implement `From` and `Into` traits from/to the original for it)

druid/src/widget/immediate.rs
xStrom: I looked more closely at druid code and my initial comment seems correct. The only place in the whole druid code base that generates `WidgetAdded` is the `WidgetPod` if it is new.

Thus saying that *all you need to do after adding a widget is call `ctx.children_changed()`* is not true. Your widgets need to also be encapsulated in a `WidgetPod` for it to work.

So let's clarify, @cmyr. Are you saying that `WidgetPod` use is mandatory and bare widgets are no longer valid druid code?

druid/src/widget/immediate.rs
xStrom: It's possible, but that way:
* Each developer needs to know about this even being a problem and possible to solve.
* Each developer needs their own bespoke implementation of `Empty`.

I think that common scenarios should be solved by the druid widget efficiently. Given that the whole idea here is to react to giant enums, not wanting to react to each and every variant seems rather common.

xStrom:
To me this `Immediate` approach seems like a quite convenient solution to showing enums.

The alternative apporach of creating a bunch of extra structs and enums to convert the original enum seems like a lot of work, both in developer time and CPU time.

druid/src/widget/immediate.rs
cmyr: sorry, this stuff really needs to be written down.

Basically: for a widget to fully participate in druid (to have an id, to have its own `hot` state, to be focusable, to correctly handle events) that widget needs to be in a `WidgetPod`. One way of thinking about `WidgetPod` is that it is the 'base class' for a widget.

In general, `WidgetPod`s are created behind the scenes; for instance a container widget creates a `WidgetPod` for each widget it manages.

You *can*, in certain circumstances, use a widget on its own. The main example of this I can think of is when a widget wants to reuse another widgets `paint` method, such as how `Button` currently owns a `Label` that is not in a pod; the button calls methods on the label directly, paints it with an offset, etc.

So yes, looking at this, if we want these widgets to correctly receive `WidgetAdded` we will need each one to be wrapped in a widgetpod.

cmyr:
I think this is interesting, but I'm not yet convinced it should be in druid proper; it feels like a significant break from our common patterns. That said, I would be curious to see an event viewer that uses this?

I would be more interested in a solution (maybe with some macro help) that generates a lens that maps from an enum to a widget, with the widget choice being based on the enum variant, and us generating the lens code that unwraps the correct data for that widget. The lens could actually be part of the widget that is returned, such that for an enum `T`, all returned widgets are `Widget<T>`. This definitely needs more sketching out, though?

(that said, I could imagine a macro that generates struct representations of enum members, like:)

```rust
// given this enum
enum MyState {
    Unit,
    Tuple(usize, String),
    Struct { index: usize, name: String },
}

// we might generate the following, which would be the `Data` type of the inner widgets:
struct MyState_Unit;
struct MyState_Tuple(usize, String);
struct MyState_Struct { index: usize, name: String };
```


So basically: I'm open to this but still a bit circumspect, but why don't you at least use it for the event viewer and we can see how that feels?

Finnerale:
> That said, I would be curious to see an event viewer that uses this?

https://github.com/Finnerale/druid/blob/event-logger/druid/examples/event_log.rs

Finnerale:
> needs more sketching out, though?

This is what I got so far
https://github.com/Finnerale/druid-enum-switcher

druid/src/widget/immediate.rs
xStrom: Thanks for the description. It matches my understanding.

Now back to what I was originally saying. One solution is to have `self.content` be encapsulated in a `WidgetPod`. It's certainly valid and also the most robust solution. However because we're really just passing events straight to `self.content` with no processing we could skip having a `WidgetPod` for performance reasons. That means `Immediate` would have to emulate the `WidgetAdded` lifecycle event for `self.content` whenever it gets re-created. It's really straightforward to do and won't cause any issues. The only threat of issues is when `WidgetPod` gets more bootstrapping functionality like `WidgetAdded` and we forget to update `Immediate`.

cmyr:
(The following is my attempt to work through this on my own; it covers some of the same ground as some of ideas in your sample project)

Okay, so matcher.rs is a step in the direction I'm thinking of. The main thing that's missing, for me, is the idea of having concrete types for the different inner widgets. So given the data,

```rust
enum Things {
    One(usize),
    Two { name: String, count: usize },
}
```

We want to have a 'base' widget that is a `Widget<Things>`, and then it will have children that are also `Widget<Things>` but use something like lens to end up being either `Widget<One>` or `Widget<Two>`.

We will need a trait on the enum, as you mentioned.


```rust
trait DataVariant<T: Data> {
    fn same_variant(&self, other: &Self) -> bool;
    fn make_widget(&self) -> Box<dyn Widget<Self>>;
}
```

And then some kind of wrapper widget:

```rust
struct VariantSwitcher<T> {
    inner: WidgetPod<T, Box<dyn Widget<T>>>,
}

impl<T: DataVariant> Widget<T> for VariantSwitcher<T> {
    fn lifecycle(&mut self, ctx: &mut LifeCycleCtx, event: &LifeCycle, data: &T, env: &Env) {
        if let LifeCycle::WidgetAdded = event {
            self.inner = WidgetPod::new(data.make_variant());
        }
        self.inner.lifcycle(ctx, event, data, env);
    }
    
    fn update(&mut self, ctx: &mut UpdateCtx, old_data: &T, data: &T, env: &Env) {
        if !old_data.same_variant(data) {
            self.inner = WidgetPod::new(data.make_variant());
            ctx.children_changed();
        } else {
            self.inner.update(ctx, old_data, data, env);
        }
    }
}
```

And here `DataVariant` is intended to be derived or otherwise generated by macro, like (we also have `LensVariant` here that generates (unwrapping?) lenses for each variant.

```rust
#[derive(Clone, Data, LensVariant)]
enum Things {
    One(usize),
    Two { name: String, count: usize },
}

data_variant!(Things,
    Things::One => || Label::dynamic(|_, data: &(usize), | format!("One {}", data.0)).lens(Things::VariantOne).boxed(),
    Things::Two => || make_variant_two(),
);

fn make_variant_two() -> Box<dyn Widget<Things>> {
    Flex::row()
        .with_child(Button::new(|_, data: &Things_Two, _| data.name.to_string()))
        .lens(Things::VariantTwo)
        .boxed()
}
```

So... this is also kind of gross. Particularly gross is the fact that you need to know the names of the generated structs in order to build the widgets for each variant. I'm not really sure how to get around that, though?

druid/src/widget/immediate.rs
cmyr: Yes, this would be possible. If we wanted our child to do anything at all fancy, though (such as receive events or set timers or even potentially animate) then we struggle to make that work.

druid/src/widget/immediate.rs
Finnerale: My idea was that druid would provide such `Empty` widget, but I agree that it is a bit of a hidden solution and, on second thought, it would probably only be useful inside of `Immediate`.

druid/src/widget/immediate.rs
xStrom: Why would we struggle? If `Immediate` can do it, then the child can do it. Because we just lend out the `Immediate` identity/ctx to the child and `Immediate` itself never does anything.

cmyr: For one thing, If the widget was a container, its children would not be appropriately registered, and so things like focus would break.

Finnerale:
I've now updated the widget to take
```rust
impl Fn(&D) -> Option<W> + 'static
```
as constructor as discussed [here](https://github.com/xi-editor/druid/pull/751#discussion_r397122277)

Also wrapped the content in a `WidgetPod` as it should not be much overhead and makes this a bit more future proof as discussed [here](https://github.com/xi-editor/druid/pull/751#discussion_r397116114)

I agree with @cmyr that we should have better means to deal with enums, however I don't see an alternative when the enum in question is not under ones control.

cmyr:
I may change my tune at some point, but I basically think that your data model should be under your control. If this means you have to do a bit of typing to massage something into shape, then I think that's an okay price to pay.

We basically tied ourselves to this when we decided to use `Data`.

cmyr:
My current understanding of this PR is that it is on hold pending #789, if anybody understands differently please correct me!

druid/src/widget/immediate.rs
```diff
@@ -0,0 +1,84 @@
...
+    fn layout(&mut self, ctx: &mut LayoutCtx, bc: &BoxConstraints, _data: &D, env: &Env) -> Size {
+        self.content
+            .as_mut()
+            .map(|c| c.layout(ctx, bc, &(), env))
+            .unwrap_or_default()
+    }
+
+    fn paint(&mut self, ctx: &mut PaintCtx, _data: &D, env: &Env) {
+        if let Some(content) = &mut self.content {
+            content.paint(ctx, &(), env);
```
justinmoon, line 81: I just tried copying this widget into my project. It didn't work, but changing `content.paint` to `content.paint_raw` worked. Perhaps Druid's API changed?

cmyr:
@Finnerale do you think it makes sense to keep this open, or should we close or now?

Finnerale:
I think this should be closed for a couple reasons:

- First, the initial intend behind this was displaying enums, which has been solve for me with [`druid-enums`](https://github.com/finnerale/druid-enums).

- Second, this Widget fails as soon as you place anything dynamic / interactive in there. Druid just does not fit the immediate mode model, because widgets have no way of retaining state when being reconstructed.

- Third, we already have this widget in our standard set, we just called it `ViewSwitcher` and I intend to write a word of warning about this "use case" in it's documentation.

Thanks for reminding me about it :+1:

Issue #628, opened 2020-03-10
Widgets with stroked outlines draw outside their bounds


cmyr:
The `stroke` method on our drawing context draws the stroke centered on the path; if the path it is passed is equal to the layout bounds, half of the stroke will be outside of those bounds.

This means that any widget that draws a stroke and does not account for this will end up drawing outside of its layout bounds.

This is illustrated well by the flex example program, from which I've taken the following zoomed screenshots:
<img width="266" alt="Screen Shot 2020-03-10 at 10 27 34 AM" src="https://user-images.githubusercontent.com/3330916/76322435-c44bdf80-62b9-11ea-808f-e29ad215641a.png">
<img width="271" alt="Screen Shot 2020-03-10 at 10 28 03 AM" src="https://user-images.githubusercontent.com/3330916/76322479-d4fc5580-62b9-11ea-86a0-f7cef64f1c53.png">
<img width="275" alt="Screen Shot 2020-03-10 at 10 28 45 AM" src="https://user-images.githubusercontent.com/3330916/76322551-ed6c7000-62b9-11ea-8415-f6ecebd53cf9.png">

The solution to this is fairly simple; we should inset our layout size before painting, increasing the layout size if necessary. When dealing with  a `Rect` this is as easy as,

```rust
let layout_rect = Rect::ZERO.with_size(paint_ctx.size());
let paint_rect = layout_rect - Insets::uniform(my_stroke_width / 2.0);
```

PR #752, opened 2020-03-23
Reimplement Button using on_click

### LINKED ISSUE #745: refactor Button in terms of Clickable
Basically: I think `Button` should just be a nice way to create a button that displays some text and has reasonable layout properties.

I think our API would be better if `button::new()` only took one parameter (the text) and that you added the action by calling an `on_click` method.

This could either _actually_ be using a `Clickable`, or it could be a separate method with the same name, if for some reason `Button` needs its own logic?

futurepaul:
closes #745 

This behaves and scales pretty much exactly like the standard button. The only change is I flipped the gradient so that it's light on top — not sure how it got flipped in the first place. I also believe I have the correct inset on the background size so the border doesn't paint outside its bounds anymore (#628).

I'm not super excited about needing this CenteredLabel widget for centering the label, but I couldn't figure out how to do it with standard layout widgets. (See [related zulip discussion](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/ConstrainedBox) for my tale of woe)

### DIFF
--- a/druid/src/tests/mod.rs
+++ b/druid/src/tests/mod.rs
@@ -38,7 +38,6 @@ fn propogate_hot() {
     let widget = Split::vertical(
         SizedBox::empty().with_id(empty),
         Button::new("hot")
-            .on_click(|_, _, _| {})
             .record(&button_rec)
             .with_id(button)
             .padding(50.)
--- a/druid/src/tests/mod.rs
+++ b/druid/src/tests/mod.rs
@@ -37,7 +37,8 @@ fn propogate_hot() {
 
     let widget = Split::vertical(
         SizedBox::empty().with_id(empty),
-        Button::new("hot", |_, _, _| {})
+        Button::new("hot")
+            .on_click(|_, _, _| {})
             .record(&button_rec)
             .with_id(button)
             .padding(50.)
--- a/druid/src/widget/button.rs
+++ b/druid/src/widget/button.rs
@@ -16,8 +16,7 @@
 use crate::theme;
 use crate::widget::{Flex, Label, LabelText, MainAxisAlignment, Painter};
 use crate::{
-    Data, Env, EventCtx, Insets, LinearGradient, Point, Rect, RenderContext, UnitPoint, Widget,
-    WidgetExt,
+    Data, Insets, LinearGradient, Point, Rect, RenderContext, UnitPoint, Widget, WidgetExt,
 };
 
 // the minimum padding added to a button.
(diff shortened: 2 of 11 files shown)


druid/src/widget/button.rs
```diff
@@ -40,12 +37,52 @@ impl<T: Data> Button<T> {
...
+    }
+
+    fn painter() -> Painter<T> {
+        Painter::new(|ctx, _, env| {
+            let is_active = ctx.is_active();
+            let is_hot = ctx.is_hot();
+            let size = ctx.size();
+
+            let rounded_rect = Rect::from_origin_size(Point::ORIGIN, size)
+                .inset(-1.0)
```
xStrom: The inset needs to be dynamic, because the border width is dynamic.
```rust
let border_width = env.get(theme::BUTTON_BORDER_WIDTH);
// ...
.inset(border_width / -2.0)
// ...
ctx.stroke([.......], border_width);
```

futurepaul: Thank you!

futurepaul:
Okay, updated as per a zulip convo. Turns out I really like using `.on_click` everywhere because passing a closure as a second argument can be a little visually confusing.

druid/src/tests/mod.rs
```diff
@@ -37,7 +37,8 @@ fn propogate_hot() {
 
     let widget = Split::vertical(
         SizedBox::empty().with_id(empty),
-        Button::new("hot", |_, _, _| {})
+        Button::new("hot")
+            .on_click(|_, _, _| {})
```
cmyr: can we just delete this `on_click` here

druid/src/widget/button.rs
```diff
@@ -13,146 +13,98 @@
...
-const LABEL_INSETS: Insets = Insets::uniform_xy(8., 2.);
+const LABEL_INSETS: Insets = Insets::uniform_xy(8., 4.);
 
 /// A button with a text label.
 pub struct Button<T> {
-    label: Label<T>,
-    label_size: Size,
-    /// A closure that will be invoked when the button is clicked.
-    action: Box<dyn Fn(&mut EventCtx, &mut T, &Env)>,
+    phantom: std::marker::PhantomData<T>,
```
cmyr: I don't think we need this? It doesn't seem like button needs any generic params if we're just sort of using it as a namespace

druid/src/widget/button.rs
```diff
@@ -13,146 +13,98 @@
...
 
 //! A button widget.
-
 use crate::theme;
-use crate::widget::{Label, LabelText};
+use crate::widget::{Flex, Label, LabelText, MainAxisAlignment, Painter};
 use crate::{
-    Affine, BoxConstraints, Data, Env, Event, EventCtx, Insets, LayoutCtx, LifeCycle, LifeCycleCtx,
-    LinearGradient, PaintCtx, Point, Rect, RenderContext, Size, UnitPoint, UpdateCtx, Widget,
+    Data, Insets, LinearGradient, Point, Rect, RenderContext, UnitPoint, Widget, WidgetExt,
```
cmyr: I don't think that button should *require* the use of `WidgetExt`; really nothing should. What this means, I think, is that button should have its own `on_click` method; this can either make a clickable or we could still have our own logic for that?

druid/src/widget/button.rs
```diff
@@ -13,146 +13,98 @@
...
-        self.label_size = self.label.layout(layout_ctx, &label_bc, data, env);
-        // HACK: to make sure we look okay at default sizes when beside a textbox,
-        // we make sure we will have at least the same height as the default textbox.
-        let min_height = env.get(theme::BORDERED_WIDGET_HEIGHT);
-
-        bc.constrain(Size::new(
-            self.label_size.width + padding.width,
-            (self.label_size.height + padding.height).max(min_height),
-        ))
+    /// Create a new button with a child widget for a label.
```
cmyr: I wouldn't say 'label', maybe

druid/src/widget/button.rs
```diff
@@ -13,146 +13,98 @@
...
-
-        ctx.fill(rounded_rect, &bg_gradient);
-
-        let label_offset = (size.to_vec2() - self.label_size.to_vec2()) / 2.0;
-
-        ctx.with_save(|ctx| {
-            ctx.transform(Affine::translate(label_offset));
-            self.label.paint(ctx, data, env);
-        });
+    fn painter() -> Painter<T> {
```
cmyr: One thing I've been thinking about is that we could use `Painter` to do button styles? This is a bit tricky because we might want to be able to assume that a button has text, i.e. we might want a text button with different styles. Is there a way we can do this? Can we have a reusable 'button style', that works with the normal button API? I know this isn't really the scope for this PR, but I'm spitballing.

futurepaul: It's not quite clear to me how, but I do think somehow Painter will give us a good way for styling Button.

futurepaul:
Okay I hear what you're saying and agree with most of it. I believe I addressed your bullet points. I do want to use Painter if possible because if I feel like it will be important for styling going forward... though I could be mistaken!

I guess some of my thinking with Button is that it should not only be a good button, but also demonstrate best practices for composing widgets when feasible / ideal. So I guess the big question is when is feasible / ideal.

druid/src/widget/button.rs
```diff
@@ -13,51 +13,104 @@
...
+    /// Create a new button with a text label.
+    ///
+    /// Use the `.on_click` method to provide a closure to be called when the
+    /// button is clicked.
     ///
     /// # Examples
     ///
     /// ```
-    /// use druid::widget::Button;
+    /// use druid::widget::{Button};
```
cmyr: oops

futurepaul:
Sure thing I can get that in tmrw

PR #755, opened 2020-03-24
Remove setup section from book

cmyr:
It makes more sense to me that this just lives in the readme.


ForLoveOfCats:
I thought that the idea was to move more of this stuff out of the readme and into the book?

cmyr:
I think "installing dependencies" is something that makes sense in the readme?

ForLoveOfCats:
I'm of the opinion that as much of the information should be stored in one place, and if that place is the book then I think that this information should be included in the book instead of the readme. The readme can "sell" the concept of druid but the book is all the information you need to use it as well as how it works under the hood.

Issue #158, opened 2019-09-17
Windows backend reports incorrect text for ctrl+key


cmyr:
As discussed [in this comment](https://github.com/xi-editor/druid/pull/149#discussion_r324437950); the goal with keyboard event handling is to report the unmodified text value for a printable key, when a printable key is pressed.

PoignardAzur:
Does this still occur?

PR #759, opened 2020-03-26
Wasm support

### LINKED ISSUE #128: wasm support
Hi Raph,

I thought i'd have a go at making druid work with wasm, since piet appears to support it (although curiously with the `web` feature rather than selecting by target architecture?).

druid-shell brings up 8 errors when i try and do it there. my gut feeling is that this is probably not the place to make the change, since the errors are for `Window`, `WindowBuilder` etc.
Trying to make the changes in in druid itself is definitely less good, with rustc giving up after 200 errors.

I know you've said it's probably a bit early to be worrying about too many backends, but given the logistical differences with this one, i thought it would be a good test of druid's design.

I'm willing to continue with the work as i'm able, but i think i'm going to need some guidance on the best way to go about it to get any further.

elrnv:
This is a continuation of #30 and includes all the changes made there + support for scrolling and keyboard input.

As mentioned in #30, there is an issue with `std::time`, which was solved with the [`instant`](https://crates.io/crates/instant) crate, but I believe there are other options available as well.

A [demo of the working widgets](https://elrnv.github.io/druid-wasm-examples/) is available.

The limitations in this PR AFAICT are:
- Loading images
- Clipboard implementation
- File dialog handling
- Menu (this may require designing a custom menu widget)

This PR also depends on a pull request in `piet` [#153](https://github.com/linebender/piet/pull/153), which makes returning the `piet` context easier.

Edit: Fixes #128


Finnerale:
I've been playing around with the demo (in Firefox and Chromium) and everything seems to work, one exception being zoom, which seems to break things quite a bit.

When loading example with zoom 100%:
- Zooming in makes content smaller
- Zooming out makes content larger
- Zooming in any way breaks cursor position

When loading example with zoom > 100%:
- Example displays at normal (100%) size

When loading example with zoom < 100%:
- White bar at the right (in the backround)
- Druid can draw above it
- Leaves artifacts behind when druid stops drawing over the white bar (flex example)

In both cases != 100%
- Zooming in any direction / resizing the window makes it jump to the chosen zoom level (for example when loading at 200% an pressing `-` sets zoom to 170% and then it looks like when loaded at 100% and then zoomed to 170%

elrnv:
Thanks @Finnerale! I believe I have resolve the issues you found with zooming. This relies on [#158 ](https://github.com/linebender/piet/pull/158) being merged as well.

It seems the zooming issue was in part because the context wasn't scaled properly during the resize callback.

elrnv:
I organized a small repo for building and hosting these examples on the web at https://github.com/elrnv/druid-wasm-examples/, hopefully it can serve as a reference on how one would go about building and deploying these examples to the web. It certainly took me some time to figure it out.

ijsnow:
Woah! Amazing timing! I just discovered druid and was hoping for wasm support and here it is.

It appears another limitation of the current state is that it doesn't capture the browser's native functionality of navigating back when you press "backspace" when a text box is focused

elrnv:
Oh thanks for finding that @ijsnow, in my setup, the browser doesn't respond to backspace so I never noticed this. I was able to reproduce this on another machine, and indeed it makes the text widgets basically unusable. I think this would be an unacceptable limitation, so I patched it by disabling the default backspace functionality altogether.

Perhaps this is too conservative and the default backspace behaviour should be enabled when the text widget is not focused, but for this PR I think this is acceptable since most likely the kinds of web applications that would require a GUI like druid would opt for a custom backspace behaviour anyway.

The [druid-wasm-examples](https://github.com/elrnv/druid-wasm-examples/) should be updated with the fix now. It would be great if you can verify if this issue is resolved there on your end :)

elrnv:
@cmyr, @raphlinus: Is there anything else that needs to be done to get this through, or is it just not a priority right now?

cmyr:
No, I'd like to get this merged while it's warm, just haven't built up the courage to give it a closer look. will do that soon!

druid/src/contexts.rs
```diff
@@ -14,8 +14,8 @@
 
 //! The context types that are passed into various widget methods.
 
+use instant::Instant;
```
cmyr, line 17: nit, but generally `std` imports at the top, extern imports below that, internal imports at bottom. 👌

druid-shell/Cargo.toml
```diff
@@ -57,3 +59,9 @@ gtk-sys = "0.9.0"
 [target.'cfg(target_os="linux")'.dependencies.gtk]
 version = "0.8.1"
 features = ["v3_20"]
+
+[target.'cfg(target_arch="wasm32")'.dependencies]
+wasm-bindgen = "0.2.59"
+js-sys = "0.3.36"
+web-sys = { version = "0.3.36", features = ["Window", "MouseEvent", "CssStyleDeclaration", "WheelEvent", "KeyEvent", "KeyboardEvent"] }
+console_log = { version = "0.1.2", optional = true }
```
cmyr, line 67: I'm okay with this being required, if we can make it work transparently with `log` and with our `use_simple_logger` method?

druid-shell/src/platform/web/window.rs
```diff
@@ -0,0 +1,633 @@
...
+use crate::dialog::{FileDialogOptions, FileDialogType, FileInfo};
+
+use crate::keyboard;
+use crate::keycodes::KeyCode;
+use crate::mouse::{Cursor, MouseButton, MouseEvent};
+use crate::window::{IdleToken, Text, TimerToken, WinHandler};
+use crate::KeyModifiers;
+
+use super::util::init_log;
+use log::{error, warn};
```
cmyr, line 43: I'm discouraging use of these macros directly; I think the code is clearer when they are qualified through the module, i.e. `log::error!("Oh no!")` instead of `error!("Oh no")`; the latter looks like it might panic or something.

druid-shell/src/platform/web/window.rs
```diff
@@ -0,0 +1,633 @@
...
+            // Prevent the browser from going back a page by default.
+            event.prevent_default();
+        }
+        let event = keyboard::KeyEvent::new(code, repeat, mods, text.as_str(), text.as_str());
+        state.handler.borrow_mut().key_down(event);
+    });
+}
+
+/// A helper to convert the key string to the text it is supposed to represent.
+fn key_to_text(key: String) -> String {
```
cmyr, line 262: total code golf but I'd do this as,

```rust
fn key_to_text(key: &str) -> &str {
    match key {
    k if k.len() == 1 => k,
    "Enter" => "\n",
    _ => "",
    }
}
```

Also: what about tab?

Also: this length check suppresses all non-ascii input. For instance I can't type é or ô in the hello example. (thanks for providing the examples!)

elrnv: Good find! I was afraid of limiting the output lifetime of the string, but it seems like this isn't problematic for this case.

As for the last two points, we can match through all of https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values and adjust the key "text" appropriately.

elrnv:
All review comments should be addressed in 3813e3c.
I should mention that this PR depends on https://github.com/linebender/piet/pull/153 and a few other piet PRs that have already been accepted but aren't in the release yet.

elrnv:
The last PR https://github.com/linebender/piet/pull/153 was merged, so this PR is just blocked on having another piet release.

cmyr:
@elrnv the piet changes have made it to druid.

elrnv:
@cmyr Thanks! The PR is ready now.

druid/src/lib.rs
```diff
@@ -125,6 +125,7 @@ pub mod lens;
 mod localization;
 mod menu;
 mod mouse;
+#[cfg(not(target_arch = "wasm32"))]
 #[cfg(test)]
 mod tests;
```
Finnerale, line 130: Is there a benefit for making the tests work on wasm?
On a quick glance this seems to test nothing platform specific @cmyr?
If it does tho, there should be an issue opened for it.

druid/examples/game_of_life.rs
```diff
@@ -12,10 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#![allow(clippy::unreadable_literal)]
+
```
Finnerale, line 16: Wow, that's odd. Does not happen without `--target wasm32-unknown-unknown` :unamused: 
I could not find out why clippy does that tho

druid-shell/src/platform/web/window.rs
```diff
@@ -0,0 +1,618 @@
...
+                    handler_borrow.timer(token);
+                }
+            };
+            state
+                .window
+                .set_timeout_with_callback_and_timeout_and_arguments_0(
+                    Closure::once_into_js(f).as_ref().unchecked_ref(),
+                    interval,
+                )
+                .expect("Failed to call setTimeout with a callback");
```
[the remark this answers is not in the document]
elrnv: From what I understand this is true for Closures created with `wrap` or `once`, but `once_into_js` is special in that it does the forgetting for us. Quoting the wasm_bindgen docs for [`once_into_js`](https://docs.rs/wasm-bindgen/0.2.60/wasm_bindgen/closure/struct.Closure.html#method.once_into_js): 

> Unlike Closure::once, this does not return a Closure that can be dropped before the function is invoked to deallocate the closure.

druid/examples/game_of_life.rs
elrnv: I could be wrong but my impression was that the examples were actually not checked by clippy before, but since the `druid-wasm-examples` is now in the workspace, they all get checked with `cargo clippy --all -- -D warnings` with or without `--target wasm32-unknown-unknown`. I remember having to make some other tweaks to the examples to satisfy clippy.

druid/examples/wasm/src/examples.rs
```diff
@@ -0,0 +1,2 @@
+//! This module is intentionally left blank.
+//! The contents are automatically generated in the "build.rs" script.
```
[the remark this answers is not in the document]
elrnv: Ok I readjusted this mechanism to avoid any changes to any committed files in e072555. Would this work?

druid/src/lib.rs
elrnv: I am not sure about this one. I left it out for the time being because it requires changes in `piet` to get this to compile, (I submitted a PR to add the missing types to piet in https://github.com/linebender/piet/pull/167) and I'm not really familiar with the testing code. I suspect it can be used for testing although it needs to be run in a special way for wasm and not through `cargo test`, but I'm not sure. Is this appropriate for a separate PR? For reference there is [some code](https://github.com/linebender/piet/tree/master/piet-web/examples/basic) testing the wasm backend in piet.

Finnerale: Sure, this can be done in a later PR, if it makes sens to do so.
Which is what I hoped @cmyr would enlighten me about.

elrnv:
Thanks for the feedback! I tried to address all of the comments as best as I can. Passing the baton back now :)

elrnv:
Regarding the first bullet, for the time being the `switch` example is renamed to `switch_demo` when building the WASM interface (the actual issue should be fixed in 3a528fa). Perhaps in another PR we could rename the example itself to switch_demo or something like that, which will clean up a bit of code from the WASM build.rs and lib.rs files.

I did a little digging about the second bullet to find the [MDN article on performance.now()]
(https://developer.mozilla.org/en-US/docs/Web/API/Performance/now) which mentions that browsers reduce the precision of time for privacy reasons.

druid/src/lib.rs
cmyr: yea, happy for this to be followup.

PR #760, opened 2020-03-26
Align Flex layout to integers.

xStrom:
This PR is a continuation of work in #669 and makes `Flex` align everything to integers.

There are multiple possible strategies in choosing which children/paddings get the extra pixels. I based my work on how Chrome 80 handles it with CSS flex. If nothing else it will be the least surprising approach to people familiar with web flex layout.

* If `Flex` is given fractional constraints then it will truncate it to integers for child layout purposes. It will still use the untouched fractional constraints to calculate its own size. This is one behavior that diverges somewhat from Chrome 80. Chrome will round its constraints which means that a 10.5px flex container will actually paint 11px. My guiding principles here were that widgets should not paint beyond what they promise and that `Flex` should always provide its children integer constraints. I'm happy for discussion on this behavior though.
* All children of `Flex` will get integer constraints and origin.
* If any child returns a fractional size, `Flex` will expand it to the nearest integer for layout purposes.
* All `Flex` alignment options now align to integers.

The general strategy for padding and child size calculation can be described as *round and carry over the remainder*. It's pretty straightforward and nicely uses up all the space. It tends to distribute the extra pixels in stripe patterns and so rows/columns won't look out of balance as easily.

With the above changes `Flex` becomes a strong ally in achieving a pixel perfect design. The only weakness that I'm aware of right now is when the `Flex` widget's own origin is fractional. With increasingly more of druid's included widgets and tools working in favor of pixel alignment, this issue is likely to only surface if someone writes their own custom layout without care for alignment. That's a scenario for which a future druid book chapter I'm thinking of writing can be of help.

### DIFF
--- a/druid/src/widget/flex.rs
+++ b/druid/src/widget/flex.rs
@@ -544,7 +544,7 @@ impl<T: Data> Widget<T> for Flex<T> {
         let loosened_bc = bc.loosen();
 
         // Measure non-flex children.
-        let mut total_non_flex = 0.0;
+        let mut major_non_flex = 0.0;
         let mut minor = self.direction.minor(bc.min());
         for child in &mut self.children {
             if child.params.flex == 0.0 {
@@ -561,8 +561,8 @@ impl<T: Data> Widget<T> for Flex<T> {
                     log::warn!("A non-Flex child has an infinite height.");
                 }
 
+                major_non_flex += self.direction.major(child_size);
                 minor = minor.max(self.direction.minor(child_size));
-                total_non_flex += self.direction.major(child_size);
                 // Stash size.
                 let rect = Rect::from_origin_size(Point::ORIGIN, child_size);
                 child.widget.set_layout_rect(rect);
@@ -570,9 +570,9 @@ impl<T: Data> Widget<T> for Flex<T> {
         }
 
         let total_major = self.direction.major(bc.max());
-        let remaining = (total_major - total_non_flex).max(0.0);
+        let remaining = (total_major - major_non_flex).max(0.0);
         let flex_sum: f64 = self.children.iter().map(|child| child.params.flex).sum();
-        let mut flex_used: f64 = 0.0;
+        let mut major_flex: f64 = 0.0;
 
         // Measure flex children.
         for child in &mut self.children {
@@ -583,7 +583,7 @@ impl<T: Data> Widget<T> for Flex<T> {
                 let child_bc = self.direction.constraints(&loosened_bc, min_major, major);
                 let child_size = child.widget.layout(layout_ctx, &child_bc, data, env);
 
-                flex_used += self.direction.major(child_size);
+                major_flex += self.direction.major(child_size);
                 minor = minor.max(self.direction.minor(child_size));
                 // Stash size.
                 let rect = Rect::from_origin_size(Point::ORIGIN, child_size);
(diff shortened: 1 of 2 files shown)


xStrom:
I've thought about this a bit as well. It would definitely simplify logic in several places.

There is one more thing to consider, which I just recently started thinking about in the context of druid. That is integer DPI scaling, i.e. a 3840x2160 screen being used with a DPI scale 2.0 as 1920x1080. Perhaps strictly "half-pixels" make sense in such an environment, because a half pixel wide stroke will actually show up physically as a single pixel. I don't really know for sure and haven't done any research/testing of this yet. If that sort of thing is possible then integer constraints would remove that potential. This current PR doesn't support it either. To make things even more interesting, that situation would be dynamic, based on OS DPI settings changing.

Somewhat related is also the question of getting access to actual pixels for some sort of media. This would apply even to fractional DPI scale like 1.25. For example to display an image or do some precise drawing. If I have a 1000x1000 image meant for a 100x100 area that is physically 125x125 then I would rather downscale the image to 125x125 then 100x100 and have it upscaled again. Maybe piet's `draw_image` already does this? I'm not sure, again I haven't tested this.

Generally speaking I think druid should at least make it possible to use the full capability of the display. Probably best to find answers to these questions sooner than later, so future work can be made with the correct foundation.

cmyr:
I was also thinking about this a bit; I know that on macOS, you can do a "half pixel" stroke that is painted as a single device pixel.

Ideally we could have integer bounding boxes but could allow half pixel stroke widths? If you're doing something fancier (like drawing a line) then you aren't forced to truncate.

on mac there's a method `CGRectIntegral` that you're supposed to use when picking layout rects, and this returns integers.

xStrom:
That might be a very reasonable approach. Convert the general layout system to integers, but allow drawing (stroke/fill/image/etc) to operate on floats.

With such a plan, converting `BoxConstraints` to integers seems like the clearest step.

What about `Size` though? An integer `Size` would definitely make sense for all the `BoxConstraints` methods that deal with `Size`. It would also make sense as the return type of `layout` as the flipside of `BoxConstraints`. On the other hand `Rect` would definitely remain float and so the `Rect` to `Size` conversion would require a float `Size`. Maybe there's some benefit in having both, e.g. `Size` and `LayoutSize` or something.

cmyr:
I think for the time being we use the current `Size`, and we just make it easy/natural to use integer values whenever possible.

xStrom:
Okay, I modified `BoxConstraints` so that it always forces itself to integers. I went with expansion for this strategy, so a `1.1x1.1 Size` ends up as `2.0x2.0 Size` in the constraints. This definitely makes sense for `constrain()` and there's value in consistency - in that `BoxConstraints` will now always `expand` its input sizes. The mac `CGRectIntegral` works the same way for what that's worth.

Also if we decide to change this strategy, doing it in `BoxConstraints` means we only have to edit those methods and won't have to dig through all the widget layout code.

druid/src/widget/flex.rs
```diff
@@ -659,54 +657,109 @@ impl CrossAxisAlignment {
...
 
 impl MainAxisAlignment {
-    fn spacing(self, extra: f64, n_children: usize) -> Spacing {
+    /// Given the provided extra space and children count,
+    /// this returns a `Vec<f64>` of spacing,
+    /// where the first element is the spacing before any children
+    /// and all subsequent elements are the spacing after children.
+    ///
+    /// If the extra space is infinite then it returns an empty `Vec<f64>`.
+    fn spacing(self, extra: f64, n_children: usize) -> Vec<f64> {
```
cmyr: An observation: this feels like something that could be very naturally expressed as an iterator. This isn't a blocker or anything, but it feels like it could be a nice win.

xStrom: I'll play around with `Iterator` a bit and see if I can construct code that is more readable.

xStrom: Okay, I pushed the iterator rewrite.

PR #681, opened 2020-03-17
Fix #265 PostQuitMessage when WM_DESTROY in Win

usagi:
Fix to exit the loop normally when the window is closed in Windows.
Related: #265 ( and #674, #438, #395, #362 )

Issue #674, opened 2020-03-16
How can I exit an app if closed window?


usagi:
Expected behavor:

![image](https://user-images.githubusercontent.com/115972/76730667-12bc1b00-679f-11ea-90c6-1664179f9b0e.png)

Actual behavor:

![image](https://user-images.githubusercontent.com/115972/76730921-95dd7100-679f-11ea-95cd-c7b12d944e40.png)

In the other words, how can I exit the `AppLauncher`'s loop?

```rs
  AppLauncher::with_window(window).use_simple_logger().launch(calc_state).expect("launch failed");
```

xStrom:
This is a known issue tracked in #265. Right now the solution is to Ctrl+C the console app.

cmyr:
indeed, this is a pretty easy fix if someone is interested, but I'm not going to get to it for a while.

usagi:
@xStrom Thank you for the guidance on duplication.

@cmyr Okay, I understand your interest. But I and several developers has interested the issue. Thus, I was PRed #681.

cmyr:
@usagi the place to fix this is https://github.com/xi-editor/druid/blob/master/druid/src/win_handler.rs#L223

usagi:
@cmyr Thanks, I'm read around that and implementation of application.rs for each platforms. And try to it:

```
    fn remove_window(&mut self, window_id: WindowId) {
        self.with_delegate(|del, data, env, ctx| del.window_removed(window_id, data, env, ctx));
        // when closing the last window:
        if let Some(mut win) = self.windows.remove(window_id) {
            if self.windows.windows.is_empty() {
                // on mac we need to keep the menu around
                self.root_menu = win.menu.take();
                //FIXME: on windows we need to shutdown the app here?
                Application::quit(); // <-- fixing for the problem!
```

It works and solve the problemfor Windows. And it also works on GTK/Linux without a troubles. But, the method also effect to OSX and I saw your old comments for OSX:

> Yea, thanks for this. It's a slightly tricky one because there are platform-specific expectations; on windows we should kill the application when the main window closes, but on mac we shouldn't.
> 
> There's a separate problem on mac where once the last window is closed the menu stops working. :/

https://github.com/xi-editor/druid/issues/265#issuecomment-550431088

Unfortunately, I have not OSX environment then I cannot fix the issue use the simple method without darty `[cfg(target_of...)]`.

giannissc:
It should be noted that the shello example is closing normally when the button is pressed

[...]

Issue #438, opened 2020-01-01
Process does not exit when window is closed on Windows


Zwo1in:
As in title, I've runned hello world and layout example. In both the process keep running after window is closed and I need to interrupt it in order to exit.
This is the toolchain I use:
```
stable-i686-pc-windows-msvc
rustc 1.40.0 (73528e339 2019-12-16)
```

raphlinus:
This is being tracked as #265. It's slightly tricky because the conventions are platform-specific, but we do plan to fix it as it's certainly annoying.

Issue #395, opened 2019-12-16
error: process didn't exit successfully


Adarma:
Most of the examples don't exit correctly for me on Windows 7 Enteprise SP1 x64.

Many `close gb` messages are printed, but when I close the GUI I have to Ctrl+c to kill the process.

Then I get the following:

    error: process didn't exit successfully: `target\debug\examples\anim.exe` (exit code: 0xc000013a, STATUS_CONTROL_C_EXIT)

shello and perftest exit correctly.

xStrom:
The problems you are seeing exist for everyone, so you're not doing anything wrong.

The exiting is tracked in #265. The `close gb` issue is tracked in [piet #65](https://github.com/linebender/piet/issues/65).

cmyr:
closing as a dupe, #265 tracks this.

PR #763, opened 2020-03-26
Terminate app run loop on Windows when all windows have closed.

### LINKED ISSUE #265: Windows 10 : running examples : closing druid main window doesn't stop cargo.exe
Thank you for the great job you are doing in the field of Rust GUI ...

When running the examples on Windows 10, closing the main window doesn't stop cargo.exe from running. The window get closed, but the process in the back doesn't stop. You have to go to the console and use Ctrl + c to stop the process. Then you receive the following message :
error: process didn't exit successfully: `D:\...\target\debug\examples\multiwin.exe` (exit code: 0xc000013a, STATUS_CONTROL_C_EXIT)
Thanks for advice.

rust version : 1.38.0

xStrom:
This PR solves one of the most reported annoyances of druid (Fixes #265, #362, #395, #438, #674, #681). When all windows have been closed on Windows, the `Application::run` loop now returns execution back to the host app.

### DIFF
--- a/druid-shell/src/application.rs
+++ b/druid-shell/src/application.rs
@@ -14,6 +14,8 @@
 
 //! The top-level application type.
 
+use std::sync::atomic::{AtomicBool, Ordering};
+
 use crate::clipboard::Clipboard;
 use crate::platform::application as platform;
 
@@ -36,14 +38,43 @@ pub trait AppHandler {
     fn command(&mut self, id: u32) {}
 }
 
+/// The top level application state.
+///
+/// This helps the application track all the state that it has created,
+/// which it later needs to clean up.
+#[derive(Clone)]
+pub struct AppState(pub(crate) platform::AppState);
+
+impl AppState {
+    /// Create a new `AppState` instance.
+    pub fn new() -> AppState {
+        AppState(platform::AppState::new())
+    }
+}
+
 //TODO: we may want to make the user create an instance of this (Application::global()?)
 //but for now I'd like to keep changes minimal.
 /// The top level application object.
 pub struct Application(platform::Application);
 
+// Used to ensure only one Application instance is ever created.
+// This may change in the future.
+// For more information see https://github.com/xi-editor/druid/issues/771
+static APPLICATION_CREATED: AtomicBool = AtomicBool::new(false);
+
 impl Application {
-    pub fn new(handler: Option<Box<dyn AppHandler>>) -> Application {
-        Application(platform::Application::new(handler))
+    /// Create a new `Application`.
+    ///
+    /// It takes the application `state` and a `handler` which will be used to inform of events.
+    ///
+    /// Right now only one application can be created. See [druid#771] for discussion.
+    ///
+    /// [druid#771]: https://github.com/xi-editor/druid/issues/771
+    pub fn new(state: AppState, handler: Option<Box<dyn AppHandler>>) -> Application {
+        if APPLICATION_CREATED.compare_and_swap(false, true, Ordering::AcqRel) {
+            panic!("The Application instance has already been created.");
+        }
+        Application(platform::Application::new(state.0, handler))
     }
 
     /// Start the runloop.
--- a/druid-shell/src/application.rs
+++ b/druid-shell/src/application.rs
@@ -14,10 +14,13 @@
 
 //! The top-level application type.
 
+use std::cell::RefCell;
+use std::rc::Rc;
 use std::sync::atomic::{AtomicBool, Ordering};
 
 use crate::clipboard::Clipboard;
 use crate::platform::application as platform;
+use crate::util;
 
 /// A top-level handler that is not associated with any window.
 ///
@@ -38,73 +41,148 @@ pub trait AppHandler {
     fn command(&mut self, id: u32) {}
 }
 
-/// The top level application state.
+/// The top level application object.
 ///
-/// This helps the application track all the state that it has created,
-/// which it later needs to clean up.
+/// This can be thought of as a reference and it can be safely cloned.
 #[derive(Clone)]
-pub struct AppState(pub(crate) platform::AppState);
-
-impl AppState {
-    /// Create a new `AppState` instance.
-    pub fn new() -> AppState {
-        AppState(platform::AppState::new())
-    }
+pub struct Application {
+    state: Rc<RefCell<State>>,
+    pub(crate) platform_app: platform::Application,
 }
 
-//TODO: we may want to make the user create an instance of this (Application::global()?)
-//but for now I'd like to keep changes minimal.
-/// The top level application object.
-pub struct Application(platform::Application);
+/// Platform-independent `Application` state.
+struct State {
+    running: bool,
+}
 
-// Used to ensure only one Application instance is ever created.
-// This may change in the future.
-// For more information see https://github.com/xi-editor/druid/issues/771
+/// Used to ensure only one Application instance is ever created.
 static APPLICATION_CREATED: AtomicBool = AtomicBool::new(false);
+
+thread_local! {
+    /// A reference object to the current `Application`, if any.
+    static GLOBAL_APP: RefCell<Option<Application>> = RefCell::new(None);
+}
 
 impl Application {
     /// Create a new `Application`.
     ///
-    /// It takes the application `state` and a `handler` which will be used to inform of events.
+    /// # Panics
     ///
-    /// Right now only one application can be created. See [druid#771] for discussion.
+    /// Panics if an `Application` has already been created.
+    ///
+    /// This may change in the future. See [druid#771] for discussion.
     ///
     /// [druid#771]: https://github.com/xi-editor/druid/issues/771
-    pub fn new(state: AppState, handler: Option<Box<dyn AppHandler>>) -> Application {
+    pub fn new() -> Application {
         if APPLICATION_CREATED.compare_and_swap(false, true, Ordering::AcqRel) {
             panic!("The Application instance has already been created.");
         }
-        Application(platform::Application::new(state.0, handler))
+        util::claim_main_thread();
+        let state = Rc::new(RefCell::new(State { running: false }));
+        let app = Application {
+            state,
+            platform_app: platform::Application::new(),
+        };
+        GLOBAL_APP.with(|global_app| {
+            *global_app.borrow_mut() = Some(app.clone());
+        });
+        app
     }
 
-    /// Start the runloop.
+    /// Get the current globally active `Application`.
     ///
-    /// This will block the current thread until the program has finished executing.
-    pub fn run(&mut self) {
-        self.0.run()
+    /// A globally active `Application` exists
+    /// after [`new`] is called and until [`run`] returns.
+    ///
+    /// # Panics
+    ///
+    /// Panics if there is no globally active `Application`.
+    /// For a non-panicking function use [`try_global`].
+    ///
+    /// This function will also panic if called from a non-main thread.
+    ///
+    /// [`new`]: #method.new
+    /// [`run`]: #method.run
+    /// [`try_global`]: #method.try_global
+    #[inline]
+    pub fn global() -> Application {
+        // Main thread assertion takes place in try_global()
+        Application::try_global().expect("There is no globally active Application")
     }
 
-    /// Terminate the application.
-    pub fn quit() {
-        platform::Application::quit()
+    /// Get the current globally active `Application`.
+    ///
+    /// A globally active `Application` exists
+    /// after [`new`] is called and until [`run`] returns.
+    ///
+    /// # Panics
+    ///
+    /// Panics if called from a non-main thread.
+    ///
+    /// [`new`]: #method.new
+    /// [`run`]: #method.run
+    pub fn try_global() -> Option<Application> {
+        util::assert_main_thread();
+        GLOBAL_APP.with(|global_app| global_app.borrow().clone())
+    }
+
+    /// Start the `Application` runloop.
+    ///
+    /// The provided `handler` will be used to inform of events.
+    ///
+    /// This will consume the `Application` and block the current thread
+    /// until the `Application` has finished executing.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the `Application` is already running.
+    pub fn run(self, handler: Option<Box<dyn AppHandler>>) {
+        // Make sure this application hasn't run() yet.
+        if let Ok(mut state) = self.state.try_borrow_mut() {
+            if state.running {
+                panic!("Application is already running");
+            }
+            state.running = true;
+        } else {
+            panic!("Application state already borrowed");
+        }
+
+        // Run the platform application
+        self.platform_app.run(handler);
+
+        // This application is no longer active, so clear the global reference
+        GLOBAL_APP.with(|global_app| {
+            *global_app.borrow_mut() = None;
+        });
+        // .. and release the main thread
+        util::release_main_thread();
+    }
+
+    /// Quit the `Application`.
+    ///
+    /// This will cause [`run`] to return control back to the calling function.
+    ///
+    /// [`run`]: #method.run
+    pub fn quit(&self) {
+        self.platform_app.quit()
     }
 
     // TODO: do these two go in some kind of PlatformExt trait?
     /// Hide the application this window belongs to. (cmd+H)
-    pub fn hide() {
+    pub fn hide(&self) {
         #[cfg(target_os = "macos")]
-        platform::Application::hide()
+        self.platform_app.hide()
     }
 
     /// Hide all other applications. (cmd+opt+H)
-    pub fn hide_others() {
+    pub fn hide_others(&self) {
         #[cfg(target_os = "macos")]
-        platform::Application::hide_others()
+        self.platform_app.hide_others()
     }
 
     /// Returns a handle to the system clipboard.
-    pub fn clipboard() -> Clipboard {
-        platform::Application::clipboard().into()
+    pub fn clipboard(&self) -> Clipboard {
+        self.platform_app.clipboard().into()
     }
 
     /// Returns the current locale string.
--- a/druid-shell/src/application.rs
+++ b/druid-shell/src/application.rs
@@ -19,6 +19,7 @@ use std::rc::Rc;
 use std::sync::atomic::{AtomicBool, Ordering};
 
 use crate::clipboard::Clipboard;
+use crate::error::Error;
 use crate::platform::application as platform;
 use crate::util;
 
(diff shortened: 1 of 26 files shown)


druid-shell/src/platform/windows/application.rs
```diff
@@ -49,6 +49,7 @@ impl Application {
                 let mut msg = mem::MaybeUninit::uninit();
                 let res = GetMessageW(msg.as_mut_ptr(), ptr::null_mut(), 0, 0);
                 if res <= 0 {
+                    // TODO: If res is -1 then should handle the error!
```
cmyr: for the time being this might at least be worth logging?

raphlinus:
> I'm a bit curious about the decision to do this entirely in druid-shell, it seems like maybe a very slight mixing of concerns?

I'd also be more inclined to have this in druid, as druid-shell is mostly mechanism, leaving policy to the higher levels, but there is a case to be made for doing it in shell: it's behavior that's platform specific, and doing it this way makes all platforms behave like mac.

xStrom:
I moved the decision to exit the run loop into `druid`. This will also allow us to make this behavior configurable later, for those who don't want to exit on Windows or do want to exit on macOS.

I also fixed the `Application::quit()` Windows implementation to actually work in most cases. The previous simple `PostQuitMessage` implementation assumed that all windows have already been closed and that it will only ever be called from the main thread.

I also made a bit of a style change for the few windows `WM_USER` message names, e.g. `XI_REQUEST_QUIT` -> `DS_REQUEST_QUIT`. The `XI` part seems historic and I went with `DS` for `druid-shell`.

druid/src/win_handler.rs
```diff
@@ -220,7 +224,11 @@ impl<T: Data> Inner<T> {
             if self.windows.windows.is_empty() {
                 // on mac we need to keep the menu around
                 self.root_menu = win.menu.take();
-                //FIXME: on windows we need to shutdown the app here?
+                // If there are even no pending windows, we quit the run loop.
+                if self.windows.count() == 0 {
+                    #[cfg(target_os = "windows")]
+                    Application::quit();
+                }
```
xStrom: This `windows.windows.is_empty()` vs `windows.count() == 0` situation has me wondering a bit still. I think the additional check for pending windows makes sense for exiting, because if there is a pending window it'll very shortly be a real window, right?

I did some minor looking for the macOS menu implementation but couldn't quickly figure out whether or where does the menu get assigned back to the window that is pending right now. Perhaps you can enlighten me a bit @cmyr.

cmyr: The window that is pending will create it's own menu. I don't think it's an issue that this menu sticks around, although it isn't helpful that it does.

This is all about handling the fact that on macOS if you close the last window you still get a menu. In druid we associate menus with windows; this stashes the menu of the last window when closing the last window.

xStrom: Okay, so this `self.root_menu` is never visible if there are actual windows, even if they are minimized/hidden?

cmyr: that is currently the case. I don't think this is the best way to do this long-term, but it was an expedient way to make sure that menus were still available after the last window had been closed on macOS.

druid-shell/src/platform/windows/application.rs
```diff
@@ -17,52 +17,86 @@
...
-                    TranslateMessage(&msg);
-                    DispatchMessageW(&msg);
+                    // We check for DS_REQUEST_QUIT here because thread messages
+                    // will not be forwarded to any window procedures by DispatchMessageW.
+                    if msg.message == DS_REQUEST_QUIT {
+                        if let Some(handler) = &mut self.handler {
+                            // We want to queue up the destruction of all open windows.
+                            // Failure to do so will lead to resource leaks
+                            // and an eventual error code exit for the process.
+                            handler.close_all_windows();
```
cmyr: I wonder if we can invert this?

My naive vision: if you want to quit a druid app in windows, the druid layer is responsible for ensuring that all windows get closed, and _then_ we call into druid-shell and ask it to quit.

The inverse here feels difficult. `quit` in shell should not be the equivalent of a user-initiated quit; it should be a final request to drud-shell to terminate the process. This still feels like druid-shell is implementing policy.

One reason some of this stuff might be a bit weird is that there is a lingering assumption in a lot of this code that there's only ever one window, and a lot of our logic is oriented around that assumption.

druid-shell/src/platform/windows/application.rs
```diff
@@ -17,52 +17,86 @@
...
                 let mut msg: MSG = msg.assume_init();
                 let accels = accels::find_accels(GetAncestor(msg.hwnd, GA_ROOT));
                 let translated = accels.map_or(false, |it| {
                     TranslateAcceleratorW(msg.hwnd, it.handle(), &mut msg) != 0
                 });
-
                 if !translated {
-                    TranslateMessage(&msg);
-                    DispatchMessageW(&msg);
+                    // We check for DS_REQUEST_QUIT here because thread messages
```
raphlinus: I'm uneasy about adding more logic in the runloop, as we're not always in control. When there's a modal dialog, or when the window is being resized, we're at the mercy of the system's runloop. This is why I prefer logic to be in the wndproc.

I'm not sure whether that's a serious concern here, but wanted to indicate my thinking.

druid-shell/src/platform/windows/application.rs
```diff
@@ -17,52 +17,86 @@
...
 impl Application {
-    pub fn new(_handler: Option<Box<dyn AppHandler>>) -> Application {
+    pub fn new(handler: Option<Box<dyn AppHandler>>) -> Application {
         Application::init();
-        Application
+        Application { handler }
     }
 
     pub fn run(&mut self) {
+        claim_main_thread();
```
raphlinus: I'm a little unclear what the motivation for this is; it seems orthogonal to quit logic. I'm also concerned about safety implications if the main thread is changed; right now we don't require `Sync` or `Send` on the wrapped wndproc, but if the wndproc can be called from a different thread than the one it was created on, it could violate soundness.

druid-shell/src/platform/windows/application.rs
```diff
@@ -17,52 +17,86 @@
...
 use winapi::um::wingdi::CreateSolidBrush;
 use winapi::um::winuser::{
-    DispatchMessageW, GetAncestor, GetMessageW, LoadIconW, PostQuitMessage, RegisterClassW,
-    TranslateAcceleratorW, TranslateMessage, GA_ROOT, IDI_APPLICATION, MSG, WNDCLASSW,
+    DispatchMessageW, GetAncestor, GetMessageW, LoadIconW, PostQuitMessage, PostThreadMessageW,
+    RegisterClassW, TranslateAcceleratorW, TranslateMessage, GA_ROOT, IDI_APPLICATION, MSG,
+    WNDCLASSW,
 };
 
+use log::{error, warn};
```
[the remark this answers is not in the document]
xStrom: What do you mean? As separate imports or no imports and just call them with fully qualified names?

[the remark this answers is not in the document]
cmyr: I mean qualified names.

druid-shell/src/platform/windows/application.rs
xStrom: You mean `window::win_proc_dispatch`? If so, then that only handles messages sent to a window. This is a message that's sent to the thread (i.e. the app) and it doesn't get delivered by `DispatchMessageW` to any window procedure.

raphlinus: Ah, right. So what will happen if this is sent during live resize or when a modal dialog is open? My guess would be dropped on the floor.

druid-shell/src/platform/windows/application.rs
xStrom: The motivation is to know the thread id where to send `DS_REQUEST_QUIT` (or even `PostQuitMessage` for completeness). The `Application::quit()` function is just a static associated function and operates on global state.

raphlinus: Ah, I see. I was confused by the naming; I thought it was changing what Windows considered the main thread, but on re-read I see that isn't the case.

druid-shell/src/platform/windows/application.rs
raphlinus: Answering my own question, the [Microsoft docs](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-postthreadmessagea) confirm this:

> Messages sent by PostThreadMessage are not associated with a window. As a general rule, messages that are not associated with a window cannot be dispatched by the DispatchMessage function. Therefore, if the recipient thread is in a modal loop (as used by MessageBox or DialogBox), the messages will be lost. To intercept thread messages while in a modal loop, use a thread-specific hook.

I'm uneasy about the complexity and fragility of posting messages to threads, but don't at the moment have a better idea.

druid-shell/src/platform/windows/application.rs
xStrom: I think the `quit` in shell should work without issues. Calling it should never lead to resource leaks or errors. If we move the window closing decision to `druid` then all other users of `druid-shell` will either have to read the documentation and be sure to do the same, or they will have a resource leak by just using the `druid-shell` `quit`.

druid-shell/src/application.rs
```diff
@@ -24,13 +24,16 @@ use crate::platform::application as platform;
...
-/// no window is open.
+/// This is currently limited in its functionality,
+/// having the ability to close all windows
+/// and handling menu commands when no window is open.
 ///
 /// It is possible that this will expand to cover additional functionality
 /// in the future.
 pub trait AppHandler {
+    /// Closes all the windows.
+    fn close_all_windows(&mut self) {}
```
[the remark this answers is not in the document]
xStrom: This function is meant to just close all windows no questions asked. I agree that it's confusing, I also found a bunch of the existing closing related code confusing. I think we should introduce some naming conventions to differentiate these things. Perhaps `close` for the path that works like the user has pressed the titlebar corner X button and `destory` for the lower path that won't ask any questions and is determined to actually get rid of the window.

druid-shell/src/platform/windows/application.rs
xStrom: Indeed, it's a problem. I guess the next best thing is to use our own flag. :thinking:

druid-shell/src/platform/windows/application.rs
xStrom: Yeah I just mimicked what the other platforms do with `util::assert_main_thread()`.

druid-shell/src/platform/windows/application.rs
xStrom: Perhaps a [message-only window](https://devblogs.microsoft.com/oldnewthing/20171218-00/?p=97595) could be the solution. I'm assuming that still receives its messages while a modal loop is active. I'll look deeper into it and try out some code.

raphlinus: Yeah, I think the concept of message-only window has come up in other contexts (maybe for system tray apps). I definitely support looking into it. I'd even (personally) be fine with merging this as an interim solution and having an issue open for migration to message-only.

druid-shell/src/platform/windows/application.rs
cmyr: It depends what the tradeoffs are.

Basically: if the tradeoff is that by making druid-shell be more difficult to use correctly, we allow druid to have a clearer and more intuitive API, it might be worth it.

`druid-shell` is hairy. It is 100% possible to misuse. In many ways, it makes sense to think of it as a C API.

If the documentation of `quit` clearly states that it should only be called when all windows are closed, I think that's reasonable. A compromise might be to have some mechanism in `druid-shell` to count windows, and then make `quit` return `Result<_>`, with an error if windows remain open; and an even better compromise might be to have the `Application` type become fuller featured, and include handles to open windows etc. But having us call `quit` and then it calling us back and asking us to close our windows feels like it is equally compromised, and it isn't clear to me that it offers any significant benefit.

druid-shell/src/application.rs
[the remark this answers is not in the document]
cmyr: I think this convention makes sense, and I agree that the existing closing related code is confusing.

druid-shell/src/platform/windows/application.rs
xStrom: I'll play around with this a bit and see if I can perform the cleanup in another way without calling back to `druid`. The solution in this PR right now was just the quickest way to do it, as druid was already doing that bookkeeping. However I agree it's a bit spaghetti.

Now more generally, the `AppHandler` is passed to `Application` and right now only used for mac root menus. What kind of features do you think would roughly fit `AppHandler`? It's only calling back to `druid` right, but if we don't want to do that, when is it correct to use it? For some sort of event passing that isn't associated with any window? Cases where `druid-shell` is informing and doesn't depend on anyone listening or behaving in any specific way?

cmyr: So the genesis here: up until recently, druid-shell was entirely window based. That is, all events were associated with a window, and were delivered to a `WinHandler`.

This means that if you didn't have any open windows, we had no way to deliver events. In many cases this isn't a huge problem; but it really is on mac, which expects the application to remain open even without any windows.

So `AppHandler` is an attempt to introduce some type 'above' `WinHandler` that receives events at the 'application scope'.

Currently, yes, this is really just doing menus on mac. I could definitely imagine it having other roles; in particular it might make sense that the `AppHandler` gets notified when windows are added or removed (there is currently an `AppDelegate` for this, which predates `AppHandler`, and could perhaps be removed) and it may also make sense for `AppHandler` to receive certain system level events where it doesn't make sense to deliver them to a particular window. I can't think of great examples right now, but one might be if your app is registered to be able to open some type of file, and the user double clicks one of those files; we might notify your `AppHandler` at launch? 

In any case this didn't get a ton of long-term planning; it felt like something we would eventually need if we wanted to support windowless applications, and it was something that I clearly needed for runebender on mac.

xStrom:
I rewrote the implementation. Now druid-shell is self-sufficient in its capability of correctly quitting on Windows. Druid simply calls `Application::quit` when there are no more open windows.

I added a new `AppState` struct to druid-shell which is given to `Application::new` and `WindowBuilder::new`. On Windows this allows druid-shell to actually know what it has done so that it can clean it all up.

I also introduced a message-only window on Windows. This can be considered the application message loop for druid-shell purposes. It is created first, dies last, and is invisible to druid. This apporach also makes it all work even during a modal loop.

xStrom:
I think you're right, having the `Application` state originate from inside of itself makes more sense.

I have pushed yet another new iteration of this work. I removed the message window from the Windows platform code and perform the quit work inside of `Application` now.

I also changed the macOS `Application` so that `quit` makes `run` return control back to the Rust application instead of killing the process. Also added some additional cleanup code that helps ensure `Drop` code is actually run.

The GTK/X11 platform code can be improved thanks to the new `Application` structure as well. I don't want this PR to balloon too much so I'm leaving that work for another PR.

druid-shell/src/application.rs
```diff
@@ -36,44 +41,148 @@ pub trait AppHandler {
...
+    /// Create a new `Application`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if an `Application` has already been created.
+    ///
+    /// This may change in the future. See [druid#771] for discussion.
+    ///
+    /// [druid#771]: https://github.com/xi-editor/druid/issues/771
+    pub fn new() -> Application {
```
cmyr, line 76: I'm trying to think a bit about what the API should look like in this new world.

Some basic questions:

- should we require *explicit* initialization, or should we just lazily initialize in `global`?
- if we _do_ require explicit initialization, should that be a method like `new` that returns the application, or should it be a method like `init` that returns `Result<(), SomeError>`?, and then the actual application is always retrieved via `global`?
- Should `Application` be passed by value, or by reference? Having `global` return `&'static Application' (and having `Application: !Clone` perhaps better expresses the semantics of the program, where there is a single shared application).

druid-shell/src/platform/mac/application.rs
```diff
@@ -32,57 +34,91 @@ use crate::application::AppHandler;
...
 impl Application {
-    pub fn new(handler: Option<Box<dyn AppHandler>>) -> Application {
+    pub fn new() -> Application {
+        // macOS demands that we run not just on one thread,
+        // but specifically the first thread of the app.
         util::assert_main_thread();
         unsafe {
             let _pool = NSAutoreleasePool::new(nil);
 
-            let delegate: id = msg_send![APP_DELEGATE.0, alloc];
```
cmyr, line 45: this stuff was very carefully organized after much trial and error, and based on a close reading of how appkit starts an application; I would leave it as is unless there's a clear problem.


edit: I think I see your rationale; you would like `run` to be symmetric. Do we imagine a world where `run` can be called multiple times? Wouldn't you need to call `new` between calls to `quit` and `run` in this case, anyway? I think this probably should be in the `run` docs, in any case.

druid-shell/src/application.rs
ratmice: About your second bullet (In particular with gtk and perhaps macOS)

I think returning a `Result` might be good for cases where we are dealing with a remote application that was previously launched.  E.g. In the MacOS case there is `NSRunningApplication`, and in GTK the application is less functional, cannot display windows, but can send signals such as to open a new window to the remote application.

Being able to differentiate these via `Result<ApplicationInitializationStatus, SomeError>` or such might work for exposing this behavior on both platforms?

druid-shell/src/platform/mac/application.rs
xStrom: The delegate was previously set after creating the app but before running it - and that is still the case now. I just moved it from `new` to `run`. There's no actual change to the order of events though. On my macOS testing it seems to work fine, including the `applicationDidFinishLaunching` event.

I can imagine `run` being called twice by accident, but that will panic on purpose. The check is done in the platform-independent `Application::run` which also documents this. Also yes, you can't call `run` on the same exact copy of `Application` because `run` consumes `self`. That's one safe-guard. However because `Application` is clonable someone might for some bizarre reason call `run` on that clone elsewhere. That's why there's a check for it which will panic.

druid-shell/src/application.rs
xStrom: I'll play around a bit with the `&'static Application` idea and see if and how well I can get it working.

xStrom: Okay I ran into a dead end with `&'static Application` but opened [a thread on zulip](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/druid.23763.20Application.20exit) for it.

Regarding initialization, I favor explicit because it gives clear control over the timing of when this significant work takes place, and if we plan on returning errors here in the future instead of panicking - then also a single place to handle those errors. The `global` function would always be fast and there wouldn't be a need to reason about whether one should check for initialization errors at every call location.

Naming wise, `new` vs `init`, I think `new` is better if we don't want to start painting ourselves into a strategy corner in terms of #771. That's because I think it makes more sense to have `new` if this function is going to be called more than once.

Return value wise I think changing it to `fn new() -> Result<Application, Error>` would be a good upgrade. This would allow the calling app to choose whether to panic or to do something else. Right now `druid-shell` is a bit too panic friendly in its initialization code.

xStrom: I pushed a new commit containing the `Result<Application, Error>` change.

Issue #723, opened 2020-03-20
split `druid` derive attribute into `data` and `lens`


cmyr:
Currently we have a single `druid` attribute in our derive macros. Eventually (I suspect) we are going to want more flexibility here, and I think it makes sense to break this up into separate `lens` and `data` attributes, aka

```rust
#[derive(Clone, Data, Lens)]
struct Before {
    #[druid(same_fn = "PartialEq::eq", lens_name = "file_path")]
    path: PathBuf,
}

#[derive(Clone, Data, Lens)]
struct After {
    #[data(same_fn = "PartialEq::eq")]
    #[lens(name = "file_count")]
    path: PathBuf,
}
```

It would also be nice if we still checked for the `druid` attribute and printed a helpful failure message.

This should bump `druid-derive` to the next minor version.

Dmitry-Borodin:
@cmyr I think this can be closed

cmyr:
Thanks!

PR #765, opened 2020-03-28
Split `druid` derive attribute into `data` and `lens`

Dmitry-Borodin:
Implemented #723 

It will not compile with old druid attribute.

Note, ignore is parsed for both (never used for lens and not parsed), but saved to the same variable. 
Should I use ignore attribute for lens? Or Remove possibility to use it? Removed of now.

I'm thinking whether we should have single Field struct for attributes, or have separate struct for each base attribute, which probably will have some duplicated code.

And I would like to add documentation of which attributes are supported, it's not clear now. But I don't see a good place where to do it.

### DIFF
--- a/druid-derive/src/attr.rs
+++ b/druid-derive/src/attr.rs
@@ -111,7 +111,7 @@ impl Field {
 
         for attr in field.attrs.iter() {
             if attr.path.is_ident(BASE_DRUID_DEPRECATED_ATTR_PATH) {
-                panic(
+                panic!(
                     "The 'druid' attribute has been replaced with separate \
                     'lens' and 'data' attributes.",
                 );
--- a/druid-derive/src/attr.rs
+++ b/druid-derive/src/attr.rs
@@ -111,11 +111,10 @@ impl Field {
 
         for attr in field.attrs.iter() {
             if attr.path.is_ident(BASE_DRUID_DEPRECATED_ATTR_PATH) {
-                return Err(Error::new(
-                    attr.path.span(),
+                panic(
                     "The 'druid' attribute has been replaced with separate \
                     'lens' and 'data' attributes.",
-                ));
+                );
             } else if attr.path.is_ident(BASE_DATA_ATTR_PATH) {
                 match attr.parse_meta()? {
                     Meta::List(meta) => {
--- a/druid-derive/src/attr.rs
+++ b/druid-derive/src/attr.rs
@@ -154,17 +154,6 @@ impl Field {
                     Meta::List(meta) => {
                         for nested in meta.nested.iter() {
                             match nested {
-                                NestedMeta::Meta(Meta::Path(path))
-                                    if path.is_ident(IGNORE_ATTR_PATH) =>
-                                {
-                                    if ignore {
-                                        return Err(Error::new(
-                                            nested.span(),
-                                            "Duplicate attribute",
-                                        ));
-                                    }
-                                    ignore = true;
-                                }
                                 NestedMeta::Meta(Meta::NameValue(meta))
                                     if meta.path.is_ident(LENS_NAME_OVERRIDE_ATTR_PATH) =>
                                 {
--- a/druid-derive/src/attr.rs
+++ b/druid-derive/src/attr.rs
@@ -111,7 +111,11 @@ impl Field {
 
         for attr in field.attrs.iter() {
             if attr.path.is_ident(BASE_DRUID_DEPRECATED_ATTR_PATH) {
-                panic!("druid attribute is deprecated, please update it, it won't work")
+                return Err(Error::new(
+                    attr.path.span(),
+                    "The 'druid' attribute has been replaced with separate \
+                    'lens' and 'data' attributes.",
+                ));
             } else if attr.path.is_ident(BASE_DATA_ATTR_PATH) {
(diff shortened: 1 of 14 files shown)


druid-derive/src/lib.rs
```diff
@@ -24,7 +24,7 @@ use proc_macro::TokenStream;
 use syn::parse_macro_input;
 
 /// Generates implementations of the `Data` trait.
-#[proc_macro_derive(Data, attributes(druid))]
```
cmyr: is there a way we can keep `druid` attribute, and print a helpful error if we see it?

druid-derive/src/attr.rs
```diff
@@ -105,52 +109,71 @@ impl Field {
...
-                    for nested in meta.nested.iter() {
-                        match nested {
-                            NestedMeta::Meta(Meta::Path(path))
-                                if path.is_ident(IGNORE_ATTR_PATH) =>
-                            {
-                                if ignore {
-                                    return Err(Error::new(nested.span(), "Duplicate attribute"));
+        for attr in field.attrs.iter() {
+            if attr.path.is_ident(BASE_DRUID_DEPRECATED_ATTR_PATH) {
+                panic!("druid attribute is deprecated, please update it, it won't work")
```
cmyr: I'm not sure this will work if we haven't annotated the derives as accepting this attribute?

I'm also not sure `panic` is best, ideally we could return an error that displays over the span?

Dmitry-Borodin: Tested - at compile time this panic is shown as compilation error.
I'm thinking about splitting attribute parsing logic per derive(right now for derive Data we will parse for lens attribute as well, and then ignore it during code generation). We can write here more more explanations with examples if splitted.
We can make it warning, but I think for this case compile error is better.

cmyr: it does panic, but it doesn't show the correct span; it just shows the `Lens` derive. I was trying to describe something like this:
```suggestion
                return Err(Error::new(
                    attr.path.span(),
                    "The 'druid' attribute has been replaced with separate \
                    'lens' and 'data' attributes.",
                ));
```

cmyr: although I also agree this seems to let the main crate still try to compile, which creates more error messages than we need. I'll try and play around a bit more..

cmyr: okay actually I take this back, I think your solution is our best bet, until something like https://github.com/rust-lang/rust/issues/54140 lands.

Dmitry-Borodin: With this we have clear error messages with links to to place that should be fixed, I like it even more, didn't know that I can pass path to error to see it in a stacktrace.
I'm more concerned about decoupling attribute parsing for different derive attributes. But I can work on it in a separate branch.

druid-derive/src/lib.rs
Dmitry-Borodin: We still printing error without it. This will additionally show error that attribute is not supported. Would you like to add it here to expected attributes, and only show error during parsing?

druid-derive/src/attr.rs
Dmitry-Borodin: I have bad internet now days, didn't see new messages.

Dmitry-Borodin:
Can we restart tests? Looks like some azure problem.

xStrom:
Tried rerunning, Azure still having issues. Will have to rerun again later.

druid-derive/src/lib.rs
cmyr: In a perfect world we could print a compiler warning but still function if `druid` was used. I think this is maybe hard to do, though, so I'm happy with this how it is.

druid-derive/src/attr.rs
```diff
@@ -105,52 +109,75 @@ impl Field {
...
-                            {
-                                if ignore {
-                                    return Err(Error::new(nested.span(), "Duplicate attribute"));
+        for attr in field.attrs.iter() {
+            if attr.path.is_ident(BASE_DRUID_DEPRECATED_ATTR_PATH) {
+                return Err(Error::new(
+                    attr.path.span(),
+                    "The 'druid' attribute has been replaced with separate \
+                    'lens' and 'data' attributes.",
+                ));
```
cmyr: okay I played around with this a bunch and I think your version was better, where we just panic; returning an error causes compilation to continue, and we get a whole bunch of subsequent error messages. :(

```suggestion
                panic(
                    "The 'druid' attribute has been replaced with separate \
                    'lens' and 'data' attributes.",
                );
```

druid-derive/src/attr.rs
```diff
@@ -105,52 +109,74 @@ impl Field {
...
-                    for nested in meta.nested.iter() {
-                        match nested {
-                            NestedMeta::Meta(Meta::Path(path))
-                                if path.is_ident(IGNORE_ATTR_PATH) =>
-                            {
-                                if ignore {
-                                    return Err(Error::new(nested.span(), "Duplicate attribute"));
+        for attr in field.attrs.iter() {
+            if attr.path.is_ident(BASE_DRUID_DEPRECATED_ATTR_PATH) {
+                panic(
```
cmyr: Oops 🙄 

```suggestion
                panic!(
```

PR #766, opened 2020-03-29
Implement WindowBuilder::set_min_size for MacOS.

mobile-bungalow:
implements #714.

### DIFF
--- a/druid-shell/src/platform/mac/window.rs
+++ b/druid-shell/src/platform/mac/window.rs
@@ -123,7 +123,6 @@ impl WindowBuilder {
     }
 
     pub fn set_min_size(&mut self, size: Size) {
-        // TODO: Use this in `self.build`
         self.min_size = Some(size);
     }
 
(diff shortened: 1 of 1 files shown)


druid-shell/src/platform/mac/window.rs
```diff
@@ -167,6 +166,16 @@ impl WindowBuilder {
                 NO,
             );
 
+            if let Some(min_size) = self.min_size {
+                let frame = NSWindow::frame(window);
+                let content_height = window.contentRectForFrameRect_(frame).size.height;
+                let window_height = frame.size.height;
+                let title_bar_height = window_height - content_height;
```
xStrom: Is `height` the only thing that ever has extra space with any style? I don't know much about macOS but it feels to me that if the function returns the `width` as well, we should calculate the `width` difference too.

cmyr: I can't think of an instance where `width` would be meaningful, but I'm also not sure it *doesn't* exist, so using it this does seem more likely to be Correct.

xStrom: Yeah, using it doesn't seem too complex. Also it would future proof us against future style changes.

mobile-bungalow: Yeah, in the case that apple adds bottom and side padding back to windows, like was common in Aqua, this would be a source of bugs. good catch, i'll add that in.

druid-shell/src/platform/mac/window.rs
```diff
@@ -168,11 +168,16 @@ impl WindowBuilder {
...
+                let excess_height = frame.size.height - content_dimensions.height;
+                let excess_width = frame.size.width - content_dimensions.width;
+
+                // setMinSize_ restricts the size of the window including padding.
+                let size = NSSize::new(
+                    min_size.width + excess_width,
+                    min_size.height + excess_height,
+                );
+
                 window.setMinSize_(size);
```
mobile-bungalow: Looking at the NSWindow docs there is actually a method for what I am trying to achieve here. SetContentMinSize_.

Issue #754, opened 2020-03-23
Missing request_paint somewhere.


totsteps:
I am not sure how to title this issue, but when incrementing or decrementing the menu item value via the context menu in `multi-win` example, the custom menu in the menu-bar gets updated instantly but there's a slight random delay when updating the label value.

**Tested on macOS.**

An example and discussion can be found in [zulip chat](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/druid-chatter/near/191421810).

As noted by Colin Rofls this is missing `request_paint` or similar somewhere.

emigr2k1:
After clicking on the context menu item, `AppState::handle_system_cmd` gets called but this function doesn't call `do_update` as other functions such as `do_window_event` do.
I'm not familiar with druid's internals so I don't know if calling `do_update` in `handle_system_cmd` is a good fix.

totsteps:
@emigr2k1 thanks for the PR.
I can confirm that this has been fixed.

This issue can be closed unless @cmyr has some other thoughts.

PR #767, opened 2020-03-29
Call update on system commands

emigr2k1:
When a system command is handled, widgets don't get updated even if the command changed the data.
I'm not sure if this is the best place to do an update.
Related issue #754


cmyr:
good find, we should be calling `self.inner.borrow_mut().do_update();` at the bottom of `handle_system_cmd` in win_handler.rs.

PR #769, opened 2020-03-29
Default to showing both scrollbars in the scroll example.

xStrom:
When testing #709 I needed both scroll directions. I think this will be a useful addition to the scroll example in general.

### DIFF
--- a/druid/examples/scroll.rs
+++ b/druid/examples/scroll.rs
@@ -33,7 +33,11 @@ fn main() {
 fn build_widget() -> impl Widget<u32> {
     let mut col = Flex::column();
     for i in 0..30 {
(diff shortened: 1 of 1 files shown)


xStrom:
Aha, I didn't even see `scroll_colors`. Now I'm thinking that perhaps these two should be combined into one example, with some sort of buttons switching between what is being drawn.

As far as this PR here, I'm fine with not merging now that I know about `scroll_colors`.

What do you think about this dual scroll example situation @cmyr?

cmyr:
I think that `scroll_colors` is the real scroll example, and that `scroll` at this point is mostly just an example of painting outside your layout bounds. Not sure if that deserves to be kept around?

xStrom:
I'll close this because `scroll_colors` already does what this PR offers.

PR #772, opened 2020-03-29
Turn off event compression for gtk.

jneem:
I don't actually know if this is the right thing to do, but I figured it was at least worth bringing up.

By default, gtk compresses mouse motion events so that it delivers at most one per frame. This is causing issues for the toy drawing program that I'm writing, because if I want smooth lines then I need mouse events at the highest possible temporal resolution.

What's the behavior of other backends? If they all compress events, then could there be a knob for turning it off? If they don't compress events, probably gtk shouldn't either.

### DIFF
--- a/druid-shell/src/platform/gtk/window.rs
+++ b/druid-shell/src/platform/gtk/window.rs
@@ -423,6 +423,11 @@ impl WindowBuilder {
         }));
 
         vbox.pack_end(&drawing_area, true, true, 0);
+        drawing_area.realize();
+        drawing_area
+            .get_window()
+            .expect("realize didn't create window")
+            .set_event_compression(false);
 
         win_state
             .handler


cmyr:
I think the behaviour of getting fine grained events is correct by default.

Someone who is using/more familiar with gtk should review this, though, will ask on zulip.

raphlinus:
Without being a (current) gtk expert, I'm pretty sure that turning off event compression is the right thing. My feeling is that we should basically count on the event handler being fast, and do aggregation at other points in the pipeline (we already do for drawing, and could to so for updates as well, if profiling showed the need).

fishrockz:
I have verified that this dose not seem to brake anything, I have had a go at game of life and calc. But I don't know how to recreate the original problematic behaviour.

+1 from me

PR #773, opened 2020-03-29
Add resizable option to mac window

emigr2k1:
Issue #633

This adds support for `resizable` option on mac.

### DIFF
--- a/druid-shell/src/platform/mac/window.rs
+++ b/druid-shell/src/platform/mac/window.rs
@@ -689,8 +689,20 @@ impl WindowHandle {
     // TODO: Implement this
     pub fn show_titlebar(&self, _show_titlebar: bool) {}
 
-    // TODO: Implement this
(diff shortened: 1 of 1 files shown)


druid-shell/src/platform/mac/window.rs
```diff
@@ -689,8 +689,20 @@ impl WindowHandle {
     // TODO: Implement this
     pub fn show_titlebar(&self, _show_titlebar: bool) {}
 
-    // TODO: Implement this
-    pub fn resizable(&self, _resizable: bool) {}
```
cmyr: Hmm must not have noticed this, but these should be called `set_`.  We should also revisit whether or not these should be marked as `&mut`; they are clearly mutating but I'm not sure how much we expose that to druid.

In any case don't worry for this PR, both these things would be future work.

It would also be nice if we had some way to test this; for now I'll just say it looks good when I skim it. 😁

PR #775, opened 2020-03-30
Use glib::idle_add instead of rolling our own.

### LINKED ISSUE #595: clean up gtk ext event handling mess
in #550 we copy-pasted some code to work around gtk-rs/gdk#304; when that issue is resolved we should figure out what the longer-term solution is supposed to be, and clean up that mess.

jneem:
Fixes #595.

This is the solution suggested in gtk-rs/gdk#304


cmyr:
Cool, thanks for this; cleanup is always appreciated.

fishrockz:
I have checked this out and run game of life, calc, and flex with no issues, But im not sure which examples would exercise this code path.

PR #779, opened 2020-03-30
Update various docs for 0.5.0

cmyr:


futurepaul:
Looks harmless and the book / docs still build! Sucks we can't test doc urls until everything is pushed.

cmyr:
yea, I'll hold off on this until we're going to pull the trigger, but should be shortly...

PR #780, opened 2020-03-30
is_focused/has_focus for EventCtx and PaintCtx

yrns:
I'd like to propose the ability to discern between a widget being the leaf focus widget and being an ancestor of the focused widget in both PaintCtx and EventCtx. Right now, only PaintCtx has the former, and EventCtx the latter. This changes the meaning of the method in PaintCtx but doesn't affect any of the existing widgets - only TextBox uses it, and it doesn't have any child widgets.

### DIFF
--- a/druid/src/contexts.rs
+++ b/druid/src/contexts.rs
@@ -262,6 +262,13 @@ impl<'a> EventCtx<'a> {
         is_child || self.focus_widget == Some(self.widget_id())
     }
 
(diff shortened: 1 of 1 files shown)


yrns:
Thanks for reviewing. I updated `is_focused` in `PaintCtx` to match. Do I need to squash?

xStrom:
No requirement to squash and we can squash before merging anyway.

PR #782, opened 2020-03-30
Reduce the flashing in ext_event and identity examples

futurepaul:
This is pretty hacky, let me know if you have a better idea! My main idea is to move only slightly toward the new random color. The result is the color stays pretty subtle, but I didn't want to introduce more state.

### DIFF
--- a/druid/examples/identity.rs
+++ b/druid/examples/identity.rs
@@ -53,23 +53,13 @@ fn color_eq(one: &Color, two: &Color) -> bool {
     one.as_rgba_u32() == two.as_rgba_u32()
 }
 
-fn split_rgba(rgba: &Color) -> (f64, f64, f64, f64) {
+fn split_rgba(rgba: &Color) -> (u8, u8, u8, u8) {
     let rgba = rgba.as_rgba_u32();
     (
-        (rgba >> 24) as f64 / 255.0,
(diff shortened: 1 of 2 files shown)


druid/examples/identity.rs
```diff
@@ -53,6 +53,26 @@ fn color_eq(one: &Color, two: &Color) -> bool {
...
+    let rgba = rgba.as_rgba_u32();
+    (
+        (rgba >> 24) as f64 / 255.0,
+        ((rgba >> 16) & 255) as f64 / 255.0,
+        ((rgba >> 8) & 255) as f64 / 255.0,
+        (rgba & 255) as f64 / 255.0,
+    )
+}
+
+fn color_average(one: &Color, two: &Color) -> Color {
```
fishrockz: This is a low pass filter rather than a average, I would suggest naming this appropriately

fishrockz:
I get @cmyr 's resivations and I expected a change to just turn the frequency down too. But the effect of this is actually quite nice as you still get the immediate change and you still get a reasonable range of colors, if anything i think the filter could let through slightly higher frequency change.

futurepaul:
Okay this is a new approach that gets rid of all the old random color logic and just mutates the color randomly. Hits a wider range of colors, still is low on flashing (my primary concern). I'll merge this later today if there are no objections.

PR #783, opened 2020-03-30
Log size warnings for Align

futurepaul:
I was inspired by #781 to get this into Align as well, which I missed in the first pass.


druid/src/widget/align.rs
```diff
@@ -137,3 +140,14 @@ impl<T: Data> Widget<T> for Align<T> {
         self.child.paint_with_offset(ctx, data, env);
     }
 }
+
+fn log_size_warnings(size: Size) {
+    if size.width.is_infinite() {
+        println!("what");
```
xStrom: :exclamation:

futurepaul: that's embarrassing! thank you for catching

Issue #774, opened 2020-03-29
Configurable label font


thecodewarrior:
It would be great if labels could have configurable fonts. One of my projects needs labels with monospaced text, but I don't want all the text in the UI to be monospaced. From what I can tell this isn't possible right now since at the moment Label's [`get_layout`](https://github.com/xi-editor/druid/blob/ddafe0dfb8cf6f03a735dc80d853e8c89167710d/druid/src/widget/label.rs#L166-L173) gets the font name from the theme. Probably the most elegant solution would be an `Option<String>` with the theme being used as the default if the font is `None`.

cmyr:
I agree this makes sense, and a PR would be welcome; font should work the same as color and size.

cc @futurepaul who has been thinking about this lately.

futurepaul:
I haven't tried anything out yet but ideally you can follow the example of the builder / setter pair we have for size and color. That way Label::new sets the font as `theme::FONT_NAME` and then the builder or setter updates that value, no need for an option. I'm curious if we can do KeyOrValue here as well, but I haven't used that yet with a String so no guarantees.

thecodewarrior:
`theme::FONT_NAME` is the key for the environment, not the actual font name. If we don't use an `Option` then we either store the env key, meaning we couldn't directly set the font name, or directly store the font name and grab the theme font name once during creation, neither of which are good ideas IMO.

PR #785, opened 2020-03-30
Add option to set label fonts

thecodewarrior:
Issue #774 

This adds an option to `Label` to configure which font it uses. It's a simple implementation using a `KeyOrValue<&'static str>`. While it's likely possible to adjust the system to not need the cumbersome `&'static str` (which also doesn't allow dynamic strings), I have neither the Rust experience nor the depth of understanding of the current system necessary to know how to do that elegantly.


druid/src/widget/label.rs
```diff
@@ -132,6 +134,16 @@ impl<T: Data> Label<T> {
         self
     }
 
+    /// Builder-style method for setting the font.
+    ///
+    /// The argument can be either a `&'static str` or a [`Key<&'static str>`].
+    ///
+    /// [`Key<&str>`]: struct.Key.html
```
xStrom: Does ```[`Key<&str>`]``` apply the link to ```[`Key<&'static str>`]```? I thought it needs to match. Also the link is incorrect and should be `../struct.Key.html`

druid/src/widget/label.rs
```diff
@@ -163,8 +175,17 @@ impl<T: Data> Label<T> {
         self.size = size.into();
     }
 
+    /// Set the text size.
+    ///
+    /// The argument can be either a `&'static str` or a [`Key<&'static str>`].
+    ///
+    /// [`Key<f64>`]: struct.Key.html
```
xStrom: Same story here.

druid/src/widget/label.rs
thecodewarrior: Oops, yeah I forgot to update that when I added the `'static`.

druid/examples/styled_text.rs
```diff
@@ -57,26 +59,39 @@ fn ui_builder() -> impl Widget<AppData> {
...
+    .with_text_color(theme::PRIMARY_LIGHT)
+    .with_text_size(MY_CUSTOM_TEXT_SIZE)
+    .background(my_painter)
+    .on_click(|_, data, _| {
+        data.size *= 1.1;
+    })
+    .env_scope(|env: &mut druid::Env, data: &AppData| {
+        env.set(MY_CUSTOM_TEXT_SIZE, data.size);
+        if data.mono {
+            env.set(theme::FONT_NAME, "monospace");
```
futurepaul: The current thinking on "idiomatic" druid (a WIP) is to use a custom key and override that (like with MY_CUSTOM_TEXT_SIZE) instead of overriding a default key

druid/src/env.rs
```diff
@@ -420,6 +420,12 @@ macro_rules! impl_value_type_borrowed {
                 Value::$var(self)
             }
         }
+
+        impl Into<Value> for &$ty {
```
futurepaul: Just flagging this as interesting, I don't understand this enough to comment usefully.

thecodewarrior: This is where we just about reach the limit of my experience and understanding of the system. Without that it's impossible to use `into` to turn a `Key<&str>` into a `KeyOrValue<&str>`, since you can't `into` a `&str` into a `Value`, and it's not possible to use `KeyOrValue<String>` since the `Key<T>` and `KeyOrValue<T>` have to match. I'm certain there are more elegant ways to do this, but I don't understand the system well enough to know how to implement them.

druid/examples/styled_text.rs
thecodewarrior: Will do.

thecodewarrior: Trying to set the custom font name to the current theme name causes errors due to borrowing the `env` when I borrow the font name. I think adjustments may need to be made to the current env system to make this work.

druid/src/env.rs
cmyr: I think you want a `Key<String>`, which can be borrowed as a `&str` with a lifetime derived from the value. This is similar to how you can use a `&str` to check if an item exists in a `HashSet<String>`?

thecodewarrior: I started a topic in zulip to discuss implementation details.

druid/examples/styled_text.rs
cmyr: you might need to clone the name before passing it back in or something?

Curious to see what your problem was.

druid/examples/styled_text.rs
```diff
@@ -57,26 +59,39 @@ fn ui_builder() -> impl Widget<AppData> {
...
+    let styled_label = Label::new(|data: &AppData, _env: &_| {
+        format!(
+            "Size {:.1}{}: {}",
+            data.size,
+            if data.mono { " mono" } else { "" },
+            data.text
+        )
+    })
+    .with_text_color(theme::PRIMARY_LIGHT)
+    .with_text_size(MY_CUSTOM_TEXT_SIZE)
```
cmyr, line 84: I think this should be trying to use the `with_font` method we've added?

cmyr:
@thecodewarrior I think this is on reasonable shape; would you like to rebase and address feedback and I can take another look?

cmyr:
@thecodewarrior ping on this? seems like it was very close to ready to go.

thecodewarrior:
Everything should be ready to go. Were there any other changes that you wanted before you merged?

PR #788, opened 2020-03-31
Add env chapter to book

cmyr:


docs/src/env.md
```diff
@@ -1 +1,98 @@
...
+    const GOOD_NAME: Key<f64> = Key::new("com.example.my-app.main-view-height");
+    ```
+1. **`Key`s must always be set before they are used.** In practice this means
+   that most keys are set when your application launches, using
+   [`AppLauncher::configure_env`]. Once a key has been added to the `Env`, it
+   cannot be deleted, although it can be overwritten.
+
+1. **Values can only be overwritten by values of the same type.** If you have a
+   `Key<f64>`, assuming that key has already been added to the `Env`, you cannot
+   replace it with any other type.
```
Finnerale, line 50: I think it would be nice to add (at least the bold parts) of this list directly to the `Env` docs as well

docs/src/env.md
```diff
@@ -1 +1,98 @@
...
+[`Value`]: https://docs.rs/druid/0.4.0/druid/struct.Value.html
+[`LocalizedString`]: https://docs.rs/druid/0.4.0/druid/struct.LocalizedString.html
+[`resolve`]: https://docs.rs/druid/0.4.0/druid/struct.LocalizedString.html#method.resolve
+[localization]: ./localization.md
+[reverse-DNS]: https://en.wikipedia.org/wiki/Reverse_domain_name_notation
+[`AppLauncher::configure_env`]: https://docs.rs/druid/0.4.0/druid/struct.AppLauncher.html#method.configure_env
+[`KeyOrValue`]: https://docs.rs/druid/0.4.0/druid/enum.KeyOrValue.html
+[`EnvScope`]: https://docs.rs/druid/0.4.0/druid/widget/struct.EnvScope.html
+[`WidgetExt`]: https://docs.rs/druid/0.4.0/druid/trait.WidgetExt.html
+[`env_scope`]: https://docs.rs/druid/0.4.0/druid/trait.WidgetExt.html#method.env_scope
```
Finnerale, line 98: Quite a few of those are dead ends (`KeyOrValue` for example), I suppose they did not yet exist in druid `0.4.0` and you added them as a reminder?

xStrom: They will probably work with `0.5.0`. I confirmed `KeyOrValue` working.

cmyr: yea, I want to keep this at 0.4.0 and then I'll bump to 0.5.0 all at once.

PR #791, opened 2020-04-01
Update fluent

cmyr:


futurepaul:
Is the best way to test if this is working (outside of running the tests, which are passing) to change my OS language?

cmyr:
Na, I think if this compiles it should be working? You could also verify that menu items have the right titles.

PR #794, opened 2020-04-02
Inset background and clip child of Container

futurepaul:
child widget before:
Screenshot from 2020-04-02 18-47-07

child after:
Screenshot from 2020-04-02 18-46-44

background before:
Screenshot from 2020-04-02 18-59-10

background after:
Screenshot from 2020-04-02 18-58-29

### DIFF
--- a/druid/src/widget/container.rs
+++ b/druid/src/widget/container.rs
@@ -144,8 +144,12 @@ impl<T: Data> Widget<T> for Container<T> {
             });
         }
 
+        let border_width = match &self.border {
+            Some(border) => border.width.resolve(env),
+            None => 0.0,
+        };
+
         if let Some(border) = &self.border {
-            let border_width = border.width.resolve(env);
             let border_rect = ctx
                 .size()
                 .to_rect()
@@ -154,6 +158,21 @@ impl<T: Data> Widget<T> for Container<T> {
             ctx.stroke(border_rect, &border.color.resolve(env), border_width);
         };
 
-        self.inner.paint_with_offset(ctx, data, env);
+        // Since we're painting this after the border we need to inset by the
+        // border width
+        if self.corner_radius > 0.0 {
+            let panel = ctx
+                .size()
+                .to_rect()
+                .inset(-1.0 * border_width)
+                .to_rounded_rect(self.corner_radius);
+
+            ctx.with_save(|ctx| {
+                ctx.clip(panel);
+                self.inner.paint_with_offset(ctx, data, env);
+            });
+        } else {
+            self.inner.paint_with_offset(ctx, data, env);
+        }
     }
 }
--- a/druid/src/widget/container.rs
+++ b/druid/src/widget/container.rs
@@ -135,19 +135,23 @@ impl<T: Data> Widget<T> for Container<T> {
     }
 
     fn paint(&mut self, ctx: &mut PaintCtx, data: &T, env: &Env) {
-        if let Some(background) = self.background.as_mut() {
-            let panel = ctx.size().to_rounded_rect(self.corner_radius);
-
-            ctx.with_save(|ctx| {
-                ctx.clip(panel);
-                background.paint(ctx, data, env);
-            });
-        }
-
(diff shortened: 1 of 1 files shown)


druid/src/widget/container.rs
```diff
@@ -154,6 +162,13 @@ impl<T: Data> Widget<T> for Container<T> {
...
 
-        self.inner.paint_with_offset(ctx, data, env);
+        if self.corner_radius > 0.0 {
+            ctx.with_save(|ctx| {
+                ctx.clip(inset_panel);
+                self.inner.paint_with_offset(ctx, data, env);
+            });
+        } else {
+            self.inner.paint_with_offset(ctx, data, env);
+        }
```
xStrom, line 172: Clipping only with rounded rect seems rather inconsistent. I think having the clip always/never would be better and lead to fewer surprises.

Is clipping something we want to do here though? Is there an easy way to remove the clip from within the child's paint method? If so then this could be viewed as `overflow: hidden;` by default.

As a comparison, web CSS doesn't hide the overflow by default. Children of a rounded rect paint happily outside the bounds. With such an approach it would still be valuable to be able to opt-in to the clipping.

Perhaps the clipping should be a configuration option on `Container`?

druid/src/widget/container.rs
```diff
@@ -135,17 +135,25 @@ impl<T: Data> Widget<T> for Container<T> {
...
+        let inset_panel = ctx
+            .size()
+            .to_rect()
+            .inset(-1.0 * border_width)
+            .to_rounded_rect(self.corner_radius);
+
+        if let Some(background) = self.background.as_mut() {
             ctx.with_save(|ctx| {
-                ctx.clip(panel);
+                ctx.clip(inset_panel);
```
xStrom, line 151: Painting the background with these insets means that the background isn't visible with dashed borders etc. Is this behavior we want?

xStrom:
Both of the changes in this PR add functionality that differs from web CSS, for which I made [an example](https://jsfiddle.net/m2hsgx93/). On the web it works like this:
* The child gets the non-rounded rect constraints and if it fills that, then it means it's happily painting over the rounded corners. This behavior can be disabled by adding `overflow: hidden;` to the container.
* The container background is painted under the border as well, so that pieces of it appear when using a dashed border style. Different behavior can be achieved by setting the background on the child instead.

Now I'm not saying that web CSS is perfect. Although it's definitely the most well known behavior, so mimicking it would be least surprising.

Is this the default behavior in some other system like Flutter? I'm not familiar with Flutter, but a quick look at [the docs](https://api.flutter.dev/flutter/painting/BorderStyle-class.html) suggests that they don't have dashed borders at all.

Whatever the default, I think being able to achieve both behaviors (like in CSS) is a good solution, unless it adds a ton of complexity. In druid it's already possible to achieve both (although it's not trivial) but this PR seems to remove options.

druid/src/widget/container.rs
futurepaul: I think you're right about this else statement being inconsistent.

futurepaul:
I'm going to mark this as waiting-on-author because you're making a lot of good points!!!

1. I would be happy with an option for show / hide overflow, as long as it's not confused with the basic layout sizing stuff. Maybe a method called `clip`? My goal was to keep the behaviour of container but only improve the handling of rounded corners for my own selfish purposes.

2. I agree if we have dashed borders we should paint under them, so I need to fix that here. My problem with the current version (see "background before" image) is we were painting outside of even a solid border due to some mismatch between the size of the border rrect and the background rrect.

3. I played around with this in Flutter: https://dartpad.dev/b6409e10de32b280b8938aa75364fa7b
Their Container doesn't clip the child corners by default (though it does clip content overflow, with a warning about overflow), and it does clip the background color correctly. To get the clipping behaviour I want I can use a ClipRRect with the same BorderRadius of the child Container and I get my desired outcome.

futurepaul:
Going to close this and continue it as a discussion: #853

PR #797, opened 2020-04-03
Actually close GTK windows

### LINKED ISSUE #464: Can't close a window with a custom button
I'm running code which in simplified version looks like this
```rust
use std::{thread, time};
use druid::widget::{Button, Flex};
use druid::{AppLauncher, Widget, WindowDesc};

// Clicking a button should close the window

//// The close button close the window
//fn main () {
//    let window = WindowDesc::new(build_widget);
//    AppLauncher::with_window(window)
//        .use_simple_logger()
//        .launch(0_u32)
//        .expect("launch failed");
//
//    dbg!("GUI killed");
//}

// The close button makes the window hang
fn main () {
    dbg!("LAUNCH_GUI");

    let mut should_gui = true;
    loop {
        dbg!("LOOP ITERATION");
        if should_gui == true {
            dbg!("SHOW GUI");
            should_gui = false;

            let window = WindowDesc::new(build_widget);
            AppLauncher::with_window(window)
                .use_simple_logger()
                .launch(0_u32)
                .expect("launch failed");

            dbg!("GUI KILLED");
        }
        let sleep_time = time::Duration::from_millis(100);
        thread::sleep(sleep_time);
    }
}

fn build_widget() -> impl Widget<u32> {
    Flex::column()
        .with_child(
            Button::sized(
                "Close",
                |evt_ctx , data: _, _env| {
                    dbg!(&data);
                    //evt_ctx.window().close();
                    evt_ctx.submit_command(druid::commands::CLOSE_WINDOW, None);
                },
                80.0,
                20.0,
            ),
            0.0,
        )
}
```

I'm expecting the window is closed properly and the program keeps looping without GUI.

Instead, I'm getting a hanging window which can't be closed, see
druid-close-broken2

Also discussed here: https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/evt_ctx.2Ewindow().2Eclose().20not.20working.3F

# Software versions
gtk3-devel.x86_64                                  3.24.13-1.fc31                       
glib2-devel.x86_64                                 2.62.3-1.fc31

Finnerale:
Instead of removing the window from the application, it now gets closed.
This Closes #464 and also fixes the `multiwin` example with GTK.
Closing a window will also remove it from the `Application`

### DIFF
--- a/druid-shell/src/platform/gtk/window.rs
+++ b/druid-shell/src/platform/gtk/window.rs
@@ -472,9 +472,7 @@ impl WindowHandle {
     /// Close the window.
     pub fn close(&self) {
         if let Some(state) = self.state.upgrade() {
-            with_application(|app| {
-                app.remove_window(&state.window);
-            });
+            state.window.close();
         }
     }
 


ForLoveOfCats:
I'm currently testing this in relation to #464 but I do have a question in the meantime. What specifically about the `multiwin` example does this fix? It doesn't fix the context menu crash and I cannot find anything else about it which happens to be broken under GTK on `master`.

Finnerale:
@ForLoveOfCats The close item in the file menu did not work (as it works exactly like a custom close button)

cmyr:
Thanks both!

PR #801, opened 2020-04-04
Simple wasm example

futurepaul:
This depends on #759 and should be merged after (and will def need some changes before that!).

Just wanted to get something up so I could get some feedback. A couple questions:

1. Should we come up with a new example specifically for the web? Or is hello fine.
2. Is this doing the Cargo workspace stuff correctly? It's bugging me about needing a license file but I don't see that warning from i.e. druid-shell?


elrnv:
FWIW, you should be able to compile and serve almost all the examples unchanged so long as `fn main` is made `pub` and `std::time::Instant` is swapped with `instant::Instant` in each example.

druid/examples/hello_wasm/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+
+[dependencies]
+druid = { path="../.."}
+wasm-bindgen = "0.2.60"
+console_error_panic_hook = { version = "0.1.6" }
+log = "0.4.8"
+instant = { version = "0.1", features = [ "wasm-bindgen" ] }
+
+[target.'cfg(not(target_arch="wasm32"))'.dependencies]
+simple_logger = { version = "1.6.0", default-features = false }
```
Finnerale, line 21: These seem to be unused?

elrnv: More specifically `simple_logger`, `instant` and `log`.

elrnv:
This builds and runs for me perfectly :)

I would add two things.
1. Should we have a CI build for this example for completeness?
2. Druid could be used for other WASM targets like Electron. This particular use case builds it for the web target to be used in browsers, but the Electron build is slightly different. With that in mind I would suggest renaming this to `hello_web` or `hello_wasm_web` for the possibility of adding `hello_npm` or something similar later for other wasm targets. I think having examples for other wasm targets is important because the behaviour of windows and menus could be different.

futurepaul:
@Finnerale @xStrom sorry for the long delay but I've finally addressed your comments. Not sure why the Cargo.lock is conflicting, it wasn't yesterday.

xStrom:
It's conflicting because it doesn't exist anymore. You need to rebase with `master` and make sure `Cargo.lock` doesn't exist in your PR.

druid/examples/hello_wasm_web/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Building
+
+You will need `cargo` and `wasm-pack` for building the code and a simple
+server like [`http`](https://crates.io/crates/https) for serving the web page.
+
+First build with.
+
+```
+> wasm-pack build --target web
```
Finnerale, line 13: It might be nice to add `--dev` here, as `wasm-pack` builds in release mode by default.

druid/examples/hello_wasm_web/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+// limitations under the License.
+
+use druid::widget::{Align, Flex, Label, TextBox};
+use druid::{AppLauncher, Data, Env, Lens, LocalizedString, Widget, WidgetExt, WindowDesc};
+
+use wasm_bindgen::prelude::*;
+
+const VERTICAL_WIDGET_SPACING: f64 = 20.0;
+const TEXT_BOX_WIDTH: f64 = 200.0;
+const WINDOW_TITLE: LocalizedString<HelloState> = LocalizedString::new("Hello World!");
```
Finnerale, line 22: Window titles can be normal strings now :tada:

druid/examples/hello_wasm_web/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+#[wasm_bindgen]
+pub fn wasm_main() {
+    std::panic::set_hook(Box::new(console_error_panic_hook::hook));
+    main()
+}
+pub fn main() {
+    // describe the main window
+    let main_window = WindowDesc::new(build_root_widget)
+        .title(WINDOW_TITLE)
+        .window_size((400.0, 400.0));
```
Finnerale, line 40: As `window_size` is ignored on wasm anyways, I think it can be removed.

druid/examples/hello_wasm_web/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+#[derive(Clone, Data, Lens)]
+struct HelloState {
+    name: String,
+}
+
+// This wrapper function is the only modification we're making to the vanilla
+// hello.rs example.
+#[wasm_bindgen]
+pub fn wasm_main() {
+    std::panic::set_hook(Box::new(console_error_panic_hook::hook));
```
Finnerale, line 33: Would be nice if this one could get a comment, explaining that this is necessary to get panic messages on wasm32.

druid/examples/hello_wasm_web/src/lib.rs
futurepaul: looks like window title isn't used by the web version so I'm leaving it off entirely now

druid/examples/hello_wasm_web/src/lib.rs
futurepaul: I was unaware of what this actually did, so I hope my copy of your explanation is sufficient.

futurepaul:
Okay, I believe I've addressed all comments. Also, I realized I hadn't updated the index.js to refer to the different package name, so that's fixed now as well.

PR #810, opened 2020-04-07
Add EditableLabel proof of concept

cmyr:
This is a fairly simple widget that displays a label that can be edited on click. The input is validated, and the data is only mutated when editing completes.

---

This is something I need in runebender. I don't want to merge this in its current state, but wanted to see what people think; it illustrates some of how I've been thinking about validation in general.

I'm not opposed to merging something like this, as a temporary measure, but I would want to at least remove the changes to the `hello` example.


ForLoveOfCats:
I'm probably being dense but in which ways does this differ from a `TextBox`?

cmyr:
it looks like a label when it isn't being edited; it has a non `String` data type (kind of like `Parse`, but more general); and the data doesn't change until the user ends editing, by pressing return or switching the focus.

PR #813, opened 2020-04-07
Use checkout@v2 for X11 tests

crsaracco:


xStrom:
Thanks but there's already #812.

PR #817, opened 2020-04-08
Adds support for invalidating a single rectangle.

jneem:
This adds support in druid-shell for invalidating a single rectangle.

Caveats/questions:
 - I checked that the win/mac changes compile, but I can't test them. Mac is particularly likely to be wrong, since I have no idea how rust-objc works, and how it maps to the apple docs online.
- Should the plain `invalidate` methods be deprecated? Or should I just keep the name and introduce the rect parameter?

### DIFF
--- a/druid-shell/examples/invalidate.rs
+++ b/druid-shell/examples/invalidate.rs
@@ -14,12 +14,12 @@
 
 use std::any::Any;
 
-use std::time::Instant;
+use std::time::{Duration, Instant};
 
 use druid_shell::kurbo::{Point, Rect};
 use druid_shell::piet::{Color, Piet, RenderContext};
 
-use druid_shell::{Application, WinHandler, WindowBuilder, WindowHandle};
+use druid_shell::{Application, TimerToken, WinHandler, WindowBuilder, WindowHandle};
 
 struct InvalidateTest {
     handle: WindowHandle,
@@ -49,13 +49,17 @@ impl InvalidateTest {
 impl WinHandler for InvalidateTest {
     fn connect(&mut self, handle: &WindowHandle) {
         self.handle = handle.clone();
+        self.handle.request_timer(Duration::from_millis(60));
+    }
+
+    fn timer(&mut self, _id: TimerToken) {
+        self.update_color_and_rect();
+        self.handle.invalidate_rect(self.rect);
+        self.handle.request_timer(Duration::from_millis(60));
     }
 
     fn paint(&mut self, piet: &mut Piet, rect: Rect) -> bool {
-        self.update_color_and_rect();
         piet.fill(rect, &self.color);
-
-        self.handle.invalidate_rect(self.rect);
         false
     }
 
--- a/druid-shell/examples/invalidate.rs
+++ b/druid-shell/examples/invalidate.rs
@@ -1,0 +1,107 @@
+// Copyright 2020 The xi-editor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::any::Any;
+
+use std::time::Instant;
+
+use piet_common::kurbo::{Point, Rect};
+use piet_common::{Color, Piet, RenderContext};
+
+use druid_shell::{Application, WinHandler, WindowBuilder, WindowHandle};
+
+struct InvalidateTest {
+    handle: WindowHandle,
+    size: (f64, f64),
+    start_time: Instant,
+    color: Color,
+    rect: Rect,
+}
+
+fn split_rgb(c: &Color) -> (u8, u8, u8) {
+    let rgba = c.as_rgba_u32();
+    (
+        (rgba >> 24 & 255) as u8,
+        (rgba >> 16 & 255) as u8,
+        (rgba >> 8 & 255) as u8,
+    )
+}
+
+impl InvalidateTest {
+    fn update_color_and_rect(&mut self) {
+        let time_since_start = (Instant::now() - self.start_time).as_nanos();
+        let (r, g, b) = split_rgb(&self.color);
+        self.color = match (time_since_start % 2, time_since_start % 3) {
+            (0, _) => Color::rgb8(r.wrapping_add(10), g, b),
+            (_, 0) => Color::rgb8(r, g.wrapping_add(10), b),
+            (_, _) => Color::rgb8(r, g, b.wrapping_add(10)),
+        };
+
+        self.rect.x0 = (self.rect.x0 + 5.0) % self.size.0;
+        self.rect.x1 = (self.rect.x1 + 5.5) % self.size.0;
+        self.rect.y0 = (self.rect.y0 + 3.0) % self.size.1;
+        self.rect.y1 = (self.rect.y1 + 3.5) % self.size.1;
+    }
+}
+
+impl WinHandler for InvalidateTest {
+    fn connect(&mut self, handle: &WindowHandle) {
+        self.handle = handle.clone();
+    }
+
+    fn paint(&mut self, piet: &mut Piet, rect: Rect) -> bool {
+        self.update_color_and_rect();
+        piet.fill(rect, &self.color);
+
+        self.handle.invalidate_rect(self.rect);
+        false
+    }
+
+    fn size(&mut self, width: u32, height: u32) {
+        let dpi = self.handle.get_dpi();
+        let dpi_scale = dpi as f64 / 96.0;
+        let width_f = (width as f64) / dpi_scale;
+        let height_f = (height as f64) / dpi_scale;
+        self.size = (width_f, height_f);
+    }
+
+    fn command(&mut self, id: u32) {
+        match id {
+            0x100 => self.handle.close(),
+            _ => println!("unexpected id {}", id),
+        }
+    }
+
+    fn as_any(&mut self) -> &mut dyn Any {
+        self
+    }
+}
+
+fn main() {
+    let mut app = Application::new(None);
+    let mut builder = WindowBuilder::new();
+    let inv_test = InvalidateTest {
+        size: Default::default(),
+        handle: Default::default(),
+        start_time: Instant::now(),
+        rect: Rect::from_origin_size(Point::ZERO, (10.0, 20.0)),
+        color: Color::WHITE,
+    };
+    builder.set_handler(Box::new(inv_test));
+    builder.set_title("Invalidate tester");
+
+    let window = builder.build().unwrap();
+    window.show();
+    app.run();
+}
--- a/druid-shell/examples/invalidate.rs
+++ b/druid-shell/examples/invalidate.rs
@@ -16,8 +16,8 @@ use std::any::Any;
 
 use std::time::Instant;
 
-use piet_common::kurbo::{Point, Rect};
-use piet_common::{Color, Piet, RenderContext};
+use druid_shell::kurbo::{Point, Rect};
+use druid_shell::piet::{Color, Piet, RenderContext};
 
 use druid_shell::{Application, WinHandler, WindowBuilder, WindowHandle};
 
@@ -29,19 +29,10 @@ struct InvalidateTest {
     rect: Rect,
 }
 
-fn split_rgb(c: &Color) -> (u8, u8, u8) {
-    let rgba = c.as_rgba_u32();
-    (
-        (rgba >> 24 & 255) as u8,
-        (rgba >> 16 & 255) as u8,
-        (rgba >> 8 & 255) as u8,
-    )
-}
-
 impl InvalidateTest {
     fn update_color_and_rect(&mut self) {
         let time_since_start = (Instant::now() - self.start_time).as_nanos();
-        let (r, g, b) = split_rgb(&self.color);
+        let (r, g, b, _) = self.color.as_rgba_u8();
         self.color = match (time_since_start % 2, time_since_start % 3) {
             (0, _) => Color::rgb8(r.wrapping_add(10), g, b),
             (_, 0) => Color::rgb8(r, g.wrapping_add(10), b),
--- a/druid-shell/src/platform/mac/window.rs
+++ b/druid-shell/src/platform/mac/window.rs
@@ -651,7 +651,7 @@ impl WindowHandle {
         );
         unsafe {
             // We could share impl with redraw, but we'd need to deal with nil.
-            let () = msg_send![*self.nsview.load(), setNeedsDisplay: rect];
+            let () = msg_send![*self.nsview.load(), setNeedsDisplayInRect: rect];
         }
     }
 
--- a/druid-shell/src/platform/mac/window.rs
+++ b/druid-shell/src/platform/mac/window.rs
@@ -38,7 +38,7 @@ use objc::{class, msg_send, sel, sel_impl};
 use cairo::{Context, QuartzSurface};
 use log::{error, info};
 
-use crate::kurbo::{Point, Size, Vec2};
+use crate::kurbo::{Point, Rect, Size, Vec2};
 use crate::piet::{Piet, RenderContext};
 
 use super::dialog;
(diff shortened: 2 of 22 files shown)


druid/src/contexts.rs
```diff
@@ -378,14 +387,24 @@ impl<'a> EventCtx<'a> {
 impl<'a> LifeCycleCtx<'a> {
     #[deprecated(since = "0.5.0", note = "use request_paint instead")]
     pub fn invalidate(&mut self) {
-        self.base_state.needs_inval = true;
+        self.request_paint();
     }
 
-    /// Request a [`paint`] pass.
+    /// Request a [`paint`] pass. This is equivalent to calling [`request_paint_rect`] for the
+    /// entire layout rectangle.
```
raphlinus, line 394: I think we want the paint rect rather than the layout rect here. But the basic idea of using existing geometry is sound.

druid/src/widget/invalidation.rs
```diff
@@ -0,0 +1,67 @@
+// Copyright 2020 The xi-editor Authors.
```
raphlinus, line 1: This feels like an example? I'm not sure it belongs as a core widget.

jneem: This is in support of `WidgetExt::debug_invalidation`. I originally tried it using `EnvScope`, but I don't think you want the invalidation regions to be drawn recursively so I ended up breaking it into another widget.

druid/src/contexts.rs
jneem: Got it.

jneem:
So I had a first stab at getting scrolling working. I'm not super happy with it (which is why I only implemented `EventCtx` so far), but it's the best I have right now.

As far as I understand, the way scrolling eventually works is the following: the scroll widget will maintain a buffer that contains the painted contents of its children. That buffer will be somewhat bigger than the viewport. When the `Scroll` is scrolled, it just blits from the buffer. It can handle asking its child to rerender the buffer on its own time, possibly even as an idle callback.

This means that:
- if a child of the scroll requests repainting some visible rect, the scroll needs to propagate that back up to the system, properly translated; but also,
- the scroll needs to store the invalidated regions somewhere, in order do do its own repainting.

AFAICT, these two constraints (and especially the second) means that the `Scroll` really needs to insert a callback somewhere. And this was the least-ugly place I could find to put it.

jneem:
By the way, the current implementation has a buglet, in that there is a situation where it invalidates too much (and I don't know how to fix it without more drastic changes): if the scroll has a `Flex` child, and that child invalidates something in the viewport and something out of the viewport, then the bounding box of those two invalidations will be quite large (but this isn't the bug yet). The bug is that the part outside the viewport won't get painted, because `Flex` doesn't paint its invisible children. That means it will never get uninvalidated (validated?), and so that large bounding box will keep getting repainted.

I'm not planning to do anything about this right away, I just wanted to right it down so I don't forget...

druid/src/core.rs
```diff
@@ -212,6 +217,27 @@ impl<T, W: Widget<T>> WidgetPod<T, W> {
         self.state.layout_rect.unwrap_or_default()
     }
 
+    /// Set the viewport offset.
+    ///
+    /// This should be called on a child widget whenever its "physical" position,
+    /// as seen by painting, mouse events, etc. differs from its "logical"
+    /// position. You probably only want this method if you are implementing a
+    /// widget like [`Scroll`].
```
cmyr, line 225: I would clarify this further. I think "physical" and "logical" are unclear, here. I would say something like,

`This should must be set by the parent in cases where the parent is modifying the position of a child view relative to the viewport; this is only relevant in special cases such as a scroll view.'

(this doesn't sound much better, either, blah)

druid/src/core.rs
```diff
@@ -721,7 +747,7 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
         self.old_data = Some(data.clone());
         self.env = Some(env.clone());
 
-        ctx.base_state.merge_up(&self.state)
+        ctx.base_state.merge_up(&self.state);
```
cmyr, line 750: nitty rust thing, but I don't feel a semicolon is necessary here? we're returning nothing, this last method returns nothing, we can just let it evaluate? (not a blocker just an observation)

druid/src/core.rs
```diff
@@ -75,11 +76,15 @@ pub(crate) struct BaseState {
     /// drop shadows or overflowing text.
     pub(crate) paint_insets: Insets,
 
-    // TODO: consider using bitflags for the booleans.
+    // The region that needs to be repainted, relative to the origin of `layout_rect`.
```
cmyr, line 79: a distinction I make sometimes (borrowed from cocoa) is between 'frame' and 'bounds'; 'frame' is a rectangle in the parent's coordinate space, and 'bounds' is a rectangle in the current widget's coordinate space. In that sense, this is a region relative to the widget's bounds.

druid/src/widget/invalidation.rs
```diff
@@ -0,0 +1,67 @@
...
+//
```
cmyr, line 2: the way I might consider implementing this is not as a separate widget, but as part of `WidgetPod`'s `paint` method? It's possible this doesn't make sense, but it's my first instinct.

druid/src/window.rs
```diff
@@ -273,8 +273,17 @@ impl<T: Data> Window<T> {
     }
 
     pub(crate) fn invalidate_and_finalize(&mut self) {
-        if self.root.state().needs_inval {
+        // Currently, if we need to re-layout then we invalidate the whole window.
+        // We could be smarter about this (only invalidating the part of the widget tree
+        // that changed its layout), but I think this would require doing the layout
+        // before the paint pass, so that the system knows which part got invalidated.
```
cmyr, line 279: we do always do layout (if needed) before paint? This interaction is subtle; a layout invalidation implies a paint invalidation, but we could be smarter about stopping invalidation when old and new layouts converge? That's definitely out of scope for this PR.

To clarify: when any widget has its layout invalidated we should be repainting the whole view.

druid/src/window.rs
```diff
@@ -295,7 +305,17 @@ impl<T: Data> Window<T> {
...
+        self.paint(piet, invalid_rect, data, env);
+
+        // If commands were submitted during anim frame, ask the handler
+        // to call us back on idle so we can process them in a new event/update pass.
+        if !queue.is_empty() {
+            if let Some(mut handle) = self.handle.get_idle_handle() {
+                handle.schedule_idle(RUN_COMMANDS_TOKEN);
+            } else {
+                log::error!("failed to get idle handle");
+            }
```
cmyr, line 317: was this code moved from elsewhere? I'm slightly confused.

druid/src/widget/invalidation.rs
```diff
@@ -0,0 +1,67 @@
...
+
+    fn layout(&mut self, ctx: &mut LayoutCtx, bc: &BoxConstraints, data: &T, env: &Env) -> Size {
+        self.inner.layout(ctx, bc, data, env)
+    }
+
+    fn paint(&mut self, ctx: &mut PaintCtx, data: &T, env: &Env) {
+        self.inner.paint(ctx, data, env);
+
+        let color = Color::hlca(self.hue, 127.0, 64.0, 0.5);
+        let rect = Rect::from_origin_size(Point::ZERO, ctx.size());
```
cmyr, line 59: you can do `ctx.size().to_rect()`, now. (This wasn't always possible, so you see a number of other ways to create a rect from a size in various examples)

druid/examples/scroll.rs
```diff
@@ -44,7 +44,13 @@ struct OverPainter(u64);
 const INSETS: Insets = Insets::uniform(50.);
 
 impl<T: Data> Widget<T> for OverPainter {
-    fn event(&mut self, _: &mut EventCtx, _: &Event, _: &mut T, _: &Env) {}
+    fn event(&mut self, ctx: &mut EventCtx, ev: &Event, _: &mut T, _: &Env) {
```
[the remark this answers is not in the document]
jneem: Partly to illustrate the bug, but mostly just to check that I was computing the invalidation offsets correctly. I wasn't planning to include this in the final PR, though -- I'll write a separate example.

druid/src/widget/invalidation.rs
jneem: I originally tried to imitate the implementation of `debug_paint_layout` (i.e. using `EnvScope` and `WidgetPod::paint` instead of a new widget). The problem is that, unlike drawing layout-rect boundaries, we don't want to recursively draw the invalidation regions for the children. It just occurred to me, though, that I can probably use `Painter` instead of a new widget.

druid/src/window.rs
jneem: I agree that this it's definitely out of scope, and I guess a github issue would be better than a comment that no one will read.
Anyway, this was triggered by the calc example, where every time you press a number the layout gets invalidated, even though only the calculator's "screen" changes its layout.

druid/src/window.rs
jneem: Sorry, it must have  been a faulty rebase.

jneem:
Thanks for the review! I'll do a round or two of clean-up and then remove the [WIP]. I'd still feel better if someone briefly tested windows/mac, though...

druid/src/core.rs
xStrom: I'm guessing the upside of the semicolon is that if `merge_up` starts returning a value then without a semicolon this will be a compile error.

jneem:
The remaining CI failures appear to be linking errors that I can't reproduce locally...

jneem:
Ok, seems like a rebase resolved it.

druid/src/widget/invalidation.rs
cmyr: I see what you mean.

druid/src/widget/mod.rs
```diff
@@ -64,6 +65,7 @@ pub use either::Either;
 pub use env_scope::EnvScope;
 pub use flex::{CrossAxisAlignment, Flex, FlexParams, MainAxisAlignment};
 pub use identity_wrapper::IdentityWrapper;
+pub use invalidation::DebugInvalidation;
```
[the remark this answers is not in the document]
jneem: Huh, I was expecting that to cause some sort of private-in-public error, but it doesn't...

druid/src/widget/invalidation.rs
```diff
@@ -0,0 +1,66 @@
...
+        self.inner.layout(ctx, bc, data, env)
+    }
+
+    fn paint(&mut self, ctx: &mut PaintCtx, data: &T, env: &Env) {
+        self.inner.paint(ctx, data, env);
+
+        let color = env.get_debug_color(self.debug_color).with_alpha(0.5);
+        let rect = ctx.region().to_rect();
+        ctx.fill(rect, &color);
+        self.debug_color += 1;
```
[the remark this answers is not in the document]
jneem: Without the color change, every time a region gets invalidated it gets painted with a blue (say) overlay. The overall effect is that the entire window is covered with a blue overlay all the time, so you can't see what was just invalidated.

druid/examples/invalidation.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    fn paint(&mut self, ctx: &mut PaintCtx, data: &Point, _env: &Env) {
+        ctx.with_save(|ctx| {
+            let rect = ctx.size().to_rect();
+            ctx.clip(rect);
+            ctx.fill(rect, &Color::WHITE);
+            ctx.fill(Circle::new(*data, RADIUS), &Color::BLACK);
+        })
+    }
+}
```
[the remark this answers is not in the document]
jneem: Ok, that's not right. Does the invalidation test in druid_shell work? It should look something like the attached gif.
invalidation_test

jneem:
I'll try out a few options for the debug visualization. I don't *think* recursive drawing will work well, because there's just one single invalidation rect for the entire window, and so any individual widget's invalidation rect is just its paint rect intersected with the global rect -- there's no interesting recursive structure like there is for the layout rects.

I'll also try doing outlines.

jneem:
By the way, this is what the (druid) invalidation test looks like for me.
druid_invalidation

jneem:
Here are two other options: recursive outlines and non-recursive outlines. I think my favorite is the non-recursive outlines.
single_outline
recursive_outline

druid/src/widget/invalidation.rs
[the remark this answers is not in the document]
cmyr: I'm confused; wouldn't that only be the case if the entire view was invalidated?

druid/examples/invalidation.rs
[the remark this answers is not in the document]
cmyr: no, on mac it is just a window with a single solid background color. I'll dig in a little bit...

druid-shell/src/platform/mac/window.rs
```diff
@@ -639,6 +643,18 @@ impl WindowHandle {
...
 
+    /// Request invalidation of one rectangle.
+    pub fn invalidate_rect(&self, rect: Rect) {
+        let rect = NSRect::new(
+            NSPoint::new(rect.x0, rect.y0),
+            NSSize::new(rect.width(), rect.height()),
+        );
+        unsafe {
+            // We could share impl with redraw, but we'd need to deal with nil.
+            let () = msg_send![*self.nsview.load(), setNeedsDisplay: rect];
```
cmyr, line 654:
```suggestion
            let () = msg_send![*self.nsview.load(), setNeedsDisplayInRect: rect];
```

oops 😅 💦

druid-shell/examples/invalidate.rs
```diff
@@ -0,0 +1,98 @@
...
+        self.handle = handle.clone();
+    }
+
+    fn paint(&mut self, piet: &mut Piet, rect: Rect) -> bool {
+        self.update_color_and_rect();
+        piet.fill(rect, &self.color);
+
+        self.handle.invalidate_rect(self.rect);
+        false
+    }
```
cmyr, line 60: on mac, invalidation calls that happen during drawing are ignored.

I would suggest changing this to use a timer, something like:
```suggestion
    fn connect(&mut self, handle: &WindowHandle) {
        self.handle = handle.clone();
        self.handle.request_timer(std::time::Duration::from_millis(60));
    }
    
    fn timer(&mut self, _id: TimerToken) {
        self.update_color_and_rect();
        self.handle.invalidate_rect(self.rect);
        self.handle.request_timer(std::time::Duration::from_millis(60));
    }

    fn paint(&mut self, piet: &mut Piet, rect: Rect) -> bool {
        piet.fill(rect, &self.color);
        false
    }
```

jneem:
Thanks! I fixed those, and also some other issues with propagating a too-large invalid region in druid (in cairo, the drawing context is automatically clipped, so I wasn't seeing the issue).

Regarding the animation, there was a reason I hadn't done it yet, but I can't remember why...

cmyr:
I wouldn't worry about the animation, it will be sort of tricky to get right with the current API. The longer-term plan is to build a higher-level API on top of what we currently have, and when we do that we can revisit.

I guess it would be possible to make some specific improvements now, though; but this would require turning `request_anim` into a `Region` or something.

cmyr:
Also I'm not sure I can tell the difference between the recursive v. non-recursive outlines; I defer to your judgement.

jneem:
Ok, so I've gone with non-recursive outlines then. It's easy enough to change later anyway.

Also, I'll open an issue about `request_anim`.

cmyr:
Oh, one last thing: it might be worth adding this to the wasm examples? in `druid/examples/wasm`. (forgive me if this is don't, I skimmed and didn't see i)

jneem:
I think it's automatic, right? `druid/examples/wasm/src/examples` seems to be symlinked to `druid/examples`.

jneem:
Ok, I think I may have fixed the wasm example. I am a bit puzzled by the two calls to `piet_ctx.finish()` in `druid_shell::platform::web::WindowState::render()` -- is that just an oversight?

I had a toy example showing improved performance for small invalidations when drawing is expensive. I can try cleaning it up and adding it as an example. But my real motivation for doing this needs more work before it can benefit...

cmyr:
I can't speak to the wasm stuff; maybe @elrnv can. In any case I'm happy to merge this now, thanks again for your patience!

elrnv:
I am not sure why we need to finish calls. For reference, I believe the code @jneem  is referring to is in the render function
```
    fn render(&self) -> bool {
        self.context
            .clear_rect(0.0, 0.0, self.get_width() as f64, self.get_height() as f64);
        let mut piet_ctx = piet_common::Piet::new(self.context.clone(), self.window.clone());
        let want_anim_frame = self.handler.borrow_mut().paint(&mut piet_ctx);
        if let Err(e) = piet_ctx.finish() {
            log::error!("piet error on render: {:?}", e);
        }
        let res = piet_ctx.finish();
        if let Err(e) = res {
            log::error!("EndDraw error: {:?}", e);
        }
        want_anim_frame
    }
```

This part is over a year old, it does look redundant to me. After trying the examples without the second `piet_ctx.finish()`, I couldn't find any difference. I think its safe to remove the second `.finish()` call.

PR #818, opened 2020-04-08
Add documentation, e.g. around bloom usage.

xStrom:
These are the simpler parts of #811 so that they can go on the merge train without waiting for more work there.

### DIFF
--- a/druid/src/bloom.rs
+++ b/druid/src/bloom.rs
@@ -71,10 +71,12 @@ impl<T: ?Sized + Hash> Bloom<T> {
         self.entry_count += 1;
     }
 
-    /// Return whether an item exists in the filter.
+    /// Returns `true` if the item may have been added to the filter.
     ///
     /// This can return false positives, but never false negatives.
-    pub fn contains(&self, item: &T) -> bool {
+    /// Thus `true` means that the item may have been added - or not,
+    /// while `false` means that the item has definitely not been added.
+    pub fn may_contain(&self, item: &T) -> bool {
         let mask = self.make_bit_mask(item);
         self.bits & mask == mask
     }
@@ -134,11 +136,11 @@ mod tests {
         let mut bloom = Bloom::default();
         for i in 0..100 {
             bloom.add(&i);
-            assert!(bloom.contains(&i));
+            assert!(bloom.may_contain(&i));
         }
         bloom.clear();
         for i in 0..100 {
-            assert!(!bloom.contains(&i));
+            assert!(!bloom.may_contain(&i));
         }
     }
 
@@ -147,18 +149,18 @@ mod tests {
         let mut bloom1 = Bloom::default();
         bloom1.add(&0);
         bloom1.add(&1);
-        assert!(!bloom1.contains(&2));
-        assert!(!bloom1.contains(&3));
+        assert!(!bloom1.may_contain(&2));
+        assert!(!bloom1.may_contain(&3));
         let mut bloom2 = Bloom::default();
         bloom2.add(&2);
         bloom2.add(&3);
-        assert!(!bloom2.contains(&0));
-        assert!(!bloom2.contains(&1));
+        assert!(!bloom2.may_contain(&0));
+        assert!(!bloom2.may_contain(&1));
 
         let bloom3 = bloom1.union(bloom2);
-        assert!(bloom3.contains(&0));
-        assert!(bloom3.contains(&1));
-        assert!(bloom3.contains(&2));
-        assert!(bloom3.contains(&3));
+        assert!(bloom3.may_contain(&0));
+        assert!(bloom3.may_contain(&1));
+        assert!(bloom3.may_contain(&2));
+        assert!(bloom3.may_contain(&3));
     }
 }
--- a/druid/src/contexts.rs
+++ b/druid/src/contexts.rs
@@ -238,7 +238,7 @@ impl<'a> EventCtx<'a> {
         self.is_handled
     }
 
-    /// The focus status of a widget.
+    /// The (tree) focus status of a widget.
     ///
     /// Focus means that the widget receives keyboard events.
     ///
@@ -251,20 +251,21 @@ impl<'a> EventCtx<'a> {
     /// hierarchy, all ancestors of that leaf widget are also invoked with
     /// `FocusChanged(true)`.
     ///
-    /// Discussion question: is "is_focused" a better name?
+    /// Discussion question: is "may_have_focus" a better name?
     ///
     /// [`request_focus`]: struct.EventCtx.html#method.request_focus
     pub fn has_focus(&self) -> bool {
+        // The bloom filter we're checking can return false positives.
         let is_child = self
             .focus_widget
-            .map(|id| self.base_state.children.contains(&id))
+            .map(|id| self.base_state.children.may_contain(&id))
             .unwrap_or(false);
         is_child || self.focus_widget == Some(self.widget_id())
     }
 
-    /// The (leaf) focus status of a widget. See [`has_focus`].
+    /// The (leaf) focus status of a widget.
     ///
-    /// [`has_focus`]: struct.EventCtx.html#method.has_focus
+    /// See [`has_focus`](struct.EventCtx.html#method.has_focus).
     pub fn is_focused(&self) -> bool {
         self.focus_widget == Some(self.widget_id())
     }
@@ -282,7 +283,7 @@ impl<'a> EventCtx<'a> {
     ///
     /// This should only be called by a widget that currently has focus.
     pub fn focus_next(&mut self) {
-        if self.focus_widget == Some(self.widget_id()) {
+        if self.is_focused() {
             self.base_state.request_focus = Some(FocusChange::Next);
         } else {
             log::warn!("focus_next can only be called by the currently focused widget");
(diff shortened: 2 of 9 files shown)


druid/src/core.rs
```diff
@@ -446,7 +446,9 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
                 Target::Window(_) => Event::Command(cmd.clone()),
                 Target::Widget(id) if *id == child_ctx.widget_id() => Event::Command(cmd.clone()),
                 Target::Widget(id) => {
-                    recurse = child_ctx.base_state.children.contains(id);
+                    // Recurse when the target widget could be our descendant.
+                    // The bloom filter we're checking can return false positives.
```
Finnerale, line 450: > The bloom filter we're checking can return false positives.

I think the renaming to `may_contain` should make it clear enough, as this comment has to be applied to too many places to stay consistent (and those were definitly to many `to`s 😅 ).

druid/src/widget/widget.rs
```diff
@@ -17,7 +17,7 @@ use std::ops::{Deref, DerefMut};
 
 use super::prelude::*;
 
-/// A unique identifier for a single [`Widget`].
+/// A unique identifier for a chain of [`Widget`]s.
```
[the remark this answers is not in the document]
xStrom: By chain I mean several widgets connected together with no other widgets inbetween. I'm open for suggestions for better wording. The problem is that *a single* isn't correct. *A group* does not communicate the strict relationship in the tree.

druid/src/widget/widget.rs
```diff
@@ -29,15 +29,23 @@ use super::prelude::*;
 /// A widget can retrieve its id via methods on the various contexts, such as
 /// [`LifeCycleCtx::widget_id`].
 ///
+/// The `WidgetId` does not necessarily identify only a single [`Widget`] in the
+/// strictest sense. Container widgets (e.g. [`WidgetPod`]) can use the same `WidgetId`
+/// as their children and they will be treated as one for the purpose of events.
+///
+/// The `WidgetId` sharing is only valid for an unbroken chain of widgets.
+/// Only the direct parent of a child can use its child's `WidgetId`.
+/// It is invalid to share a `WidgetId` with a widget in any other spot in the hierarchy.
```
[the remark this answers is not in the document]
xStrom: By *share* I mean that two separate widgets use the same id. The `WidgetPod` is an excellent example.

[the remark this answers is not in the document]
xStrom: Another example is that the window state and the root widget of the window have the same id.

druid/src/contexts.rs
```diff
@@ -251,20 +251,21 @@ impl<'a> EventCtx<'a> {
     /// hierarchy, all ancestors of that leaf widget are also invoked with
     /// `FocusChanged(true)`.
     ///
-    /// Discussion question: is "is_focused" a better name?
+    /// Discussion question: is "maybe_has_focus" a better name?
     ///
     /// [`request_focus`]: struct.EventCtx.html#method.request_focus
     pub fn has_focus(&self) -> bool {
+        // The bloom filter we're checking can return false positives.
```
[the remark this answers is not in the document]
xStrom: `may_have_focus` would be the spelling. Anyway I left the `has_focus` method name and comments unchanged because those will be tackled in #811.

druid/src/core.rs
xStrom: Just the name `may_contain` will make it clear enough to someone who already knows the function or will go read its docs. Others who are just glancing at the code at the call site won't necessarily know.

I also don't think the three places this comment was added to is too many. Once we're up to 10 or something we can reconsider.

druid/src/widget/widget.rs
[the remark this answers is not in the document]
xStrom: I changed this to *A unique identifier for `Widget`s.* It's more subtle and won't cause immediate confusion. Then deeper into the description people can read about the multiple widgets.

druid/src/widget/widget.rs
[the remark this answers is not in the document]
xStrom: I rewrote this section and removed the word *share* which was causing confusion.

xStrom:
On further thinking I moved the `WidgetId` docs into a more basic internal comment. It's useful info for druid internal development but can be viewed as an implementation detail and we shouldn't advertise this as something people can do, because then we'll have to support it.

Finnerale:
The force-pushing makes it a bit difficult to see what you changed 😓 
Did you remove your changes to the `WidgetId` docs? (except for the not-doc comment)
I think the chain explanation you had there was quite useful to understand the internal use.
Making it an internal comment is better than nothing even though I'm kinda sure I would miss it when looking for an explanation 😅 . Wish Rust had an 'implementation-doc` feature.

xStrom:
Yeah I removed the public doc changes for `WidgetId` and replaced it with an internal code comment that's quite brief.

druid/src/widget/widget.rs
```diff
@@ -46,6 +46,11 @@ use super::prelude::*;
 /// [`LifeCycleCtx::widget_id`]: struct.LifeCycleCtx.html#method.widget_id
 /// [`WidgetExt::with_id`]: trait.WidgetExt.html#method.with_id
 /// [`IdentityWrapper`]: widget/struct.IdentityWrapper.html
+//
+// The widget id does not necessarily identify only a single widget
```
cmyr, line 50: This is true, but might be confusing.

`WidgetPod` is not really a widget. Or: it _can_ be, but that's sort of also confusing.

The best way to think of `WidgetPod` is that it is the base class of `Widget`. In certain instances, we can have a widget that has another widget as a child, without a `WidgetPod`; but I would also almost think of this as a sort of subclassing, a way of modifying the child's behaviour in some way; the two widgets are really acting as a single until.

In a sense this would be more clear if `WidgetPod` was named `Widget`, and `Widget` was named `Component`; conceptually, it can be helpful to think of a 'widget' as being all of the `Widget`s between one `WidgetPod` and it's closest descendent `WidgetPod`.

druid/src/widget/widget.rs
```diff
@@ -46,6 +46,11 @@ use super::prelude::*;
...
+// in the strictest sense. Internally WidgetPod uses its child's id,
```
cmyr, line 51: Kind of. In the general case, the `WidgetPod` is the source of the id; it only uses the child's id if that child has been given an explicit id via `IdentityWrapper`, which is only done when the user wants to know the id ahead of time.

So it's more like: The `WidgetId` will be the same for all `Widget`s that share the same nearest ancestor `WidgetPod`.

druid/src/bloom.rs
```diff
@@ -71,10 +71,12 @@ impl<T: ?Sized + Hash> Bloom<T> {
         self.entry_count += 1;
     }
 
-    /// Return whether an item exists in the filter.
+    /// Returns `true` if the item may have been added to the filter.
     ///
     /// This can return false positives, but never false negatives.
-    pub fn contains(&self, item: &T) -> bool {
+    /// Thus `true` means that the item may have been added - or not,
+    /// while `false` means that the item has definitely not been added.
```
cmyr, line 78: I think these two lines are a strict restating of the preceding line; I think between the changed method name and the changed first line comment, this should be clear to our expected audience.

druid/src/contexts.rs
```diff
@@ -238,7 +238,7 @@ impl<'a> EventCtx<'a> {
         self.is_handled
     }
 
-    /// The focus status of a widget.
+    /// The (tree) focus status of a widget.
```
cmyr, line 241: Maybe also something like,

```suggestion
    /// The (tree) focus status of a widget.
    ///
    /// This is used principally for event routing; it is `true` iff
    /// this widget or one of its descendants may have focus.
```

druid/src/widget/widget.rs
xStrom: Yeah my understanding about this is getting refined every day. I was thinking about proposing renaming `WidgetPod` to `Widget` just earlier today. I think this is something worth discussing in length in zulip some day.

The origin story of this comment is me finding that the window base state copies the root `WidgetPod` id which means there's a duplicated id in the event path. Maybe I'll just change this comment here to mention only that, or maybe even move that comment to window instead.

druid/src/bloom.rs
xStrom: I think it's worth stating it twice with different wording. The mistakes that result from not understanding this are extremely subtle and hard to catch.

druid/src/widget/widget.rs
cmyr: yea, that's slightly hacky; iirc the root window's basestate really just exists because we need _some_ initial basestate in order to create the various contexts; we do check some things on it after a cycle, but we could just as easily check those on the root widget.

xStrom:
I removed the note I originally had added to `WidgetId` about multiple use. Instead I may look into improving the implementation in the future, but for documentation I think it's good to promote strictly unique usage.

I also rewrote the documentation for focus related methods/variants. Some of that stuff was strictly expired and no longer reflected reality.

cmyr:
We could also do something like prefix them with `_`.

Finnerale:
I've created a new issue (#826) regarding the internal variants of `LifeCycle` so that this can get merged.

PR #819, opened 2020-04-09
Make has_focus reliable.

xStrom:
This is a simpler implementation of parts of #811.

* `has_focus` no longer returns false positives and can be relied on.
* Focus cycling now works even when starting from a non-registered-for-focus widget that merely called `request_focus`. The cycle is still only between registered-for-focus widgets.
* Fixed propagation of `RouteFocusChanged` and `FocusChanged`. Previously they were propagated only to the focus losing widget, unless there was no previously focused widget. This behavior stumbled into somewhat working because the `request_focus` in `BaseState` didn't get cleared and was merged up again generating a new `RouteFocusChanged` where `old` and `new` are the same. This allowed clearing of `request_focus` but also generated a `FocusChanged(false)` to the widget that never received `FocusChanged(true)`. When a 3rd widget gets focus, the 2nd one receives another `FocusChanged(false)`.

### DIFF
--- a/druid/src/core.rs
+++ b/druid/src/core.rs
@@ -86,6 +86,10 @@ pub(crate) struct BaseState {
     /// Any descendant is active.
     has_active: bool,
 
+    /// In the focused path, starting from window and ending at the focused widget.
+    /// Descendants of the focused widget are not in the focused path.
+    pub(crate) has_focus: bool,
+
     /// Any descendant has requested an animation frame.
     pub(crate) request_anim: bool,
 
@@ -514,19 +518,25 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
                 };
 
                 if let Some(change) = this_changed {
+                    self.state.has_focus = change;
                     let event = LifeCycle::FocusChanged(change);
                     self.inner.lifecycle(ctx, &event, data, env);
                     false
                 } else {
+                    self.state.has_focus = false;
                     // Recurse when the target widgets could be our descendants.
                     // The bloom filter we're checking can return false positives.
-                    old.map(|id| self.state.children.may_contain(&id))
-                        .or_else(|| new.map(|id| self.state.children.may_contain(&id)))
-                        .unwrap_or(false)
+                    match (old, new) {
+                        (Some(old), _) if self.state.children.may_contain(old) => true,
+                        (_, Some(new)) if self.state.children.may_contain(new) => true,
+                        _ => false,
+                    }
                 }
             }
             LifeCycle::FocusChanged(_) => {
-                self.state.request_focus = None;
+                // We are a descendant of a widget that has/had focus
+                self.state.has_focus = false; // Descendants don't inherit focus
+                self.state.request_focus = None; // Clear the handled request
                 true
             }
             #[cfg(test)]
@@ -624,6 +634,7 @@ impl BaseState {
             needs_layout: false,
             is_active: false,
             has_active: false,
+            has_focus: false,
             request_anim: false,
             request_timer: false,
             request_focus: None,
@@ -640,6 +651,7 @@ impl BaseState {
         self.request_anim |= child_state.request_anim;
         self.request_timer |= child_state.request_timer;
         self.has_active |= child_state.has_active;
+        self.has_focus |= child_state.has_focus;
         self.children_changed |= child_state.children_changed;
         self.request_focus = self.request_focus.or(child_state.request_focus);
     }
--- a/druid/src/core.rs
+++ b/druid/src/core.rs
@@ -518,9 +518,12 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
                 };
 
                 if let Some(change) = this_changed {
-                    self.state.has_focus = change;
-                    let event = LifeCycle::FocusChanged(change);
-                    self.inner.lifecycle(ctx, &event, data, env);
+                    // Only send FocusChanged in case there's actual change
+                    if old != new {
+                        self.state.has_focus = change;
+                        let event = LifeCycle::FocusChanged(change);
+                        self.inner.lifecycle(ctx, &event, data, env);
+                    }
                     false
                 } else {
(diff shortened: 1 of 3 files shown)


xStrom:
Added another fix for when there's only a single widget registered for focus.

Cycling to the "next" widget would cause a `RouteFocusChanged` with the same `old` and `new` but only a single `FocusChanged` was generated per event, `FocusChanged(false)`.

Now a `FocusChanged` event gets generated only when there's actual change.

xStrom:
I added a change that `FocusChanged` now only gets sent to the widget that specifically has their `is_focused` state changed. That is, if that widget just passes all their events down to children, `WidgetPod` will make sure to stop the propagation of `FocusChanged`. More discussion about this on [zulip](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/Focus.20improvements).

druid/src/core.rs
```diff
@@ -513,18 +517,26 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
                 };
 
                 if let Some(change) = this_changed {
-                    let event = LifeCycle::FocusChanged(change);
-                    self.inner.lifecycle(ctx, &event, data, env);
+                    // Only send FocusChanged in case there's actual change
+                    if old != new {
```
cmyr, line 521: if this is happening, maybe we should suppress it earlier, like at the window level?

druid/src/window.rs
```diff
@@ -324,22 +324,33 @@ impl<T: Data> Window<T> {
...
+                    let new_idx = if forward {
+                        (idx + 1) % len
+                    } else {
+                        (idx + len - 1) % len
+                    };
+                    self.focus_chain()[new_idx]
+                })
+                .or_else(|| {
+                    // If the currently focused widget isn't in the focus chain,
+                    // then we'll just return the first entry of the chain, if any.
```
cmyr, line 350: this highlights a definite little rough spot. What should `tab` do if nothing is focused? A reasonable argument is that it should send focus to the first item in the focus chain.

This is related to what should happen when something that isn't in the focus chain is focused, and we end up in this code path.

Realistically, this shouldn't happen. To get here, a widget has to explicitly request next or previous focus. A widget that is doing that has also had the opportunity to register to participate in focus. Doing one without the other feels like maybe an API violation?

This again touches, to me, on the idea of 'ambient focus'. How I interpret this is basically that if no other widget has focus, the root widget does. A `<tab>` from the root widget should focus the next focusable widget, but the root widget itself should not be in the focus chain; it should only receive focus if another widget resigns focus without trying to pass it on, as might happen when someone presses `<esc>`.

So with all that said, I think that this implementation is reasonable. A widget has asked for next/previous focus, so we should focus something.

This would also then work for the root widget, if it had focus; if it sent `focus_next` then that would just grab the first item in the chain.

Given this, should 'previous' maybe take the last item?

druid/src/window.rs
```diff
@@ -324,22 +324,33 @@ impl<T: Data> Window<T> {
...
-                }),
-            FocusChange::Previous => self
-                .focus
-                .and_then(|id| self.focus_chain().iter().position(|i| i == &id))
+            FocusChange::Next => self.widget_from_focus_chain(true),
+            FocusChange::Previous => self.widget_from_focus_chain(false),
+        }
+    }
+
+    fn widget_from_focus_chain(&self, forward: bool) -> Option<WidgetId> {
```
cmyr, line 332: 👍

druid/src/core.rs
```diff
@@ -513,18 +517,26 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
...
+                        (Some(old), _) if self.state.children.contains(old) => true,
+                        (_, Some(new)) if self.state.children.contains(new) => true,
+                        _ => false,
+                    }
                 }
             }
             LifeCycle::FocusChanged(_) => {
-                self.state.request_focus = None;
-                true
+                // We are a descendant of a widget that has/had focus.
```
cmyr, line 537: should we be clearing this somewhere else, now?

druid/src/core.rs
xStrom: The problem is that we need to come down here to clear `state.request_focus`.

druid/src/core.rs
xStrom: There's nothing to clear because `request_focus` only gets set starting from the widget that does the request. This code here only executes for descendants of those widgets.

cmyr: ah I see and we were already clearing it above.

xStrom:
Regarding tests, yes I've been thinking about that as well. Because some of this stuff is hard to grasp from just glancing at the code.

I'm planning on writing test code for widget id verification soon, because I know for a fact that some of that stuff is broken and I want it to remain fixed.

After I've had the experience with the widget id tests I can rotate back to *focus* and write some more robust tests.

druid/src/window.rs
xStrom: Regarding `focus_prev` yes I think it should take the last item, I'll change that. The reason is that I just checked that Chrome does work that way, so it's probably what users expect.

As for a widget being focused without registering for focus, I think this is not an API violation and something that is useful to have. It exists in HTML and can be seen in [this example](https://jsfiddle.net/vtjd961k/) I set up. The use case would be a form that has rarely used fields. Then you can have a textbox which can be focused when clicked to change its contents, but in 95% of use cases the user doesn't need to modify it so it doesn't need to be in the tab cycle. However this would make life more difficult for keyboard only users. 

This starts touching upon accessibility in general and being able to tab through buttons and dropdowns etc. I think there's major value in supporting keyboard navigation automatically in all of druid's included widgets. This is a rather large topic though. I think we should start a new issue to talk about this. It's not super clear what the answer is to add any ambient focus support in this PR right now.

cmyr: your logic here makes sense to me; and I think ambient focus makes sense as a separate patch.

I agree about better support for keyboard navigation, and also yes it's a big one.

PR #820, opened 2020-04-09
Make request_focus a no-op when the widget is already focused.

xStrom:
It doesn't really make sense to request focus when you're already focused, but previously you had to check for `is_focused` yourself like `TextBox` or you would run into bugs because `FocusChanged(false)` and `FocusChanged(true)` wouldn't be sent to the same widget during one event run.

This PR makes calling `request_focus` care-free.

### DIFF
--- a/druid/src/contexts.rs
+++ b/druid/src/contexts.rs
@@ -280,11 +280,16 @@ impl<'a> EventCtx<'a> {
 
     /// Request keyboard focus.
     ///
+    /// Calling this when the widget is already focused does nothing.
+    ///
     /// See [`is_focused`] for more information about focus.
     ///
     /// [`is_focused`]: struct.EventCtx.html#method.is_focused
     pub fn request_focus(&mut self) {
-        self.base_state.request_focus = Some(FocusChange::Focus(self.widget_id()));
+        let id = self.widget_id();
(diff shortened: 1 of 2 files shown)


druid/src/contexts.rs
```diff
@@ -271,11 +271,16 @@ impl<'a> EventCtx<'a> {
...
     ///
+    /// Calling this when the widget is already focused does nothing.
+    ///
     /// See [`has_focus`] for more information.
     ///
     /// [`has_focus`]: struct.EventCtx.html#method.has_focus
     pub fn request_focus(&mut self) {
-        self.base_state.request_focus = Some(FocusChange::Focus(self.widget_id()));
+        let id = self.widget_id();
+        if self.focus_widget.filter(|focus| focus == &id) == None {
```
cmyr, line 281: not a blocker, but I might write this as `if Some(self.widget_id()) == self.focus_widget {`?

cmyr: ... or we can wait until we have `is_focused`.

xStrom: You mean `if Some(self.widget_id()) != self.focus_widget {` I presume, because we want to request focus when `self.focus_widget` is either `None` or `Some(other_id)`.

PR #821, opened 2020-04-09
Add Event::MouseLeave for windows

teddemunnik:
Adds a new new event `MouseLeft`, and implemented it for windows.
The purpose for this event is to give a possibility of the OS to notifiy the druid app that it is no longer going to receive mouse events for some time. Either because the mouse has left the window rect, or because another app is blocking the mouse cursor.

### DIFF
--- a/druid-shell/src/platform/windows/window.rs
+++ b/druid-shell/src/platform/windows/window.rs
@@ -607,7 +607,7 @@ impl WndProc for MyWndProc {
                     let y = HIWORD(lparam as u32) as i16 as i32;
 
                     // When the mouse first enters the window client rect we need to register for the
-                    // WM_MOUSELEAVE event. Note that WM_MOUSEMOVE is called even when the change
+                    // WM_MOUSELEAVE event. Note that WM_MOUSEMOVE is also called even when the
                     // window under the cursor changes without moving the mouse, for example when
                     // our window is first opened under the mouse cursor.
                     if !s.has_mouse_focus && is_point_in_client_rect(hwnd, x, y) {
--- a/druid-shell/src/platform/windows/window.rs
+++ b/druid-shell/src/platform/windows/window.rs
@@ -173,6 +173,8 @@ struct WndState {
     // capture. When the first mouse button is down on our window we enter
     // capture, and we hold it until the last mouse button is up.
     captured_mouse_buttons: u32,
+    // Is this window the topmost window under the mouse cursor
+    has_mouse_focus: bool,
     //TODO: track surrogate orphan
 }
 
@@ -226,6 +228,22 @@ fn get_mod_state() -> KeyModifiers {
             mod_state.shift = true;
         }
         mod_state
+    }
+}
+
+fn is_point_in_client_rect(hwnd: HWND, x: i32, y: i32) -> bool {
+    unsafe {
+        let mut client_rect = mem::MaybeUninit::uninit();
+        if GetClientRect(hwnd, client_rect.as_mut_ptr()) == FALSE {
+            warn!(
+                "failed to get client rect: {}",
+                Error::Hr(HRESULT_FROM_WIN32(GetLastError()))
+            );
+            return false;
+        }
+        let client_rect = client_rect.assume_init();
+        let mouse_point = POINT { x, y };
+        PtInRect(&client_rect, mouse_point) != FALSE
     }
 }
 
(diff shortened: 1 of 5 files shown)


xStrom:
Maybe there's a better name than `MouseLeft`? That name just immediately makes me think of the left mouse button. :thinking: 

`MouseLost` could be similar length but less ambiguous. There's also `MouseExited` which seems a bit more cumbersome but still clear.

teddemunnik:
Good point, didn't think of the confusion with left mouse button.. I started with `MouseLeave`, but then noticed that `MouseMoved` is past tense. `MouseLost` sounds good to me!

raphlinus:
I think we generally follow web terminology (in this case, [mouseleave](https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseleave_event)) unless there is compelling reason to do otherwise.

xStrom:
Following web terms would be fine, except we don't with e.g. `MouseMoved` while web has [mousemove](https://developer.mozilla.org/en-US/docs/Web/API/Element/mousemove_event).

raphlinus:
> Following web terms would be fine, except we don't with e.g. `MouseMoved` while web has [mousemove](https://developer.mozilla.org/en-US/docs/Web/API/Element/mousemove_event).

Indeed. Now I regret that.

xStrom:
Well it's not too late to change it to `MouseMove` but then there's the question of `Event::WindowConnected`. Might as well change that to `WindowConnect` then and have everything consistent.

teddemunnik:
Alright! `MouseMove` and `MouseLeave` sounds perfect to me as well. How about I change this PR to be `MouseLeave` and we potentially rename `MouseMove` and `WindowConnected` in a future change then?

xStrom:
Yes I agree with that naming plan @teddemunnik, split it into two PRs.

druid-shell/src/platform/windows/window.rs
```diff
@@ -595,6 +613,29 @@ impl WndProc for MyWndProc {
                     let s = s.as_mut().unwrap();
                     let x = LOWORD(lparam as u32) as i16 as i32;
                     let y = HIWORD(lparam as u32) as i16 as i32;
+
+                    // When the mouse first enters the window client rect we need to register for the
+                    // WM_MOUSELEAVE event. Note that WM_MOUSEMOVE is called even when the change
+                    // window under the cursor changes without moving the mouse, for example when
```
xStrom, line 619: Is this an accidental extra word? *when the ~change~ window under the cursor changes*

teddemunnik: You're right, fixed

druid-shell/src/platform/windows/window.rs
```diff
@@ -595,6 +613,29 @@ impl WndProc for MyWndProc {
...
                     let y = HIWORD(lparam as u32) as i16 as i32;
+
+                    // When the mouse first enters the window client rect we need to register for the
+                    // WM_MOUSELEAVE event. Note that WM_MOUSEMOVE is also called even when the
+                    // window under the cursor changes without moving the mouse, for example when
+                    // our window is first opened under the mouse cursor.
+                    if !s.has_mouse_focus && is_point_in_client_rect(hwnd, x, y) {
+                        s.has_mouse_focus = true;
+                        let mut desc = TRACKMOUSEEVENT {
+                            cbSize: std::mem::size_of::<TRACKMOUSEEVENT>() as DWORD,
```
xStrom, line 624: Let's use just `mem::size_of` instead of `std::mem::size_of`.

druid-shell/src/platform/windows/window.rs
```diff
@@ -595,6 +613,29 @@ impl WndProc for MyWndProc {
...
+                            dwFlags: TME_LEAVE,
+                            hwndTrack: hwnd,
+                            dwHoverTime: HOVER_DEFAULT,
+                        };
+                        unsafe {
+                            if TrackMouseEvent(&mut desc) == FALSE {
+                                warn!(
+                                    "failed to TrackMouseEvent: {}",
+                                    Error::Hr(HRESULT_FROM_WIN32(GetLastError()))
+                                );
```
xStrom, line 634: This is an interesting failure possibility. I think it might be better to set `s.has_mouse_focus` to `false` so that we'll try again on the next move event. Otherwise we'll forever be without this tracking.

druid/src/event.rs
```diff
@@ -84,6 +84,11 @@ pub enum Event {
     ///
     /// [`set_cursor`]: struct.EventCtx.html#method.set_cursor
     MouseMoved(MouseEvent),
+    /// Called when the mouse has left the application area.
+    ///
+    /// The `MouseLeave` event is propagated to the active widget, if
+    /// there is one, otherwise to hot widgets (see `HotChanged`).
+    MouseLeave,
```
xStrom, line 91: *~the application area~ the window*

Also the event will be propagated to both active and hot widgets. Multiple can be active, and hot widgets will receive this regardless of whether there are active.

teddemunnik: It's true, I copied the comment and behavior from `MouseMove`, but I think they don't match there either?

druid-shell/src/platform/windows/window.rs
teddemunnik: I'm curious in what circumstances this call would fail, besides being passed wrong arguments. But good call with not setting `has_mouse_focus`.

druid/src/event.rs
xStrom: Probably remnants of an era where druid was single window or something.

xStrom:
As per the discussion in zulip let's make `MouseLeave` internal. Pending #826 this doesn't actually mean anything in terms of code. However let's at least add a note to the `Event::MouseLeave` documentation stating that it is internal and `LifeCycle::HotChanged` should be used instead.

PR #824, opened 2020-04-09
Update to piet v0.0.12

cmyr:
This gives us blurred rects as well as multi-line text.

this unblocks #759

### DIFF
--- a/druid/src/widget/textbox.rs
+++ b/druid/src/widget/textbox.rs
@@ -89,7 +89,7 @@ impl TextBox {
             .unwrap();
 
         piet_text
-            .new_text_layout(&font, &text.to_string())
+            .new_text_layout(&font, &text.to_string(), std::f64::INFINITY)
             .build()
             .unwrap()
     }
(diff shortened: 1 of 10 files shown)


xStrom:
In tests/harness.rs:

```rust
        let mut harness = Harness {
            piet,
            inner,
            window_size: DEFAULT_SIZE,
        };
        f(&mut harness);
        harness.piet.finish().expect("piet finish failed"); // <--- Add this line
```

druid/src/widget/textbox.rs
```diff
@@ -89,7 +89,7 @@ impl TextBox {
             .unwrap();
 
         piet_text
-            .new_text_layout(&font, &text.to_string())
+            .new_text_layout(&font, &text.to_string(), std::f64::INFINITY)
```
futurepaul, line 92: When I hit enter in the hello.rs example, the cursor gets into a weird state (previously it would just print the box character). I'm on linux haven't tried it on mac yet.

xStrom: For me the cursor goes to the beginning of the textbox and the label is multi-line. Certainly broken, but probably as simple as disabling enter for now.

cmyr: okay, I'm going to patch that here as well; we will suppress enter in the textbox widget. (this isn't perfect, since you could still paste one in, but I think it's a good initial step. We need to figure out a whole whack of new text stuff, now.)

PR #825, opened 2020-04-10
Rename Event::MouseMoved to Event::MouseMove

teddemunnik:
As discussed in #821, this renames `MouseMoved` to `MouseMove` for consistency with web API's.

### DIFF
--- a/druid/src/contexts.rs
+++ b/druid/src/contexts.rs
@@ -177,7 +177,7 @@ impl<'a> EventCtx<'a> {
     /// event handler, but can also be called in response to other events,
     /// for example pressing a key to change the behavior of a widget.
     ///
-    /// [`MouseMove`]: enum.Event.html#variant.MouseDown
+    /// [`MouseMove`]: enum.Event.html#variant.MouseMove
     pub fn set_cursor(&mut self, cursor: &Cursor) {
         *self.cursor = Some(cursor.clone());
     }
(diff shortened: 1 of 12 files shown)


druid/src/contexts.rs
```diff
@@ -173,11 +173,11 @@ impl<'a> EventCtx<'a> {
     /// only has the effect of the last one (ie no need to worry about
     /// flashing).
     ///
-    /// This method is expected to be called mostly from the [`MouseMoved`]
+    /// This method is expected to be called mostly from the [`MouseMove`]
     /// event handler, but can also be called in response to other events,
     /// for example pressing a key to change the behavior of a widget.
     ///
-    /// [`MouseMoved`]: enum.Event.html#variant.MouseDown
+    /// [`MouseMove`]: enum.Event.html#variant.MouseDown
```
xStrom, line 180: Let's fix this link to `enum.Event.html#variant.MouseMove` while we're here.

teddemunnik: Sure, fixed :)

cmyr:
I wonder if we should start keeping a list of breaking changes somewhere for release notes. 🤔

xStrom:
Yeah I think it might be a good idea to start a `CHANGELOG.md` which gets modified with PRs. Every PR can then add a short description of its change to the changelog with a link to the PR. Changes can be grouped by e.g. breaking changes/new features/bug fixes. Generally would be pretty easy to gather the changelog this way. The only somewhat tricky thing is that whenever some behavior gets reverted or modified and there hasn't been a druid release yet, then this revert/modify PR should also remove the older changelog entry.

cmyr:
for now i'm going to use a label for breaking changes, which will solve one of my problems.

PR #830, opened 2020-04-12
Log if no WidgetPod::set_layout_rect before event

cmyr:
This will hopefully make an easy mistake harder to make.

motivated by [this discussion on zulip](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/children.20widgets.20could.20not.20receive.20events.20from.20parent/near/193672231)

### DIFF
--- a/druid/src/core.rs
+++ b/druid/src/core.rs
@@ -66,7 +66,10 @@ pub struct WidgetPod<T, W> {
 #[derive(Clone)]
 pub(crate) struct BaseState {
     pub(crate) id: WidgetId,
-    pub(crate) layout_rect: Rect,
+    /// The frame of this widget in its parents coordinate space.
+    /// This should always be set; it is only an `Option` so that we
+    /// can more easily track (and help debug) if it hasn't been set.
+    layout_rect: Option<Rect>,
     /// The insets applied to the layout rect to generate the paint rect.
     /// In general, these will be zero; the exception is for things like
     /// drop shadows or overflowing text.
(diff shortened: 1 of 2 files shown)


druid/src/core.rs
```diff
@@ -377,7 +380,18 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
...
+        let rect = match child_ctx.base_state.layout_rect {
+            Some(rect) => rect,
+            None => {
+                log::warn!(
+                    "WidgetPod received an event without having been laid out. \
+                This likely indicates a missed call to set_layout_rect."
+                );
+                Rect::ZERO
+            }
+        };
```
Finnerale, line 393: How about
```suggestion
        if child_ctx.base_state.layout_rect.is_none() {
            log::warn!(
                "WidgetPod received an event without having been laid out. \
                This likely indicates a missed call to set_layout_rect."
            );
        }
        let rect = child_ctx.base_state.layout_rect.unwrap_or_default();
```

druid/src/core.rs
```diff
@@ -176,20 +179,20 @@ impl<T, W: Widget<T>> WidgetPod<T, W> {
     /// Intended to be called on child widget in container's `layout`
     /// implementation.
     pub fn set_layout_rect(&mut self, layout_rect: Rect) {
-        self.state.layout_rect = layout_rect;
+        self.state.layout_rect = layout_rect.into();
```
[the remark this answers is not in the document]
cmyr: fair point!

druid/src/core.rs
cmyr: I don't think I have clearly prefer your version, but since you did all that typing I'm happy to take it. ;)

cmyr:
@Finnerale good catch, I've updated so that we don't log on `WindowConnected` and `Size`, which can arrive before a layout pass 😅

Finnerale:
@cmyr Did you push it?

cmyr:
just now. :)

PR #831, opened 2020-04-12
Associate timers with widget ids

### LINKED ISSUE #494: associate timers with widget ids
currently we route timers to all widgets that have ever requested a timer; we can now store the id of the requesting widget along with the timer, and route things that way.

If someone is interested in taking this, it shouldn't be too tricky, and it will be a nice excuse to explore some of druid's internals.

sjoshid:
Closes #494


xStrom:
I haven't looked at the code closely yet, but can't we do this association without `BaseState`? `BaseState` gets replicated a lot. I'm thinking of something like sending the `WidgetId` in with `self.window.request_timer()` and then the `WidgetId` gets stored close to wherever the `TimerToken` is stored. When the timer fires, we can have a `RouteTimer{WidgetId, TimerToken}` internal event that gets translated into the classic `Timer` event when it reaches the target widget.

I'll have more time to look into this tomorrow hopefully, but just wanted to get that thought out there.

druid/src/contexts.rs
```diff
@@ -349,7 +352,13 @@ impl<'a> EventCtx<'a> {
     /// request with the event.
     pub fn request_timer(&mut self, deadline: Instant) -> TimerToken {
         self.base_state.request_timer = true;
-        self.window.request_timer(deadline)
+        let timer_token = self.window.request_timer(deadline);
+        self.base_state.add_timer(timer_token);
+        timer_token
+    }
+
+    pub fn remove_timer(&mut self, timer_token: &TimerToken) {
```
sjoshid: After widgets are done processing timer events, should we all ask them to remove map entry?
This is not used anywhere in examples. 
I think framework needs to remove it rather than forcing widgets to remove it themselves.

sjoshid:
> WidgetId gets stored close to wherever the TimerToken is stored.

That was my first attempt. `TimerToken` is stored in platform window. After some discussion, we decided not to change druid shell. 

> When the timer fires, we can have a RouteTimer{WidgetId, TimerToken} internal event that gets translated into the classic Timer event when it reaches the target widget

The objective of this was to do aggressive traversing of `Timer` event. We can't let the event reach every widget to find out if it is a target or not. So I let 'timer -> widget id' entries `merge_up` to root. And when the event is fired, we use this map along with `maybe_contains` to determine if we need to recurse.

xStrom:
Yeah druid-shell shouldn't be modified, but we could probably store the `HashMap<TimerToken, WidgetId>` in the druid window. :thinking:

As for agressive traversing, you can still do that if you just know the `WidgetId` at the root, because you can still use `may_contain` to check if a path may contain the `WidgetId`.

sjoshid:
Whatever you said is what this PR does.

druid/src/contexts.rs
sjoshid: After merging up, Im clearing the map of BaseState. So remove is not needed when event is handled.

sjoshid:
You are right. I cleaned up `BaseState` timer but not `Window`. Will fix this asap.

sjoshid:
To remove timers from window, I check `is_handled` flag at the end of `event()` in window. 
This has one drawback, all handlers need to call `ctx.set_handled()`.

Is that acceptable?

xStrom:
I think it's fine to set `is_handled` when the timer event reaches its destination.
Something like the following:
```rust
recurse = if *widget_id != child_ctx.base_state.id {
    child_ctx.base_state.children.may_contain(widget_id);
} else {
    child_ctx.is_handled = true;
    true
};
```

That can be useful for preventing further propagation, however this is not how the removal should happen in the window, because the widget who requested the timer may not exist anymore. I think the window `HashMap` entry should be removed every time a `Timer` event comes to the window, regardless if it was handled by a widget or the widget no longer exists.

sjoshid:
> I think the window HashMap entry should be removed every time a Timer event comes to the window, regardless if it was handled by a widget or the widget no longer exists.

Sounds good. With that in mind, this should work at the end of `event` in window. 
```
if let Event::Timer(token) = event {
    self.timers.remove(&token);
}
```

sjoshid:
I spoke too early. With the change I suggested, the `timer` example crashes. 
I will look in to more details.

xStrom:
`TimerToken` is reused when available, so make sure you do the removal before you extend it with the new timer requests, because that new timer request set may contain the same id that you're removing.

sjoshid:
You are absolutely right. Looks like tokens are reused only on Windows. Linux/Mac seem to work as token aren't reused there.

xStrom:
Okay this looks functional now.

There is some more work opportunity here though. Instead of the window sending `Event::Timer` it should create a new `Event::Internal(InternalEvent::Timer{ widget_id, token })` and send that, similarly to how `LifeCycle::Internal(InternalLifeCycle::RouteFocusChanged{ widget_id, widget_id })` works. Then there's no need to pass the window's `HashMap` reference via `EventCtx` and inside `WidgetPod` we can handle the internal event. When it finally reaches the target widget we synthesize the real `Event::Timer`.

Let me know if this is something you wish to continue working on. It's completely fine if you want to just get this merged and be done for now. :+1:

sjoshid:
I prefer doing this the right way rather than just merging and creating a technical debt. Besides hopefully I'll learn something more.
I will look more in to your suggestion. 
Thanks for asking.

cmyr:
This approach is pretty good, but It isn't quite what i had in mind; as mentioned by @xStrom, I think it would be even cleaner if we used an internal `RouteTimer` event to deliver the timer to only the correct widget. The advantage of this method would be that widgets would no longer see timers that were not intended for them, and so they could be less careful about always stashing and checking timer tokens.

sjoshid:
I agree. Im working on the approach you and @xStrom are talking about. But isn't
> widgets would no longer see timers that were not intended for them

already working? I ask because that was THE goal I was working towards and I thought I implemented it. 🤔

cmyr:
not if a widget has children; we will recurse into a container widget with the `Timer` event, which is not actually for the container but for the child.

druid/src/event.rs
```diff
@@ -147,6 +147,7 @@ pub enum InternalEvent {
     MouseLeave,
     /// A command still in the process of being dispatched.
     TargetedCommand(Target, Command),
+    RouteTimer(TimerToken, WidgetId),
```
xStrom, line 150: Add a comment here describing this variant. Something simple like *Used to route timer events.* would work.

druid/src/core.rs
```diff
@@ -591,10 +602,7 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
                 recurse = had_active || child_ctx.base_state.is_hot;
                 Event::Zoom(*zoom)
             }
-            Event::Timer(id) => {
-                recurse = child_ctx.base_state.request_timer;
-                Event::Timer(*id)
-            }
+            Event::Timer(token) => Event::Timer(*token),
```
xStrom, line 605: Add `recurse = false;` here because if we get this then we're already a descendant of the targeted widget.

druid/src/window.rs
```diff
@@ -215,6 +226,17 @@ impl<T: Data> Window<T> {
 
         self.post_event_processing(queue, data, env, false);
 
+        //In some platforms, timer tokens are reused. So it is necessary to remove token from
+        //window's timer before adding base state's timers to it.
```
xStrom, line 230: *~In~ **On** some platforms, timer tokens are reused. So it is necessary to remove **the** token from **the** window's timer **map** before adding ~base state's timers~ **new tokens** to it.*

druid/src/window.rs
```diff
@@ -215,6 +226,17 @@ impl<T: Data> Window<T> {
...
+        if let Event::Internal(InternalEvent::RouteTimer(token, _)) = event {
+            self.timers.remove(&token);
+        }
+
+        //If at least one widget requested timer, collect those timers from widgets and add to window's timers map.
```
xStrom, line 235: *If at least one widget requested **a** timer, ~collect those timers from widgets and~ add **all the requested timers** to window's timers map.*

druid/src/core.rs
sjoshid: Good catch. Missed this important change.  :(
Thanks.

Issue #787, opened 2020-03-31
Add .on_keyup method, similar to .on_click


futurepaul:
I want to do a search-as-you-type interaction. Currently I'm sending a Command with the query using a Button, so I'd like to change that to sending a Command every time a key is pressed. I also want to intercept keyboard "enter" events and use that to send a different command.

Would it be appropriate to implement this in a similar way as the `.on_click` method? I imagine there's also a way to do this with lenses but I'll admit I haven't fully explored that yet. That wouldn't help with intercepting enter, however.

cmyr:
Play around with it? it isn't immediately clear to me, since focus is more complicated than mouse stuff. What you're describing here does sound like a custom widget, though?

One thing I've been thinking about is whether you should be able to request 'ambient' focus, that is have some root widget that will receive key events when nobody else is focused.

Finnerale:
I'm not sure how useful `.on_keyup` would be.
The use cases that come to my mind are pressing enter, which would probably deserve its own method on `TextBox`, and keyboard shortcuts which also should be a separate concept (and would probably require such 'ambient' focus)?
For something like an instant search `Controller` should be perfectly adequate.

PR #832, opened 2020-04-12
Add docs for Painter & Controller

cmyr:
This is the first part of the major section going over custom
widgets and the general widget model.

### DIFF
--- a/docs/book_examples/src/custom_widgets_md.rs
+++ b/docs/book_examples/src/custom_widgets_md.rs
@@ -1,5 +1,9 @@
 use druid::widget::{Controller, Label, Painter, SizedBox, TextBox};
-use druid::{Color, Env, Event, EventCtx, KeyCode, PaintCtx, RenderContext, Widget, WidgetExt};
+use druid::{
+    Color, Env, Event, EventCtx, KeyCode, PaintCtx, RenderContext, Selector, TimerToken, Widget,
+    WidgetExt,
+};
+use std::time::{Duration, Instant};
 
 const CORNER_RADIUS: f64 = 4.0;
 const STROKE_WIDTH: f64 = 2.0;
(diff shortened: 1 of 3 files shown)


docs/book_examples/src/custom_widgets_md.rs
```diff
@@ -0,0 +1,60 @@
...
+            self.suppress_next = !self.suppress_next;
+            if self.suppress_next {
+                return;
+            }
+        }
+
+        // if we want our child to receive this event, we must send it explicitly.
+        child.event(ctx, event, data, env);
+    }
+}
```
Finnerale, line 59: This is what came to my mind based on #787 .
A `TextBox` that fires an action (e.g. for search) when hitting enter or 300 ms after the last key press.
```rust
const ACTION: Selector = Selector::new("hello.textbox-action");

struct TextBoxActionController {
    timer: Option<TimerToken>,
}

impl TextBoxActionController {
    pub fn new() -> Self {
        TextBoxActionController { timer: None }
    }

    // Fire ACTION after 300 ms
    fn deadline() -> Instant {
        Instant::now() + Duration::from_millis(300)
    }
}

impl Controller<String, TextBox> for TextBoxActionController {
    fn event(
        &mut self,
        child: &mut TextBox,
        ctx: &mut EventCtx,
        event: &Event,
        data: &mut String,
        env: &Env,
    ) {
        match event {
            Event::KeyDown(k) if k.key_code == KeyCode::Return => {
                ctx.submit_command(ACTION, None);
            }
            Event::KeyUp(k) if k.key_code != KeyCode::Return => {
                self.timer = Some(ctx.request_timer(Self::deadline()));
                child.event(ctx, event, data, env);
            }
            Event::Timer(token) if Some(*token) == self.timer => {
                ctx.submit_command(ACTION, None);
            }
            _ => child.event(ctx, event, data, env),
        }
    }
}
```

cmyr: thanks, I've changed the example to this.

docs/book_examples/src/custom_widgets_md.rs
```diff
@@ -32,12 +36,25 @@ fn background_label() -> impl Widget<Color> {
...
-impl Controller<String, TextBox> for AnnoyingController {
+impl TextBoxActionController {
+    pub fn new() -> Self {
+        TextBoxActionController { timer: None }
+    }
+
+    // Fire ACTION after 300 ms
+    fn deadline() -> Instant {
+        Instant::now() + DELAY
+    }
```
Finnerale, line 54:
```suggestion
```

docs/book_examples/src/custom_widgets_md.rs
```diff
@@ -46,15 +63,19 @@ impl Controller<String, TextBox> for AnnoyingController {
...
-        if matches!(event, Event::KeyDown(k) if k.key_code == KeyCode::Backspace) {
-            self.suppress_next = !self.suppress_next;
-            if self.suppress_next {
-                return;
+        match event {
+            Event::KeyDown(k) if k.key_code == KeyCode::Return => {
+                ctx.submit_command(ACTION, None);
+            }
+            Event::KeyUp(k) if k.key_code != KeyCode::Return => {
+                self.timer = Some(ctx.request_timer(Self::deadline()));
```
Finnerale, line 71:
```suggestion
                self.timer = Some(ctx.request_timer(Instant::now() + DELAY));
```

PR #833, opened 2020-04-13
Move internal events to a separate enum.

### LINKED ISSUE #826: Hide internal variants of LifeCycle from docs
I've been thinking about adding doc(hidden) to `RouteFocusChanged` and `RouteWidgetAdded`.
If I understand correctly, those are never really of interest to the user.

One disadvantage could be that they will still be completed by IDEs like Rust Analyzer, same applies however to the Debug... ones so this problem exists already.

Maybe hiding them and renaming to `InternalRouteFocusChanged` would be an option?

By @cmyr:
> We could also do something like prefix them with _.

Also adding a `LifeCycle::Internal(InternalLifeCycle)` came to my mind.

xStrom:
As discussed in #826 this PR does the following:
* Moves all internal `Event` variants into a `InternalEvent` enum and adds the `Event::Internal(InternalEvent)` variant.
* Moves all internal `LifeCycle` variants into a `InternalLifeCycle` enum and adds the `LifeCycle::Internal(InternalLifeCycle)` variant.

Closes #826

### DIFF
--- a/druid/src/event.rs
+++ b/druid/src/event.rs
@@ -56,6 +56,13 @@ pub enum Event {
     ///
     /// [`LifeCycle::WidgetAdded`]: enum.LifeCycle.html#variant.WidgetAdded
     WindowConnected,
+    /// Called on the root widget when the window size changes.
+    ///
+    /// Discussion: it's not obvious this should be propagated to user
+    /// widgets. It *is* propagated through the RootWidget and handled
+    /// in the WindowPod, but after that it might be considered better
+    /// to just handle it in `layout`.
+    WindowSize(Size),
     /// Called when a mouse button is pressed.
     MouseDown(MouseEvent),
     /// Called when a mouse button is released.
@@ -133,8 +140,6 @@ pub enum Event {
 /// [`Event`]: enum.Event.html
 #[derive(Debug, Clone)]
 pub enum InternalEvent {
-    /// Sent to the root widget when the window size changes.
-    Size(Size),
     /// Sent in some cases when the mouse has left the window.
     ///
     /// This is used in cases when the platform no longer sends mouse events,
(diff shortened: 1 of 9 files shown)


druid/src/widget/scroll.rs
```diff
@@ -388,8 +388,6 @@ impl<T: Data, W: Widget<T>> Widget<T> for Scroll<T, W> {
                         self.reset_scrollbar_fade(ctx, &env);
                     }
                 }
-                // Show the scrollbars any time our size changes
-                Event::Size(_) => self.reset_scrollbar_fade(ctx, &env),
```
xStrom, line 392: This was a `Scroll` bug, but it's not easy to fix. More info in #834.

druid/src/event.rs
```diff
@@ -131,9 +116,31 @@ pub enum Event {
...
+///
+/// These events are translated into regular [`Event`]s
+/// and should not be used directly.
+///
+/// [`WidgetPod`]: struct.WidgetPod.html
+/// [`Event`]: enum.Event.html
+#[derive(Debug, Clone)]
+pub enum InternalEvent {
+    /// Sent to the root widget when the window size changes.
+    Size(Size),
```
cmyr, line 137: There's an ongoing question about whether or not this should actually be more widely used, and sent down the tree by default; and even if not I do think of it as public API, custom widgets would regularly want to handle it I think?

xStrom:
Well `Size` is only sent to the root widget and it's window resize, not layout size changes. By changing its behavior, you mean that every widget should be informed about window resizing? In that case, it should at least be renamed to `WindowSize`. Or you're thinking abouit changing its meaning?

cmyr:
yes, it could be `WindowSize`. Basically I think that this is an event that applications might want to handle explicitly, such as by recentering the contents of a scroll view.

xStrom:
Okay, what about propagation? Still send it only to the root or to everyone?

Finnerale:
Wouldn't it be more useful if widgets could get an event when their own size changes?
Not sure what the window size would really be useful for.

xStrom:
I'm somewhat skeptical of the value of the window size event as well. However for now I brought it back to the public `Event` enum under the better describing `WindowSize` name. We can discuss its value further, but I don't want that holding up this merge.

PR #843, opened 2020-04-14
Add support for more mouse buttons and tracking holding.

xStrom:
This PR adds mouse support for more than just two buttons.

**Update:** It also adds `MouseButtons` to the `MouseEvent` to track which buttons are being held down.

### DIFF
--- a/druid-shell/src/mouse.rs
+++ b/druid-shell/src/mouse.rs
@@ -18,32 +18,41 @@ use crate::kurbo::Point;
 
 use crate::keyboard::KeyModifiers;
 
-/// The state of the mouse for a click, mouse-up, or move event.
+/// Information about the mouse event.
 #[derive(Debug, Clone, PartialEq)]
 pub struct MouseEvent {
     /// The location of the mouse in the current window.
     ///
-    /// This is in px units, that is, adjusted for hi-dpi.
+    /// This is in px units not device pixels, that is, adjusted for hi-dpi.
     pub pos: Point,
-    /// Keyboard modifiers at the time of the mouse event.
+    /// Mouse buttons being held down during a move or after a click event.
+    /// Thus it will contain the `button` that triggered a mouse-down event,
+    /// and it will not contain the `button` that triggered a mouse-up event.
+    pub buttons: MouseButtons,
+    /// Keyboard modifiers at the time of the event.
     pub mods: KeyModifiers,
     /// The number of mouse clicks associated with this event. This will always
-    /// be `0` for a mouse-up event.
-    pub count: u32,
-    /// The currently pressed button in the case of a move or click event,
-    /// or the released button in the case of a mouse-up event.
+    /// be `0` for a mouse-up and mouse-move events.
+    pub count: u8,
+    /// The button that was pressed down in the case of mouse-down,
+    /// or the button that was released in the case of mouse-up.
+    /// This will always be `MouseButton::None` in the case of mouse-move.
     pub button: MouseButton,
 }
 
 /// An indicator of which mouse button was pressed.
 #[derive(PartialEq, Eq, Clone, Copy, Debug)]
+#[repr(u8)]
 pub enum MouseButton {
+    /// No mouse button.
+    // MUST BE FIRST (== 0)
+    None,
     /// Left mouse button.
     Left,
-    /// Middle mouse button.
-    Middle,
     /// Right mouse button.
     Right,
+    /// Middle mouse button.
+    Middle,
     /// First X button.
     X1,
     /// Second X button.
@@ -51,16 +60,166 @@ pub enum MouseButton {
 }
 
 impl MouseButton {
-    /// Returns `true` if this is the left mouse button.
-    #[inline(always)]
+    /// Returns `true` if this is [`MouseButton::Left`].
+    ///
+    /// [`MouseButton::Left`]: #variant.Left
+    #[inline]
     pub fn is_left(self) -> bool {
         self == MouseButton::Left
     }
 
-    /// Returns `true` if this is the right mouse button.
-    #[inline(always)]
+    /// Returns `true` if this is [`MouseButton::Right`].
+    ///
+    /// [`MouseButton::Right`]: #variant.Right
+    #[inline]
     pub fn is_right(self) -> bool {
         self == MouseButton::Right
+    }
+
+    /// Returns `true` if this is [`MouseButton::Middle`].
+    ///
+    /// [`MouseButton::Middle`]: #variant.Middle
+    #[inline]
+    pub fn is_middle(self) -> bool {
+        self == MouseButton::Middle
+    }
+
+    /// Returns `true` if this is [`MouseButton::X1`].
+    ///
+    /// [`MouseButton::X1`]: #variant.X1
+    #[inline]
+    pub fn is_x1(self) -> bool {
+        self == MouseButton::X1
+    }
+
+    /// Returns `true` if this is [`MouseButton::X2`].
+    ///
+    /// [`MouseButton::X2`]: #variant.X2
+    #[inline]
+    pub fn is_x2(self) -> bool {
+        self == MouseButton::X2
+    }
+}
+
+/// A set of [`MouseButton`]s.
+///
+/// [`MouseButton`]: enum.MouseButton.html
+#[derive(PartialEq, Eq, Clone, Copy, Default)]
+pub struct MouseButtons(u8);
+
+impl MouseButtons {
+    /// Create a new empty set.
+    #[inline]
+    pub fn new() -> MouseButtons {
+        MouseButtons(0)
+    }
+
+    /// Add the `button` to the set.
+    #[inline]
+    pub fn insert(&mut self, button: MouseButton) {
+        self.0 |= 1.min(button as u8) << button as u8;
+    }
+
+    /// Remove the `button` from the set.
+    #[inline]
+    pub fn remove(&mut self, button: MouseButton) {
+        self.0 &= !(1.min(button as u8) << button as u8);
+    }
+
+    /// Builder-style method for adding the `button` to the set.
+    #[inline]
+    pub fn with(mut self, button: MouseButton) -> MouseButtons {
+        self.0 |= 1.min(button as u8) << button as u8;
+        self
+    }
+
+    /// Builder-style method for removing the `button` from the set.
+    #[inline]
+    pub fn without(mut self, button: MouseButton) -> MouseButtons {
+        self.0 &= !(1.min(button as u8) << button as u8);
+        self
+    }
+
+    /// Returns `true` if the `button` is in the set.
+    #[inline]
+    pub fn contains(self, button: MouseButton) -> bool {
+        (self.0 & (1.min(button as u8) << button as u8)) != 0
+    }
+
+    /// Returns `true` if the set is empty.
+    #[inline]
+    pub fn is_empty(self) -> bool {
+        self.0 == 0
+    }
+
+    /// Returns `true` if all the `buttons` are in the set.
+    #[inline]
+    pub fn is_superset(self, buttons: MouseButtons) -> bool {
+        self.0 & buttons.0 == buttons.0
+    }
+
+    /// Returns `true` if [`MouseButton::Left`] is in the set.
+    ///
+    /// [`MouseButton::Left`]: enum.MouseButton.html#variant.Left
+    #[inline]
+    pub fn has_left(self) -> bool {
+        self.contains(MouseButton::Left)
+    }
+
+    /// Returns `true` if [`MouseButton::Right`] is in the set.
+    ///
+    /// [`MouseButton::Right`]: enum.MouseButton.html#variant.Right
+    #[inline]
+    pub fn has_right(self) -> bool {
+        self.contains(MouseButton::Right)
+    }
+
+    /// Returns `true` if [`MouseButton::Middle`] is in the set.
+    ///
+    /// [`MouseButton::Middle`]: enum.MouseButton.html#variant.Middle
+    #[inline]
+    pub fn has_middle(self) -> bool {
+        self.contains(MouseButton::Middle)
+    }
+
+    /// Returns `true` if [`MouseButton::X1`] is in the set.
+    ///
+    /// [`MouseButton::X1`]: enum.MouseButton.html#variant.X1
+    #[inline]
+    pub fn has_x1(self) -> bool {
+        self.contains(MouseButton::X1)
+    }
+
+    /// Returns `true` if [`MouseButton::X2`] is in the set.
+    ///
+    /// [`MouseButton::X2`]: enum.MouseButton.html#variant.X2
+    #[inline]
+    pub fn has_x2(self) -> bool {
+        self.contains(MouseButton::X2)
+    }
+
+    /// Adds all the `buttons` to the set.
+    pub fn extend(&mut self, buttons: MouseButtons) {
+        self.0 |= buttons.0;
+    }
+
+    /// Returns a union of the values in `self` and `other`.
+    #[inline]
+    pub fn union(mut self, other: MouseButtons) -> MouseButtons {
+        self.0 |= other.0;
+        self
+    }
+
+    /// Clear the set.
+    #[inline]
+    pub fn clear(&mut self) {
+        self.0 = 0;
+    }
+}
+
+impl std::fmt::Debug for MouseButtons {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "MouseButtons({:05b})", self.0 >> 1)
     }
 }
 
--- a/druid-shell/src/platform/gtk/window.rs
+++ b/druid-shell/src/platform/gtk/window.rs
@@ -41,7 +41,7 @@ use super::util::assert_main_thread;
 use crate::common_util::IdleCallback;
 use crate::dialog::{FileDialogOptions, FileDialogType, FileInfo};
 use crate::keyboard;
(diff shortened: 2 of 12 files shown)


xStrom:
Okay I added `MouseButton::Unknown` and use it as a fallback on all the platforms.

xStrom:
After some further testing and thinking I also changed mouse events to contain `Option<MouseButton>`, because in some cases we know for sure that no button was pressed and that is different from an unknown button.

One slight downside of using an `Option` is that people now need to import the `MouseButtonExt` trait to get the convenience `is_left` etc methods.

cmyr:
Would it make more sense to have `MouseButton::None`?

cmyr:
An additional (probably useless?) thought: mouse buttons are really a mask, multiple can be down at a time etc. I'm probably over thinking this though, it definitely isn't what we would want to surface as the principle API (although we might want to surface it at some point, for advanced cases?)

druid-shell/src/mouse.rs
```diff
@@ -48,22 +48,49 @@ pub enum MouseButton {
     X1,
     /// Second X button.
     X2,
+    /// Unknown button.
+    Unknown,
 }
 
-impl MouseButton {
+/// Adds convenience methods to `MouseButton` and `Option<MouseButton>`.
+pub trait MouseButtonExt {
```
cmyr, line 56: I get the motivation but this doesn't feel great to me. I do think I might prefer `MouseButton::None`, and we could sort of redefine `MouseButton` to be 'the mouse button that triggered a given event, if one exists.' really this is all about mouse moved. Down and up must be associated with a specific button, but moved can have no or all buttons, etc. I wonder if it's worth representing differently in the API?

I've also experimented with ways to build a [higher-level API on top of this](https://github.com/linebender/runebender/blob/master/src/mouse.rs) in runebender; I'm not totally happy with the results, but I think something like that _could_ be useful?

druid-shell/src/platform/gtk/window.rs
```diff
@@ -713,28 +713,26 @@ fn make_gdk_cursor(cursor: &Cursor, gdk_window: &gdk::Window) -> Option<gdk::Cur
...
-        4 => MouseButton::X1,
-        5 => MouseButton::X2,
-        _ => MouseButton::Left,
+        0 => None,
+        1 => Some(MouseButton::Left),
+        2 => Some(MouseButton::Middle),
+        3 => Some(MouseButton::Right),
+        4 => Some(MouseButton::X1),
+        5 => Some(MouseButton::X2),
+        _ => Some(MouseButton::Unknown),
```
cmyr, line 724: I might consider making this `Unknown(u32)`? it doesn't matter too much either way, but if someone wants that info we might as well provide it?

But only if it's not a super big hassle..

xStrom:
I went with `Option` because that's *the rust way* but I don't consider that too important in this case.

How about a custom struct `MouseButtons` which is basically a type-safe bitmask. Something like this:
```rust
impl MouseButtons {
    fn has(&self, button: MouseButton) -> bool;

    fn has_left(&self) -> bool;
    fn has_right(&self) -> bool;
    // ....
    fn has_x2(&self) -> bool;

    fn has_any(&self) -> bool;
}
```

cmyr:
The problem is that these have different uses; so that would work, but would not replace `MouseButton`.

on `MouseDown` and `MouseUp`, it matters *which* button triggered the event; the mask would be supplementary at best. On `MouseMoved`, the idea of a `MouseButton` is meaningless, and something like `MouseButtons` makes sense.

For all this, though, what's most important is a good API that covers the 99% case; to me this means that `MouseDown` and `MouseUp` *always* have an associated mouse button, and that `MouseMoved` probably shouldn't. 

One option:

`MouseEvent` does not have a `MouseButton`, it has only something like the `MouseButton`s you describe; and then `MouseDown` and `MouseUp` are a `MouseEvent` *plus* a `MouseButton`, but `MouseMoved` is only a `MouseEvent`? This is the most reasonable thing I can think of right now...

xStrom:
That sounds reasonable. During `MouseDown` you would have druid-managed state that tells you which other buttons are currently already held down, plus the one new button separately.

```rust
Event::MouseDown((MouseEvent, MouseButton)) // Something like this tuple?
```

xStrom:
After thinking about this some more, especially in relation to #842 which is headed towards adding a new `focused` bool to the event for clicks, I'm thinking of splitting up the event structs into `MouseMoveEvent` and `MouseClickEvent`.

Having the single `button` info and also `focused` info only makes sense for clicks. This would also allow for more easy divergence between the event data in the future. It should also break less code because there's no tuple and the following should still work:

```rust
Event::MouseUp(e) => {
    if e.button.is_left() { do stuff; }
}
```

cmyr:
a few options I think:
```rust
MouseDown { event: MouseEvent, button: MouseButton }

// or,
MouseClickEvent {
    mouse: MouseEvent,
    button: MouseButton,
    count: u32,
}
// with
Event::MousDown(MouseClickEvent),
Event::MouseUp(MouseClickEvent),
Event::MouseMove(MouseEvent),
```

(I realized in the course of writing this that the click count is also irrelevant for the purposes of `MouseMove`).

I think maybe one reason for the current design is that I was thinking about mac, and NSEvent is a monolith; the same underlying 'type' can represent tons of different events, so there are lots of fields that exist that aren't actually meaningful for a given specific event.

xStrom:
I pushed a new iteration of this work. Holding down multiple mouse buttons is now nicely tracked. It works perfectly on macOS and Windows and mostly on GTK. GTK has all sorts of issues due to fragmentation, which will need solving separately. This PR at the very least only improves the situation in druid-shell and doesn't make it worse.

For this iteration I went with `ClickEvent` and `MoveEvent`. It's not perfect but doesn't seem too bad either. The benefit is that the `struct` doesn't contain useless fields, however it does make `match` code a bit more verbose if it wants to do something on a common field.
```rust
// What used to be this ..
Event::MouseDown(e) | Event::MouseUp(e) | Event::MouseMove(e) => {
    self.last_mouse_pos = Some(e.pos)
}
// is now this.
Event::MouseDown(ClickEvent { pos, .. })
| Event::MouseUp(ClickEvent { pos, .. })
| Event::MouseMove(MoveEvent { pos, .. }) => self.last_mouse_pos = Some(*pos),
```

I'm not 100% convinced that splitting them is the way to go. Maybe the superior ergonomics in handling these three events together is worth the price of wasting those extra bytes of memory and having useless fields in the code editor autocomplete for `MouseMove`?

druid-shell/src/mouse.rs
```diff
@@ -18,50 +18,240 @@ use crate::kurbo::Point;
...
     pub mods: KeyModifiers,
     /// The number of mouse clicks associated with this event. This will always
     /// be `0` for a mouse-up event.
-    pub count: u32,
-    /// The currently pressed button in the case of a move or click event,
-    /// or the released button in the case of a mouse-up event.
+    pub count: u8,
+    /// The button that was pressed down in the case of mouse-down,
+    /// or the button that was released in the case of mouse-up.
     pub button: MouseButton,
```
cmyr, line 52: okay, thinking about the API:

I think that separating out `MouseButtons` from `MouseButton` solves most of our issue here, and shows a viable way to have a single type cover these three events without any clear inconsistency. There will be two fields, `button` and `count`, that are not used in `MouseMove`, and I think that's fine.

`count` is actually a funny one; the docstring says "will always be 0 on mouse-up", but I'm not sure I actually like that as an implementation; this is something we're going to have to revisit at some point, in that some platforms might provide their own double/triple click counting, and on some platforms we might need to do it ourselves. We probably do want to trust the platform, though, so that we are respecting whatever settings the user has on that platform.

My gut tells me that mouse-up should include count, though? (but this is unrelated to current PR).

druid-shell/src/mouse.rs
```diff
@@ -18,50 +18,240 @@ use crate::kurbo::Point;
...
     /// Right mouse button.
     Right,
+    /// Middle mouse button.
+    Middle,
     /// First X button.
     X1,
     /// Second X button.
     X2,
+    /// Some other unknown button.
+    Other,
```
cmyr, line 70: I think if we're reworking this API we should include precise arbitrary buttons, which means `Other` should be `Other(u8)`. We will end up needing to do this eventually for something, so we might as well do it now.

druid-shell/src/mouse.rs
```diff
@@ -18,50 +18,240 @@ use crate::kurbo::Point;
...
+    /// [`MouseButton::Other`]: enum.MouseButton.html#variant.Other
+    #[inline]
+    pub fn has_other(self) -> bool {
+        self.has(MouseButton::Other)
+    }
+
+    /// Adds all the [`MouseButton`]s in `other` to the set.
+    ///
+    /// [`MouseButton`]: enum.MouseButton.html
+    pub fn union(&mut self, other: MouseButtons) {
```
cmyr, line 231: I would expect this to return a new `MouseButtons`

druid-shell/src/mouse.rs
```diff
@@ -18,50 +18,240 @@ use crate::kurbo::Point;
...
+    /// Builder-style method for adding the `button` to the set.
+    #[inline]
+    pub fn with(mut self, button: MouseButton) -> MouseButtons {
+        self.0 |= 1 << button as u8;
+        self
+    }
+
+    /// Builder-style method for removing the `button` from the set.
+    #[inline]
+    pub fn without(mut self, button: MouseButton) -> MouseButtons {
```
cmyr, line 157: I'm getting some YAGNI vibes here; it isn't clear when or where these would be used? Are they just by us, or do we expect external users to need to synthesize mouse buttons?

druid-shell/src/mouse.rs
```diff
@@ -18,50 +18,240 @@ use crate::kurbo::Point;
...
+    #[inline]
+    pub fn has_x2(self) -> bool {
+        self.has(MouseButton::X2)
+    }
+
+    /// Returns `true` if [`MouseButton::Other`] is in the set.
+    ///
+    /// [`MouseButton::Other`]: enum.MouseButton.html#variant.Other
+    #[inline]
+    pub fn has_other(self) -> bool {
```
cmyr, line 224: if we make `Other` include a specific id, we should allow this to check for that value.

druid-shell/src/mouse.rs
```diff
@@ -18,50 +18,240 @@ use crate::kurbo::Point;
...
+    pub fn is_other(self) -> bool {
+        self == MouseButton::Other
+    }
+}
+
+/// A set of [`MouseButton`]s.
+///
+/// [`MouseButton`]: enum.MouseButton.html
+#[derive(PartialEq, Eq, Clone, Copy, Default)]
+pub struct MouseButtons(u8);
```
cmyr, line 127: we probably have more than 8 bits of possible buttons; I would just make this a u32 now, that should be enough for anybody.

druid-shell/src/mouse.rs
xStrom: `MouseMove` would also have a useless `focused` that will be introduced in #842, however even so I'm also thinking a single event is probably the way to go.

As for `count`, I have additional thoughts on it that I describe in #859. The short of it is that yes I think `count` should be there for `MouseUp` but it's not there right now on any platfrom besides macOS so I just wrote the docs to reflect that reality and made macOS compatible with that. This would be a short-lived change in light of #859.

druid-shell/src/mouse.rs
xStrom: Having it be `Other(u8)` would change the `MouseButtons` implementation because it couldn't just be a bitmask and would need to store arbitrary `u8`s in the set as well.

Also it doesn't really look like any of the platforms (Windows, macOS, GTK, web) support more specific buttons via the APIs that we're currently using. GTK is struggling with X1/X2 already. It's of course possible to tap into the mouse driver and get more of them, but I feel like that's a rather large project and probably not even that useful for general GUI apps.

So given the large scope of cross-platfrom support for more buttons, I think we can't reasonably prepare our API for them. When someone wants to take that task on and knows the demands of a practical implementation, then it would be much more reasonable to change the API.

druid-shell/src/mouse.rs
xStrom: You shouldn't, because builder methods start with `with`.

druid-shell/src/mouse.rs
xStrom: Well this specific `without` method is used in the GTK platform code already.

More generally these functions can be used by external users. Yes they might synthesize `MouseButtons` from some sort of hotkey configuration file and then do a check during a mouse event against that.

druid-shell/src/mouse.rs
cmyr: this wouldn't be a builder method, it would just be a method that happens to return the union of two sets; similar to the `union` method on collection types like `HashSet`. (These actually return an iterator over the union, but since our type is `Copy` I wouldn't worry about that).

If what you want to do is add all the items from one set to the other, the only `std` api for this would be the [`Extend`](http://doc.rust-lang.org/1.42.0/std/iter/trait.Extend.html#) trait.

Anyway not a blocker or anything, just something that jumps out at me.

druid-shell/src/mouse.rs
cmyr: on macOS the mouse buttons are presented as a u32 bitmask. Windows seems to support up to five buttons. I haven't checked elsewhere.

Okay, so my last question would be, is `Other` even a useful distinction? If the user can't get any more information I think it could be problematic, since two separate buttons could send `Other`, and you can't tell them apart for the purposes of up & down.

In the very long term, I've been imagining we want to have some API to get the underlying event from the platform in these cases, but that's also another can of worms.

xStrom:
A new iteration has been pushed. I went back to a single `MouseEvent` struct and changed the `union` / `extend` naming for `MouseButtons`.

I also thought about your question of whether `MouseButton::Other` is useful. It can have a bit of value but it does also introduce a lot of problems.
- It isn't really cross-platform in a portable way, so hot-key configurations made with `Other` in druid wouldn't be portable either.
- It represents multiple buttons so events can get rather funky with up/down and we couldn't really perform double-click detection on it. #859
- It isn't supported in our `MouseButtons` in more than one way. Not just the set itself, but even figuring out what to add to the set. There isn't a clear way to figure out the holding state on all platforms.

I think druid should strive for a consistent API that is reliable across platforms. Given these issues I don't think `Other` passes the quality bar for druid. For that reason I have removed it from this PR. Now `druid-shell` just ignores click events on buttons that we don't know.

In the long term, if we want to support more buttons, that can of course be done. However it should be done in a qualitative way and I don't think it can be achieved with the platform mouse APIs `druid-shell` is using right now. More lower level APIs would need to be used.

druid-shell/src/mouse.rs
```diff
@@ -18,50 +18,218 @@ use crate::kurbo::Point;
...
+
+impl MouseButtons {
```
cmyr, line 109: just sort of riffing on this type; if the idea is to construct it from `MouseButton`s, one possible thing to do is implement `FromIterator<MouseButton>`, and then you could write, say,

```rust
let buttons: MouseButtons = [MouseButton::Left, MouseButton::X1].iter().copied().collect();
```

That doesn't feel great really; probably simpler to have a constructor that just takes a `&[MouseButton]`.

In any case I'm just daydreaming, this isn't necessary or anything.

druid-shell/src/mouse.rs
```diff
@@ -18,50 +18,218 @@ use crate::kurbo::Point;
...
+
+    /// Returns `true` if the `button` is in the set.
+    #[inline]
+    pub fn has(self, button: MouseButton) -> bool {
+        (self.0 & (1 << button as u8)) != 0
+    }
+
+    /// Returns `true` if any button is in the set.
+    #[inline]
+    pub fn has_any(self) -> bool {
```
cmyr, line 150: I might prefer to use `is_empty()` for this, if we're treating it like a set? Not totally sure, just a thought.

druid-shell/src/platform/x11/window.rs
```diff
@@ -424,7 +424,7 @@ impl WindowHandle {
 
     pub fn get_dpi(&self) -> f32 {
         // TODO(x11/dpi_scaling): figure out DPI scaling
-        log::warn!("WindowHandle::get_dpi is currently unimplemented for X11 platforms.");
+        //log::warn!("WindowHandle::get_dpi is currently unimplemented for X11 platforms.");
```
cmyr, line 427: if this is spammy we can remove it?

druid-shell/src/platform/x11/application.rs
```diff
@@ -139,6 +141,8 @@ impl Application {
                                 button_release.event_x() as f64,
                                 button_release.event_y() as f64,
                             ),
+                            // TODO: Fill with held down buttons
```
cmyr, line 144: let's open an issue for this when we merge

druid-shell/src/mouse.rs
xStrom: I named as many as I could with `has_` prefix because it seemed nicer to me. However it's probably more ergonomic to follow `HashSet` more closely in naming, because that's what Rust programmers will know. I will rename all the methods that don't match but have equals in `HashSet`.

druid-shell/src/mouse.rs
xStrom: The following one liner isn't too bad either:
```rust
let buttons = MouseButtons::new().with(MouseButton::Left).with(MouseButton::Right);
```

There can be value in adding more methods, however I think I would leave these slice/iterator  constructors out of this PR.

druid-shell/src/mouse.rs
cmyr: yea, this is my general thinking; people are more likely to try and use method names they know from elsewhere.

PR #845, opened 2020-04-14
Add hot glow option to multiwin example.

xStrom:
This PR adds an option to the context menu of the multiwin example. When enabled the background turns red when the window is hot.

I've used it successfully to test the workings of hot state and I think it can be interesting to others as well. Also it will be useful for demonstrating various hot state issues that I plan to tackle in upcoming PRs.

### DIFF
--- a/druid/examples/multiwin.rs
+++ b/druid/examples/multiwin.rs
@@ -14,11 +14,11 @@
 
 //! Opening and closing windows and using window and context menus.
 
-use druid::widget::{Align, Button, Flex, Label, Padding};
+use druid::widget::prelude::*;
+use druid::widget::{Align, BackgroundBrush, Button, Flex, Label, Padding};
 use druid::{
-    commands as sys_cmds, AppDelegate, AppLauncher, Command, ContextMenu, Data, DelegateCtx, Env,
-    Event, EventCtx, LocalizedString, MenuDesc, MenuItem, Selector, Target, Widget, WindowDesc,
-    WindowId,
+    commands as sys_cmds, AppDelegate, AppLauncher, Color, Command, ContextMenu, Data, DelegateCtx,
+    LocalizedString, MenuDesc, MenuItem, Selector, Target, WidgetPod, WindowDesc, WindowId,
 };
 
 use log::info;
@@ -26,11 +26,13 @@ use log::info;
 const MENU_COUNT_ACTION: Selector = Selector::new("menu-count-action");
 const MENU_INCREMENT_ACTION: Selector = Selector::new("menu-increment-action");
 const MENU_DECREMENT_ACTION: Selector = Selector::new("menu-decrement-action");
+const MENU_SWITCH_GLOW_ACTION: Selector = Selector::new("menu-switch-glow");
 
 #[derive(Debug, Clone, Default, Data)]
 struct State {
     menu_count: usize,
     selected: usize,
+    glow_hot: bool,
 }
 
 pub fn main() {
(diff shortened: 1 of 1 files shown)


druid/examples/multiwin.rs
```diff
@@ -78,7 +80,58 @@ fn ui_builder() -> impl Widget<State> {
...
+
+impl<W: Widget<State>> Widget<State> for Glow<W> {
+    fn event(&mut self, ctx: &mut EventCtx, event: &Event, data: &mut State, env: &Env) {
+        self.inner.event(ctx, event, data, env);
+    }
+
+    fn lifecycle(&mut self, ctx: &mut LifeCycleCtx, event: &LifeCycle, data: &State, env: &Env) {
+        if matches!(event, LifeCycle::HotChanged(_)) {
+            ctx.request_paint();
+        }
```
Finnerale, line 106: For consistency with other examples
```suggestion
        if let LifeCycle::HotChanged(_) = event {
            ctx.request_paint();
        }
```

druid/examples/multiwin.rs
```diff
@@ -78,7 +80,58 @@ fn ui_builder() -> impl Widget<State> {
...
+    fn layout(
+        &mut self,
+        ctx: &mut LayoutCtx,
+        bc: &BoxConstraints,
+        data: &State,
+        env: &Env,
+    ) -> Size {
+        let size = self.inner.layout(ctx, bc, data, env);
+        self.inner.set_layout_rect(size.to_rect());
+        bc.max()
```
Finnerale, line 126: Returning size would make this container 'neutral' regarding layout, which seems more idiomatic to me
```suggestion
        size
```

xStrom: Probably better with `size` indeed.

druid/examples/multiwin.rs
```diff
@@ -79,7 +81,58 @@ fn ui_builder() -> impl Widget<State> {
...
     col.add_flex_child(Align::centered(row), 1.0);
-    col
+    Glow::new(col)
+}
+
+struct Glow<W> {
+    inner: WidgetPod<State, W>,
+}
+
+impl<W: Widget<State>> Glow<W> {
```
cmyr, line 91: just a note: I think you could have achieved this using a `Painter`?

druid/examples/multiwin.rs
```diff
@@ -79,7 +81,58 @@ fn ui_builder() -> impl Widget<State> {
     row.add_child(Padding::new(5.0, inc_button));
     row.add_child(Padding::new(5.0, dec_button));
     col.add_flex_child(Align::centered(row), 1.0);
-    col
+    Glow::new(col)
+}
+
+struct Glow<W> {
+    inner: WidgetPod<State, W>,
```
cmyr, line 88: if a widget is entirely transparent for the purposes of event handling and layout (it is just passing things through) then WidgetPod is unnecessary. It doesn't hurt in any meaningful way, but it is unnecessary.

druid/examples/multiwin.rs
```diff
@@ -79,7 +81,58 @@ fn ui_builder() -> impl Widget<State> {
...
+    fn layout(
+        &mut self,
+        ctx: &mut LayoutCtx,
+        bc: &BoxConstraints,
+        data: &State,
+        env: &Env,
+    ) -> Size {
+        let size = self.inner.layout(ctx, bc, data, env);
+        self.inner.set_layout_rect(ctx, data, env, size.to_rect());
+        size
```
cmyr, line 127: a general note; I think it's good to call `bc.constrain(size)` anytime you return from layout, as a way of ensuring all returned sizes are appropriate for the constraints.

Finnerale: Well, my inspiration came from [`ControllerHost`](https://github.com/xi-editor/druid/blob/master/druid/src/widget/controller.rs) which does not do that, maybe open an Issue to make this consistent for all container?

druid/examples/multiwin.rs
Finnerale: This really needs to be written down :sweat_smile:

druid/examples/multiwin.rs
xStrom: Perhaps in some way, but it wasn't super obvious. Like the `Painter` docs state: *If you would like it to repaint at other times (such as when hot or active state changes) you will need to call `request_paint` further up the tree.* There's nothing further up the tree, so it wouldn't just be a `Painter` at least.

cmyr: oh huh for some reason I thought it did repaint on hot. My mistake!

I wonder if it makes sense to just make 'paint on hot' be an option for painter? but that's a small thing.

PR #846, opened 2020-04-15
Fix the instructions for serving the wasm exmaples

elrnv:
This PR builds an `index.html` file which links to all the examples in the html folder and fixes the serving instructions in the README.md.


druid/examples/wasm/build.rs
```diff
@@ -61,14 +74,22 @@ mod examples {
                 example_str.to_string()
             };
 
+            // Add an entry to the index.html file.
+            let index_entry = format!(
+                "<li><a href=\"./html/{name}.html\">{name}</a></li>",
+                name = js_entry_fn_name
```
Finnerale, line 80: The html file is named after the example name
```suggestion
                name = example_str
```

PR #847, opened 2020-04-15
Change request_timer to work with Duration

Finnerale:
Since the wasm PR I've been thinking about changing `request_timer` to take `Duration` instead of `Instant`, considering that I've never seen it being called in a different way than
```rust
ctx.request_timer(Instant::now() + some_duration)
```
With this it's just
```rust
ctx.request_timer(some_duration)
```
That would allow us to get rid of Instant as part of our API and thus `instant::Instant` would not be necessary in user code when targeting wasm.

This should also fix the wasm CI


Finnerale:
Looks like it does indeed fix the pipeline :smile:

PR #850, opened 2020-04-15
Add clippy checks for examples and tests.

xStrom:
The current CI configuration wasn't checking examples and tests for clippy mistakes.

### DIFF
--- a/druid-derive/Cargo.toml
+++ b/druid-derive/Cargo.toml
@@ -17,4 +17,4 @@ proc-macro2 = "1.0.9"
 
 [dev-dependencies]
 druid = { path = "../druid", version = "0.6.0" }
-float-cmp = "0.6.0"
+float-cmp = { version = "0.6.0", default-features = false }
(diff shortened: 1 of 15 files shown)


druid-derive/Cargo.toml
```diff
@@ -17,3 +17,4 @@ proc-macro2 = "1.0.9"
 
 [dev-dependencies]
 druid = { path = "../druid", version = "0.6.0" }
+float-cmp = "0.6.0"
```
Finnerale, line 20: Can we avoid the `num-traits` dependency?
```suggestion
float-cmp = { version = "0.6.0", default-features = false }
```

druid/Cargo.toml
```diff
@@ -47,3 +47,6 @@ version = "0.4.0"
 
 [target.'cfg(target_arch="wasm32")'.dependencies]
 console_log = "0.1.2"
+
+[dev-dependencies]
+float-cmp = "0.6.0"
```
Finnerale, line 52: Same as above
```suggestion
float-cmp = { version = "0.6.0", default-features = false }
```

druid-derive/Cargo.toml
xStrom: Sure, we don't need it for this PR.

xStrom:
I reduce the dependency graph!

PR #857, opened 2020-04-17
Add proper CI testing coverage and optimize it for speed.

xStrom:
I spent some time really reading through the cargo docs and inspecting our workspace structure. This PR is the result of that. It covers more code and is faster.

* Made sure everything is tested on all target combinations. There were some holes with examples, including book examples. Also nightly and wasm testing were limited.
* Removed redundant `cargo rustc` steps as `cargo clippy` already compiles the code and fails on the same warnings. I verified this and it's stated in [the clippy readme](https://github.com/rust-lang/rust-clippy/blob/master/README.md):
> Note that adding `-D warnings` will cause your build to fail if any warnings are found in your code. That includes warnings found by rustc (e.g. `dead_code`, etc.). 
* Brought X11 testing into the main job as an end phase on Ubuntu, because it shares a lot of dependencies that are already built.
* Added various comments to the CI file for future editors.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   rustfmt:
     runs-on: ubuntu-latest
-    name: rustfmt
+    name: cargo fmt
     steps:
       - uses: actions/checkout@v2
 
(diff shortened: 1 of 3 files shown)


cmyr:
Wonderful, the problem with this stuff is definitely maintenance; it seems to consistently degrade over any serious set of changes. Hopefully we don't have too many additional targets coming in the near future, though? 😬

PR #858, opened 2020-04-17
Fix wasm examples not building on Windows with a standard user.

xStrom:
The wasm example build script used [std::os::windows::fs::symlink_dir](https://doc.rust-lang.org/std/os/windows/fs/fn.symlink_dir.html) but that requires administrator privileges to work. The one exception is that if it's a new enough Windows that also has developer mode enabled.

A fine alternative is a [junction](https://stackoverflow.com/a/48586946/138826). However there doesn't seem to be a Rust standard library way to do it. There is [std::sys::windows::fs::symlink_junction](https://github.com/rust-lang/rust/blob/ce93331e2cf21ac4b72a53854b105955919114e7/src/libstd/sys/windows/fs.rs#L886) but that is a private function that is not exposed for public use. Given that this is a test build file, I didn't bother too much and went with a somewhat hacky solution of just calling out to a shell.

So the solution is to attempt to create a symbolic link and if that fails just launch a Command Prompt shell and use it's inbuilt **mklink** command to create the junction.

### DIFF
--- a/druid/examples/wasm/build.rs
+++ b/druid/examples/wasm/build.rs
@@ -8,6 +8,28 @@ const EXCEPTIONS: &[&str] = &[
     "ext_event", // WASM doesn't currently support spawning threads.
 ];
 
+/// Create a platform specific link from `src` to the `dst` directory.
+fn link_dir(src: &std::path::Path, dst: &std::path::Path) {
+    // First delete any existing link, as it might be stale.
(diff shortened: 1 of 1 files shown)


elrnv:
It's probably worth mentioning another alternative -- we could simply copy all the examples during the build step instead of making the symbolic link.

PR #863, opened 2020-04-21
Document new steps needed for new examples

elrnv:
Additional steps are required since the addition of the WASM target.

Ideally these would be automated. To achieve that goal we would need to

1. Require all example main function signatures to be the same (all returning
Results or none). Alternatively determine if a main function returns a
Result in order to produce the appropriate impl_example! macro call in
lib.rs of the wasm example.

2. Rename the switch example to switch_demo or something similar.


xStrom:
Regarding the `switch` example, one short rename possibility would be `switches`.

PR #869, opened 2020-04-23
Make window title LabelText instead of LocalizedString

cmyr:
It isn't really possible for code outside of druid to use
LocalizedString in any reasonable way at the moment, and so
this makes it much easier to set a window title, especially
a title that is generated dynamically from the data.


zaynetro:
Thanks! This is a great change. I always needed to copy-paste setting window title from another example before.

cmyr:
yes, sorry for making it wait so long 😅

PR #876, opened 2020-04-24
Add debug option to display widget ids

cmyr:
This will paint the id of each widget of some subtree in the bottom
right of that widget's bounds; useful for debugging various bits
of event flow and widget behaviour.

Screen Shot 2020-04-24 at 1 32 24 PM

### DIFF
--- a/druid/src/core.rs
+++ b/druid/src/core.rs
@@ -18,9 +18,9 @@ use std::collections::{HashMap, VecDeque};
 
 use crate::bloom::Bloom;
 use crate::kurbo::{Affine, Insets, Point, Rect, Shape, Size, Vec2};
-use crate::piet::RenderContext;
+use crate::piet::{FontBuilder, RenderContext, Text, TextLayout, TextLayoutBuilder};
 use crate::{
-    BoxConstraints, Command, Data, Env, Event, EventCtx, InternalEvent, InternalLifeCycle,
+    BoxConstraints, Color, Command, Data, Env, Event, EventCtx, InternalEvent, InternalLifeCycle,
     LayoutCtx, LifeCycle, LifeCycleCtx, PaintCtx, Region, Target, TimerToken, UpdateCtx, Widget,
     WidgetId, WindowId,
 };
@@ -347,11 +347,11 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
         ctx.z_ops.append(&mut inner_ctx.z_ops);
 
         if env.get(Env::DEBUG_PAINT) {
-            const BORDER_WIDTH: f64 = 1.0;
-            let rect = inner_ctx.size().to_rect().inset(BORDER_WIDTH / -2.0);
-            let id = self.id().to_raw();
-            let color = env.get_debug_color(id);
-            inner_ctx.stroke(rect, &color, BORDER_WIDTH);
+            self.debug_paint_layout_bounds(&mut inner_ctx, env);
+        }
+
+        if env.get(Env::DEBUG_WIDGET_ID) {
+            self.debug_paint_widget_ids(&mut inner_ctx, env);
         }
 
         self.state.invalid = Region::EMPTY;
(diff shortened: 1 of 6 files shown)


Finnerale:
Good idea :+1: 

But this can cause overlapping ids:


Not sure how to deal with this but I would not like it to stay like this.
I almost did not notice that little red between the numbers and thought it would be `115`.

To reproduce:
1. enable this for the `view_switcher` example
2. choose `View 1` "Simple Button"
3. choose `View 3`
4. repeat from `2.` until the ids are >100
5. choose `View 1` or `View 4` to get the overlapping ids

cmyr:
yea, I was just working on that. It's kind of tricky; the easiest thing I can do is use z-order painting, so at least we prioritize the topmost widget. I'm also experimenting with only painting if we're currently `hot`. Not sure what combination of these things makes the most sense...

cmyr:
Okay, I've changed the behaviour to paint using z-order, so children overwrite parents; this should also fix the issue of smaller labels being written over larger labels, because children should always have larger ids than parents. (`const` ids might violate this, but I don't want to worry about that yet).

In addition, we now only paint the label if the widget is hot.

Finnerale:
The new approach feels better :+1: 
It does seem like it draws a bit outside the widget tho, which causes visual artifacts as only the widget gets invalidated.

Finnerale:
The invalidation (I guess) also causes problems with the `Current view: x` label in the `view_switcher` example. It sometimes does not get a label when hovering or the label is not removed when leaving the label.

cmyr:
right, we're stroking the boundary so it may spill. can adjust that.

about the view_switcher example.... that's weird. My spidey sense says this is an existing bug that we're just seeing for the first time, but I'll dig in a bit more timorrow.

cmyr:
Okay fixed this, was an issue with how I was handling hot changes.

druid/src/core.rs
```diff
@@ -335,25 +344,32 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
...
-        }
+        let debug_layout = env.get(Env::DEBUG_PAINT);
+        let debug_ids = env.get(Env::DEBUG_WIDGET_ID) && inner_ctx.is_hot();
 
+        if debug_layout {
+            self.debug_paint_layout_bounds(&mut inner_ctx, env);
+        }
+        if debug_ids {
+            self.debug_paint_widget_ids(&mut inner_ctx, env);
+        }
```
xStrom, line 371: The `debug_paint_widget_ids` method might call `debug_paint_layout_bounds`, so I think we can optimize it into something like this:
```rust
// Check the bool first, because most widgets aren't hot and we won't need to do the map lookup
let debug_ids = inner_ctx.is_hot() && env.get(Env::DEBUG_WIDGET_ID);

let debug_layout = if debug_ids {
    // Change debug_paint_widget_ids to return true if it called debug_paint_layout_bounds.
    // Then we won't need to call it again or even do the env map lookup.
    !self.debug_paint_widget_ids(&mut inner_ctx, env) && env.get(Env::DEBUG_PAINT)
} else {
    env.get(Env::DEBUG_PAINT)
};
if debug_layout {
    self.debug_paint_layout_bounds(&mut inner_ctx, env);
}
```

cmyr: okay, I think we can assume that debug_paint_widget_ids always paints the layout so I can simplify this a bit further, overall looks reasonable though.

Finnerale:
So, after that merge conflict irritated me, I forcefully reset my local git repo and now the rendering bug is gone :see_no_evil: . But it still does not seem to work entirely, some widgets do not show their id, or have it offset (which could cause some ids to be offscreen)

Finnerale:
After comparing the behavior with X11 in the calc example, it seems like ids are drawn to far below and right of the widget they belong to. Also they sometimes do not show up at all, probably because the place where they draw is not being invalidated.

xStrom:
I tested this on my Ubuntu 20.04 just now and both X11 and GTK work just fine. There are no rendering bugs and the ids are drawn in the correct location.

Maybe there's something wrong with your setup @Finnerale? Give `cargo clean` a try.

Finnerale:
Neither `cargo clean` nor a complete re-clone of the repo changed it.

Finnerale:
I can even reproduce this on my Ubuntu 20.04 laptop.

cmyr:
hm, so something like this *was* happening,  but I think I've addressed it in an update? is it possible you're on a commit other than 3396138?

Finnerale:
Nope, it is 3396138 . Like I said, I've deleted my local copy of druid and cloned it again, compiled it fresh and the issue still is there, on my Arch desktop and on my Ubuntu 20.04 laptop, both with up to date system and Rust.

cmyr:
that's really strange, because I had an issue similar to what you'd described, but it's now fixed. Can you share a screencap?

Finnerale:
So, I've found the reason why only I got it: It happens on Wayland, not on X11.

I've added two videos, the first is on Wayland, the second (same computer) on X11.
[wayland.webm](https://www.dropbox.com/s/22r5cwa7rjl6dri/wayland.webm?dl=1)
[x11.webm](https://www.dropbox.com/s/obrgbocwczalht3/x11.webm?dl=1)

xStrom:
I managed to reproduce the issue with Wayland. This problem has nothing to do with this PR though, the same problem already exists with the `anim` example. It seems to be related to `paint_with_z_index` and can be addressed separately from this PR here.

cmyr:
ah that makes sense, I had a similar issue on the coregraphics background where I couldn't trust the current transform returned by the platform, and have to track it separately myself.

PR #878, opened 2020-04-24
Fix incorrect LifeCycleCtx with FocusChanged event

cmyr:
We were handling this event separately, and were accidentally
reusing the parent context, instead of the child context;
this has the concrete problem that the context had
an incorrect WidgetId.

This fix feels like the lesser of two evils; we add a mechanism
for explicitly substituting a new event for the one received,
ensuring that we only call the child's lifecycle method in one place.

(this explains what I was doing with #876 😅)

### DIFF
--- a/druid/src/core.rs
+++ b/druid/src/core.rs
@@ -608,6 +608,10 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
     }
 
     pub fn lifecycle(&mut self, ctx: &mut LifeCycleCtx, event: &LifeCycle, data: &T, env: &Env) {
+        // in the case of an internal routing event, if we are at our target
+        // we may replace the routing event with the actual event
+        let mut substitute_event = None;
+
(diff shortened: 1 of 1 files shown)


druid/src/core.rs
```diff
@@ -641,10 +645,9 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
                         // Only send FocusChanged in case there's actual change
                         if old != new {
                             self.state.has_focus = change;
-                            let event = LifeCycle::FocusChanged(change);
-                            self.inner.lifecycle(ctx, &event, data, env);
+                            substitute_event = Some(LifeCycle::FocusChanged(change));
                         }
-                        false
+                        true
```
xStrom, line 650: The `true` should be inside the `if old != new` block and `false` otherwise. An `else` block with `false` would work.

cmyr: good catch

PR #879, opened 2020-04-27
Update more_information.md

derekdreery:
Add another project using druid.


derekdreery:
Cool I'll try to keep it up to date with druid so it serves as an example to learn from.

PR #882, opened 2020-04-28
Improve Windows invalidation logging.

xStrom:
I was reviewing the invalidation PR #817 and found room for a bit of improvement.
* More extensive logging in the Windows platform code.
* Quit the druid-shell example when the window is closed, like the other druid-shell examples.

### DIFF
--- a/druid-shell/examples/invalidate.rs
+++ b/druid-shell/examples/invalidate.rs
@@ -78,6 +78,10 @@ impl WinHandler for InvalidateTest {
         }
     }
 
+    fn destroy(&mut self) {
+        Application::global().quit()
+    }
+
     fn as_any(&mut self) -> &mut dyn Any {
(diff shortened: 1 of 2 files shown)


druid-shell/examples/invalidate.rs
```diff
@@ -78,6 +78,10 @@ impl WinHandler for InvalidateTest {
         }
     }
 
+    fn destroy(&mut self) {
+        Application::global().quit()
+    }
+
```
Zarenor, line 84: This looks unrelated - I assume this is for some future further additions to this example?

xStrom: Well this `invalidate` example never quits when the window is closed. The other druid-shell examples quit on window closing. This simple change fixes that.

PR #885, opened 2020-04-28
Fix partial invalidation on windows.

### LINKED ISSUE #875: Intense black flickering and artifacts on Windows
To check out what the invalidation example looks like on Windows for #872 and noticed intense flickering and rendering artifacts since partial invalidation (#817) has been merged.
Resetting druid to the last commit before the PR was merged gets rid of the flickering.

Does this happen on normal Windows? Or can someone else reproduce this for a VM?

jneem:
Fixes #875

Or, at least, it does on my win10 virtual machine.

### DIFF
--- a/druid-shell/src/platform/windows/window.rs
+++ b/druid-shell/src/platform/windows/window.rs
@@ -275,6 +275,9 @@ impl WndState {
         let anim;
         {
             let mut piet_ctx = Piet::new(d2d, dw, rt);
+            // The documentation on DXGI_PRESENT_PARAMETERS says we "must not update any
+            // pixel outside of the dirty rectangles."
+            piet_ctx.clip(invalid_rect);
             anim = self.handler.paint(&mut piet_ctx, invalid_rect);
             if let Err(e) = piet_ctx.finish() {
                 error!("piet error on render: {:?}", e);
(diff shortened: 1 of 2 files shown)


raphlinus:
It runs but the invalidation debug rect is the whole screen (in the `invalidation` example). I'm not sure if that's the expected behavior.

xStrom:
You should also delete the [temporary workaround](https://github.com/xi-editor/druid/blob/e21123c33afa7d2be2265e7d5422abe24bf159fe/druid-shell/src/platform/windows/window.rs#L1213-L1214).

druid-shell/src/platform/windows/window.rs
```diff
@@ -378,8 +381,14 @@ impl WndProc for MyWndProc {
...
                     if let Some(ref mut ds) = s.dcomp_state {
+                        let params = DXGI_PRESENT_PARAMETERS {
+                            DirtyRectsCount: 1,
+                            pDirtyRects: &mut rect as *mut RECT,
+                            pScrollRect: null_mut(),
+                            pScrollOffset: null_mut(),
+                        };
                         if !ds.sizing {
-                            (*ds.swap_chain).Present(1, 0);
+                            (*ds.swap_chain).Present1(1, 0, &params as *const _);
```
xStrom, line 391: Let's not hide the type with `_`. Can we do `*const DXGI_PRESENT_PARAMETERS`?

raphlinus: Why not just `(*ds.swap_chain).Present1(1, 0, &params);`? Seems to work for me. I personally consider `as *const _` to be idiomatically, as it reads "make this into the type winapi expects". But it's far nicer if the type just works without casting.

druid-shell/src/platform/windows/window.rs
```diff
@@ -378,8 +381,14 @@ impl WndProc for MyWndProc {
                         self.handle.borrow().rect_to_px(rect),
                     );
                     if let Some(ref mut ds) = s.dcomp_state {
+                        let params = DXGI_PRESENT_PARAMETERS {
+                            DirtyRectsCount: 1,
+                            pDirtyRects: &mut rect as *mut RECT,
```
raphlinus, line 386: Same, this doesn't seem to need the explicit cast to `*mut`.

PR #886, opened 2020-04-28
Fix gtk menu shortcuts.

jneem:
The key code returned by `KeyCode::into()` is already a gdk keyval, not a unicode value (not every keycode even has a unicode value).

Prior to this change, hotkeys like Alt+F4 were broken in gtk.


xStrom:
The menu shortcuts work fine on my Ubuntu 19.10 even without this change. This change doesn't break them either.

jneem:
Interesting. This is what it looks like, pre-patch, for me on Arch linux. Note the "Exit" shortcut.
Screenshot_2020-04-30_07-42-35

PR #889, opened 2020-04-29
Add a Changelog

Finnerale:
This introduces an in-tree changelog base on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).

Whenever a new PR ist created it should contain changelog entries for everything it changes, thus keeping the changelog always in sync with `master`.

Every changelog entry is annotated with the PR that introduced it and the author for attribution.

The attribution is also the reason why the `Maintenance` section exists, even though entries in it have no *direct* impact on users.

Annotating entries with a link to the introducing PR causes a chicken-and-egg problem tho.
When creating the PR you wont know the number yet, which you need to refer to it.
Having a quick link to the related PR seems very useful however as it provides much more information about the change than the changelog ever could.

See [Zulip](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/Bookkeeping.20about.20breaking.20changes) for previous discussion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Highlights
+
+- Basic X11 backend for druid-shell. ([#599] by [@crsaracco])
+
 ### Added
 
 - `TextBox` can receive `EditAction` commands. ([#814] by [@cmyr])
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,15 +9,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- `TextBox` can receive `EditAction` commands ([#814] by [@cmyr])
+- `TextBox` can receive `EditAction` commands. ([#814] by [@cmyr])
+
+- Added `Split::min_splitter_area(f64)` to add padding around the split bar. ([#738] by [@xStorm])
+
+- The `druid::text` module is public now. ([#816] by [@cmyr])
+
+- Basic X11 backend for druid-shell. ([#599] by [@crsarcco])
 
 ### Changed
 
-- `Split` constructors are now `Split::rows` and `columns`. ([#738] by [@xStorm])
+- `Split` constructors are now called `Split::rows` and `columns`. ([#738] by [@xStorm])
+
+- `Split::splitter_size` no longer includes padding. ([#738] by [@xStorm])
+
+- `Event::MouseMoved` has been renamed to `MouseMove`. ([#825] by [@teddemunnik])
 
 ### Deprecated
 
 ### Removed
+
+- The optional GTK feature for non-Linux platforms. ([#611] by [@pyroxymat])
 
 ### Fixed
 
@@ -27,12 +39,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - GTK windows get actually closed. ([#797] by [@finnerale])
 
+- Windows now respects the minimum window size. ([#727] by [@teddemunnik])
+
+- Windows now respects resizability. ([#712] by [@teddemunnik])
+
+- Mouse capturing on Windows. ([#695] by [@teddemunnik])
+
 ### Visual
 - Improved `Split` accuracy. ([#738] by [@xStorm])
 
 ### Docs
 
 - `Env` got a new example and usage hints. ([#796] by [@finnerale])
+
+- Usage of boom filters got documented. ([#818] by @xStorm)]
 
 ### Cleanups
 
@@ -42,6 +62,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - A new project using druid: [Kondo](https://github.com/tbillington/kondo) Save disk space by cleaning unneeded files from software projects.
 
+[#599]: https://github.com/xi-editor/druid/pull/599
+[#611]: https://github.com/xi-editor/druid/pull/611
+[#695]: https://github.com/xi-editor/druid/pull/695
+[#712]: https://github.com/xi-editor/druid/pull/712
+[#727]: https://github.com/xi-editor/druid/pull/727
 [#738]: https://github.com/xi-editor/druid/pull/738
 [#782]: https://github.com/xi-editor/druid/pull/782
 [#796]: https://github.com/xi-editor/druid/pull/796
@@ -49,11 +74,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#798]: https://github.com/xi-editor/druid/pull/798
 [#808]: https://github.com/xi-editor/druid/pull/808
 [#814]: https://github.com/xi-editor/druid/pull/814
+[#816]: https://github.com/xi-editor/druid/pull/816
+[#818]: https://github.com/xi-editor/druid/pull/818
+[#825]: https://github.com/xi-editor/druid/pull/825
 
 ## [0.5] - 2020-04-01
 
 Last release without a changelog :(
 
+
+[@pyroxymat]: https://github.com/pyroxymat
+
+[@crsaracco]: https://github.com/crsaracco
+
+[@teddemunnik]: https://github.com/teddemunnik
 
 [@xStorm]: https://github.com/xStorm
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,7 +60,7 @@ While still incomplete this lays the foundation for running druid on Linux witho
 
 - `Env` got a new example and usage hints. ([#796] by [@finnerale])
 
-- Usage of bloom filters got documented. ([#818] by @xStorm)]
+- Usage of bloom filters got documented. ([#818] by [@xStorm])
 
 ### Cleanups
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,8 @@ While some features like the clipboard, menus or file dialogs are not yet availa
 
 ### Deprecated
 
+- Nothing
+
 ### Removed
 
 - The optional GTK feature for non-Linux platforms. ([#611] by [@pyroxymat])
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -182,6 +182,7 @@ Last release without a changelog :(
 [@elrnv]: https://github.com/elrnv
 [@tedsta]: https://github.com/tedsta
 [@kindlychung]: https://github.com/kindlychung
+[@jneem]: https://github.com/jneem
 [@fishrockz]: https://github.com/fishrockz
 [@thecodewarrior]: https://github.com/thecodewarrior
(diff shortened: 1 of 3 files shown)


xStrom:
The chicken and egg problem isn't that bad because github issue numbers are sequential and you can guess what number you will get. Sometimes you lose the race, but that should be rare.

Finnerale:
Right, they are sequential that makes it a bit better but then you would have to figure out what the current count is at, I wonder if there is like a 'next PR number' badge that we could add to the readme for example.

CHANGELOG.md
```diff
@@ -0,0 +1,162 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
```
xStrom, line 6: Do we really want this preface text at the top? Other notable projects like cargo don't have anything like that in [their changelog](https://github.com/rust-lang/cargo/blob/master/CHANGELOG.md).

If anything I think an adaptation of this section could live in the main readme. Perhaps something like the following at the bottom of the first introduction section:
```markdown
Druid adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html) and documents
all its changes in [the changelog](https://github.com/xi-editor/druid/blob/master/CHANGELOG.md).
```

CHANGELOG.md
```diff
@@ -0,0 +1,162 @@
...
+
+#### Basic X11 backend for druid-shell. ([#599])
+
+[@crsaracco] has implemented basic support to run druid on bare-metal X11!
+
+While still incomplete this lays the foundation for running druid on Linux without relying on GTK.
+
+#### Mostly complete Wasm backend for druid-shell. ([#759])
+
+[@elrnv] continued the work of [@tedsta] and implemented a mostly complete Wasm backend and enabled all druid example to [run in the browser](https://elrnv.github.io/druid-wasm-examples/).
```
xStrom, line 20: ~example~ examples

CHANGELOG.md
```diff
@@ -0,0 +1,162 @@
...
+
+While some features like the clipboard, menus or file dialogs are not yet available, all fundamental features are there.
+
+### Added
+
+- `TextBox` can receive `EditAction` commands. ([#814] by [@cmyr])
+- `Split::min_splitter_area(f64)` to add padding around the split bar. ([#738] by [@xStorm])
```
xStrom, line 27: ~split bar~ splitter bar

CHANGELOG.md
```diff
@@ -0,0 +1,162 @@
...
+- Published `druid::text` module. ([#816] by [@cmyr])
+- Basic X11 backend for druid-shell. ([#599] by [@crsaracco])
```
xStrom, line 29: Maybe already highlighted changes don't need to show up in the listing?

CHANGELOG.md
```diff
@@ -0,0 +1,162 @@
...
+- `InternalEvent::MouseLeave` signalling the cursor leaved the window. ([#821] by [@teddemunnik])
```
xStrom, line 30: ~leaved~ left

CHANGELOG.md
```diff
@@ -0,0 +1,162 @@
...
+- `children_changed` now always includes layout and paint request. ([#839] by [@xStorm])
+- Mostly complete Wasm backend for druid-shell. ([#759] by [@elrnv])
```
xStrom, line 32: Maybe already highlighted changes don't need to show up in the listing?

CHANGELOG.md
```diff
@@ -0,0 +1,162 @@
...
+
+- GTK: Use the system locale. ([#798] by [@finnerale])
+- GTK: Actually close windows ([#797] by [@finnerale])
+- Windows: Respect the minimum window size. ([#727] by [@teddemunnik])
+- Windows: Respect resizability. ([#712] by [@teddemunnik])
+- Windows: Capture mouse for drag actions. ([#695] by [@teddemunnik])
+- Start focus cycling from non-registered-for-focus widgets. ([#819] by [@xStorm])
+- Propagate `Event::FocusChanged` to focus gaining widgets as well. ([#819] by [@xStorm])
+- GTK: Prevent crashing on pop-ups. ([#837] by [@finnerale])
+- Keep hot state  consistent with mouse position. ([#841] by [@xStorm])
```
xStrom, line 66: Remove double space.

CHANGELOG.md
```diff
@@ -0,0 +1,162 @@
...
+- Open file menu item works again. ([#851] by [@kindlychung])
+- Supply correct `LifeCycleCtx` to `Event::FocusChanged`. ([#878] by [@cmyr])
+- Windows: Termiate app when all windows have closed. ([#763] by [@xStorm])
```
xStrom, line 69: This same #763 ended up containing a macOS fix too. Perhaps add the following line:
```markdown
- macOS: `Application::quit` now quits the run loop instead of killing the process. ([#763] by [@xStrom])
```

CHANGELOG.md
```diff
@@ -0,0 +1,162 @@
...
+- Basic X11 backend for druid-shell. ([#599] by [@crsaracco])
+- `InternalEvent::MouseLeave` signalling the cursor leaved the window. ([#821] by [@teddemunnik])
+- `children_changed` now always includes layout and paint request. ([#839] by [@xStorm])
+- Mostly complete Wasm backend for druid-shell. ([#759] by [@elrnv])
+- `UpdateCtx::submit_command`. ([#855] by [@cmyr])
+- `request_paint_rect` for partial invalidation. ([#817] by [@jneem])
+- Window title can be any `LabelText` (such as a simple `String`). ([#869] by [@cmyr])
+- `Label::with_font` and `set_font`. ([#785] by [@thecodewarrior])
+
+### Changed
```
xStrom, line 38: There's another significant change:

```markdown
- Global `Application` associated functions are now instance methods instead, e.g. `Application::global().quit()` instead of the old `Application::quit()`. ([#763] by [@xStrom])
```

CHANGELOG.md
```diff
@@ -0,0 +1,162 @@
...
+- GTK: Prevent crashing on pop-ups. ([#837] by [@finnerale])
+- Keep hot state  consistent with mouse position. ([#841] by [@xStorm])
+- Open file menu item works again. ([#851] by [@kindlychung])
+- Supply correct `LifeCycleCtx` to `Event::FocusChanged`. ([#878] by [@cmyr])
+- Windows: Termiate app when all windows have closed. ([#763] by [@xStorm])
+
+### Visual
+
+- Improved `Split` accuracy. ([#738] by [@xStorm])
+- Build-in widgets no longer stroke outside their `paint_rect`. ([#861] by [@jneem])
```
xStrom, line 74: ~Build-in~ Built-in

CHANGELOG.md
```diff
@@ -0,0 +1,162 @@
...
+- Added hot glow option to multiwin example. ([#845] by [@xStorm])
+
+### Maintenance
+
+- Replace `#[macro_use]` with normal `use`. ([#808] by [@totsteps])
+- Clippy checks are now enabled for all targets. ([#850] by [@xStorm])
+- Added rendering tests. ([#784] by [@fishrockz])
+- CI testing has been revamped. ([#857] by [@xStorm])
+- Associate timers with widget ids. ([#831] by [@sjoshid])
+- Add a changelog. ([#889] by [@finnerale])
```
xStrom, line 91: I don't think future tense makes sense for a changelog. So this should be *Added* not *Add*. Should probably check other lines too. Generally speaking I guess both past tense (*Added a changelog*) and present tense (*There is now a changelog*) are fine and both can be used, but yeah not future.

Finnerale: Yea I thing I've mixed tenses a bit :sweat_smile: 
I'm not quite sure what fits better (present or past tense).

CHANGELOG.md
Finnerale: I've removed the preface.
Linking the changelog seams like a good idea.
Considering that semantic versioning is standard for Rust anyways I don't think that has to be mentioned.

Finnerale: I've now also added a link to the readme.

Finnerale:
@xStrom I've fixed your name (sorry for that), added some other things I've missed and hopefully corrected the tenses.

I'm not sure if we should include changes to internal-but-public API such as `InternalEvent`. For now I've just added them.

xStrom:
We should also add a section/note to `CONTRIBUTING.md` about needing to change the changelog in PRs.

I think documenting the internal stuff is fine while we're in the 0.x versions. We can reconsider this later, e.g. once we hit 1.0.

Also #890 got merged.

Finnerale:
I've added it to `CONTRIBUTING.md`.

I've not added anything regarding #890 as it is a follow-up for a feature that is not released yet.
I think keeping the changelog compact and only freezing entries once the version is released makes it easier to read. For example I've merged the `Event::MouseLeave` addition with your effort to collect `InternalEvent`s. Thus there is only one entry mentioning the addition of `InternalEvent::MouseLeave`.

xStrom:
I think that makes sense. We don't want the changelog to be a mirror copy of the merge commit log. Multiple PRs working on the same conceptual feature can be streamlined into a single entry in the changelog.

CHANGELOG.md
```diff
@@ -0,0 +1,159 @@
...
+[#850]: https://github.com/xi-editor/druid/pull/850
+[#851]: https://github.com/xi-editor/druid/pull/851
+[#855]: https://github.com/xi-editor/druid/pull/855
+[#857]: https://github.com/xi-editor/druid/pull/857
+[#861]: https://github.com/xi-editor/druid/pull/861
+[#869]: https://github.com/xi-editor/druid/pull/869
+[#878]: https://github.com/xi-editor/druid/pull/878
+[#889]: https://github.com/xi-editor/druid/pull/899
+
+## [0.5] - 2020-04-01
```
xStrom, line 137: I think we should always use the full semver version. `0.5` -> `0.5.0`.

xStrom:
For completeness sake, let's add [all the previous versions](https://crates.io/crates/druid/versions) to the changelog as well. I know we don't have the the actual changes for them right now, but we could at least have the version numbers and their dates.

CHANGELOG.md
```diff
@@ -134,10 +134,17 @@ While some features like the clipboard, menus or file dialogs are not yet availa
...
 [#889]: https://github.com/xi-editor/druid/pull/899
 
-## [0.5] - 2020-04-01
+## [0.5.0] - 2020-04-01
 
 Last release without a changelog :(
 
+## [0.4.0] - 2019-12-28
+## [0.3.2] - 2019-11-05
+## [0.3.1] - 2019-11-04
```
xStrom, line 143: Shouldn't this be 2019-11-05 as well? :thinking:

Finnerale: @xStrom I'm not sure what you mean, it seems correct to me.

xStrom: Haha, well if it's not a typo then both GitHub and crates.io show the date in my timezone. That's why I see 2019-11-05.

PR #895, opened 2020-04-30
Provide additional modifiers as part of the mouse wheel event

teddemunnik:
As part of the prototype I'm working on I wanted to be able to access the current mouse position as part of mouse wheel events, in order to implement zoom to mouse cursor. Since we already supply shared modifiers between other mouse events, and from following the discussion in #843, it seemed somewhat reasonable to do this by changing `WinHandler::wheel` to take a `MouseEvent` like the other mouse events, and adding `wheel_delta` to `MouseEvent`. 

I've implemented the changes I am suggesting above as a draft on the windows and GTK backends, but work on the other backends is remaining.


druid-shell/src/platform/windows/window.rs
```diff
@@ -649,31 +649,51 @@ impl WndProc for MyWndProc {
...
-                    let delta = Vec2::new(delta_x, 0.0);
-                    let mods = get_mod_state();
-                    s.handler.wheel(delta, mods);
+
+                    let mut p = POINT {
+                        x: LOWORD(lparam as u32) as i16 as i32,
+                        y: HIWORD(lparam as u32) as i16 as i32,
+                    };
+                    unsafe {
+                        if ScreenToClient(hwnd, &mut p) == 0 {
```
xStrom, line 671: Seems to return `BOOL` so let's check for `FALSE`.

druid-shell/src/platform/windows/window.rs
```diff
@@ -649,31 +649,51 @@ impl WndProc for MyWndProc {
...
+                            warn!(
```
xStrom, line 672: I know the Windows platform code isn't consistent about it but we're moving towards not importing functions directly and qualifying them. So let's change this from `warn!` to `log::warn!`.

druid-shell/src/platform/windows/window.rs
```diff
@@ -649,31 +649,51 @@ impl WndProc for MyWndProc {
...
+                                "ScreenToClient failed: {}",
+                                Error::Hr(HRESULT_FROM_WIN32(GetLastError()))
+                            );
```
xStrom, line 675: I'm not sure that returning bad data is better than ignoring the event. Perhaps we shouldn't create the event when we can't get the correct point.

druid-shell/src/platform/windows/window.rs
```diff
@@ -649,31 +649,51 @@ impl WndProc for MyWndProc {
...
+                    }
+
+                    let (px, py) = self.handle.borrow().pixels_to_px_xy(p.x, p.y);
+                    let pos = Point::new(px as f64, py as f64);
+                    let mods = KeyModifiers {
+                        shift: wparam & MK_SHIFT != 0,
+                        alt: get_mod_state_alt(),
+                        ctrl: wparam & MK_CONTROL != 0,
+                        meta: get_mod_state_win(),
+                    };
```
xStrom, line 686: We should move this up to replace the previous `let mods = get_mod_state();`. Also we should probably be specific about using the `LOWORD` of `wparam` as the docs state.

druid-shell/src/platform/windows/window.rs
```diff
@@ -649,31 +649,51 @@ impl WndProc for MyWndProc {
...
+                    let buttons = get_buttons(wparam);
```
xStrom, line 687: Let's `LOWORD` extract here too.

teddemunnik:
Thanks for the review @xStrom! I implemented all the changes you requested in the windows shell, let me know if I understood correctly the changes with LOWORD extracting.

xStrom:
Yeah the `LOWORD` stuff is mostly fine, but I think let's do it only once.

```rust
let system_delta = HIWORD(wparam as u32) as i16 as f64;
let down_state = LOWORD(wparam as u32) as usize;
```

Then the subsequent use can just be on `down_state`.

cmyr:
I think this is reasonable. The only alternative I can think of is to have some general mechanism that exposes the mouse position, and plumb that through `EventCtx`.

CHANGELOG.md
```diff
@@ -43,6 +43,7 @@ While some features like the clipboard, menus or file dialogs are not yet availa
 - `request_timer` uses `Duration` instead of `Instant`. ([#847] by [@finnerale])
 - Global `Application` associated functions are instance methods instead, e.g. `Application::global().quit()` instead of the old `Application::quit()`. ([#763] by [@xStrom])
 - Timer events will only be delivered to the widgets that requested them. ([#831] by [@sjoshid])
+- `Event::Wheel` now contains a `MouseEvent` structure. ([#895] by [@teddemunnik])
```
xStrom, line 46: You also need to add the following to the bottom of the file with the other links:
```md
[#894]: https://github.com/xi-editor/druid/pull/894
[#895]: https://github.com/xi-editor/druid/pull/895   <------ Add there
[#897]: https://github.com/xi-editor/druid/pull/897
```

We're hoping to automate generating these links soon.

teddemunnik:
@cmyr getting the mouse position through `EventCtx`, sounds pretty neat too, and I could imagine that being useful for getting the mouse position at other points in the application, even if this PR would be merged. Looking at Web API `WheelEvent` inherits from `MouseEvent`, but adding the wheel event info to the `MouseEvent` seemed like a simpler approach for rust.

PR #898, opened 2020-05-02
Add UpdateCtx::request_timer and request_anim_frame.

Finnerale:
This adds `UpdateCtx::request_timer` and `request_anim_frame`.
Also the `Switch` widget animates external changes of its state, the `switch` example now has a checkbox next to the `Switch` to test this.


druid/src/widget/switch.rs
```diff
@@ -181,6 +181,8 @@ impl Widget<bool> for Switch {
     fn update(&mut self, ctx: &mut UpdateCtx, old_data: &bool, data: &bool, _env: &Env) {
         if old_data != data {
             ctx.request_paint();
```
zaynetro, line 183: `request_paint` is no longer necessary here as we call it from `request_anim_frame` now.

PR #903, opened 2020-05-06
Add file filtering to gtk dialogs.

### LINKED ISSUE #390: Implement file dialog default extension in mac & gtk
The `FileDialogOptions` struct has a field `default_type: Option<FileSpec>` that specifies which of the types in `allowed_types: Option<Vec<FileSpec>>` is the default. If no `default_type` is specified or if `default_type` isn't actually among `allowed_types` then the first entry of `allowed_types` is the default.

PR #387 implemented it for Windows. In Windows `allowed_types` translate into a dropdown control in the file picker dialog and `default_type` determines which item in that dropdown is preselected.

This behavior needs to also be implemented in the `mac` and `gtk` platform code. If this behavior can't be implemented or doesn't make sense on the platform, then comments describing that are definitely welcome.

jneem:
Fixes GTK part of #390.

### DIFF
--- a/druid-shell/src/platform/gtk/dialog.rs
+++ b/druid-shell/src/platform/gtk/dialog.rs
@@ -16,10 +16,19 @@
 
 use std::ffi::OsString;
 
-use gtk::{FileChooserAction, FileChooserExt, NativeDialogExt, ResponseType, Window};
+use gtk::{FileChooserAction, FileChooserExt, FileFilter, NativeDialogExt, ResponseType, Window};
 
-use crate::dialog::{FileDialogOptions, FileDialogType};
+use crate::dialog::{FileDialogOptions, FileDialogType, FileSpec};
 use crate::Error;
+
+fn file_filter(fs: &FileSpec) -> FileFilter {
+    let ret = FileFilter::new();
+    ret.set_name(Some(fs.name));
+    for ext in fs.extensions {
+        ret.add_pattern(&format!("*.{}", ext));
+    }
+    ret
+}
 
 pub(crate) fn get_file_dialog_path(
     window: &Window,
@@ -44,6 +53,15 @@ pub(crate) fn get_file_dialog_path(
     dialog.set_show_hidden(options.show_hidden);
 
     dialog.set_select_multiple(options.multi_selection);
+
+    if let Some(ref file_types) = options.allowed_types {
+        for f in file_types {
+            dialog.add_filter(&file_filter(f));
+        }
+    }
+    if let Some(ref default_file_type) = options.default_type {
+        dialog.set_filter(&file_filter(default_file_type));
+    }
 
     let result = dialog.run();
 
--- a/druid-shell/src/platform/gtk/dialog.rs
+++ b/druid-shell/src/platform/gtk/dialog.rs
@@ -54,13 +54,34 @@ pub(crate) fn get_file_dialog_path(
 
     dialog.set_select_multiple(options.multi_selection);
 
-    if let Some(ref file_types) = options.allowed_types {
+    let mut found_default_filter = false;
+    if let Some(file_types) = &options.allowed_types {
         for f in file_types {
-            dialog.add_filter(&file_filter(f));
+            let filter = file_filter(f);
+            dialog.add_filter(&filter);
+
+            if let Some(default) = &options.default_type {
+                if default == f {
+                    // Note that we're providing the same FileFilter object to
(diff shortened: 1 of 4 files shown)


druid/examples/open_save.rs
```diff
@@ -0,0 +1,105 @@
...
+        .launch(data)
+        .expect("launch failed");
+}
+
+fn ui_builder() -> impl Widget<String> {
+    let txt = FileSpec::new("Text file", &["txt"]);
+    let other = FileSpec::new("Bogus file", &["foo", "bar", "baz"]);
+    let save_dialog_options = FileDialogOptions::new()
+        .allowed_types(vec![txt, other])
+        .default_type(txt);
```
xStrom, line 39: Let's reverse the vector to `[other, txt]` so that this sample can show the default type working. Does it work for you? On my Ubuntu 19.10 it doesn't seem to actually choose the text file if `other` is first in the vector.

druid-shell/src/platform/gtk/dialog.rs
```diff
@@ -45,6 +54,15 @@ pub(crate) fn get_file_dialog_path(
 
     dialog.set_select_multiple(options.multi_selection);
 
+    if let Some(ref file_types) = options.allowed_types {
+        for f in file_types {
+            dialog.add_filter(&file_filter(f));
+        }
+    }
+    if let Some(ref default_file_type) = options.default_type {
```
xStrom, line 62: Let's remove the `ref` keywords and use `&options....` instead. `ref` is [a legacy pattern](https://doc.rust-lang.org/1.33.0/book/ch18-03-pattern-syntax.html#legacy-patterns-ref-and-ref-mut) and can be confusing to newer Rust users.

druid/examples/open_save.rs
```diff
@@ -0,0 +1,105 @@
...
+    col.add_child(save);
+    col.add_child(open);
+    Align::centered(col)
+}
+
+impl AppDelegate<String> for Delegate {
+    fn command(
+        &mut self,
+        _ctx: &mut DelegateCtx,
+        _target: &Target,
```
xStrom, line 74: Rebase with `master` and change `&Target` to `Target`.

druid/examples/open_save.rs
jneem: Indeed, it wasn't working with the other order. I've added some extra checks in dialog.rs, and now all the combinations I've tried seem to work.

druid-shell/src/platform/gtk/dialog.rs
```diff
@@ -45,6 +54,36 @@ pub(crate) fn get_file_dialog_path(
...
+        }
+    }
+
+    if let Some(default_file_type) = &options.default_type {
+        if options.allowed_types.is_some() && !found_default_filter {
+            // It's ok to set a default file filter without providing a list of
+            // allowed filters, but it's not ok (or at least, doesn't work in gtk)
+            // to provide a default filter that isn't in the (present) list
+            // of allowed filters.
+            log::warn!("default file type not found in allowed types");
```
xStrom, line 81: This warning should be removed, because there is [an explictly documented behavior](https://github.com/xi-editor/druid/blob/af20c0906b6d62e4dc6e2568446f5bbda0de98d7/druid-shell/src/dialog.rs#L109-L110) for this case.

*If it's `None` or not present in `allowed_types` then the first entry in `allowed_types` will be used as default.*

Finnerale: But it seems unreasonable for the user to provide a default that is not even available.
If this happens, it most definitely seems to be an accident and should log a warning (on all platforms), at least I think so.

xStrom: I guess that's reasonable. We can keep the warning then.

jneem: Note that (and I can change this if it isn't desired druid behavior), if allowed_types is empty then the default type *is* allowed to be present: it applies a filter that cannot be changed.

jneem: Just to be clear, my previous comment is about gtk's behavior.

Finnerale: So if I interpret that code correctly, on Windows `Some(default_type)` is ignored if `allowed_types` is `None`. On mac the default type seems to be ignored all together.

I'm not sure what the 'correct' behavior would be (at least not the mac one).
Maybe we should allow having default be `Some` while allowed is `None`?

PR #904, opened 2020-05-07
Refactor DPI scaling and fix the GTK implementation.

### LINKED ISSUE #862: every 4 or 5 pixels, widget gets sized slightly smaller than the window
Running 0.5.0 on linux (openbox window manager, compositing on), a single widget with a layout implementation that just returns `bc.max()` will sometimes (even after resize events have ceased) be left lagged a little bit behind the real size of the window. A single pixel void sits open between the widget and the edge of the window.

to reproduce, get simple program with maximised widget:
```rust
use druid::{ AppLauncher, WindowDesc, LocalizedString, widget::{ Button } };
fn main(){
	AppLauncher::with_window(WindowDesc::new(|| Button::new("uh"))).launch(()).unwrap();
}
```

Resize it quickly, you'll see a grey edge flicker in and out when increasing the size. If you stop abruptly enough, sometimes it will stay visible.

xStrom:
## Background
This work got started due to the bug report in #862. @makoConstruct ran into an issue with GTK where there would occasionally be a single line of unused space in the window. I confirmed that this was an issue related to DPI scaling. @makoConstruct had a DPI of 75.0 but the issue appears with a bunch of other DPIs too.

## Reproducing the bug
On my Ubuntu 19.10 I could easily reproduce these issues by changing my DPI via [gnome-tweaks](https://askubuntu.com/a/1034071/586162). Changing the text scaling factor is what changes the DPI. Entering `0.78` will roughly match a DPI of 75.0. Other fractional numbers work too. In my testing the GTK DPI handling was completely broken.

## The calculator with `master` and with this PR
before after

## Getting more value out of time
As I already had researched the DPI code thoroughly to understand how to fix GTK, I figured I might as well improve the `druid-shell` DPI scaling system in general.

I added the `Scale` struct, which is a platform independent resolution scaling helper. It can do proper scaling per axis and has a bunch of helper methods for converting between logical and platform pixels.

Previously `druid-shell` was leaking some platform pixels via `WinHandler::size`, which were handled by e.g. `druid`. I moved that into `druid-shell` so that all DPI related work is a `druid-shell` implementation detail. If someone really wants then they can still get information about the platform pixels and the scaling by fetching the `Scale` via `WindowHandle::get_scale`.

I also updated the Windows and web backends to make use of the new `Scale` system.

## Pixel terminology
Druid has been using the web terminology where a pixel and a pixel aren't the same thing. This is confusing to say the least. There is the CSS pixel unit (abbreviated **px**) and the device pixel (abbreviated **px**). I know we've been generally following web naming, but I'm convinced this is not a case where it makes sense. The situation in the web world with px and px isn't the result of some wise design, it's just an overloading of a term in the name of backwards compatibility because all the existing docs/code already use pixel/px. We have a chance to be much less ambiguous and confusing.

Unfortunately there isn't really industry consensus about what to call logical pixels.
- Microsoft's [device independent pixel](https://docs.microsoft.com/en-us/windows/win32/learnwin32/dpi-and-device-independent-pixels) - **dip**
- Google's [density independent pixel](https://material.io/design/layout/pixel-density.html#density-independence) - ~**dip**~ - they're in the process of rebranding as **dp** (pronounced, I kid you not, **dip**)
- Apple's [point](https://developer.apple.com/design/human-interface-guidelines/ios/icons-and-images/image-size-and-resolution/) - **pt**

Are there any others with good abbreviation?

I think it's important that we move away from calling both platform pixels and logical pixels - pixels. I think we should keep platform pixels as pixels and use something else for logical pixels, something that isn't confusing and has a good abbreviation.

The problem with **dip**/**dp** is that for a newcomer it's easy to imagine it meaning *device pixel*. It's not even that wrong, that's what the **d** truly stands for! We want the opposite effect.

The problem with **pt** is that it would be overloading a different existing unit, from the text/print world. It also has a bit of a clash with `Point`.

It's not super clear to me what the best choice is. Suggestions and discussion is welcome on this. What I do know is that I want to stop the use of pixel for both concepts.

For this PR I went with point/**pt**, but it can be changed if we think another choice is superior.

## Future work
This PR addresses the DPI issue on the shell side, but there's still a remaining issue on the druid side. While the shell issue revealed itself as an empty pixel line at the edge, the druid issue reveals itself as a clipped pixel line at the same edges. It has to do with `BoxConstraints` expanding to integers. The solution however isn't as trivial as just removing the expansion, because some other widgets like `Flex` depend on that behavior. A more thorough solution is needed, which is out of scope for this PR here.

Fixes #862


druid-shell/src/scale.rs
```diff
@@ -0,0 +1,289 @@
...
+    /// Converts from points into pixels, using the y axis scale factor.
+    #[inline]
+    pub fn pt_to_px_y<T: Into<f64>>(&self, y: T) -> f64 {
+        y.into() * self.scale_y
+    }
+
+    /// Converts from points into pixels,
+    /// using the x axis scale factor for `x` and the y axis scale factor for `y`.
+    #[inline]
+    pub fn pt_to_px_xy<T: Into<f64>>(&self, x: T, y: T) -> (f64, f64) {
```
cmyr, line 179: this is a very annoying suggestion from me, _but_: if we had a trait `Scaleable`, which we implemented for our geometry types, we could just have,

```rust
trait Scaleable {
    fn to_display_point(&self, scale: &Scale) -> Self;
    fn to_pixel(&self, scale: &Scale) -> Self;
}

impl Scale {
    fn to_display_point<T: Scaleable>(&self, geom: &T) -> T {
        geom.to_display_point(self)
    }
    // etc
}
```

This would at least reduce the overall API surface?

xStrom:
I think having a `Scalable` trait is an improvement and will allow for easier extensions, so I implemented it.

I think *display point* / **dp** is a fine choice, so I changed everything to that. Using `dp` in method names will also allow for some backwards compatible rebranding if the need arises.

druid-shell/src/mouse.rs
```diff
@@ -21,9 +21,7 @@ use crate::keyboard::KeyModifiers;
 /// Information about the mouse event.
 #[derive(Debug, Clone, PartialEq)]
 pub struct MouseEvent {
-    /// The location of the mouse in the current window.
-    ///
-    /// This is in px units not device pixels, that is, adjusted for hi-dpi.
+    /// The location of the mouse in display points in relation to the current window.
```
cmyr, line 24: I would include a section in the docs of `scale.rs` that explicitly explains our use of 'pixel/px' and 'display point/dp', and then we can link to that section everywhere we use one of those terms?

It would be nice if we had some canonical doc explanation of the dp/px distinction, and then we could link to that everywhere we use one of those terms? I'd make this super obvious, something with a section in the `//! top level` doc comment called like '## Pixels (px) and display points (dp)'. Or something? I'd just like it to be easy for someone reading docs (more important in druid than druid-shell I guess) to be able to easily access an explanation.

druid-shell/src/platform/web/window.rs
```diff
@@ -230,16 +221,19 @@ fn setup_resize_callback(ws: &Rc<WindowState>) {
...
-        let physical_height = (dpr * css_height) as u32;
-        state.dpr.replace(dpr);
-        state.canvas.set_width(physical_width);
-        state.canvas.set_height(physical_height);
-        let _ = state.context.scale(dpr, dpr);
-        state
-            .handler
-            .borrow_mut()
-            .size(physical_width, physical_height);
+        let size_dp = state.scale.try_borrow_mut().map_or(None, |mut scale| {
```
cmyr, line 224: if this is converting an error to an option I would use `.ok().map(_)`.

druid-shell/src/platform/web/window.rs
```diff
@@ -230,16 +221,19 @@ fn setup_resize_callback(ws: &Rc<WindowState>) {
...
+            *scale = Scale::from_scale(dpr, dpr);
+            let size_px = scale.set_size_dp(Size::new(css_width, css_height));
+            state.canvas.set_width(size_px.width as u32);
+            state.canvas.set_height(size_px.height as u32);
+            let _ = state.context.scale(scale.scale_x(), scale.scale_y());
+            Some(scale.size_dp())
+        });
```
cmyr, line 231: does it make sense to just move this to a method on the state type? `try_update_scale` or something? Not sure, just seems like maybe.

druid-shell/src/platform/web/window.rs
```diff
@@ -544,34 +541,16 @@ impl WindowHandle {
...
+    /// Get the `Scale` of the window.
+    pub fn get_scale(&self) -> Result<Scale, ShellError> {
+        if let Some(state) = self.0.upgrade() {
+            match state.scale.try_borrow() {
+                Ok(scale) => Ok(scale.clone()),
+                Err(_) => Err(BorrowError::new("WindowHandle::get_scale", "scale", false).into()),
+            }
+        } else {
+            Err(ShellError::WindowDropped)
+        }
```
cmyr, line 553: just for fun, I probably write this as:

```rust
self.0
    .upgrade()
    .ok_or(ShellError::WindowDropped)?
    .scale
    .try_borrow()
    .map_err(|_| BorrowError::new("WindowHandle::get_scale", "scale", false).into())
    .map(|s| s.clone())
```

druid-shell/src/platform/windows/util.rs
```diff
@@ -103,6 +105,28 @@ impl FromWide for [u16] {
     }
 }
 
+/// Converts a `Rect` to a winapi `RECT`.
+#[inline]
+pub fn rect_to_recti(rect: Rect) -> RECT {
```
cmyr, line 110: `pub(crate)` here, probably; the main advantage is we'll get compiler warnings in the future if these are unused. (Probably not an issue in practice since this module shouldn't be public, but good general strategy I think)

xStrom:
Non-uniform x/y scaling is definitely not common. It is supported by drawing, but the platform scale factor provided by Windows/macOS/web is always uniform from what I've gathered. However Linux supports it. So I guess this is mostly a question if we want to support it in Linux. The cost of supporting it isn't that high as it's mostly an implementation detail of `Scale`, especially if you do conversions on thigs that implement `Scalable`.

I haven't really looked into DPI scaling images yet, because I felt this PR was getting big as it is already. However that's definitely the next significant step on the road to DPI success. In general yes the app could provide a set of images and then we should select the smallest that is equal or bigger than the target area.

Having `Scale` in `Env` might be useful. I'll look more into that and see what can be done and if it makes sense.

druid-shell/src/mouse.rs
xStrom: Sure, I'll write some more docs about it. I also already wrote a book chapter about it in #913.

druid-shell/src/platform/web/window.rs
xStrom: Sure, why not. I moved it to its own method.

xStrom:
Okay I made changes based on the feedback. I added some more documentation around display points and a bunch of doc links pointing to it. I surfaced the `Scale` type in druid, which also surfaces that documentation.

I looked into adding `Scale` to `Env` but there was no clear way to do it, because `Env` seems global. However `Scale` can be different per window. Not sure if `Env` should grow the capability to have per-window state or if we should have some sort of `Event::Scale(Scale)` event.

I did add the `scale` method to `WinHandler` so the `druid-shell` part of informing changes to this is done. The druid side will take more thinking and currently it will ignore that change. I think it may make more sense to figure out the bitmap scaling story first - as in does scaling images even work right now and how to do it. Then we can figure out how to best surface scale changes.

In that sense I think this PR can just be the shell part, which is done. We can deal with the druid side in a different PR as we have more answers.

druid-shell/src/scale.rs
```diff
@@ -0,0 +1,388 @@
...
+
+    /// Returns the y axis platform DPI associated with this `Scale`.
+    #[inline]
+    pub fn dpi_y(&self) -> f64 {
+        self.dpi_y
+    }
+
+    /// Returns `true` if the specified DPI is approximately equal to the `Scale` DPI.
+    #[inline]
+    pub fn dpi_approx_eq(&self, dpi_x: f64, dpi_y: f64) -> bool {
```
cmyr, line 153: is there a specific rationale for this, instead of say making the caller create a `Scale` and having an `approx_eq` for that?

druid-shell/src/scale.rs
```diff
@@ -0,0 +1,388 @@
...
+    /// Converts from pixels into display points, using the y axis scale factor.
+    #[inline]
+    pub fn px_to_dp_y<T: Into<f64>>(&self, y: T) -> f64 {
+        y.into() / self.scale_y
+    }
+
+    /// Converts from pixels into display points,
+    /// using the x axis scale factor for `x` and the y axis scale factor for `y`.
+    #[inline]
+    pub fn px_to_dp_xy<T: Into<f64>>(&self, x: T, y: T) -> (f64, f64) {
```
cmyr, line 212: can't we just rely on the `Scaleable` impl on `Vec2`/`Point` etc?

druid-shell/src/scale.rs
```diff
@@ -0,0 +1,388 @@
...
+    /// using the x axis scale factor for coordinates on the x axis
+    /// and the y axis scale factor for coordinates on the y axis.
+    #[inline]
+    pub fn to_px<T: Scalable>(&self, item: &T) -> T {
+        item.to_px(self)
+    }
+
+    /// Converts from pixels into display points, using the x axis scale factor.
+    #[inline]
+    pub fn px_to_dp_x<T: Into<f64>>(&self, x: T) -> f64 {
```
cmyr, line 199: do we anticipate these two methods be used? If it's an obscure case, I would be happy to require the user to construct a point, convert that, and then take the `x`.

druid-shell/src/scale.rs
```diff
@@ -0,0 +1,388 @@
...
+            dpi_x: SCALE_TARGET_DPI * scale_x,
+            dpi_y: SCALE_TARGET_DPI * scale_y,
+            scale_x,
+            scale_y,
+        }
+    }
+
+    /// Returns the x axis platform DPI associated with this `Scale`.
+    #[inline]
+    pub fn dpi_x(&self) -> f64 {
```
cmyr, line 141: I would probably just have a single `dpi` method that returns either a `(f64, f64)` or `Vec2`.

druid-shell/src/scale.rs
```diff
@@ -0,0 +1,388 @@
...
+
+    /// Converts a `Point` from pixels into display points,
+    /// using the x axis scale factor for `x` and the y axis scale factor for `y`.
+    #[inline]
+    fn to_dp(&self, scale: &Scale) -> Point {
+        Point::new(self.x / scale.scale_x, self.y / scale.scale_y)
+    }
+}
+
+impl Scalable for Line {
```
cmyr, line 257: Is this used? I would have thought that scaling is happening to points, sizes, and rects, but not other shapes.

druid-shell/src/scale.rs
```diff
@@ -0,0 +1,388 @@
...
+        Rect::new(
+            self.x0 / scale.scale_x,
+            self.y0 / scale.scale_y,
+            self.x1 / scale.scale_x,
+            self.y1 / scale.scale_y,
+        )
+    }
+}
+
+impl Scalable for Insets {
```
cmyr, line 319: ditto this; do we use insets to store like window chrome size or something?

druid-shell/src/scale.rs
xStrom: The rationale has vanished. This used to be useful with an initial implementation that used a different strategy and attempted to always achieve integer logical display points. That no longer being the case combined with `Scale` now just being the scale factors, we can remove this. The new `Scale` implements `PartialEq` which will be just fine for our needs.

xStrom:
I pushed a new version which fixes the typo and removes some code.
- `dpi_approx_eq` removed because it's not really needed anymore.
- `px_to_dp_xy` / `px_to_dp_x` / `px_to_dp_y` were adaptations of the old DPI methods in `druid-shell`. They're no longer used thanks to `Scalable` and so yeah we can delete them.

I think there's value in keeping the `dpi_x` and `dpi_y` methods. They're used right now like this:
```rust
let props = D2D1_RENDER_TARGET_PROPERTIES {
    _type: D2D1_RENDER_TARGET_TYPE_DEFAULT,
    pixelFormat: D2D1_PIXEL_FORMAT {
        format: DXGI_FORMAT_B8G8R8A8_UNORM,
        alphaMode: D2D1_ALPHA_MODE_IGNORE,
    },
    dpiX: scale.dpi_x() as f32,
    dpiY: scale.dpi_y() as f32,
    usage: D2D1_RENDER_TARGET_USAGE_NONE,
    minLevel: D2D1_FEATURE_LEVEL_DEFAULT,
};
```
It's possible to achieve this even with a unified method of course, but it would be an extra line.

`Line` and `Insets` are not used right now, but could be useful if say the X11 backend implements its own menu. I don't think the cost of having them is too high and they provide some completeness.

PR #906, opened 2020-05-08
Add more entries to gitignore.

xStrom:
The `.DS_Store` files showing up in `git status` were annoying me on macOS. I added that and a few others to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,23 +1,32 @@
-## Rust specific files
+## Rust
 /target/
 
-## Intellij Idea project configuration files
+## Intellij Idea
 .idea/
 
-## VScode project configuration files
+## Visual Studio Code
 .vscode/
 
-## Sublime Text project configuration files
+## Sublime Text
 *.sublime-workspace
 *.sublime-project
 
-## macOS system files
+## Vim
+.vim
+tags
+Session.vim
+Sessionx.vim
+
+## macOS
 .DS_Store
 .AppleDouble
 .LSOverride
 
-## Windows system files
+## Windows
 Thumbs.db
 ehthumbs.db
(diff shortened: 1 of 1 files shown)


Finnerale:
Can we add `.vim/` as well while we are at it?

alerque:
@Finnerale what version of VIM puts anything in a `.vim` folder in project directories? I'm a long time vimmer and not aware that that's a normal usage thing.

I might suggest `*~`, `.*~`, and `tags` as a couple common patterns and a generated file that often gets in my way.

Finnerale:
@alerque Well, I have added an autocmd to load `.vim/init.vim` whenever I enter a directory :sweat_smile: 
That way I can setup keybindings to run build commands on the druid package and execute useful examples.

xStrom:
Okay I added some Vim/Linux entries.

alerque:
Okay. That's fine, but I think I would add that to your personal setup, either through your global `$XDG_CONFIG_HOME/git/ignore` or to your clone of this project via `.git/info/exclude` files. What you are doing is very strongly discouraged by vim in general because it has significant security implications. You or course can do what you want, but this is basically the same as a modeline vulnerability and I wouldn't suggest encouraging it with a shared gitignore.

Finnerale:
@alerque Interesting, didn't know that. Doing what I did is recommended by the fandom wiki (which scores pretty good with search engines) and quite a few plugins. But it should be fine if I make the file name more complex plus a keybinding to generate the file if it does not yet exist. And thanks for the idea with a the local `git/ignore`, did that as well.

xStrom:
Okay I removed `.vim` then.

PR #907, opened 2020-05-08
Fix macOS mouse events with multiple windows and active states.

xStrom:
This PR replaces `acceptsMouseMovedEvents` with `NSTrackingArea`. This fixes a bunch of issues with not receiving mouse events in a multiple window situation.

The behavior is perhaps best testable with the `multiwin` example with the hot glow option turned on via the context menu. For example on `master` if you overlap multiple druid windows partially and then move the mouse not-too-fast over both of them, the window that's not active might get the hot status and remain stuck with it.

For another example:
- Position the two druid windows partially on top of each other.
- Press and hold a mouse button on the active window where the windows would overlap.
- `Cmd`+``` ` ``` to the bottom window so that it becomes the top window and is below the mouse cursor.
- Move the mouse a tiny bit, but not out of the area where it's on top of the new bottom window but in an obscured manner.
- Release the mouse button and move the mouse away.

On `master` the bottom window gets stuck in a hot state, while there's a custom fix for it in this PR.

### DIFF
--- a/druid-shell/src/platform/mac/appkit.rs
+++ b/druid-shell/src/platform/mac/appkit.rs
@@ -1,0 +1,75 @@
+// Copyright 2020 The xi-editor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! macOS AppKit bindings.
+
+#![allow(non_snake_case)]
+#![allow(non_upper_case_globals)]
+
+use bitflags::bitflags;
+use cocoa::base::id;
+use cocoa::foundation::NSRect;
+use objc::{class, msg_send, sel, sel_impl};
+
+bitflags! {
+    pub struct NSTrackingAreaOptions: i32 {
+        const MouseEnteredAndExited = 1;
+        const MouseMoved = 1 << 1;
+        const CursorUpdate = 1 << 2;
+        // What's 1 << 3?
+        const ActiveWhenFirstResponder = 1 << 4;
+        const ActiveInKeyWindow = 1 << 5;
+        const ActiveInActiveApp = 1 << 6;
+        const ActiveAlways = 1 << 7;
+        const AssumeInside = 1 << 8;
+        const InVisibleRect = 1 << 9;
+        const EnabledDuringMouseDrag = 1 << 10;
+    }
+}
+
+pub trait NSTrackingArea: Sized {
+    unsafe fn alloc(_: Self) -> id {
+        msg_send![class!(NSTrackingArea), alloc]
+    }
+
+    unsafe fn initWithRect_options_owner_userInfo(
+        self,
+        rect: NSRect,
+        options: NSTrackingAreaOptions,
+        owner: id,
+        userInfo: id,
+    ) -> id;
+}
+
+impl NSTrackingArea for id {
+    unsafe fn initWithRect_options_owner_userInfo(
+        self,
+        rect: NSRect,
+        options: NSTrackingAreaOptions,
+        owner: id,
+        userInfo: id,
+    ) -> id {
+        msg_send![self, initWithRect:rect options:options owner:owner userInfo:userInfo]
+    }
+}
+
+pub trait NSView: Sized {
+    unsafe fn addTrackingArea(self, trackingArea: id) -> id;
+}
+
+impl NSView for id {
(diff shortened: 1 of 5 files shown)


cmyr:
I have some preliminary thoughts & questions:

- on mac, windows that are not active are generally not interactive; we don't want them to be drawing hover state etc. I think the way to handle this overall is with some sort of "enabled" flag, but I'm not sure how that interacts with this work?

for the second example you mention, this does seem like a bug; the issue as far as I can tell is that the second window isn't receiving the mouse up event, and it should be? I presume we're just dropping this somewhere?

xStrom:
The second example bug on `master` is because non-active windows don't receive mouse events when no mouse buttons are held down. The bug persists even with visible tracking areas because they still won't receive events when the mouse leaves via an obscured part. That's why it needed special attention.

---

As for non-active windows drawing hot state, I did a short survey of apps on macOS Catalina.

App | Hot state description | When active | When not active
--- | --- | --- | ---
Calculator | None | - | -
Calendar | None | - | -
Contacts | When hovering over lines of info they either get colored background or an icon appears. Top titlebar gets extra edit button that appears on hover. | **Yes** | **Yes**
Dictionary | Top toolbar tabs change color behind text on hover. | **Yes** | **Yes**
Finder | On the left sidebar Favourites/Tags show *Hide* button | **Yes** | **Yes**
Notes | Hovering over top toolbar icons changes their color | No !! | **Yes**
Photos | On the left sidebar Library / Albums / Projects show *Hide* button, and My Albums / My Projects show a *+* button | **Yes** | ***Hide* buttons**
Reminders | Hovering over items shows a blue icon | **Yes** | No
Safari | Address bar shows plus sign on left side when hovered, and when hovering that plus sign the text *Add page to reading list* appears | **Yes** | **Yes**
TextEdit | Cursor changes over text | **Yes** | No

- Of the 10 apps tested, 8 had some sort of hot effect.
- Of the 8 apps with hot effects, 6 of them had hot effects even when the window is not active.
- One of them had hot effects **only** when not active!

Based on my testing it seems that although hover effects aren't used nearly as often in macOS apps as say Windows apps, they are still widely used in most apps, even if sparingly. Also the majority of them had hover effects even when the window is not active.

Thus to have a macOS app that does that have hot effects when hovering would put it into the minority. I'm not sure how 3rd party apps on macOS tend to behave, but the apps I tested came with macOS Catalina and are promoted in the bottom toolbar so I assume Apple considers them well enough designed.

I think the case is rather clear that hot state for non-active windows should be there.

Now what I will say is that during my survey I also noticed that several apps change their design based on whether the window is active or not. To do that the app needs to know whether the window is active or not.

Thus I think for druid to be powerful enough to mimic the behavior of say the Notes app on macOS, it needs to have both non-active mouse events that this PR here provides, but also an additional capability added via a different PR that allows the app to know whether the window is currently active.

cmyr:
Okay, I think that's a good argument for keeping these concepts separated. Thanks for the survey; I'll do a review focused on the implementation when I have a few window.

druid-shell/src/platform/mac/appkit.rs
```diff
@@ -0,0 +1,75 @@
...
+        const ActiveInKeyWindow = 1 << 5;
+        const ActiveInActiveApp = 1 << 6;
+        const ActiveAlways = 1 << 7;
+        const AssumeInside = 1 << 8;
+        const InVisibleRect = 1 << 9;
+        const EnabledDuringMouseDrag = 1 << 10;
+    }
+}
+
+pub trait NSTrackingArea: Sized {
```
cmyr, line 41: I would consider trying to upstream any of this if we want it. I've also personally found that these abstractions around cocoa stuff really aren't super helpful, and generally I lean towards just using `msg_send!` directly everywhere. (this is fine, though)

druid-shell/src/platform/mac/appkit.rs
```diff
@@ -0,0 +1,75 @@
...
+
+impl NSTrackingArea for id {
+    unsafe fn initWithRect_options_owner_userInfo(
+        self,
+        rect: NSRect,
+        options: NSTrackingAreaOptions,
+        owner: id,
+        userInfo: id,
+    ) -> id {
+        msg_send![self, initWithRect:rect options:options owner:owner userInfo:userInfo]
```
cmyr, line 63: it's possible we want an `autorelease` call on this, would have to run the memory profiler.

druid-shell/src/platform/mac/appkit.rs
xStrom: Yeah the `appkit.rs` stuff I would like to get upstreamed. I've done very minimal macOS Rust programming to have an opinion on what's helpful or not, but the stuff I wrote here mimics the servo appkit crate so at least it's consistent.

druid-shell/src/platform/mac/appkit.rs
xStrom: The `autorelease` stuff I'm least familiar with because the Apple documentation doesn't directly translate to the Rust world. Anyway for this case I did add an `autorelease` call in `window.rs` after calling this method. In my testing adding that `autorelease` at least doesn't seem to cause any issues.

PR #908, opened 2020-05-08
Statically typed selectors

Finnerale:
This changes `Selector` to `Selector<T>` where `T` is the selectors payload.

There was a [Zulip](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/commands.20and.20'side.20effects') discussion about why `Command`s rely on purely dynamic typing while `Env` uses statically typed `Key`s.

I found myself typing 'Must have ThisType as payload` for pretty much every selector I've defined, so adding a type to it seemed like the logical consequence.

This also fixes some issues with menu items submitting wrong commands or showing wrong labels.

This still lacks in documentation and I haven't thoroughly checked if everything works, but at least for `multiwin` it does work better now.

Also I have plans to introduce `OneShotSelector<T>` for one-shot commands and I'm not sure if it should be part of this PR or go into a separate one.

I've opened this as a draft for now so that I have a better overview of the changes made and where I have to add documentation. Also I can see if there are errors on mac or windows and  if someone wants to give some feedback, that would be welcome as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ While some features like the clipboard, menus or file dialogs are not yet availa
 - Timer events will only be delivered to the widgets that requested them. ([#831] by [@sjoshid])
 - `Event::Wheel` now contains a `MouseEvent` structure. ([#895] by [@teddemunnik])
 - `AppDelegate::command` now receives a `Target` instead of a `&Target`. ([#909] by [@xStrom])
+- `SHOW_WINDOW` and `OPEN_WINDOW` no longer require a `WindowId` as payload, but they must `Target` the correct window. ([#???] by [@finnerale])
 
 ### Deprecated
 
--- a/druid/examples/identity.rs
+++ b/druid/examples/identity.rs
@@ -40,8 +40,8 @@ use druid::{
 
 const CYCLE_DURATION: Duration = Duration::from_millis(100);
 
-const FREEZE_COLOR: Selector = Selector::new("identity-example.freeze-color");
-const UNFREEZE_COLOR: Selector = Selector::new("identity-example.unfreeze-color");
+const FREEZE_COLOR: Selector<Color> = Selector::new("identity-example.freeze-color");
+const UNFREEZE_COLOR: Selector<()> = Selector::new("identity-example.unfreeze-color");
 
 /// Honestly: it's just a color in fancy clothing.
 #[derive(Debug, Clone, Data, Lens)]
@@ -114,15 +114,10 @@ impl Widget<OurData> for ColorWell {
                 self.token = ctx.request_timer(CYCLE_DURATION);
             }
 
-            Event::Command(cmd) if cmd.selector == FREEZE_COLOR => {
-                self.frozen = cmd
-                    .get_object::<Color>()
-                    .ok()
-                    .cloned()
-                    .expect("payload is always a Color")
-                    .into();
+            Event::Command(cmd) if cmd.is(FREEZE_COLOR) => {
+                self.frozen = cmd.get(FREEZE_COLOR).ok().cloned();
             }
-            Event::Command(cmd) if cmd.selector == UNFREEZE_COLOR => self.frozen = None,
+            Event::Command(cmd) if cmd.is(UNFREEZE_COLOR) => self.frozen = None,
             _ => (),
         }
     }
--- a/druid/examples/identity.rs
+++ b/druid/examples/identity.rs
@@ -115,7 +115,7 @@ impl Widget<OurData> for ColorWell {
             }
 
             Event::Command(cmd) if cmd.is(FREEZE_COLOR) => {
-                self.frozen = cmd.get(FREEZE_COLOR).ok().cloned();
+                self.frozen = cmd.get(FREEZE_COLOR).cloned();
             }
             Event::Command(cmd) if cmd.is(UNFREEZE_COLOR) => self.frozen = None,
             _ => (),
--- a/druid/examples/identity.rs
+++ b/druid/examples/identity.rs
@@ -41,7 +41,7 @@ use druid::{
 const CYCLE_DURATION: Duration = Duration::from_millis(100);
 
 const FREEZE_COLOR: Selector<Color> = Selector::new("identity-example.freeze-color");
-const UNFREEZE_COLOR: Selector<()> = Selector::new("identity-example.unfreeze-color");
+const UNFREEZE_COLOR: Selector = Selector::new("identity-example.unfreeze-color");
 
 /// Honestly: it's just a color in fancy clothing.
 #[derive(Debug, Clone, Data, Lens)]
--- a/druid/src/app.rs
+++ b/druid/src/app.rs
@@ -66,7 +66,7 @@ pub struct AppStateWindowDesc {
     type_name: &'static str,
 }
 
-impl<T: 'static> WindowDesc<T> {
+impl<T: Any> WindowDesc<T> {
     /// This turns a typed `WindowDesc<T>` into an untyped `AppStateWindowDesc`.
     /// Doing so allows sending `WindowDesc` through `Command`s.
     /// It is up to you, to ensure that this `T` represents your application
@@ -80,7 +80,7 @@ impl<T: 'static> WindowDesc<T> {
 }
 
 impl AppStateWindowDesc {
-    pub(crate) fn realize<T: 'static>(self) -> Result<WindowDesc<T>, AppStateTypeError> {
+    pub(crate) fn realize<T: Any>(self) -> Result<WindowDesc<T>, AppStateTypeError> {
         let inner: Result<Box<WindowDesc<T>>, _> = self.inner.downcast();
         if let Ok(inner) = inner {
             Ok(*inner)
--- a/druid/src/app.rs
+++ b/druid/src/app.rs
@@ -21,8 +21,10 @@ use crate::widget::LabelText;
 use crate::win_handler::{AppHandler, AppState};
 use crate::window::WindowId;
 use crate::{
-    theme, AppDelegate, Data, DruidHandler, Env, LocalizedString, MenuDesc, Widget, WidgetExt,
+    command::AppStateTypeError, theme, AppDelegate, Data, DruidHandler, Env, LocalizedString,
+    MenuDesc, Widget, WidgetExt,
 };
+use std::any::{self, Any};
 
 /// A function that modifies the initial environment.
 type EnvSetupFn<T> = dyn FnOnce(&mut Env, &T);
@@ -52,6 +54,43 @@ pub struct WindowDesc<T> {
     /// This can be used to track a window from when it is launched and when
     /// it actually connects.
     pub id: WindowId,
+}
+
+/// A description of a window to be instantiated. The user has to guarantee that this
+/// represents a `WindowDesc<T>` where `T` is the users `AppState`.
+///
+/// This includes a function that can build the root widget, as well as other
+/// window properties such as the title.
+pub struct AppStateWindowDesc {
+    inner: Box<dyn Any>,
+    type_name: &'static str,
+}
+
+impl<T: 'static> WindowDesc<T> {
+    /// This turns a typed `WindowDesc<T>` into an untyped `AppStateWindowDesc`.
+    /// Doing so allows sending `WindowDesc` through `Command`s.
+    /// It is up to you, to ensure that this `T` represents your application
+    /// state that you passed to `AppLauncher::launch`.
+    pub fn into_app_state_menu_desc(self) -> AppStateWindowDesc {
+        AppStateWindowDesc {
+            inner: Box::new(self),
+            type_name: any::type_name::<Self>(),
+        }
+    }
+}
+
+impl AppStateWindowDesc {
+    pub(crate) fn realize<T: 'static>(self) -> Result<WindowDesc<T>, AppStateTypeError> {
+        let inner: Result<Box<WindowDesc<T>>, _> = self.inner.downcast();
+        if let Some(inner) = inner.ok() {
+            Ok(*inner)
+        } else {
+            Err(AppStateTypeError::new(
+                any::type_name::<WindowDesc<T>>(),
+                self.type_name,
+            ))
+        }
+    }
 }
 
 impl<T: Data> AppLauncher<T> {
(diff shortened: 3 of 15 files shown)


druid/src/command.rs
```diff
@@ -87,15 +122,11 @@ enum Arg {
     OneShot(Arc<Mutex<Option<Box<dyn Any>>>>),
 }
 
-/// Errors that can occur when attempting to retrieve the a command's argument.
+/// Errors that can occur when attempting to retrieve the a `OneShotCommand`s argument.
```
xStrom, line 125: Remove the "a" before OneShotCommand.

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ While some features like the clipboard, menus or file dialogs are not yet availa
 - Timer events will only be delivered to the widgets that requested them. ([#831] by [@sjoshid])
 - `Event::Wheel` now contains a `MouseEvent` structure. ([#895] by [@teddemunnik])
 - `AppDelegate::command` now receives a `Target` instead of a `&Target`. ([#909] by [@xStrom])
+- `SHOW_WINDOW` and `OPEN_WINDOW` no longer require a `WindowId` as payload, but they must `Target` the correct window. ([#???] by [@finnerale])
```
cmyr, line 50: maybe this can just be a separate patch?

druid/examples/identity.rs
```diff
@@ -114,15 +114,10 @@ impl Widget<OurData> for ColorWell {
                 self.token = ctx.request_timer(CYCLE_DURATION);
             }
 
-            Event::Command(cmd) if cmd.selector == FREEZE_COLOR => {
-                self.frozen = cmd
-                    .get_object::<Color>()
-                    .ok()
-                    .cloned()
-                    .expect("payload is always a Color")
-                    .into();
```
cmyr, line 123: this sort of diff does make this feel like a pretty compelling change!

druid/src/app.rs
```diff
@@ -54,6 +56,43 @@ pub struct WindowDesc<T> {
     pub id: WindowId,
 }
 
+/// A description of a window to be instantiated. The user has to guarantee that this
+/// represents a `WindowDesc<T>` where `T` is the users `AppState`.
+///
+/// This includes a function that can build the root widget, as well as other
+/// window properties such as the title.
+pub struct AppStateWindowDesc {
```
cmyr, line 64: Naming: I would call this `AnyWindowDesc`.

Philosophy:  I get that this is a problem that needs to be solved for this PR to work, but I also think it's worth thinking about on its own.

What we need:

- some kind of type erasure or 'implicit `T`' for certain types of objects, like window descriptions and menus
- not to confuse the user
- to clutter the namespace as little as possible

What's the best way to do this? I'm not totally sure.

If we could have some internal type that did what this type is doing, that would be great. One way this might work is if we had convenience methods for things like making a new window or setting a menu; for instance creating a new window might just use a method on `EventCtx`, and then the user never needs to create the command or actually interact with the type erased type?

It might also be that we require a public type (that appears in the signature of the selector, for instance) but we don't actually need the user to ever interact with it?

druid/src/command.rs
```diff
@@ -15,19 +15,69 @@
...
+/// by druid, and have special meaning to the framework; these are listed in the
+/// [`druid::commands`] module.
+///
+/// [`druid::commands`]: commands/index.html
+#[derive(Debug, PartialEq, Eq)]
+pub struct Selector<T = ()>(SelectorSymbol, PhantomData<*const T>);
+
+// This has do be done explicitly, to avoid the Copy bound on `T`.
+// See https://doc.rust-lang.org/std/marker/trait.Copy.html#how-can-i-implement-copy .
+impl<T> Copy for Selector<T> {}
```
cmyr, line 40: just in general I tend to throw these sorts of impl blocks down at the bottom of the file, for a better reading experience; this is a matter of personal preference though.

druid/src/command.rs
```diff
@@ -15,19 +15,69 @@
...
+    }
+}
+
 /// An identifier for a particular command.
 ///
 /// This should be a unique string identifier. Certain `Selector`s are defined
 /// by druid, and have special meaning to the framework; these are listed in the
 /// [`druid::commands`] module.
 ///
 /// [`druid::commands`]: commands/index.html
```
cmyr, line 53: we should indicate what makes this different from just selector, above.

druid/src/command.rs
```diff
@@ -15,19 +15,69 @@
...
+// This has do be done explicitly, to avoid the Copy bound on `T`.
+// See https://doc.rust-lang.org/std/marker/trait.Copy.html#how-can-i-implement-copy .
+impl<T> Copy for OneShotSelector<T> {}
+impl<T> Clone for OneShotSelector<T> {
+    fn clone(&self) -> Self {
+        *self
+    }
+}
+
+pub trait AnySelector {
```
cmyr, line 66: I would try and get rid of this in favour of using some std trait, like maybe `AsRef`?

druid/src/command.rs
```diff
@@ -73,19 +122,42 @@ enum Arg {
...
-/// Errors that can occur when attempting to retrieve the a command's argument.
+/// Errors that can occur when attempting to retrieve the `OneShotCommand`s argument.
 #[derive(Debug, Clone, PartialEq)]
 pub enum ArgumentError {
-    /// The command did not have an argument.
-    NoArgument,
-    /// The argument was expected to be reusable and wasn't, or vice-versa.
-    WrongVariant,
-    /// The argument could not be downcast to the specified type.
-    IncorrectType,
```
cmyr, line 84: this is still possible in the case of a name collision, isn't it?

druid/src/command.rs
```diff
@@ -73,19 +122,42 @@ enum Arg {
...
+    WrongSelector,
     /// The one-shot argument has already been taken.
     Consumed,
 }
 
+/// This error can occur when wrongly promising that a type erased
+/// variant of some generic item represents the application state.
+/// Examples are `MenuDesc<T>` and `AppStateMenuDesc`.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct AppStateTypeError {
```
cmyr, line 138: do we have an 'errors' module? I might move that there, I'm not sure.

I'd also maybe be more general, and call it a `DowncastError` or something?

druid/src/command.rs
```diff
@@ -15,19 +15,69 @@
...
 ///
 /// This should be a unique string identifier. Certain `Selector`s are defined
 /// by druid, and have special meaning to the framework; these are listed in the
 /// [`druid::commands`] module.
 ///
 /// [`druid::commands`]: commands/index.html
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub struct Selector(&'static str);
+#[derive(Debug, PartialEq, Eq)]
+pub struct OneShotSelector<T>(SelectorSymbol, PhantomData<*const T>);
```
cmyr, line 55: I'm not totally sure about this; what it comes down to is whether or not we expect external people to need to use this, or whether it's an implementation detail used in a few places inside druid. Again, I think that API simplicity is important, and I think the more concepts we introduce the more work we need to do to explain what they're for and how to use them. My _feeling_ is that 95+% of the time, the user doesn't need to know about one-shot commands?

A kind of gross way we could encode one-shotness would be in the selector symbol itself, like we could have some keyword that prefaced one-shot commands...

druid/src/menu.rs
```diff
@@ -120,6 +125,35 @@ pub struct MenuDesc<T> {
     items: Vec<MenuEntry<T>>,
 }
 
+/// A platform-agnostic description of an application, window, or context
+/// menu. The user has to guarantee that this represents a `MenuDesc<T>` where `T` is the users
+/// `AppState`.
+pub struct AppStateMenuDesc {
```
cmyr, line 131: ditto, just `AnyMenuDesc` if we're going this route

druid/src/menu.rs
```diff
@@ -120,6 +125,35 @@ pub struct MenuDesc<T> {
...
+    pub fn into_app_state_menu_desc(self) -> AppStateMenuDesc {
+        AppStateMenuDesc {
+            inner: Box::new(self),
+            type_name: any::type_name::<Self>(),
+        }
+    }
+}
+
+impl AppStateMenuDesc {
+    pub(crate) fn realize<T: Any>(&self) -> Result<&MenuDesc<T>, AppStateTypeError> {
```
cmyr, line 150: don't mind this name, but I might just call it `downcast` or `to_concrete`? (ditto the other one)

druid/src/app.rs
Finnerale: I though about some kind of wrapper type but I don't think it is possible with the Rust type system.
What I wanted to make more obvious: `MenuDesc` and others **must** represent the application state.
Adding methods for this to `EventCtx` seems like a possible solution.
Doing so could even allow equipping `EventCtx` with the `TypeId`s of eg. `MenuDesc<AppState>` to check it right away and panic if the type is wrong. That way the back trace will show exactly where the mismatch happened.
But if we want to keep it as a `Command` I doubt we can get around some kind of specific wrapper type.

druid/src/command.rs
Finnerale: I do want to reorder a couple things anyways, currently they are just where I could see the context right away.

druid/src/command.rs
Finnerale: Right, I forgot about this one.

druid/src/app.rs
cmyr: the nice  thing is that on `EventCtx` the method could take a concrete `MenuDesc`.

druid/src/command.rs
Finnerale: Well the ideas I had where:

 Leave it as it is.

- IMO the worst choice as it makes it less obvious how a command is supposed to be used, and would require you to look up every commands doc in order to find out how the hell you are supposed to submit it and how to retrieve the object. Also `Command::get` and `Command::take` will either have to panic or require an additional error, which would add to the interface at least as much noise as a new type.

 `Selector(key: &str, is_once: bool)` and a `Selector::once` constructor.

- Not much better than the previous one.

Encoding it in the selector symbol.

- Again, does not add anything really.

Having a separate type for it.

- Allows `Command::get` to return `Option` instead of `Result`.
- Submitting and retrieving becomes a no-brainer as the compiler checks it for you.
- Could allow merging `Command::new` and `Command::once` if we want to.
- No need to look up a selectors docs to know how to use it (unless it does something very special)
- Those who do not care about one shot commands do not have to use this type

Which is why I very much like this design.

druid/src/app.rs
Finnerale: How so? I mean `EventCtx` does not know what the `AppState` is, so how is it supposed to be typed for it?
If this could be done statically, than that would be absolutely my favorite.

druid/src/command.rs
Finnerale: Yes, if `AnySelector` is only used for the symbol, than that could make more sens.
I used a custom trait with the intend to merge `Command::new` and `Command::once`.
If we don't do that, `AsRef` could be fine, even though I like that `AnySelector` makes it clear that it expects a selector.

druid/src/command.rs
Finnerale: Yes, and it will panic just like `Env` does.

druid/src/command.rs
Finnerale: I don't think we have one, but it could be interesting to see if it would be beneficial.
As for the name, `DowncastError` is less specific but would be more in line with the other names.

druid/src/menu.rs
Finnerale: `downcast` seams reasonable.

druid/src/command.rs
```diff
@@ -73,19 +122,42 @@ enum Arg {
...
-    WrongVariant,
-    /// The argument could not be downcast to the specified type.
-    IncorrectType,
+    /// The command represented a different selector.
+    WrongSelector,
     /// The one-shot argument has already been taken.
     Consumed,
 }
 
+/// This error can occure when wrongly promising that a type ereased
```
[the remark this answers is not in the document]
Finnerale: Because the user still has to 'promise' that any menus or windows they create match the application state. This error occurs when the promise is broken.

CHANGELOG.md
Finnerale: I suppose this could be split into
- Updating `SHOW_WINDOW` and `OPEN_WINDOW` to use `Target`
- Fix open and save menu items
- Make `Selector` type safe
- Split `Selector` into `Selector` and `SelectorOnce`
- Add type-safe? methods to `EventCtx` to create menus and windows.

Splitting the last two out would cause quite some duplicated / wasted work tho, as `SelectorOnce` influences error and return types, while menus creation decides over adding wrapper types or new methods.

Finnerale: Even the first one means that I'll have to first declare it to take `WidgetId` and than remove that in a separate PR again.

druid/src/command.rs
cmyr: I wonder if another option is to encode it in the type param? Like have some type `SingleUse<T>`, and then you're required to explicitly `take` from that... I don't know if I like this, just spitballing.

druid/src/app.rs
cmyr: mm you're right I was thinking that `EventCtx` was paramaterized over `T`, but even in that case it doesn't tell us what the root data is.

CHANGELOG.md
cmyr: yea, if things are truly entwined than doing it together makes sense, but if there are good dividing points that does make review a lot easier.

druid/src/command.rs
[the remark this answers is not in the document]
cmyr: oh this comment was out of date, sorry

druid/src/command.rs
xStrom: The `AsRef` implementation isn't too confusing I think if we add some documentation that it takes a `Selector` / `SelectorOnce`.

```rust
impl<T> AsRef<SelectorSymbol> for Selector<T> {
    fn as_ref(&self) -> &SelectorSymbol {
        &self.0
    }
}

impl Command {
    pub fn is(&self, selector: impl AsRef<SelectorSymbol>) -> bool {
        self.selector == *selector.as_ref()
    }
}
```

Finnerale:
Splitting this up was a great idea, and resulted in a far better design.

Thanks a lot @cmyr and @xStrom for your help and reviews!

This was completed with #993 an can thus be closed as successful.

PR #909, opened 2020-05-09
Optimize clippy trivially_copy_pass_by_ref for 64-bit.

xStrom:
This PR changes `trivially_copy_pass_by_ref` to 16 bytes, which is the number of bytes that fit into two 64-bit CPU registers.

The default clippy value for this is 8 bytes, which is chosen to improve performance on 32-bit. Given that druid is being designed for the future and already even mobile phones have 64-bit CPUs, it makes sense to optimize for 64-bit and accept the performance hits on 32-bit. 

This also changes command `Target` to be passed by value. It's a breaking change, but given that commands are getting redone in #908 anyway, I think this is about as perfect of a time to change this as possible.

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,0 +1,5 @@
+# The default clippy value for this is 8 bytes, which is chosen to improve performance on 32-bit.
+# Given that druid is being designed for the future and already even mobile phones have 64-bit CPUs,
+# it makes sense to optimize for 64-bit and accept the performance hits on 32-bit.
+# 16 bytes is the number of bytes that fits into two 64-bit CPU registers.
+trivial-copy-size-limit = 16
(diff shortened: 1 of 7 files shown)


clippy.toml
```diff
@@ -0,0 +1,5 @@
+# The default clippy value for this is 8 bytes, which is chosen to improve performance on 32-bit.
+# Given that druid is being designed for the future and already even mobile phones have 64-bit CPUs,
```
Finnerale, line 2: > and already even mobile phones have 64-bit CPUs

Is it correct to place `already` there?

> and even mobile phones already have 64-bit CPUs

feels better to me.

xStrom: Both are fine and you could even put it at the end. Check out this [Cambridge dictionary page](https://dictionary.cambridge.org/grammar/british-grammar/already) about it.

PR #911, opened 2020-05-09
Implement Data for Vec<Data>

chris-zen:
This PR implements the `Data` trait for a `Vec<T> where T: Data`, which solves the following case without having to use `#[data(same_fn = "PartialEq::eq")]`:

```rust
#[derive(Data)]
struct A {
  // ...
}

#[derive(Data)]
struct B {
  v: Vec<A>
  // ...
}

```

### DIFF
--- a/druid/src/data.rs
+++ b/druid/src/data.rs
@@ -178,6 +178,14 @@ impl<T: Data, U: Data> Data for Result<T, U> {
     }
 }
 
+impl<D: Data> Data for Vec<D> {
+    fn same(&self, other: &Self) -> bool {
+        self.iter()
+            .zip(other.iter())
+            .all(|(a, b)| a.same(b))
+    }
+}
+
 impl Data for () {
     fn same(&self, _other: &Self) -> bool {
         true
@@ -386,6 +394,13 @@ impl_data_for_array! { 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 }
 #[cfg(test)]
 mod test {
     use super::Data;
+
+    #[test]
+    fn vec_data() {
+        let input = vec![1u8, 0, 0, 1, 0];
+        assert!(input.same(&vec![1u8, 0, 0, 1, 0]));
+        assert!(!input.same(&vec![1u8, 1, 0, 1, 0]));
+    }
 
     #[test]
     fn array_data() {
--- a/druid/src/data.rs
+++ b/druid/src/data.rs
@@ -180,9 +180,7 @@ impl<T: Data, U: Data> Data for Result<T, U> {
 
 impl<D: Data> Data for Vec<D> {
     fn same(&self, other: &Self) -> bool {
-        self.iter()
-            .zip(other.iter())
-            .all(|(a, b)| a.same(b))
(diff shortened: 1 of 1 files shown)


cmyr:
so long-term we don't want data for vec; we want data-aware collection types that are cheap to clone; and currently, if you're using a vec for data, it should probably be in an `Arc`.

xStrom:
Should this be reverted? :thinking:

Finnerale:
Well, I think it is nice for ergonomics, as `Arc<Vec<T>>` is quite annoying, especially when ending up with nesting `Arc<Vec<Arc<Vec<T>>>>`.

But some good collection types, maybe similar to the `Im` ones would be a much better solution, as using `Vec<T>` could lead to surprisingly* bad performance.

Thus I'm not to happy about this change as it pretty much will have to be reverted at some point.

*Mostly for people new to druid

xStrom:
Somewhat tangential, but I'm also wondering under what scenario the data goes on multiple threads that it needs an `Arc` instead of `Rc`. :thinking: I haven't looked at the code too deeply with this question in mind, but most of the platform code which generates events seems pretty single-threaded.

Finnerale:
Yes, I think there is a note somewhere about the possibility about using `Rc` but that there is no guarantee for it to work .

xStrom:
After giving it some further thought, even if the underlying platform is single-threaded, druid could potentially do its own event handling in a fan-out way. Global events could be ran on a separate thread per window. Even a single window could split up the widget tree, running the event on half of the tree on one thread and the other half on another.

It's not a simple change though and it's an open question how big a widget tree would have to be for multi-threaded work to be faster. This deserves its seperate topic, probably on zulip to start with.

cmyr:
Yea, the choice of `Arc` is that it should be no slower than `Rc` in the uncontested case, and it sets us up for a future where we're doing multi-threaded stuff. 🤷. (I'm imagining more passing things back and forth explicitly to offload work).

My gut feeling is that this should be reverted, but maybe also that we should speed up work on adding some immutable collection types; we've also had performance issues with people using `String` in their `Data`.

medium term I'd like to discourage use of `String` (and bring over the rope from xi) and I'd also like good general purpose immutable set and list collections; ideally this would be based on some ecosystem crate like [im.rs](https://docs.rs/im/14.3.0/im/index.html).

cmyr:
Looking at source, it seems like the `im` crate has `ptr_eq` methods on its collections, so it should be trivial to implement `Data` for these types, and encourage their use in druid? We could even reexport them if we wanted, since we'll need to import them anyway?

derekdreery:
I think some notes in the docs would be good, as people will instinctively use `Vec` for their `Data` impls. Maybe it could go in the *Data* section of the book?

chris-zen:
Thanks for the discussion, I understand now why I should avoid direct use of Vec. I'll update my code to use a clone-friendly data structure. Feel free to revert.
Keep the great work 💪

cmyr:
@chris-zen Thanks for actually lighting the fire to get us to bring in some druid-appropriate collections. ;)

Issue #914, opened 2020-05-10
Window white flash on init 


Adarma:
Is it possible to prevent the white flash that happens before the GUI is initially drawn?
Maybe just being able to choose black rather then a white flash is enough.
Seeing it in the examples on Windows.

raphlinus:
I'm not personally seeing it, and at one time I spent considerably effort to avoid it. Are you on Windows 10? Is this a regression, ie do you still see it on older revs such as 9d7f3ab281f44188f075e0ed9baa586b3b025482?

Adarma:
Sorry, should have said I'm seeing it on Win7x64.
It's 3am here in Germany, so I will check that commit tomorrow for regression

xStrom:
I can confirm that I also have the white flash, both during startup and during resize at the edges, with my Windows 7. It's not there on my Windows 10 machine.

xStrom:
I've been seeing this behavior for quite some time, so it's not a recent regression at least.

It seems to be performance related, because I see the initial white flash when cold starting an app, but not on restarts. I can force it by sleeping in `paint` of the root widget, including on Windows 10.

xStrom:
I managed to fix the white flash during resize in #915. It doesn't have an impact on the sleep based reproduction, however maybe the initial flash is caused by an issue that can't be represented by sleep. It will take time until I'll be able to know from my own experience, because I can't really reproduce cold starts easily.

@Adarma how often do you see the white flash? Always? Does it still happen with the change in #915?

xStrom:
Okay I managed to get a cold start again and #915 does not fix the white flash. I'm still leaning towards thinking that this is performance related. Giving this some further thought, perhaps going to a full event cycle at the beginning is a mistake. All the caches are going to be cold and we need to go through full layout etc. Perhaps a solution is to have some sort of `initial_paint` event coming from `druid-shell`. It has the window size and `druid` can just do a simple fill with the background color. The idea being to get that first frame out with the correct background color, and then the second frame can be delayed due to it having to do peak work with cold caches.

Adarma:
I tested 9d7f3ab and it also happens with that.
I see it with every window create: every example, and every time I click "New window" in the `multiwin` example.
Yes, it still happens in #915 for me too.

xStrom:
[...]

PR #916, opened 2020-05-11
Move HWND render target creation to WM_CREATE.

xStrom:
This PR should help with the initial white flash reported in #914. On my Windows 7 desktop I measured the HWND render target creation at 39ms. This is definitely too long to lazily create in `WM_PAINT` so I moved it to `WM_CREATE`.

Even if it doesn't fully fix the white flash, it cuts down the first frame time by 39ms on my machine, so it's worth a merge regardless.

I left the check in `WM_PAINT` intact, because there are failure paths with dcomp that may need it.


cmyr:
I don't have access to a windows machine right now, so I can't test this. :(

PR #917, opened 2020-05-11
Add a size method to UpdateCtx and LifeCycleCtx.

jneem:
I found myself needing this, for a widget that wants to update its internal state based on changes to data, but needs the size to do so (think of something like a scroll bar, whose state depends on the viewport size).

More generally, is there a rationale somewhere for which contexts have which functions? For example, I see that `LifeCycleCtx` is also missing `size`...


xStrom:
Yeah I think this is something that's fine to have on `LifeCycleCtx` as well. Feel free to add it as well.

jneem:
Ok, I just added it to LifeCycleCtx as well.

PR #920, opened 2020-05-11
Handle keyboard and button state in X11 events.

### LINKED ISSUE #893: X11: Add support for held down mouse buttons
The `druid-shell` `MouseEvent` has a `buttons: MouseButtons` field which is currently not being correctly set in X11. This field shows which buttons are being held down when the event happens.

jneem:
Fixes #893.

### DIFF
--- a/druid-shell/src/platform/x11/window.rs
+++ b/druid-shell/src/platform/x11/window.rs
@@ -522,22 +522,65 @@ impl Window {
         self.render(rect)
     }
 
+    fn mouse_button(&self, button: u8) -> MouseButton {
+        match button {
+            1 => MouseButton::Left,
+            2 => MouseButton::Middle,
+            3 => MouseButton::Right,
+            // buttons 4 through 7 are for scrolling.
+            8 => MouseButton::X1,
+            9 => MouseButton::X2,
+            _ => {
+                log::warn!("unknown mouse button code {}", button);
+                MouseButton::None
+            }
+        }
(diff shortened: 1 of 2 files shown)


druid-shell/src/platform/x11/window.rs
```diff
@@ -522,22 +522,65 @@ impl Window {
...
+    }
+
+    fn mouse_buttons(&self, mods: u16) -> MouseButtons {
+        let mut buttons = MouseButtons::new();
+        let button_masks = &[
+            (xcb::xproto::BUTTON_MASK_1, MouseButton::Left),
+            (xcb::xproto::BUTTON_MASK_2, MouseButton::Middle),
+            (xcb::xproto::BUTTON_MASK_3, MouseButton::Right),
+            (xcb::xproto::BUTTON_MASK_4, MouseButton::X1),
+            (xcb::xproto::BUTTON_MASK_5, MouseButton::X2),
```
xStrom, line 547: `BUTTON_MASK_4` and `BUTTON_MASK_5` don't actually work. Let's at least copy the code comment from GTK `get_mouse_buttons_from_modifiers` (remove the word "GDK" from the comment though).

jneem:
Thanks for the review! I've moved the functions in question to the module scope.

xStrom:
Yeah that looks better. Need a rebase with `master` and then this will be ready for merge.

PR #923, opened 2020-05-12
Fix paint_with_z_index with pre-existing transforms.

### LINKED ISSUE #922: GTK: paint_with_z_index is broken
`paint_with_z_index` internally calls into piet `render_ctx.current_transform()` which returns an unexpected value when running on Wayland. This can be seen in the `anim` example.

Piet internally seems to just call `cairo_get_matrix` so is this an issue with cairo on Wayland or what?

The returned matrix seems to have a static +26x / +60y translation added to it on Wayland. Indeed the GTK `connect_draw` handler receives a cairo context which has the following transform applied to it already: `Affine([1.0, 0.0, 0.0, 1.0, 26.0, 60.0])`.

Faulty behavior originally discovered by @Finnerale in #876.

xStrom:
The transform state isn't necessarily empty when it arrives to druid. Right now it's GTK that applies a translate when there are menus or even just always in the case of Wayland, presumably for some sort of border space.

However soon enough we'll get the core-graphics backend for macOS which also has its own transforms applied. Hopefully this prepares druid for that too.

Fixes #922

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,7 @@ While some features like the clipboard, menus or file dialogs are not yet availa
 - X11: Support `Application::quit`. ([#900] by [@xStrom])
 - GTK: Support file filters in open/save dialogs. ([#903] by [@jneem])
 - X11: Support key and mouse button state. ([#920] by [@jneem])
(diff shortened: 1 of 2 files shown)


xStrom:
Okay, so better to solve this in piet by doing the same for cairo?

cmyr:
yea, I think that makes sense.

PR #925, opened 2020-05-13
Fix for RouteFocusChanged and focusable descendants

yrns:
My project uses a hierarchy of focusable widgets and I ran into some problems which I've attempted to fix here.

* Route RouteFocusChanged all the way down since a focusable widget
may contain descendants which can also have focus

* Instead of replacing RouteFocusChanged with FocusChanged, send it
after the former is completely routed, so request_focus is cleared

* Added a new test case to demonstrate cases that would previously
fail, along with some test harness fixes/additions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,7 @@ While some features like the clipboard, menus or file dialogs are not yet availa
 - X11: Support `Application::quit`. ([#900] by [@xStrom])
 - GTK: Support file filters in open/save dialogs. ([#903] by [@jneem])
 - X11: Support key and mouse button state. ([#920] by [@jneem])
+- Routing `Event::FocusChanged` for descendant widgets. ([#925] by [@yrns])
 
 ### Visual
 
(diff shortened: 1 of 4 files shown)


xStrom:
I'll take a look at this more closely later, but for now I just have one quick comment.

> * Route RouteFocusChanged all the way down since a focusable widget
>   may contain descendants which can also have focus

This does not sound correct to me. Only a single widget has focus.

CHANGELOG.md
```diff
@@ -79,6 +79,7 @@ While some features like the clipboard, menus or file dialogs are not yet availa
 - X11: Support `Application::quit`. ([#900] by [@xStrom])
 - GTK: Support file filters in open/save dialogs. ([#903] by [@jneem])
 - X11: Support key and mouse button state. ([#920] by [@jneem])
+- Routing `Event::FocusChanged` for descendant widgets. ([#925] by [@yrns])
```
xStrom, line 82: ~Event::FocusChanged for~ LifeCycle::FocusChanged to

Also define your username link at the bottom where all the other username links are.

druid/src/core.rs
```diff
@@ -790,15 +788,17 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
...
         if recurse {
-            let mut child_ctx = LifeCycleCtx {
-                command_queue: ctx.command_queue,
-                base_state: &mut self.state,
-                window_id: ctx.window_id,
-            };
+            self.inner.lifecycle(&mut child_ctx, event, data, env);
+        }
+
+        if let Some(event) = substitute_event.as_ref() {
```
xStrom, line 801: Given that this is no longer a substitute event, it should be renamed to something like `extra_event`.

yrns:
I think I addressed everything, and fixed the conflict.

yrns:
Thank you for reviewing!

PR #929, opened 2020-05-15
Clip the invalidation example.

### LINKED ISSUE #872: The invalidation example behaves differently with wasm
When testing the wasm examples after #817 I noticed that it behaves differently to the native version.
While the left side seems to work perfectly fine, on the right only one rect is visible at a time.
This seems to indicate that the entire right half gets invalidated?
Maybe @jneem has an Idea.

jneem:
Fixes #872.

This makes the invalidation example act the same on all platforms, and it also explains what's going on in a comment.

### DIFF
--- a/druid/examples/invalidation.rs
+++ b/druid/examples/invalidation.rs
@@ -79,7 +79,12 @@ impl Widget<Point> for CircleView {
 
     fn paint(&mut self, ctx: &mut PaintCtx, data: &Point, _env: &Env) {
         ctx.with_save(|ctx| {
-            let rect = ctx.size().to_rect();
+            let rect = ctx.region().to_rect();
+            // Clip to the rect that was invalidated. This is usually not
+            // necessary, but if there's lots of drawing involved then there's a
+            // chance that clipping can be faster. Also, drawing outside the
+            // invalid region has backend-dependent behavior: for example, it
+            // has no effect on GTK but it does have an effect on wasm.
             ctx.clip(rect);
             ctx.fill(rect, &Color::WHITE);
             ctx.fill(Circle::new(*data, RADIUS), &Color::BLACK);


Finnerale:
Totally forgot that issue :stuck_out_tongue_closed_eyes: 
I'm not sure this is the solution we want however.

One of druids goals is to abstract platform wonkiness away where possible and having clipping on some platforms but not on others actually adds custom druid wokiness.

IMO anything that is not explicitly platform specific should behave exactly the same on all platforms we support. Thus we should either have clipping on all platforms (seems preferable) or on none.

jneem:
It should be easy enough to enable clipping on all platforms. I think the other one isn't possible, though. For example, the windows docs explicitly say that on the paint callback you aren't allowed to paint outside the invalidated rect, so I think we have to do clipping there.

Finnerale:
Gotta Clip 'em All seems like the best choice then :+1:

jneem:
Cool, I've opened #941

cmyr:
yea this is... tricky, and feels like policy. We haven't really codified what we should do when a widget draws outside of its bounds; my hunch is that we should clip by default, but allow for opting in to overflow?

xStrom:
I don't think there's a way to opt-in to overflow with GTK for what it's worth.

Finnerale:
Due to what @jneem said I think clipping always is the best choice:
>For example, the windows docs explicitly say that on the paint callback you aren't allowed to paint outside the invalidated rect, so I think we have to do clipping there.

cmyr:
this is explicitly about the partial invalidation case? clipping makes more sense there, I agree.

xStrom:
Yeah it's about partial invalidation. Widgets can still overpaint as much as they want inside the window's invalidation rect union.

Finnerale:
So if one wants a widget to draw outside of its bounds, it can just invalidated that area as well.

jneem:
Ignoring partial invalidation and just talking about widgets painting outside their bounds: isn't this what `paint_insets` is for? Since a widget can use that to gain complete control over their own bounds, I feel like it isn't restrictive to enforce paint bounds via clipping. On the other hand, I'm not sure if there's a performance cost for saving and restoring the clip state so frequently...

xStrom:
I think there's a performance impact for doing so much clipping, but I haven't measured it.

PR #930, opened 2020-05-15
Correct built-in open and save menu items.

Finnerale:
Second part of #908 .

Based on the [Zulip](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/commands.20and.20'side.20effects'/near/196882766) discussion here.

On Windows / Linux:
- `save_ellipsis` now actually displays an ellipsis.

On mac:
- `open_file` shows the dialog instead of submitting an invalid `OPEN_FILE` command.

All:
- `save_ellipsis` shows the dialog and thus behaves like `save_as` instead of `save`.

### DIFF
--- a/druid/src/menu.rs
+++ b/druid/src/menu.rs
@@ -519,8 +519,8 @@ pub mod sys {
                     .append(new())
                     .append(open())
                     .append(close())
-                    .append(save().disabled())
-                    .append(save_as().disabled())
+                    .append(save_ellipsis())
+                    .append(save_as())
                     // revert to saved?
                     .append(print().disabled())
                     .append(page_setup().disabled())
@@ -566,8 +566,8 @@ pub mod sys {
             /// The 'Save' menu item.
             pub fn save_ellipsis<T: Data>() -> MenuItem<T> {
                 MenuItem::new(
-                    LocalizedString::new("common-menu-file-save"),
-                    commands::SAVE_FILE,
+                    LocalizedString::new("common-menu-file-save-ellipsis"),
+                    commands::SHOW_SAVE_PANEL,
                 )
                 .hotkey(RawMods::Ctrl, "s")
             }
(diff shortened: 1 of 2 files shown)


druid/src/menu.rs
```diff
@@ -566,8 +566,8 @@ pub mod sys {
             /// The 'Save' menu item.
```
xStrom, line 566: Update the comment here for Windows. Just take the macOS comment I think.

Finnerale: Thanks! I knew I would forget something when splitting things up :sweat_smile:

cmyr:
Just for the historical record: a lot of this menu stuff is specifically motivated by my wanting, on the mac, to mimic the look and behaviour of Cocoa apps.

In particular this means that if you are an app that manages documents, and you have a new unsaved document, the "save" menu item is titled "Save..." and will bring up the save panel; that is it behaves like "save as". If you've already saved the document, it is titled "Save", does *not* bring up the dialog, and just saves it at its existing location.

In Cocoa menus, any menu item that will ask for additional options has a trailing ellipsis; menu items that work without further input do not.

So I'm not totally sure if this makes sense on windows or elsewhere, or not?

Finnerale:
@cmyr We had a rather long Zulip discussion (linked in the PR comment) where we discovered that basically *all* platforms recommend doing this, it's just that they do not adhere to their own guidelines.

xStrom:
The Windows and Gnome/KDE docs recommend that exact same approach. In practice nobody follows it.

That said, I think it's fine (and simplifies things) if we just recommend the "Save..." -> "Save" situation on all platforms.

cmyr:
One thing I'd like to do in the future is to rework menus to work more like widgets; that is they can be computed based on the data, and updated automatically. In this world if we can just have some way to check that a document has a path, we can do this automatically.

PR #931, opened 2020-05-15
Replace generic commands with methods on EventCtx and DelegateCtx

Finnerale:
Third part of #908 .

Replaces `NEW_WINDOW`, `SET_MENU` and `SHOW_CONTEXT_MENU` commands with methods on `EventCtx` and `DelegateCtx`.

This will get rid of any generic commands which caused quite some trouble for typed selectors.

When calling any of the methods with the wrong data type, it will  `panic!` in debug mode and `log::error!` in a release build.

I've not added `show_context_menu` to `DelegateCtx` as I could not imagine a real world usecase.
Thus I had to change `multiwin` a bit to use a `Controller` for the context menu, which seemed to be much closer to actual usage of a context menu.

### DIFF
--- a/druid/src/app_delegate.rs
+++ b/druid/src/app_delegate.rs
@@ -59,7 +59,7 @@ impl<'a> DelegateCtx<'a> {
                 Target::Global,
             );
         } else {
-            const MSG: &str = "All application windows must represent the same type of state.";
+            const MSG: &str = "WindowDesc<T> - T must match the application state.";
             if cfg!(debug_assertions) {
                 panic!(MSG);
             } else {
@@ -79,7 +79,7 @@ impl<'a> DelegateCtx<'a> {
                 Target::Window(window),
             );
         } else {
-            const MSG: &str = "Menus must represent the application state.";
+            const MSG: &str = "MenuDesc<T> - T must match the application state.";
             if cfg!(debug_assertions) {
                 panic!(MSG);
             } else {
--- a/druid/src/app_delegate.rs
+++ b/druid/src/app_delegate.rs
@@ -49,7 +49,7 @@ impl<'a> DelegateCtx<'a> {
     }
 
     /// Create a new window.
-    /// `T` must represent the application state provided to [`AppLauncher::launch`].
+    /// `T` must be the application's root `Data` type (the type provided to [`AppLauncher::launch`]).
     ///
     /// [`AppLauncher::launch`]: struct.AppLauncher.html#method.launch
     pub fn new_window<T: Any>(&mut self, desc: WindowDesc<T>) {
@@ -68,8 +68,8 @@ impl<'a> DelegateCtx<'a> {
         }
     }
 
-    /// Set the windows menu.
-    /// `T` must represent the application state provided to [`AppLauncher::launch`].
+    /// Set the window's menu.
+    /// `T` must be the application's root `Data` type (the type provided to [`AppLauncher::launch`]).
     ///
     /// [`AppLauncher::launch`]: struct.AppLauncher.html#method.launch
     pub fn set_menu<T: Any>(&mut self, menu: MenuDesc<T>, window: WindowId) {
--- a/druid/src/app_delegate.rs
+++ b/druid/src/app_delegate.rs
@@ -14,7 +14,10 @@
 
 //! Customizing application-level behaviour.
 
-use std::{any::Any, collections::VecDeque};
+use std::{
+    any::{Any, TypeId},
+    collections::VecDeque,
+};
 
 use crate::{
     commands, contexts::StateTypes, Command, Data, Env, Event, MenuDesc, Target, WindowDesc,
@@ -53,7 +56,7 @@ impl<'a> DelegateCtx<'a> {
     ///
     /// [`AppLauncher::launch`]: struct.AppLauncher.html#method.launch
     pub fn new_window<T: Any>(&mut self, desc: WindowDesc<T>) {
-        if self.state_types.check_window_desc::<T>() {
+        if self.state_types.window_desc == TypeId::of::<WindowDesc<T>>() {
             self.submit_command(
                 Command::one_shot(commands::NEW_WINDOW, desc),
                 Target::Global,
@@ -73,7 +76,7 @@ impl<'a> DelegateCtx<'a> {
     ///
     /// [`AppLauncher::launch`]: struct.AppLauncher.html#method.launch
     pub fn set_menu<T: Any>(&mut self, menu: MenuDesc<T>, window: WindowId) {
-        if self.state_types.check_menu_desc::<T>() {
+        if self.state_types.menu_desc == TypeId::of::<MenuDesc<T>>() {
             self.submit_command(
                 Command::new(commands::SET_MENU, menu),
                 Target::Window(window),
--- a/druid/src/app_delegate.rs
+++ b/druid/src/app_delegate.rs
@@ -14,15 +14,19 @@
 
 //! Customizing application-level behaviour.
 
-use std::collections::VecDeque;
+use std::{any::Any, collections::VecDeque};
 
-use crate::{Command, Data, Env, Event, Target, WindowId};
+use crate::{
+    commands, contexts::StateTypes, Command, Data, Env, Event, MenuDesc, Target, WindowDesc,
+    WindowId,
+};
 
 /// A context passed in to [`AppDelegate`] functions.
 ///
 /// [`AppDelegate`]: trait.AppDelegate.html
 pub struct DelegateCtx<'a> {
     pub(crate) command_queue: &'a mut VecDeque<(Target, Command)>,
+    pub(crate) state_types: StateTypes,
 }
 
 impl<'a> DelegateCtx<'a> {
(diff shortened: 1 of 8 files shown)


druid/src/contexts.rs
```diff
@@ -242,6 +276,60 @@ impl<'a> EventCtx<'a> {
...
+        }
+    }
+
+    /// Set the menu of the window containing the event handling widget.
+    /// `T` must represent the application state provided to [`AppLauncher::launch`].
+    ///
+    /// [`AppLauncher::launch`]: struct.AppLauncher.html#method.launch
+    pub fn set_menu<T: Any>(&mut self, menu: MenuDesc<T>) {
+        if self.state_types.check_menu_desc::<T>() {
+            self.submit_command(Command::new(commands::SET_MENU, menu), None);
```
xStrom, line 305: I know `None` defaults to the current window, but I think we should be explicit in internal code when we can. Should just use `self.window_id` instead of `None`. Same for `show_context_menu`.

druid/src/contexts.rs
```diff
@@ -242,6 +276,60 @@ impl<'a> EventCtx<'a> {
...
+    ///
+    /// [`AppLauncher::launch`]: struct.AppLauncher.html#method.launch
+    pub fn new_window<T: Any>(&mut self, desc: WindowDesc<T>) {
+        if self.state_types.check_window_desc::<T>() {
+            self.submit_command(
+                Command::one_shot(commands::NEW_WINDOW, desc),
+                Target::Global,
+            );
+        } else {
+            const MSG: &str = "All application windows must represent the same type of state.";
```
xStrom, line 290: This message is confusing me.
Perhaps something like `WindowDesc<T> - T must match the application state.`

druid/src/contexts.rs
```diff
@@ -242,6 +276,60 @@ impl<'a> EventCtx<'a> {
...
+
+    /// Set the menu of the window containing the event handling widget.
+    /// `T` must represent the application state provided to [`AppLauncher::launch`].
+    ///
+    /// [`AppLauncher::launch`]: struct.AppLauncher.html#method.launch
+    pub fn set_menu<T: Any>(&mut self, menu: MenuDesc<T>) {
+        if self.state_types.check_menu_desc::<T>() {
+            self.submit_command(Command::new(commands::SET_MENU, menu), None);
+        } else {
+            const MSG: &str = "Menus must represent the application state.";
```
xStrom, line 307: I don't think *represent* is the right word here.
I would go with something more specific like `MenuDesc<T> - T must match the application state.`

druid/src/contexts.rs
```diff
@@ -242,6 +276,60 @@ impl<'a> EventCtx<'a> {
...
+
+    /// Show the context menu in the window containing the event handling widget.
+    /// `T` must represent the application state provided to [`AppLauncher::launch`].
+    ///
+    /// [`AppLauncher::launch`]: struct.AppLauncher.html#method.launch
+    pub fn show_context_menu<T: Any>(&mut self, menu: ContextMenu<T>) {
+        if self.state_types.check_context_menu::<T>() {
+            self.submit_command(Command::new(commands::SHOW_CONTEXT_MENU, menu), None);
+        } else {
+            const MSG: &str = "Context menus must represent the application state.";
```
xStrom, line 324: I don't think *represent* is the right word here.
I would go with something more specific like `ContextMenu<T> - T must match the application state.`

druid/src/app_delegate.rs
```diff
@@ -43,6 +47,46 @@ impl<'a> DelegateCtx<'a> {
...
+    ///
+    /// [`AppLauncher::launch`]: struct.AppLauncher.html#method.launch
+    pub fn set_menu<T: Any>(&mut self, menu: MenuDesc<T>, window: WindowId) {
+        if self.state_types.check_menu_desc::<T>() {
+            self.submit_command(
+                Command::new(commands::SET_MENU, menu),
+                Target::Window(window),
+            );
+        } else {
+            const MSG: &str = "Menus must represent the application state.";
```
xStrom, line 82: I don't think *represent* is the right word here.
I would go with something more specific like `MenuDesc<T> - T must match the application state.`

druid/src/app_delegate.rs
```diff
@@ -43,6 +47,46 @@ impl<'a> DelegateCtx<'a> {
...
+    ///
+    /// [`AppLauncher::launch`]: struct.AppLauncher.html#method.launch
+    pub fn new_window<T: Any>(&mut self, desc: WindowDesc<T>) {
+        if self.state_types.check_window_desc::<T>() {
+            self.submit_command(
+                Command::one_shot(commands::NEW_WINDOW, desc),
+                Target::Global,
+            );
+        } else {
+            const MSG: &str = "All application windows must represent the same type of state.";
```
xStrom, line 62: This message is confusing me.
Perhaps something like `WindowDesc<T> - T must match the application state.`

Finnerale:
Thanks for the suggestions!

druid/src/app_delegate.rs
```diff
@@ -43,6 +47,46 @@ impl<'a> DelegateCtx<'a> {
...
+            const MSG: &str = "WindowDesc<T> - T must match the application state.";
+            if cfg!(debug_assertions) {
+                panic!(MSG);
+            } else {
+                log::error!("DelegateCtx::new_window: {}", MSG)
+            }
+        }
+    }
+
+    /// Set the windows menu.
```
cmyr, line 71:
```suggestion
    /// Set the window's menu.
```

druid/src/app_delegate.rs
```diff
@@ -43,6 +47,46 @@ impl<'a> DelegateCtx<'a> {
         let target = target.into().unwrap_or(Target::Global);
         self.command_queue.push_back((target, command))
     }
+
+    /// Create a new window.
+    /// `T` must represent the application state provided to [`AppLauncher::launch`].
```
cmyr, line 52: I would say,


```suggestion
    /// `T` must be the application's root `Data` type (the type provided to [`AppLauncher::launch`].)
```

druid/src/app_delegate.rs
```diff
@@ -43,6 +47,46 @@ impl<'a> DelegateCtx<'a> {
...
+            if cfg!(debug_assertions) {
+                panic!(MSG);
+            } else {
+                log::error!("DelegateCtx::new_window: {}", MSG)
+            }
+        }
+    }
+
+    /// Set the windows menu.
+    /// `T` must represent the application state provided to [`AppLauncher::launch`].
```
cmyr, line 72:
```suggestion
    /// `T` must be the application's root `Data` type (the type provided to [`AppLauncher::launch`].)
```

druid/src/contexts.rs
```diff
@@ -15,16 +15,27 @@
 //! The context types that are passed into various widget methods.
 
 use std::ops::{Deref, DerefMut};
-use std::time::Duration;
+use std::{
```
cmyr, line 18: if we're making this change I'd go ahead and combine `ops` as well 🤷

druid/src/contexts.rs
```diff
@@ -242,6 +276,66 @@ impl<'a> EventCtx<'a> {
         &self.window
     }
 
+    /// Create a new window.
+    /// `T` must represent the application state provided to [`AppLauncher::launch`].
```
cmyr, line 280:
```suggestion
    /// `T` must be the application's root `Data` type (the type provided to [`AppLauncher::launch`].)
```

druid/src/contexts.rs
```diff
@@ -242,6 +276,66 @@ impl<'a> EventCtx<'a> {
...
+            const MSG: &str = "WindowDesc<T> - T must match the application state.";
+            if cfg!(debug_assertions) {
+                panic!(MSG);
+            } else {
+                log::error!("EventCtx::new_window: {}", MSG)
+            }
+        }
+    }
+
+    /// Set the menu of the window containing the event handling widget.
```
cmyr, line 299:
```suggestion
    /// Set the menu of the window containing the current widget.
```

druid/src/contexts.rs
```diff
@@ -242,6 +276,66 @@ impl<'a> EventCtx<'a> {
...
+    /// `T` must represent the application state provided to [`AppLauncher::launch`].
```
cmyr, line 300:
```suggestion
    /// `T` must be the application's root `Data` type (the type provided to [`AppLauncher::launch`].)
```

druid/src/app_delegate.rs
```diff
@@ -43,6 +47,46 @@ impl<'a> DelegateCtx<'a> {
         let target = target.into().unwrap_or(Target::Global);
         self.command_queue.push_back((target, command))
     }
+
+    /// Create a new window.
+    /// `T` must represent the application state provided to [`AppLauncher::launch`].
+    ///
+    /// [`AppLauncher::launch`]: struct.AppLauncher.html#method.launch
+    pub fn new_window<T: Any>(&mut self, desc: WindowDesc<T>) {
+        if self.state_types.check_window_desc::<T>() {
```
cmyr, line 56: since we know this is a `WIndowDesc` and we're only worried about `T`, can we just store the typeid of `T` and compare that?

druid/src/contexts.rs
```diff
@@ -242,6 +276,66 @@ impl<'a> EventCtx<'a> {
...
+    /// [`AppLauncher::launch`]: struct.AppLauncher.html#method.launch
+    pub fn show_context_menu<T: Any>(&mut self, menu: ContextMenu<T>) {
+        if self.state_types.check_context_menu::<T>() {
+            self.submit_command(
+                Command::new(commands::SHOW_CONTEXT_MENU, menu),
+                Target::Window(self.window_id),
+            );
+        } else {
+            const MSG: &str = "ContextMenu<T> - T must match the application state.";
+            if cfg!(debug_assertions) {
```
cmyr, line 331: if this becomes a pattern we could consider a custom macro, like `log_or_panic!(...)`.

Finnerale: I would like it if this becomes a pattern (or `log_or_panic!`) as `panic!` makes debugging easier due to the back trace while `log::error!` does not crash the app once you ship it.

Finnerale:
@cmyr Thanks for the review, I think I've addressed everything.

PR #940, opened 2020-05-15
Add im::Vector support to the List widget.

xStrom:
This PR adds `im::Vector` support for the `List` widget when the `im` feature is enabled. It seems to work real nicely with `ListIter<T>` where we can just pass references. The shared data `ListIter<(T1, T)>` implementation is less nice as we still need to clone data due to the signature. It may be worth looking into changing the signature for shared data lists in the future so that we could use just references even with shared data.

The `Vector` usage seems much more ergonomical than `Arc<Vec>`. I also updated the `list` example to use `Vector`. The real functional diff to the example looks great, with basically just removing code. However due to having to add feature guards the code diff looks a lot noisier.

I also removed an unnecessary variable from the `Arc<Vec>` implementation, which was the remnant of a previous refactoring.

This PR also reverts the `Data` implementation of `Vec<T: Data>` as discussed in #911. The implementation was broken as well.

### DIFF
--- a/druid/src/widget/list.rs
+++ b/druid/src/widget/list.rs
@@ -93,16 +93,17 @@ impl<T: Data> ListIter<T> for Vector<T> {
     }
 }
 
+// S == shared data type
 #[cfg(feature = "im")]
-impl<T1: Data, T: Data> ListIter<(T1, T)> for (T1, Vector<T>) {
-    fn for_each(&self, mut cb: impl FnMut(&(T1, T), usize)) {
+impl<S: Data, T: Data> ListIter<(S, T)> for (S, Vector<T>) {
+    fn for_each(&self, mut cb: impl FnMut(&(S, T), usize)) {
         for (i, item) in self.1.iter().enumerate() {
             let d = (self.0.to_owned(), item.to_owned());
             cb(&d, i);
         }
     }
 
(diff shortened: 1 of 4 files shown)


druid/src/widget/list.rs
```diff
@@ -71,6 +74,53 @@ pub trait ListIter<T>: Data {
...
+        }
+    }
+
+    fn data_len(&self) -> usize {
+        self.len()
+    }
+}
+
+#[cfg(feature = "im")]
+impl<T1: Data, T: Data> ListIter<(T1, T)> for (T1, Vector<T>) {
```
Finnerale, line 97: Maybe call them `Shared` and `T` instead? Or at least `S` and `T`.
I still do not understand why people like useless names for generic parameters,
this is really not something where we should let mathematicians inspire us.

xStrom: I just copied that from the `Arc<Vec>` implementation, but yeah it took me a while to grasp what the `T1` even is, because I hadn't used `List` before. I changed it to `S` and added a comment.

Finnerale: Thanks! At first I got it the wrong way, thought `T1` would be the iterable due to the `1`.

cmyr:
Something to reiterate:  the current `List` widget is a proof of concept, and may need to be reconsidered or otherwise rearchitected.

PR #941, opened 2020-05-15
On wasm, clip to the invalid rect.

### LINKED ISSUE #872: The invalidation example behaves differently with wasm

jneem:
Fixes #872.


jneem:
With this change, gtk, wasm, and windows all clip to the invalid rect. I'm not sure what mac does (I can't test it), and X11 doesn't support partial invalidation at all yet.

PS: What's the changelog policy on something like this? It's a fix to a feature that didn't exist in the last release, so does it need an entry?

PR #942, opened 2020-05-15
Clean up cargo manifests and generate more docs.

xStrom:
This PR does the following:
* Enables the `im` feature for [docs.rs/druid](https://docs.rs/druid/).
  The `svg` and `image` features run into cargo bugs unfortunately and will have to wait.
* Adds CI docs generation for all the crates with their features.
* Cleans up the `Cargo.toml` files so that they follow a unified style.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -320,11 +320,38 @@ jobs:
           profile: minimal
           override: true
 
-      - name: check docs in druid/
+      # Doc packages in deeper-to-higher dependency order
+      - name: cargo doc druid-shell
         uses: actions-rs/cargo@v1
         with:
           command: doc
-          args: --document-private-items
+          args: --manifest-path=druid-shell/Cargo.toml --document-private-items
+
+      - name: cargo doc druid
+        uses: actions-rs/cargo@v1
+        with:
+          command: doc
+          args: --manifest-path=druid/Cargo.toml --features=svg,image,im --document-private-items
+
+      - name: cargo doc druid-derive
+        uses: actions-rs/cargo@v1
+        with:
+          command: doc
+          args: --manifest-path=druid-derive/Cargo.toml --document-private-items
+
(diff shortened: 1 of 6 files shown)


druid/Cargo.toml
```diff
@@ -11,45 +11,41 @@ keywords = ["gui", "ui", "toolkit"]
...
+# the "svg" and "image" features should be enabled for the docs.rs output.
+features = ["im"]
 default-target = "x86_64-pc-windows-msvc"
 
 [features]
 x11 = ["druid-shell/x11"]
 svg = ["usvg"]
 
-[badges]
-travis-ci = { repository = "xi-editor/druid" }
```
Finnerale, line 21: I suppose this was removed intentionally? I could not find this badge anywhere on crates.io or docs.rs :thinking:

.github/workflows/ci.yml
```diff
@@ -320,11 +320,38 @@ jobs:
...
+        uses: actions-rs/cargo@v1
+        with:
+          command: doc
+          args: --manifest-path=druid-derive/Cargo.toml --document-private-items
+
+      - name: cargo doc book examples
+        uses: actions-rs/cargo@v1
+        with:
+          command: doc
+          args: --manifest-path=docs/book_examples/Cargo.toml --document-private-items
```
Finnerale, line 346: Seems consequent, but I'm not sure we need this one?

druid/Cargo.toml
xStrom: Yeah that's a remnant of a time when druid still used Travis CI.

.github/workflows/ci.yml
xStrom: Well might as well make sure that the doc format in the book examples is functional.

xStrom:
The book docs build takes around 20s on the CI, which you can see from the logs. I don't think it's an issue.

PR #948, opened 2020-05-16
Make focus request handling predictable.

xStrom:
Focus request handling in `master` is somewhat unpredictable:
- If a parent requests focus to start with, that wins regardless of further focus requests by children.
- If a paren't doesn't request focus immediately, then the first child to request focus wins - maybe.
- If a parent requests focus after passing events to children, then the parent wins.

This PR makes it a lot simpler to understand:
- Whoever calls `request_focus` last will win.

This change makes it easier to build widget groups where both the parent and children are interested in focus. It will allow the parent to request focus before passing down the event if it's ok with not winning. With `master` the parent would have to track the children's interest in focus by its own means to not override it.

### DIFF
--- a/druid/src/core.rs
+++ b/druid/src/core.rs
@@ -15,6 +15,7 @@
 //! The fundamental druid types.
 
 use std::collections::{HashMap, VecDeque};
+use std::mem;
 
 use crate::bloom::Bloom;
 use crate::kurbo::{Affine, Insets, Point, Rect, Shape, Size, Vec2};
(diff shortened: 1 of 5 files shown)


druid/src/core.rs
```diff
@@ -962,8 +957,8 @@ impl BaseState {
         self.has_active |= child_state.has_active;
         self.has_focus |= child_state.has_focus;
         self.children_changed |= child_state.children_changed;
-        self.request_focus = self.request_focus.or(child_state.request_focus);
-        self.timers.extend(&child_state.timers);
+        self.request_focus = child_state.request_focus.take().or(self.request_focus);
+        self.timers.extend(&mut child_state.timers.drain());
```
cmyr, line 961: Was just looking at this yesterday; there's a small optimization we could make where if child has timers and we don't, we just do a `mem::swap`, which will save us an allocation?

PR #949, opened 2020-05-16
Fix progress bar rendering

scholtzan:
Fixes https://github.com/xi-editor/druid/issues/919

The problem causing this rendering issue is that insets get applied to a `Rect` with zero width. This results in a `Rect` of some negative width getting displayed when the progress bar is at zero.

### DIFF
--- a/druid/src/widget/progress_bar.rs
+++ b/druid/src/widget/progress_bar.rs
@@ -60,6 +60,7 @@ impl Widget<f64> for ProgressBar {
     fn paint(&mut self, ctx: &mut PaintCtx, data: &f64, env: &Env) {
         let clamped = data.max(0.0).min(1.0);
         let stroke_width = 2.0;
+        let inset = -stroke_width / 2.0;
 
         let rounded_rect = Rect::from_origin_size(
             Point::ORIGIN,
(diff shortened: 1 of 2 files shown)


druid/src/widget/progress_bar.rs
```diff
@@ -86,18 +87,20 @@ impl Widget<f64> for ProgressBar {
         );
         ctx.fill(rounded_rect, &background_gradient);
 
-        //Paint the bar
+        // Paint the bar
         let calculated_bar_width = clamped * rounded_rect.width();
+
         let rounded_rect = Rect::from_origin_size(
-            Point::ORIGIN,
+            Point::new(-inset, 0.),
```
scholtzan, line 94: We need to translate the rect to respect the x inset.

xStrom:
Thanks for the fix!

Please also add an entry to [`CHANGELOG.md`](https://github.com/xi-editor/druid/blob/master/CHANGELOG.md) - [more instructions here](https://github.com/xi-editor/druid/blob/master/CONTRIBUTING.md).

scholtzan:
Added an entry to the changelog

cmyr:
Thanks @scholtzan!

PR #952, opened 2020-05-17
Fix simple logger init checks.

xStrom:
Converting the `Result` into an `Option` and not doing anything with it seems like a mistake. I think `unwrap` was intended. I went with `expect` with a bit more info.


xStrom:
I added a note about the panic.

Issue #834, opened 2020-04-13
Briefly show the scrollbars any time the Scroll widget's size changes


xStrom:
There's a `reset_scrollbar_fade(&mut self, ctx: &mut EventCtx, env: &Env)` function already that does this. The problem is that it requires the `EventCtx` because it uses a timer to achieve its goal. The `Scroll` only really knows its size has changed when `layout` gets called.

Is this an example that we need some new druid event for when our own size changes? Is there any way to plumb the timer request at all right now? It's not even possible to send commands from `layout` or `paint`.

Previously `Scroll` was listening to `Event::Size` but that was a mistake, because that event only gets sent to the root widget and means that the window size has changed. Window size changing doesn't necessarily mean the widget's size has changed even if we propagate that to every widget.

PR #953, opened 2020-05-17
Add the LifeCycle::Size event.

xStrom:
There have been several occurrences now where a widget would like to know when its size changes and do some non-layout work. This was [disccused in zulip](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/send.20command.20to.20sibling.20when.20resize) and this would also help solve an animation issue with the `Scroll` widget as explained in #834.

Thus this PR adds the `LifeCycle::Size` event which gets sent every time the size of the widget changes.

### DIFF
--- a/druid/src/core.rs
+++ b/druid/src/core.rs
@@ -188,12 +188,36 @@ impl<T, W: Widget<T>> WidgetPod<T, W> {
         self.state.id
     }
 
-    /// Set layout rectangle.
+    /// Set the layout [`Rect`].
     ///
-    /// Intended to be called on child widget in container's `layout`
-    /// implementation.
+    /// A container widget should call the [`Widget::layout`] method on its children in
+    /// its own [`Widget::layout`] implementation, then possibly modify the returned [`Size`], and
+    /// finally call this `set_layout_rect` method on the child to set the final layout [`Rect`].
+    ///
+    /// The child will receive the [`LifeCycle::Size`] event informing them of the final [`Size`].
+    ///
+    /// [`Widget::layout`]: trait.Widget.html#tymethod.layout
+    /// [`Rect`]: struct.Rect.html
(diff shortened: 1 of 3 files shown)


druid/src/core.rs
```diff
@@ -215,9 +243,12 @@ impl<T, W: Widget<T>> WidgetPod<T, W> {
         self.layout_rect()
     }
 
-    /// The layout rectangle.
+    /// Returns the layout [`Rect`].
+    ///
+    /// This will be same value as set by [`set_layout_rect`].
```
Finnerale, line 248: Should this be something like "This will be *the* same value as set by `set_layout_rect`"?

druid/src/core.rs
```diff
@@ -188,13 +188,37 @@ impl<T, W: Widget<T>> WidgetPod<T, W> {
...
+    /// [`Rect`]: struct.Rect.html
+    /// [`Size`]: struct.Size.html
+    /// [`LifeCycle::Size`]: enum.LifeCycle.html#variant.Size
     pub fn set_layout_rect(&mut self, ctx: &mut LayoutCtx, data: &T, env: &Env, layout_rect: Rect) {
+        let mut needs_merge = false;
+
+        let old_size = self.state.layout_rect.map(|r| r.size());
+        let new_size = layout_rect.size();
+
         self.state.layout_rect = Some(layout_rect);
```
cmyr, line 209: this feels like an opportunity for `Option::replace`, maybe.

druid/src/core.rs
```diff
@@ -188,13 +188,37 @@ impl<T, W: Widget<T>> WidgetPod<T, W> {
...
 
+        if old_size.is_none() || old_size.unwrap() != new_size {
```
cmyr, line 211: and I'd have written this as `if old_size != Some(new_size)`, probably

PR #954, opened 2020-05-17
Add the request_timer method to LifeCycleCtx.

xStrom:
This PR adds the `request_timer` method to `LifeCycleCtx` which will allow to fix the `Scroll` widget as explained in #834.

To achieve this I had to introduce the `WindowHandle` pointer to both `LifeCycleCtx` and `LayoutCtx`.

### DIFF
--- a/druid/src/window.rs
+++ b/druid/src/window.rs
@@ -250,16 +250,17 @@ impl<T: Data> Window<T> {
         env: &Env,
         process_commands: bool,
     ) {
-        let mut base_state = BaseState::new(self.root.id());
-        let mut ctx = LifeCycleCtx {
-            command_queue: queue,
-            window_id: self.id,
-            base_state: &mut base_state,
-        };
-
         if let LifeCycle::AnimFrame(_) = event {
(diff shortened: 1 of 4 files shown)


druid/src/window.rs
```diff
@@ -250,24 +250,33 @@ impl<T: Data> Window<T> {
...
-            window_id: self.id,
-            base_state: &mut base_state,
-        };
-
         if let LifeCycle::AnimFrame(_) = event {
-            self.do_anim_frame(&mut ctx, data, env)
+            self.do_anim_frame(queue, data, env)
         } else {
+            let mut base_state = BaseState::new(self.root.id());
+            let mut ctx = LifeCycleCtx {
```
cmyr, line 257: hmm, I think I'm missing the reason we can't reuse the context here?

xStrom: It's the borrow checker. `LifeCycleCtx` holds an immutable reference to `self` while `do_anim_frame` wants a mutable reference to `self`.

PR #957, opened 2020-05-17
Fix GTK folder selection with filters.

### LINKED ISSUE #947: Unable to filter directories on GTK file dialogs.
This was reported in runebender (https://github.com/linebender/runebender/issues/108); apparently the patch that added file type filtering to gtk (#903?) means we're no longer able to open directories, even with the `select_directories` option set.

xStrom:
This PR disables file filters when opening folders, because the filter makes it impossible to traverse in the folder structure.

Fixes #947.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ You can find its changes [documented below](#060---2020-06-01).
 
 ### Fixed
 
+- GTK: Directory selection now properly ignores file filters. ([#957] by [@xStrom])
+
 ### Visual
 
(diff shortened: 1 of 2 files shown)


xStrom:
After looking more deeply into it, filters don't really work on directories even on macOS. They work on [packages](https://en.wikipedia.org/wiki/Package_(macOS)) instead. I cleaned up the API and added better docs in #960.

This PR now focuses on matching the newly documented API in behavior.
- Directories don't get filtered here, and they don't get filtered on other platforms.
- There's no logging of warnings, because specifying filters in a directory selection mode is legal. They are used to determine which paths are packages and which are directories. (The opposite of what was happening here.)

PR #959, opened 2020-05-18
Remove one shot commands.

Finnerale:
Fourth part of #908 .

I gave the suggestion by @cmyr a try and replaced one shot commands with a `SingleUse` wrapper type and it seems to work really well.

To summarize the change:
```rust
ctx.submit_command(
    Command::one_shot(commands::NEW_WINDOW, desc),
    Target::Global,
);

let desc = cmd.take_object::<WindowDesc<T>>().unwrap();
```
is now
```rust
ctx.submit_command(
    Command::new(commands::NEW_WINDOW, SingleUse::new(desc)),
    Target::Global,
);

let desc = cmd.get_object::<SingleUse<WindowDesc<T>>>().unwrap().take().unwrap();
```

which can easily be utilized by typed selectors.

### DIFF
--- a/druid/src/app_delegate.rs
+++ b/druid/src/app_delegate.rs
@@ -19,7 +19,9 @@ use std::{
     collections::VecDeque,
 };
 
-use crate::{commands, Command, Data, Env, Event, MenuDesc, Target, WindowDesc, WindowId};
+use crate::{
+    command::SingleUse, commands, Command, Data, Env, Event, MenuDesc, Target, WindowDesc, WindowId,
+};
 
 /// A context passed in to [`AppDelegate`] functions.
 ///
@@ -55,7 +57,7 @@ impl<'a> DelegateCtx<'a> {
     pub fn new_window<T: Any>(&mut self, desc: WindowDesc<T>) {
         if self.app_data_type == TypeId::of::<T>() {
             self.submit_command(
-                Command::one_shot(commands::NEW_WINDOW, desc),
+                Command::new(commands::NEW_WINDOW, SingleUse::new(desc)),
                 Target::Global,
             );
         } else {
--- a/druid/src/app_delegate.rs
+++ b/druid/src/app_delegate.rs
@@ -20,7 +20,7 @@ use std::{
 };
 
 use crate::{
-    command::SingleUse, commands, Command, Data, Env, Event, MenuDesc, Target, WindowDesc, WindowId,
+    commands, Command, Data, Env, Event, MenuDesc, SingleUse, Target, WindowDesc, WindowId,
 };
 
 /// A context passed in to [`AppDelegate`] functions.
--- a/druid/src/command.rs
+++ b/druid/src/command.rs
@@ -35,7 +35,7 @@ pub struct Selector(&'static str);
 /// and an optional argument, that can be used to pass arbitrary data.
 ///
 /// If the payload can't or shouldn't be cloned,
-/// wrapping it with [`SingleUse`] allows `take`ing the object.
+/// wrapping it with [`SingleUse`] allows you to `take` the object.
 ///
 /// # Examples
 /// ```
--- a/druid/src/command.rs
+++ b/druid/src/command.rs
@@ -34,17 +34,8 @@ pub struct Selector(&'static str);
 /// A `Command` consists of a [`Selector`], that indicates what the command is,
 /// and an optional argument, that can be used to pass arbitrary data.
 ///
-///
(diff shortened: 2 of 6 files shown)


druid/src/command.rs
```diff
@@ -34,17 +34,8 @@ pub struct Selector(&'static str);
...
-/// Commands come in two varieties, 'reusable' and 'one-shot'.
-///
-/// Regular commands are created with [`Command::new`], and their argument
-/// objects may be accessed repeatedly, via [`Command::get_object`].
-///
-/// One-shot commands are intended for cases where an object should only be
-/// used once; an example would be if you have some resource that cannot be
-/// cloned, and you wish to send it to another widget.
+/// If the payload can't or shouldn't be cloned,
+/// wrapping it with [`SingleUse`] allows `take`ing the object.
```
xStrom, line 38: Given that *takeing* is a misspelling, perhaps write it as: *allows you to `take` the object.*

druid/src/win_handler.rs
```diff
@@ -591,7 +591,10 @@ impl<T: Data> AppState<T> {
     }
 
     fn new_window(&mut self, cmd: Command) -> Result<(), Box<dyn std::error::Error>> {
-        let desc = cmd.take_object::<WindowDesc<T>>()?;
+        let desc = cmd.get_object::<SingleUse<WindowDesc<T>>>()?;
+        // The NEW_WINDOW command is private and only druid can receive it by normal means,
+        // thus unwrapping can be considered save or deserves a panic.
```
xStrom, line 596: ~save or~ safe and

druid/src/command.rs
```diff
@@ -59,31 +50,27 @@ pub struct Selector(&'static str);
...
 
-#[derive(Debug, Clone)]
-enum Arg {
-    Reusable(Arc<dyn Any>),
-    OneShot(Arc<Mutex<Option<Box<dyn Any>>>>),
-}
+/// `SingleUse` is intended for cases where a command payload should only be
+/// used once; an example would be if you have some resource that cannot be
+/// cloned, and you wish to send it to another widget.
+pub struct SingleUse<T>(Mutex<Option<T>>);
```
cmyr, line 65: this type isn's actually exposed anywhere, so I don't think external crates can access it?

Also I would expect this to need to be in an `Arc`, since the `Event` itself still needs to be cloneable?

Finnerale: Good catch, I've reexported it from `lib` now.

Not sure what you mean regarding the `Arc`. When using `SingleUse` the payload will be stored the exact same way as it used to with `Arg::OneShot`, minus the `Box<Any>` wrapper as it is now directly stored as `Arc<Any>`.

cmyr: ah okay I missed that the `Arc` has moved into the signature of the `object` field on `Command`.

druid/src/command.rs
```diff
@@ -59,31 +50,27 @@ pub struct Selector(&'static str);
...
-    object: Option<Arg>,
+    object: Option<Arc<dyn Any>>,
 }
 
-#[derive(Debug, Clone)]
-enum Arg {
-    Reusable(Arc<dyn Any>),
-    OneShot(Arc<Mutex<Option<Box<dyn Any>>>>),
-}
+/// `SingleUse` is intended for cases where a command payload should only be
```
cmyr, line 62: huh I think I lost a comment here?? anyway, to try again:

Doc comments are kind of like git commit messages; you want the first paragraph to be a short general overview that doesn't need a lot of context, and then a blank line, and then a more detailed explanation.

This is because the first paragraph is displayed beside the type name in the module or crate level docs.

For this type, for instance, I would try a first paragraph of something to the effect of,

"A wrapper type for `Command` arguments that should only be used once."

Finnerale: Thanks, I should have actually built the docs :sweat_smile: 

I've added the suggested head paragraph, and while I was at it also an example how to make use of `SingleUse`.

druid/src/app_delegate.rs
```diff
@@ -19,7 +19,9 @@ use std::{
     collections::VecDeque,
 };
 
-use crate::{commands, Command, Data, Env, Event, MenuDesc, Target, WindowDesc, WindowId};
+use crate::{
+    command::SingleUse, commands, Command, Data, Env, Event, MenuDesc, Target, WindowDesc, WindowId,
```
cmyr, line 23: since this is exported at root we don't need the module qualification.

PR #960, opened 2020-05-18
Add more file dialog configuration and macOS support.

### LINKED ISSUE #390: Implement file dialog default extension in mac & gtk

### LINKED ISSUE #902: Allow customization of Save/Cancel buttons in file dialogs
For example, if my application has an "Export' functionality, then I want to be able to open a file dialog where the buttons are "Cancel" and "Export" instead of "Cancel" and "Save".

xStrom:
This PR adds a bunch of stuff to the `FileDialogOptions` struct.
- `default_name` to specify the file name that should be pre-filled.
- `name_label` to specify text that should appear next to the file name textbox.
- `title` to set the title of the dialog.
- `button_text` to set the text on the main save/open button.
- `force_starting_directory` to support niche use cases where the app wants to override the OS chosen starting directory in the dialog.

To get things going I started with the macOS implementation. However these are all things that can also be supported on Windows, and most if not all also on GTK.

I updated the `select_directories` documentation to specify that this means that only folders are being selected. There is no way to select both folders and files on Windows. So for a cross-platform API that doesn't make sense. If there is some use case where you want both folders and files to be selected then that should be a new mode called `select_files_and_directories` or something and have clear documentation about being available only on macOS (and..?).

I also made some macOS specific changes in relation to the existing API:
- File packages (directories with known extensions) are now treated as plain directories. File packages are a macOS concept and don't translate in a cross-platform way. This is related to [runebender#108](https://github.com/linebender/runebender/issues/108) because runebender expects directories to be treated as files when they have an extension, but that's a macOS-only behavior. The cross-platform way to do it would be to enable `select_directories`.
- Implemented `default_type` which will be used e.g. in the save dialog when you don't type in an extension.
- Fixed the save dialog not showing up at all when `select_directories` or `multi_selection` are set.

Miscellaneous:
- Changed `FileSpec` to use a non-static lifetime. It works the same way with `'static` but now also supports non-static use cases.

Fixes #390.
Fixes #902.

### DIFF
--- a/druid-shell/src/dialog.rs
+++ b/druid-shell/src/dialog.rs
@@ -40,7 +40,8 @@ pub enum FileDialogType {
 /// By default the file dialogs operate in *files mode* where the user can only choose files.
 /// Importantly these are files from the user's perspective, but technically the returned path
 /// will be a directory when the user chooses a package. Read more about [packages] below.
-/// Thus it is important to verify that all the returned paths match your expectations.
+/// It's also possible for users to manually specify a path which they might otherwise not be able
+/// to choose. Thus it is important to verify that all the returned paths match your expectations.
 ///
 /// The open dialog can also be switched to *directories mode* via [`select_directories`].
 ///
@@ -51,13 +52,16 @@ pub enum FileDialogType {
 /// First there are packages that have been defined at the OS level, and secondly there are
 /// packages that are defined at the file dialog level based on [`allowed_types`].
 /// A package behaves similarly to a regular file in many contexts, including the file dialogs.
-/// This means that it can be chosen in the file open dialog when operating in *files mode*,
-/// all the file filters will apply to it, and the user won't be able to traverse into the package.
-/// It won't be selectable in *directories mode* and the user won't be able to traverse into the
-/// package. In the file save dialog packages that are only defined in [`allowed_types`] can be
-/// traversed into. However packages that are defined at the OS level can't be traversed into
-/// even if they are also defined in [`allowed_types`]. Keep in mind though that the file dialog
-/// may start inside any package if the user had traversed into one previously.
+/// This package concept can be turned off in the file dialog via [`packages_as_directories`].
+///
+/// &#xFEFF; | Packages as files. File filters apply to packages. | Packages as directories.
+/// -------- | -------------------------------------------------- | ------------------------
+/// Open directory | Not selectable. Not traversable. | Selectable. Traversable.
+/// Open file | Selectable. Not traversable. | Not selectable. Traversable.
+/// Save file | OS packages [clickable] but not traversable.<br/>Dialog packages traversable but not selectable. | Not selectable. Traversable.
+///
+/// Keep in mind that the file dialog may start inside any package if the user had traversed
+/// into one previously. The user might also manually specify a path inside a package.
 ///
 /// Generally this behavior should be kept, because it's least surprising to macOS users.
 /// However if your application requires selecting directories with extensions as directories
@@ -65,10 +69,36 @@ pub enum FileDialogType {
 /// then you can change the default behavior via [`packages_as_directories`]
 /// to force macOS to behave like other platforms and not give special treatment to packages.
 ///
+/// # macOS
+///
+/// **Use the save dialog only for new paths**
+///
+/// Existing files can be clicked on in the save dialog, but that only copies their base file name.
+/// If the clicked file's extension is different than the first extension of the default type
+/// then the returned path does not actually match the path of the file that was clicked on.
+/// Clicking on a file doesn't change the base path either. So if a user has traversed into
+/// `/Users/Joe/foo/` and then clicks on an existing file `/Users/Joe/old.txt` then the returned
+/// path will actually be `/Users/Joe/foo/old.rtf` if the default type's first extension is `rtf`.
+///
+/// Thus it's not a good idea to direct the user to choose an existing file with the save dialog.
+/// The much more optimized flow is to have the user select a file with the open dialog
+/// and then keep saving to that file without showing a save dialog.
+/// Use the save dialog only for selecting a new location.
+///
+/// **Have a really good save dialog default type**
+///
+/// There is no way for the user to choose which extension they want to save a file as via the UI.
+/// They have no way of knowing which extensions are even supported and must manually type it out.
+///
+/// *Hopefully it's a temporary problem and we can find a way to show the file formats in the UI.
+/// This is being tracked in [druid#998].*
+///
+/// [Clickable]: #macos
 /// [packages]: #packages
 /// [`select_directories`]: #method.select_directories
 /// [`allowed_types`]: #method.allowed_types
 /// [`packages_as_directories`]: #method.packages_as_directories
+/// [druid#998]: https://github.com/xi-editor/druid/issues/998
 #[derive(Debug, Clone, Default)]
 pub struct FileDialogOptions {
     pub(crate) show_hidden: bool,
@@ -141,6 +171,8 @@ impl FileDialogOptions {
     }
 
     /// Set [packages] to be treated as directories instead of files.
+    ///
+    /// This allows for writing more universal cross-platform code at the cost of user experience.
     ///
     /// This is only relevant on macOS.
     ///
--- a/druid-shell/src/dialog.rs
+++ b/druid-shell/src/dialog.rs
@@ -33,19 +33,17 @@ pub enum FileDialogType {
 
 /// Options for file dialogs.
 #[derive(Debug, Clone, Default)]
-pub struct FileDialogOptions<'a> {
-    pub show_hidden: bool,
-    pub allowed_types: Option<Vec<FileSpec<'a>>>,
-    pub default_type: Option<FileSpec<'a>>,
-    pub select_directories: bool,
-    pub multi_selection: bool,
-    pub default_name: Option<&'a str>,
-    pub name_label: Option<&'a str>,
-    pub title: Option<&'a str>,
-    pub button_text: Option<&'a str>,
-    pub starting_directory: Option<&'a Path>,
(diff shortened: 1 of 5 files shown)


xStrom:
I was thinking some more about these macOS packages and if some of it could be surfaced through a new `select_packages` option. I don't see a clear way forward though.

For opening, this `select_packages` mode could fall back to `select_directories` on other platforms. However on macOS it could have a special path where it's a file open dialog with packages treated as files. This would allow the file filters to work on macOS in this mode. Although even in this mode the user could still just choose a plain old file instead of a package and I don't see a way to turn that off in the Apple docs.

The real trouble starts with saving though. On macOS it would work fine in that same packages-as-files mode. However on other platforms you can only save files. This would work if you're creating a new package. Because you can just treat that file location as a directory location instead. However what if you want to overwrite an existing package? Well that's a directory and you can't select it in the save dialog. One workaround would be to save a file called "foo.pkg" in the parent directory of the "foo.pkg" package. However that would require manual typing, the user can't click on the package to get that result.

So for now I haven't added any `select_packages` mode as I can't see a way to do it in a reasonable way.

I do wonder though, how does runebender deal with overwriting existing packages on Windows right now? Requires manual typing of the name?

cmyr:
Something we're going to need, eventually, is a common pattern for implementing platform specific behaviour, especially around things like this. One bit of prior art we might look at are the various `Ext` traits in the stdlib, like `OsStringExt`, which adds methods to `OsString` for the current platform?

In runebender, if I have to use `cfg` to create different sets of flags for different platforms, that's fine for now; I just want to make sure we don't *decrease* the set of things that can be expressed. 

I think on mac that file packages should count as files, not directories, and it's up to the user to use `cfg` to get things to work cross-platform. I think this just because that's conceptually how the mac treats them; they're single 'things' for the purposes of the UI, and a normal user would never know that they're actually folders on the file system.

druid-shell/src/dialog.rs
```diff
@@ -33,12 +33,17 @@ pub enum FileDialogType {
 
 /// Options for file dialogs.
 #[derive(Debug, Clone, Default)]
-pub struct FileDialogOptions {
+pub struct FileDialogOptions<'a> {
```
cmyr, line 36: I expect this to cause problems? In particular I think we need to be able to `Box` this type in order to use it with `Command`.

I _believe_ this was the rationale behind the `'static` lifetimes in `FileSpec` as well.

druid-shell/src/platform/mac/dialog.rs
```diff
@@ -28,42 +28,97 @@ use crate::dialog::{FileDialogOptions, FileDialogType};
...
     unsafe {
         let panel: id = match ty {
             FileDialogType::Open => msg_send![class!(NSOpenPanel), openPanel],
             FileDialogType::Save => msg_send![class!(NSSavePanel), savePanel],
         };
 
-        // set options
+        // Directories with extensions need to be treated as directories.
+        // File packages are a macOS concept and don't translate in a cross-platform way.
+        let () = msg_send![panel, setTreatsFilePackagesAsDirectories: YES];
```
cmyr, line 44: discussed in a separate comment, I'm not sure about this.

xStrom:
Okay so in terms of packages and what the default should be. I can see the value in user expectations as far as browsing the file dialog goes. Being able to traverse into packages is probably not common.

Treating them as files means though that druid apps need to always check if a `OPEN_FILE`/`SAVE_FILE` provided `FileInfo` is a file or a directory(package).

So me defaulting to treating them as directories was optimizing for less error checking in app code.

I don't have any strong opinions here though, so I'm open to defaulting back to treating them as files and just adding a bunch of documentation warning people that file paths may actually be packages on macOS.

Changing this default back doesn't really help out runebender though. It makes it more difficult for runebender. `SHOW_OPEN_PANEL` should be used only on macOS. (Alternatively looking for `metainfo.plist` instead, but the point is that the code isn't universal.) This is also one of the reasons I changed the default, so that apps like runebender could have one cross-platform logic for dealing with these. I do understand though that the macOS experience suffers when treating packages as directories. That's why I mentioned the potential `select_packages` mode. This mode in theory would help deal with packages using a single code path while utilizing all of macOS strengths. However like I pointed out in a previous post, overwriting existing "packages" on Windows looks really messy, so this mode might be rather difficult to implement. (*It's probably possible via shell extensions added to the file dialog on Windows. Less sure about options with GTK.*)

druid-shell/src/dialog.rs
xStrom: You're right. Lifetimes won't work here. However `'static` isn't going to work either, because these need to be able to be generated at runtime.

Could just convert them to `String` - or maybe some sort of Enum of `&'static str` and `String`? Or is there a better standard way? Surely this is a common issue.

xStrom:
Okay this is ready for another look.
- The lifetime issue is resolved by using `Into<String>` instead.
- macOS packages are back to being treated as files by default, but can be optionally treated as directories.
- Added a bunch of docs around packages and their behavior.

Finnerale:
Hm, this is super irritating to me as I've never worked with macOS.

This is how I understood the `FileDialogOptions` docs:
- files mode
  - not macOS: all matching files
  - macOS: all matching files or packages
    (entering non-system packages is possible)
- files mode +  packages_as_directories
  - all matching files - same on all platforms, packages don't exist
- directory mode
  - not macOS: all directories, ignoring filters
  - macOS: all directories or packages matching the filters
    (entering non-system packages is possible)
- directory mode +  packages_as_directories
  - not macOS: all directories
  - macOS: all directories matching the filters, packages don't exist

xStrom:
It's definitely convoluted and you didn't quite get it right. Maybe I should rewrite the behavior as a table similar to what you created here.

First, in files mode, you can enter non-system packages only in the save dialog. Not in the open dialog. I'll at the very least add an extra clarification about this even if I don't rewrite the whole paragraph.

In directory mode, you only get directories. You can't interact with packages at all. The file filters decide which directories lose their directory status and are now considered packages. Packages can't be chosen or traversed into.

In directory mode + packages_as_directories you can select and traverse into all directories, packages don't exist. Filters don't do anything, like on other platforms.

xStrom:
Okay I think I improved the documentation some more. This includes adding a table to describe how packages interact with the various dialogs.

druid-shell/src/dialog.rs
```diff
@@ -32,25 +34,98 @@ pub enum FileDialogType {
...
+/// Keep in mind that the file dialog may start inside any package if the user has traversed
+/// into one just recently. The user might also manually specify a path inside a package.
+///
+/// Generally this behavior should be kept, because it's least surprising to macOS users.
+/// However if your application requires selecting directories with extensions as directories
+/// or the user needs to be able to traverse into them to select a specific file,
+/// then you can change the default behavior via [`packages_as_directories`]
+/// to force macOS to behave like other platforms and not give special treatment to packages.
+///
+/// # macOS
```
Finnerale, line 72: This differentiation between `Packages` and `macOS` kind of makes it look as if `Packages` would be relevant on other platforms as well.

druid-shell/src/dialog.rs
```diff
@@ -32,25 +34,98 @@ pub enum FileDialogType {
...
+///
+/// **Use the save dialog only for new paths**
+///
+/// Existing files can be clicked on in the save dialog, but that only copies their base file name.
+/// If the clicked file's extension is different than the first extension of the default type
+/// then the returned path does not actually match the path of the file that was clicked on.
+/// Clicking on a file doesn't change the base path either. So if a user has traversed into
+/// `/Users/Joe/foo/` and then clicks on an existing file `/Users/Joe/old.txt` then the returned
```
Finnerale, line 80: Did you mean `/Users/Joe/foo/old.txt` instead of `/Users/Joe/old.txt`?

xStrom: No, what I wrote is correct. The macOS file dialog can have multiple directories open at once. Take a look at [this screenshot](https://intellij-support.jetbrains.com/hc/user_images/dwX1BVg1MYJXOzDreC8SqA.png).

druid-shell/src/dialog.rs
xStrom: Good point, I moved it all under the `macOS` header.

druid-shell/src/dialog.rs
xStrom: I added a bit of extra clarification to this section, pointing out that the macOS file dialog can have several directories open at once.

xStrom:
Yes this is not ideal. Generally I would prefer to make things automatically cross-platform so that users wouldn't have to worry about things like this.

I do sympathize with macOS users liking their packages and not wanting to lose that. It is a concept that doesn't really translate in a cross-platform way though. So I personally would use something like a zip file to emulate a package and that would be cross-platform.

For developers I guess the guideline would be to avoid using directories with extensions unless they love `#[cfg]` guards. If your app doesn't need to interact with directories with extensions, then you don't even really need to know that packages are a thing. Packages can appear as packages for macOS users when they browse around in the file dialog, but they're never relevant for your app.

I added a note at the top of the docs informing users of this cross-platform universal path.

xStrom:
You're right, the docs should be giving better actionable advice. I reworded some of that and offer it as general cross-platform advice.

cmyr:
closed in favour of #1257

PR #961, opened 2020-05-18
Support mouse scroll in X11.

### LINKED ISSUE #921: X11: Add support for the mouse wheel.
Currently the X11 backend only supports mouse move and clicks, however it should also support mouse wheel events. There's probably some useful inspiration in the GTK backend implementation.

jneem:
Fixes #921.

### DIFF
--- a/druid-shell/src/platform/x11/window.rs
+++ b/druid-shell/src/platform/x11/window.rs
@@ -604,10 +604,13 @@ impl Window {
 
     pub fn handle_wheel(&self, event: &xcb::ButtonPressEvent) -> Result<(), Error> {
         let button = event.detail();
-        // The scroll delta of 120.0 was stolen from the GTK code, which apparently came from the
-        // windows docs.
+        let mods = key_mods(event.state());
+
+        // We use a delta of 120 per tick to match the behavior of windows.
         let delta = match button {
+            4 if mods.shift => (-120.0, 0.0),
             4 => (0.0, -120.0),
+            5 if mods.shift => (120.0, 0.0),
             5 => (0.0, 120.0),
             6 => (-120.0, 0.0),
             7 => (120.0, 0.0),
--- a/druid-shell/src/platform/x11/window.rs
+++ b/druid-shell/src/platform/x11/window.rs
@@ -606,7 +606,7 @@ impl Window {
         let button = event.detail();
         let mods = key_mods(event.state());
 
-        // We use a delta of 120 per tick to match the behavior of windows.
+        // We use a delta of 120 per tick to match the behavior of Windows.
         let delta = match button {
             4 if mods.shift => (-120.0, 0.0),
             4 => (0.0, -120.0),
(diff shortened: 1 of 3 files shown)


druid-shell/src/platform/x11/window.rs
```diff
@@ -602,6 +602,42 @@ impl Window {
...
+        let button = event.detail();
+        // The scroll delta of 120.0 was stolen from the GTK code, which apparently came from the
+        // windows docs.
+        let delta = match button {
+            4 => (0.0, -120.0),
+            5 => (0.0, 120.0),
+            6 => (-120.0, 0.0),
+            7 => (120.0, 0.0),
+            _ => {
+                log::error!("unexpected mouse wheel button: {}", button);
```
xStrom, line 615: This is a good spot to return an error instead. Perhaps something like:
```rust
_ => return Err(Error::Generic(format!("unexpected mouse wheel button: {}", button))),
```

druid-shell/src/platform/x11/window.rs
```diff
@@ -602,6 +602,42 @@ impl Window {
         }
     }
 
+    pub fn handle_wheel(&self, event: &xcb::ButtonPressEvent) -> Result<(), Error> {
+        let button = event.detail();
+        // The scroll delta of 120.0 was stolen from the GTK code, which apparently came from the
+        // windows docs.
+        let delta = match button {
```
xStrom, line 609: We should also check for whether the shift key is being held down and if so convert vertical scrolling into horizontal scrolling. This was introduced with #709 and you can copy the logic from the GTK code probably.

druid-shell/src/platform/x11/window.rs
```diff
@@ -602,6 +602,42 @@ impl Window {
         }
     }
 
+    pub fn handle_wheel(&self, event: &xcb::ButtonPressEvent) -> Result<(), Error> {
+        let button = event.detail();
+        // The scroll delta of 120.0 was stolen from the GTK code, which apparently came from the
+        // windows docs.
```
xStrom, line 608: I think we should rewrite this comment as more to the point. Something like:
*We use a delta of 120 per tick to match the behavior of Windows.*

druid-shell/src/platform/x11/window.rs
jneem: I'm not sure if this affects your preferred error handling technique, but hitting this case represents a bug in druid-shell. It's impossible for now because we check that the button is between 4 and 7 before calling `handle_wheel`.

xStrom: Yeah I think returning an error in case of invalid input parameters is just fine.

druid-shell/src/platform/x11/window.rs
```diff
@@ -602,6 +602,45 @@ impl Window {
         }
     }
 
+    pub fn handle_wheel(&self, event: &xcb::ButtonPressEvent) -> Result<(), Error> {
+        let button = event.detail();
+        let mods = key_mods(event.state());
+
+        // We use a delta of 120 per tick to match the behavior of windows.
```
xStrom, line 609: Let's call it *Windows* with a capital W to make it clear we're talking about the OS and not the concept, given that all the windows (as in the concept) code is in this same file.

PR #964, opened 2020-05-21
Fix timer requests in update and lifecycle.

### LINKED ISSUE #834: Briefly show the scrollbars any time the Scroll widget's size changes

xStrom:
This PR fixes the handling of timer requests. Previously they were only collected in the window's `event` but I moved the collection to the general post processing step. I also removed the legacy `request_timer` `bool` which no longer served a purpose and was never being reset.

This also finally allowed to address #834. Scrollbars now briefly animate when the container is resized. This can be seen with e.g. the `scroll_colors` example when you resize the window.

Fixes #834.

### DIFF
--- a/druid/src/util.rs
+++ b/druid/src/util.rs
@@ -1,0 +1,52 @@
+// Copyright 2020 The xi-editor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Miscellaneous utility functions.
+
+use std::collections::HashMap;
+use std::hash::Hash;
+use std::mem;
+
+/// Fast path for equal type extend + drain.
+pub trait ExtendDrain {
+    /// Extend the collection by draining the entries from `source`.
+    ///
+    /// This function may swap the underlying memory locations,
+    /// so keep that in mind if one of the collections has a large allocation
+    /// and it should keep that allocation.
+    fn extend_drain(&mut self, source: &mut Self);
+}
+
+impl<K, V> ExtendDrain for HashMap<K, V>
+where
+    K: Eq + Hash + Copy,
+    V: Copy,
+{
+    // Benchmarking this vs just extend+drain with a 10k entry map.
+    //
+    // running 2 tests
+    // test bench_extend       ... bench:       1,971 ns/iter (+/- 566)
+    // test bench_extend_drain ... bench:           0 ns/iter (+/- 0)
+    fn extend_drain(&mut self, source: &mut Self) {
+        if !source.is_empty() {
+            if self.is_empty() {
+                // If the target is empty we can just swap the pointers.
+                mem::swap(self, source);
+            } else {
+                // Otherwise we need to fall back to regular extend-drain.
+                self.extend(source.drain());
+            }
+        }
+    }
+}
--- a/druid/src/widget/scroll.rs
+++ b/druid/src/widget/scroll.rs
@@ -176,12 +176,15 @@ impl<T, W: Widget<T>> Scroll<T, W> {
     }
 
     /// Makes the scrollbars visible, and resets the fade timer.
-    pub fn reset_scrollbar_fade(&mut self, ctx: &mut EventCtx, env: &Env) {
+    pub fn reset_scrollbar_fade<F>(&mut self, request_timer: F, env: &Env)
+    where
+        F: FnOnce(Duration) -> TimerToken,
+    {
         // Display scroll bars and schedule their disappearance
         self.scrollbars.opacity = env.get(theme::SCROLLBAR_MAX_OPACITY);
         let fade_delay = env.get(theme::SCROLLBAR_FADE_DELAY);
         let deadline = Duration::from_millis(fade_delay);
-        self.scrollbars.timer_id = ctx.request_timer(deadline);
+        self.scrollbars.timer_id = request_timer(deadline);
     }
 
     /// Returns the current scroll offset.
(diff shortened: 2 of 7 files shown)


druid/src/window.rs
```diff
@@ -288,6 +293,8 @@ impl<T: Data> Window<T> {
         if ctx.base_state.request_anim {
             self.last_anim = Some(now);
         }
+
+        self.post_event_processing(&mut base_state, queue, data, env, process_commands);
```
cmyr, line 297: Is there a specific reason we think this is necessary? do_anim_frame is part of paint, and paint calls this already.

druid/src/widget/scroll.rs
```diff
@@ -176,12 +176,15 @@ impl<T, W: Widget<T>> Scroll<T, W> {
     }
 
     /// Makes the scrollbars visible, and resets the fade timer.
-    pub fn reset_scrollbar_fade(&mut self, ctx: &mut EventCtx, env: &Env) {
+    pub fn reset_scrollbar_fade<F>(&mut self, request_timer: F, env: &Env)
```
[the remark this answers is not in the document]
xStrom: Yeap, I was thinking of the same thing based on the amount of code duplication we have in `context.rs`. Then having to do this `request_timer` closure here really boosted that thinking. I have it in my notes to attempt to better the situation with traits. (*Would be a separate PR.*)

druid/src/window.rs
xStrom: - It's necessary because a widget might call `submit_command` or `request_timer` in response to `AnimFrame`.
- `do_anim_frame` is currently part of `lifecycle` not `paint` and so even if `paint` called `post_event_processing` that wouldn't be guaranteed for other `AnimFrame` calls to `lifecycle`.
- `paint` doesn't call `post_event_processing`.
- `AnimFrame` does not guarantee `paint` in the future as discussed [on zulip](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/purpose.20of.20request_anim).

druid/src/util.rs
```diff
@@ -0,0 +1,47 @@
...
+    /// and it should keep that allocation.
+    fn extend_drain(&mut self, source: &mut Self);
+}
+
+impl<K, V> ExtendDrain for HashMap<K, V>
+where
+    K: Eq + Hash + Copy,
+    V: Copy,
+{
+    fn extend_drain(&mut self, source: &mut Self) {
```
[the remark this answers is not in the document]
xStrom: There was a funny surprise alright. The empty target case is 0ns with this optimized function, which is a win. However unexpectedly even the slow path is faster! The slow path is faster only with `drain` though. If I bench extending with just `iter` vs `iter`, then the slow paths are equal. Some sort of non-generic code win I guess.

druid/src/widget/scroll.rs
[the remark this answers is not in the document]
cmyr: I am accidentally working on this right now.

druid/src/window.rs
cmyr: in order:

- to my mind, `AnimFrame` should only be used to update state in preparation for an imminent paint call. This is a weird case;  `AnimFrame` is one of the bits of event flow that is really a poor fit to the current architecture, and when (like.... a year ago? 😬) I was playing around with a higher-level animation API, one of the changes was to just make a `prepare_for_animation` method on `Widget` that replaced `AnimFrame`. `submit_command` and `request_timer` should be discouraged here, and it's annoying that this isn't enforced by the type system.

- `do_anim_frame` sort of is part of paint; `LifeCycle::AnimFrame` should only be sent from `Window::do_paint`, if it happens anywhere else that's a mistake.

- this checks out, I must've misread again. That said, it looks like we  call `post_event_processing` in Window::lifecycle`, which is what calls `do_anim_frame`?

- I think this is false? it's not really that anim guarantees `paint`, it's that `AnimFrame` is only called from the painting code path.

druid/src/util.rs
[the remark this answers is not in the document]
cmyr: neat, thanks for measuring!

druid/src/window.rs
xStrom: * Yes we do call `post_event_processing` in `Window::lifecycle` right now, which means it also applies to `do_anim_frame`. However due to borrow checking limitations, in order to keep that functionality, I had to add it to `do_anim_frame` because the `Window::lifecycle` call moved into an `if` block.

* Yeah right now `AnimFrame` is sent just before `paint`. However based on that zulip discussion this may change. @raphlinus said *It's up to the handler of that lifecycle event to request painting* - from which I derive that if the handler doesn't request painting, the painting might not happen at all. (This is for future behavior, today that's not true.)

More generally, moving `AnimFrame` out of `lifecycle` makes sense if it's supposed to have different limitations. However as long as it's part of `lifecycle` I think it should behave like a `lifecycle` event. That includes a working `request_timer` and `submit_command`.

cmyr: On the second point: I think there's a confusion about raph's comments. `AnimFrame` _does_ mean you're going to paint; there's no short-circuiting. If you see it, it means a call to paint is next. To continue the animation, you *do* need to request another frame.

I'll take a look at the code you're talking about, I might've missed that.

druid/src/window.rs
```diff
@@ -260,13 +260,18 @@ impl<T: Data> Window<T> {
             };
 
             self.root.lifecycle(&mut ctx, event, data, env);
+            self.post_event_processing(&mut base_state, queue, data, env, process_commands);
```
cmyr, line 263: we can move this out of the `if` branch if we move the `base_state` decl to above the block, this should solve our other little confusion?

xStrom: Yes! That will work.

PR #969, opened 2020-05-21
Rename BaseState to WidgetState

cmyr:
This is motivated by a desire to avoid some confusion with a
future patch that adds a RootState struct to unify some of the
other stuff shared between various contexts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -138,6 +138,7 @@ This means that druid no longer requires cairo on macOS and uses Core Graphics i
 - X11: Refactored `Window` to support some reentrancy and invalidation. ([#894] by [@xStrom])
 - Refactored DPI scaling. ([#904] by [@xStrom])
 - Added docs generation testing for all features. ([#942] by [@xStrom])
+- Renamed `BaseState` to `WidgetState` ([#969] by [@cmyr])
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -70,6 +70,7 @@ This means that druid no longer requires cairo on macOS and uses Core Graphics i
 - `SHOW_WINDOW` and `CLOSE_WINDOW` commands now only use `Target` to determine the affected window. ([#928] by [@finnerale])
 - Replaced `NEW_WINDOW`, `SET_MENU` and `SHOW_CONTEXT_MENU` commands with methods on `EventCtx` and `DelegateCtx`. ([#931] by [@finnerale])
 - Replaced `Command::one_shot` and `::take_object` with a `SingleUse` payload wrapper type. ([#959] by [@finnerale])
+- Renamed `BaseState` to `WidgetState` ([#969] by [@cmyr])
```
Finnerale, line 73: `BaseState` / `WidgetState` are not public API right? In that case this is not really a change for users.
I believe it should go into the `Maintenance` section instead.

cmyr: cool, I think this should be ready to go now, pending CI.

PR #970, opened 2020-05-21
Common context state

cmyr:
This is two related commits:

- one adds the `set_menu` method to the lifecycle and update contexts
- one refactors contexts so that all of the state that is A) shared and B) comes from the window and never changes is in one place. This will let us put common implementations there, and reduce duplicate code.

This is based on #969, which should go in first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ This means that druid no longer requires cairo on macOS and uses Core Graphics i
 - `im::Vector` support for the `List` widget. ([#940] by [@xStrom])
 - `LifeCycle::Size` event to inform widgets that their size changed. ([#953] by [@xStrom])
 - `Button::dynamic` constructor. ([#963] by [@totsteps])
+- `set_menu` method on `UpdateCtx` and `LifeCycleCtx` ([#970] by [@cmyr])
 
 ### Changed
 
@@ -219,6 +220,7 @@ This means that druid no longer requires cairo on macOS and uses Core Graphics i
 [#963]: https://github.com/xi-editor/druid/pull/963
 [#967]: https://github.com/xi-editor/druid/pull/967
 [#969]: https://github.com/xi-editor/druid/pull/969
+[#970]: https://github.com/xi-editor/druid/pull/970
 
 ## [0.5.0] - 2020-04-01
 
--- a/druid/src/contexts.rs
+++ b/druid/src/contexts.rs
@@ -594,6 +594,27 @@ impl<'a> LifeCycleCtx<'a> {
         let target = target.into().unwrap_or_else(|| self.window_id.into());
         self.command_queue.push_back((target, command.into()))
     }
+
+    /// Set the menu of the window containing the current widget.
+    /// `T` must be the application's root `Data` type (the type provided
+    /// to [`AppLauncher::launch`]).
+    ///
+    /// [`AppLauncher::launch`]: struct.AppLauncher.html#method.launch
+    pub fn set_menu<T: Any>(&mut self, menu: MenuDesc<T>) {
+        if self.app_data_type == TypeId::of::<T>() {
+            self.submit_command(
+                Command::new(commands::SET_MENU, menu),
+                Target::Window(self.window_id),
+            );
+        } else {
+            const MSG: &str = "MenuDesc<T> - T must match the application data type.";
+            if cfg!(debug_assertions) {
+                panic!(MSG);
+            } else {
+                log::error!("EventCtx::set_menu: {}", MSG)
+            }
+        }
+    }
 }
 
 impl<'a> UpdateCtx<'a> {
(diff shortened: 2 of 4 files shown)


druid/src/contexts.rs
```diff
@@ -87,12 +76,10 @@ pub struct UpdateCtx<'a> {
 /// As of now, the main service provided is access to a factory for
 /// creating text layout objects, which are likely to be useful
 /// during widget layout.
-pub struct LayoutCtx<'a, 'b: 'a> {
-    pub(crate) command_queue: &'a mut CommandQueue,
+pub struct LayoutCtx<'a, 'b: 'a, 'c> {
+    pub(crate) root_state: &'a mut RootState<'c>,
```
Finnerale, line 80: This might be a bit annoying to change but it would be nice if all `'b`s refer to the `RootState`s lifetime.

druid/src/contexts.rs
```diff
@@ -708,15 +680,15 @@ impl<'a> UpdateCtx<'a> {
     }
 }
 
-impl<'a, 'b> LayoutCtx<'a, 'b> {
+impl<'a, 'b, 'c> LayoutCtx<'a, 'b, 'c> {
```
Finnerale, line 683: Might be useful for the macro introduced by #972 :
This `'b` and `RootState`s  `'a` are the only required lifetimes in this file (for two functions only), all others could be elided using `'_`.

druid/src/contexts.rs
cmyr: yea that would be nice but doesn't work unfortunately. :(

druid/src/contexts.rs
```diff
@@ -883,6 +855,49 @@ impl<'a, 'b: 'a> PaintCtx<'a, 'b> {
     }
 }
 
+impl<'a> RootState<'a> {
```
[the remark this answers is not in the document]
cmyr: this is fair, I think the impl really should be here so maybe I'll move `RootState`/`ContextState` back here as well.

druid/src/contexts.rs
cmyr: interesting, I'll play around with that.

druid/src/contexts.rs
Finnerale: How so? The only 'issue' I see is `PainCtx` as it does not have `RootState` :thinking:

cmyr: oh sorry I thought you meant that we reuse `'b` here for both inner lifetimes.

druid/src/contexts.rs
cmyr: cool, I was able to simplify the lifetimes quite a bit, i'd forgotten they'd relaxed that rule :)

Finnerale: I also still keep adding them and then only realize afterwards that it was completely unnecessary :sweat_smile:

CHANGELOG.md
```diff
@@ -70,6 +70,7 @@ This means that druid no longer requires cairo on macOS and uses Core Graphics i
 - `SHOW_WINDOW` and `CLOSE_WINDOW` commands now only use `Target` to determine the affected window. ([#928] by [@finnerale])
 - Replaced `NEW_WINDOW`, `SET_MENU` and `SHOW_CONTEXT_MENU` commands with methods on `EventCtx` and `DelegateCtx`. ([#931] by [@finnerale])
 - Replaced `Command::one_shot` and `::take_object` with a `SingleUse` payload wrapper type. ([#959] by [@finnerale])
+- Add  `set_menu` method to `UpdateCtx` and `LifeCycleCtx` ([#970] by [@cmyr])
```
xStrom, line 73: Double space between *Add* and `set_menu`.

druid/src/contexts.rs
```diff
@@ -28,25 +28,30 @@ use crate::{
...
 ///
 /// Widgets should call [`request_paint`] whenever an event causes a change
 /// in the widget's appearance, to schedule a repaint.
 ///
 /// [`request_paint`]: #method.request_paint
-pub struct EventCtx<'a> {
+pub struct EventCtx<'a, 'b> {
     // Note: there's a bunch of state that's just passed down, might
     // want to group that into a single struct.
+    pub(crate) root_state: &'a mut ContextState<'b>,
```
xStrom, line 48: Since we renamed the type, how about we also rename the variables from `root_state` to `context_state` or given that they're part of contexts perhaps just `state`.

druid/src/core.rs
```diff
@@ -331,13 +328,10 @@ impl<T, W: Widget<T>> WidgetPod<T, W> {
...
     ///
     /// The provided `child_state` should be merged up if this returns `true`.
-    #[allow(clippy::too_many_arguments)]
     fn set_hot_state(
         child: &mut W,
-        command_queue: &mut CommandQueue,
         child_state: &mut WidgetState,
-        window_id: WindowId,
-        window: &WindowHandle,
+        root_state: &mut ContextState,
```
xStrom, line 334: Missed this `root_state`.

CHANGELOG.md
```diff
@@ -51,6 +51,7 @@ This means that druid no longer requires cairo on macOS and uses Core Graphics i
 - `im::Vector` support for the `List` widget. ([#940] by [@xStrom])
 - `LifeCycle::Size` event to inform widgets that their size changed. ([#953] by [@xStrom])
 - `Button::dynamic` constructor. ([#963] by [@totsteps])
+- Add `set_menu` method to `UpdateCtx` and `LifeCycleCtx` ([#970] by [@cmyr])
```
xStrom, line 54: Given how all the other items under the *Added* section are written, probably better to remove the word *Add* from the start here.

PR #974, opened 2020-05-22
Share context code via default trait implementations.

xStrom:
This is an alternative approach to #972 which aims to reduce copied code. Here, instead of macros, I implement it using traits.

Code inside of macros can't be formatted using `cargo fmt`. This means that over time the code formatting will be all over the place and will make it harder to read the code. Alternatively we would have to spend time manually checking formatting during code reviews, which seems almost worse.

I've also had some really annoying debugging experiences with macros due to compiler errors and line numbers not being meaningful. I'm not well versed in Rust macros so I can't say for sure if this will apply here, but I can say that I've had the issue with other macros.

Thus I propose we solve the issue without using macros. One way I can think of is using default trait implementations, whcih I did in this PR. That solves the problems that macros introduce. Traits also allow for more powerful function signatures in applications where the function is willing to take any context that implements a specific method.

The trait approach may have a bit of downside. The documentation for the context structs won't contain trait method docs. I've done a bit of googling but not thorough. Other people are having the same issue, but it's unclear if it's already possible to inline the docs to the struct, or if that's something that might become possible in a future version of `cargo doc`.

*PS. This PR still needs some cleanup, but I wanted to get the general idea out there as it is working already.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,7 @@ This means that druid no longer requires cairo on macOS and uses Core Graphics i
 - `SHOW_WINDOW` and `CLOSE_WINDOW` commands now only use `Target` to determine the affected window. ([#928] by [@finnerale])
 - Replaced `NEW_WINDOW`, `SET_MENU` and `SHOW_CONTEXT_MENU` commands with methods on `EventCtx` and `DelegateCtx`. ([#931] by [@finnerale])
 - Replaced `Command::one_shot` and `::take_object` with a `SingleUse` payload wrapper type. ([#959] by [@finnerale])
+- Renamed `BaseState` to `WidgetState` ([#969] by [@cmyr])
 
 ### Deprecated
 
@@ -215,6 +216,7 @@ This means that druid no longer requires cairo on macOS and uses Core Graphics i
 [#959]: https://github.com/xi-editor/druid/pull/959
 [#961]: https://github.com/xi-editor/druid/pull/961
 [#963]: https://github.com/xi-editor/druid/pull/963
+[#969]: https://github.com/xi-editor/druid/pull/969
 
 ## [0.5.0] - 2020-04-01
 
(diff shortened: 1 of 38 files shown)


cmyr:
This is definitely another approach. One of my main motivations for using a macro, despite the downsides, is to make the docs clear; each context's docs will list all the available methods in one place; it also avoids introducing too many new types. I do think we will want _some_ traits, though, so maybe it makes sense to do it all this way? not a clear call.

cmyr:
Another thing I'd been thinking about: even if we have traits that we use for different context behaviours, my thinking was that those would just be additive; all the methods would still be defined on the contexts, and the methods on the traits would just forward to those, so that you wouldn't even really need to know about the traits until you needed them; they'd be like a bonus thing.

Again, this is just how I'd been thinking about things..

jneem:
One point in favor of traits over macros is that it can enable code reuse in druid users. This is something that actually came up a little while ago, although I don't remember the exact details:

I have some code in a widget that wants to check a condition and possibly submit a command, and I want to run this code in both `lifecycle` and `event`. The natural thing would be to factor out the common code into another method, but this common code would need a context in order to submit the command. If `LifetimeCtx` and `EventCtx` implemented a common trait with a `submit_command` method, it would let me factor out that common code.

xStrom:
@jneem yeah that same issue is in the `Scroll` widget. A [closure can help](https://github.com/xi-editor/druid/blob/342264d125f5c3fd3aff950e5b594755d6642961/druid/src/widget/scroll.rs#L179-L181), but it's not ideal.

cmyr:
fwiw I did not close this intentionally.

I definitely agree that traits would be nice as a way to describe some common behaviours, but I'm not sure we want them to be the *basis* of how contexts are defined; but really the big thing I'm thinking about are docs, and making it easy to see what methods are available on what contexts.

PR #975, opened 2020-05-22
Update bloat action

cmyr:
We've removed the 'textbox' example, so this wasn't working anymore;
I've removed that from the test list and added the flex and
styled_text examples.


cmyr:
That makes one of us 😬

PR #976, opened 2020-05-22
Add more ctx.set_handled() to build-in widgets

Azorlogh:
Adds calls to `set_handled` to build-in widgets when they handle `MouseDown` and `KeyDown` (for `TextBox`).
This is so that custom widgets can store a `WidgetPod`, and choose to process events themselves if their children didn't.

Let me know if I missed anything or if changes are needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,7 @@ This means that druid no longer requires cairo on macOS and uses Core Graphics i
 - `SHOW_WINDOW` and `CLOSE_WINDOW` commands now only use `Target` to determine the affected window. ([#928] by [@finnerale])
 - Replaced `NEW_WINDOW`, `SET_MENU` and `SHOW_CONTEXT_MENU` commands with methods on `EventCtx` and `DelegateCtx`. ([#931] by [@finnerale])
 - Replaced `Command::one_shot` and `::take_object` with a `SingleUse` payload wrapper type. ([#959] by [@finnerale])
+- `is_handled` now returns true on `MouseDown` or `KeyDown` events if already handled by a built-in widget. ([#976] by [@Azorlogh])
 
 ### Deprecated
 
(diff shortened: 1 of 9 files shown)


xStrom:
If we're doing this for `MouseDown` and `KeyDown` then I think we should also do it for `MouseUp` and `KeyUp`. Perhaps even more generally, we should do it for any mouse/wheel/key event if the widget actually does something with the event that should be the only result of that event.

Azorlogh:
My reasoning was that since a `MouseUp` event could be happening on another different widget (if the user drags their mouse away after a click), it could lead to some inconsistencies.
(Ex: If the user releases their mouse on widget B after clicking widget A, if widget B comes before A in the tree, it would see the event, but not if it comes after.)

But now that I think about it, it shouldn't matter since widgets should not rely on a `MouseUp` without having first received a `MouseDown`, so I'll add it for `MouseUp` and `MouseMove` as well if you think I should.

cmyr:
I think consistency here is good, and I also expect that we will uncover some weird behaviour; `is_handled` just hasn't been used very consistently, and I suspect this will flush out some bugs.

Azorlogh:
I added it to `MouseMove` and `MouseUp`

xStrom:
Yeah indeed this will likely show some issues. I wouldn't be surprised to see hot state issues because of it. I'll think about that soon.

Also @Azorlogh, please rebase with `master` so that the CI can pass.

xStrom:
This is generally fine, however I'm in the process of reviewing the deeper effect. Don't want to break things too much just before 0.6.0 which is scheduled for June 1st.

Azorlogh:
Agreed, and there is always the possibility of temporarily copying the widgets' code in one's project anyway, for those who need this.

jneem:
Is there any further reason to wait on this?

xStrom:
> Is there any further reason to wait on this?

I've been meaning to add better tests around this stuff. Currently it's stuck behind #986.

We could merge it now, as the code here isn't bad per se. However the `handled` mechanics in general are buggy I'm pretty sure.

cmyr:
yea I'm just *slightly* scared that this PR is going to break a bunch of stuff; we do not have clearly expressed invariants around when `set_handled` should be called.

cmyr:
Going to close this as stale, we should revisit this but i'd like to do that on a case-by-case basis. Thanks for the PR  in any case @Azorlogh!

PR #982, opened 2020-05-23
Convert internal X11 errors to anyhow.

jneem:
This isn't really a serious PR, but I wanted to discuss something along these lines. Trying to add present support for X11 Present, I had lots of `RefCell` borrowing and, as a result, hundreds of lines of error handling boilerplate. This is an attempt to reduce that.

I'm not sure that adding a new dependency is the way to go, but I do think we want more ergonomic "generic" errors in druid-shell, particularly in the platform code. There are so many different errors that can come from the system, where it doesn't make sense to do much more than print a backtrace. Since many of these errors can never actually make it back to druid, it doesn't make much sense to have nice enums for them; that's why I thought of giving anyhow a try.


druid-shell/Cargo.toml
```diff
@@ -38,6 +38,7 @@ glib = { version = "0.9.3", optional = true }
 glib-sys = { version = "0.9.0", optional = true }
 gtk-sys = { version = "0.9.0", optional = true }
 xcb = { version = "0.9.0", features = ["thread", "xlib_xcb", "randr"], optional = true }
+anyhow = "1.0.31"
```
xStrom, line 41: This shouldn't be in the optional dependencies bunch.

druid-shell/src/platform/x11/window.rs
```diff
@@ -101,15 +102,9 @@ impl WindowBuilder {
...
-            Ok(reply) => reply.atom(),
-            Err(err) => {
-                return Err(Error::Generic(format!(
-                    "failed to get WM_PROTOCOLS: {}",
-                    err
-                )))
-            }
-        };
+        let wm_protocols = xcb::intern_atom(conn, false, "WM_PROTOCOLS")
+            .get_reply()?
```
xStrom, line 106: Should add context here.

druid-shell/src/platform/x11/window.rs
```diff
@@ -120,15 +115,9 @@ impl WindowBuilder {
...
-            Ok(reply) => reply.atom(),
-            Err(err) => {
-                return Err(Error::Generic(format!(
-                    "failed to get WM_DELETE_WINDOW: {}",
-                    err
-                )))
-            }
-        };
+        let wm_delete_window = xcb::intern_atom(conn, false, "WM_DELETE_WINDOW")
+            .get_reply()?
```
xStrom, line 119: Should add context here.

druid-shell/src/platform/x11/window.rs
```diff
@@ -171,14 +160,10 @@ impl WindowBuilder {
...
-        );
-        match cairo_surface {
-            Ok(cairo_surface) => Ok(RefCell::new(cairo::Context::new(&cairo_surface))),
-            Err(err) => Err(Error::Generic(format!(
-                "Failed to create cairo surface: {}",
-                err
-            ))),
-        }
+        )
+        .map_err(|status| anyhow!("Failed to create cairo surface: {}", status))?;
```
xStrom, line 164: What's the benefit of `map_err` over `context`?

druid-shell/src/platform/x11/window.rs
```diff
@@ -289,20 +274,13 @@ struct WindowState {
...
-                handler.size(size);
-                Ok(())
-            }
-            Err(err) => Err(Error::BorrowError(format!(
-                "Window::connect handler: {}",
-                err
-            ))),
-        }
+    fn connect(&self, handle: WindowHandle) -> Result<(), anyhow::Error> {
+        let mut handler = self.handler.try_borrow_mut().context("Window::connect")?;
```
xStrom, line 278: The context should contain the word *handler* like the old error so we know which variable this errored on.

druid-shell/src/platform/x11/window.rs
```diff
@@ -314,64 +292,37 @@ impl Window {
...
-                err
-            ))),
-        }
+    fn cairo_surface(&self) -> Result<XCBSurface, anyhow::Error> {
+        self.cairo_context
+            .try_borrow()
+            .context("Window::cairo_surface cairo_context")?
+            .get_target()
+            .try_into()
+            .map_err(|_| anyhow!("Window::cairo_surface try_into"))
```
xStrom, line 301: What's the benefit of `map_err` over `context`? Especially as we're dropping the underlying error.

druid-shell/src/platform/x11/window.rs
```diff
@@ -314,64 +292,37 @@ impl Window {
...
     }
 
     fn size(&self) -> Result<Size, Error> {
-        match self.state.try_borrow() {
-            Ok(state) => Ok(state.size),
-            Err(err) => Err(Error::BorrowError(format!("Window::size state: {}", err))),
-        }
+        Ok(self.state.try_borrow().context("Window::size")?.size)
```
xStrom, line 305: Let's keep the variable names in the context message like in the old error. Yes right now it's the only varible, but this will allow code growth without having to change things.

druid-shell/src/platform/x11/window.rs
```diff
@@ -314,64 +292,37 @@ impl Window {
...
-            match self.handler.try_borrow_mut() {
-                Ok(mut handler) => handler.size(size),
-                Err(err) => {
-                    return Err(Error::BorrowError(format!(
-                        "Window::set_size handler: {}",
-                        err
-                    )))
-                }
-            }
+        let mut state = self.state.try_borrow_mut().context("Window::set_size")?;
```
xStrom, line 310: Lets keep the variable name in the context message.

druid-shell/src/platform/x11/window.rs
```diff
@@ -314,64 +292,37 @@ impl Window {
...
+        if state.size != size {
+            state.size = size;
+            self.cairo_surface()?
+                .set_size(size.width as i32, size.height as i32)
+                .map_err(|status| {
```
xStrom, line 315: `map_err` vs `context`

druid-shell/src/platform/x11/window.rs
```diff
@@ -314,64 +292,37 @@ impl Window {
...
+                    anyhow!(
+                        "Failed to update cairo surface size to {:?}: {}",
+                        size,
+                        status
+                    )
+                })?;
+            self.handler
```
xStrom, line 322: This changes the lifetime of the `state` borrow. In the previous implementation it gets dropped as quickly as possible and certainly before calling the `size` handler which might call a method on `WindowHandle` which will want to borrow `state`. At the very least we should use `mem::drop` or something to ensure we don't hold it when calling this handler.

druid-shell/src/platform/x11/window.rs
```diff
@@ -536,19 +487,17 @@ impl Window {
...
-                Ok(())
-            }
-            Err(err) => Err(Error::BorrowError(format!(
-                "Window::handle_key_press handle: {}",
-                err
-            ))),
-        }
+        self.handler
+            .try_borrow_mut()
+            .context("Window::handle_key_press")?
```
xStrom, line 492: Lets keep the variable name in the context message. (`handler` instead of `handle` though)

druid-shell/src/platform/x11/window.rs
```diff
@@ -562,16 +511,11 @@ impl Window {
...
-                Ok(())
-            }
-            Err(err) => Err(Error::BorrowError(format!(
-                "Window::handle_button_press handle: {}",
-                err
-            ))),
-        }
+        self.handler
+            .try_borrow_mut()
+            .context("Window::handle_button_press")?
```
xStrom, line 516: Lets keep the variable name in the context message. (`handler` instead of `handle` though)

druid-shell/src/platform/x11/window.rs
```diff
@@ -593,19 +537,14 @@ impl Window {
...
-                Ok(())
-            }
-            Err(err) => Err(Error::BorrowError(format!(
-                "Window::handle_button_release handle: {}",
-                err
-            ))),
-        }
+        self.handler
+            .try_borrow_mut()
+            .context("Window::handle_button_release")?
```
xStrom, line 542: Lets keep the variable name in the context message. (`handler` instead of `handle` though)

druid-shell/src/platform/x11/window.rs
```diff
@@ -634,22 +568,17 @@ impl Window {
...
-                Ok(())
-            }
-            Err(err) => Err(Error::BorrowError(format!(
-                "Window::handle_wheel handle: {}",
-                err
-            ))),
-        }
+        self.handler
+            .try_borrow_mut()
+            .context("Window::handle_wheel")?
```
xStrom, line 573: Lets keep the variable name in the context message. (`handler` instead of `handle` though)

druid-shell/src/platform/x11/window.rs
```diff
@@ -662,16 +591,11 @@ impl Window {
...
-                Ok(())
-            }
-            Err(err) => Err(Error::BorrowError(format!(
-                "Window::handle_motion_notify handle: {}",
-                err
-            ))),
-        }
+        self.handler
+            .try_borrow_mut()
+            .context("Window::handle_motion_notify")?
```
xStrom, line 596: Lets keep the variable name in the context message. (`handler` instead of `handle` though)

druid-shell/src/platform/x11/window.rs
```diff
@@ -693,16 +617,11 @@ impl Window {
...
-                Ok(())
-            }
-            Err(err) => Err(Error::BorrowError(format!(
-                "Window::handle_destroy_notify handle: {}",
-                err
-            ))),
-        }
+        self.handler
+            .try_borrow_mut()
+            .context("Window::handle_destroy_notify")?
```
xStrom, line 622: Lets keep the variable name in the context message. (`handler` instead of `handle` though)

druid-shell/src/platform/x11/window.rs
jneem: `context` is only available on a `Result<T, E> where E: std::error::Error`, and several of the cairo "errors" don't implement `Error`

druid-shell/src/platform/x11/window.rs
jneem: I've made these even a little more compact with some macros that include not only the variable that caused the error, but also the source file and line where it happened.

druid-shell/src/platform/x11/window.rs
jneem: The underlying "error" here is just cairo returning the surface back to us, so it doesn't contain any useful information.

druid-shell/src/platform/x11/window.rs
jneem: This (and all the other places where I didn't include a variable name) now uses the macro.

druid-shell/src/platform/x11/window.rs
jneem: Good point, I've gone back to the old control flow.

druid-shell/src/platform/x11/window.rs
```diff
@@ -401,30 +355,23 @@ impl Window {
...
-            if let Ok(mut handler) = self.handler.try_borrow_mut() {
-                anim = handler.paint(&mut piet_ctx, invalid_rect);
-            } else {
-                log::error!("Window::render - handler already borrowed");
-            }
-            if let Err(e) = piet_ctx.finish() {
-                log::error!("Window::render - piet finish failed: {}", e);
-            }
+
+            anim = borrow_mut!(self.handler)?.paint(&mut piet_ctx, invalid_rect);
```
xStrom, line 370: We can't return an error here this easily. This will crash the program, because `piet_ctx` needs to be finished before it's dropped.

Probably also need to reset the cairo clip even if finishing fails.

I think it still makes sense to return the error from the function, however we can't do it immediately and need to try to perform all the cleanup we can.

jneem: Ok, good point. I've redone this part.

druid-shell/src/platform/x11/mod.rs
```diff
@@ -20,9 +20,22 @@
 
 pub mod application;
 pub mod clipboard;
-pub mod error;
 pub mod keycodes;
 pub mod menu;
 pub mod window;
 
+// TODO: This is just a placeholder, because we haven't removed all the other platform errors yet.
```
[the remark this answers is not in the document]
jneem: I could go either way on this, but let me explain my rationale anyway:

The only point (AFAICT) of having an error enum is to allow users to match on it, and I have trouble seeing a use-case for matching on a platform-specific error from druid-shell.

One reason for this is that druid-shell actually handles most of its own errors, because it's in charge of the run loop. I had a quick look through the API, and I only saw `Application::new` and `WindowBuilder::build` as places that the user gets an error.

The platform-specific nature of a platform error also makes me think that it's unlikely to be handled by the user, together with the fact that -- looking at the current set of errors -- I don't see what they would do with one besides logging it and exiting.

Having said that, `anyhow::Error` *does* allow the user to get the underlying error if they really want it, because it supports downcasting. The main disadvantage compared to an error enum is ergonomics and discoverability.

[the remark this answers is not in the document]
xStrom: Not a lot of errors being surfaced outside of `druid-shell` is more about it just not having been done yet. However I agree with your point that it's unlikely anyone will want to match against platform specific errors.

However the platform code itself might want to match against its own errors. Error handling isn't always going to be exactly where it happens.

Also the Windows `Error::Hr(HRESULT)` continues to be very useful. There's no base error that can just be contexted. Thus I still don't see the platform specific `Error` type going away.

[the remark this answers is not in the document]
jneem: Fair enough; I've reinstated the X11 error module.

druid-shell/src/platform/x11/error.rs
```diff
@@ -16,24 +16,13 @@
...
-            Error::Generic(msg) => write!(f, "Error: {}", msg),
-            Error::ConnectionError(err) => write!(f, "Connection error: {}", err),
-            Error::BorrowError(msg) => write!(f, "Failed to borrow: {}", msg),
-        }
+    fn fmt(&self, _: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+        Ok(())
     }
 }
-
-impl std::error::Error for Error {}
```
xStrom, line 39: Can we keep this impl line as well?

druid-shell/src/platform/x11/application.rs
```diff
@@ -183,150 +161,126 @@ impl Application {
         self.screen_num
     }
 
+    // Returns Ok(true) if we want to exit the main loop.
```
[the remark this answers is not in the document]
xStrom: Did this change get lost?

druid-shell/src/platform/x11/application.rs
```diff
@@ -14,6 +14,7 @@
 
 //! X11 implementation of features at the application scope.
 
+use anyhow::{anyhow, Context, Error};
```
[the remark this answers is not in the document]
xStrom: This is still there too.

jneem:
Sorry, I think I made a mistake with `git stash`. These should be fixed now.

CHANGELOG.md
```diff
@@ -143,6 +143,7 @@ This means that druid no longer requires cairo on macOS and uses Core Graphics i
 - Refactored DPI scaling. ([#904] by [@xStrom])
 - Added docs generation testing for all features. ([#942] by [@xStrom])
 - Renamed `BaseState` to `WidgetState` ([#969] by [@cmyr])
+- X11: Reworked error handling ([#982] by [@jneem])
```
xStrom, line 146: Need to also add the link to the bottom of the file.

druid-shell/src/platform/x11/window.rs
```diff
@@ -401,30 +355,37 @@ impl Window {
...
+                    anim = handler.paint(&mut piet_ctx, invalid_rect);
+                    err = piet_ctx
+                        .finish()
+                        .map_err(|e| anyhow!("Window::render - piet finish failed: {}", e));
+                }
+                Err(e) => {
+                    err = Err(e);
+                    let _ = piet_ctx.finish();
+                }
+            };
```
Finnerale, line 385: The second `finish` call could fail as well, right? If we can add it to `err`, than that would seem useful, otherwise we could just keep `log`ing this one.

jneem: The only way I can find to return a double error is `context`, which isn't really appropriate IMO because it's for when one error is caused by another. I added a log, though.

cmyr:
I think this is an improvement but I don't think adding a dependency is the best route long-term; in general I think that crates like anyhow make sense for applications, but less so for frameworks, where they can force downstream crates into a particular error handling path.

There's some discussion of a longer-term error strategy somewhere; but basically I think that each platform should have an internal error type that closely matches whatever the native error type on that platform is, and then druid-shell should have a set of errors for common situations and then should also wrap the platform error type for more detail.

xStrom:
I understand the usual dependency issues like compile time, quality issues etc. However when directly talking about the error types, I'm not sure I follow how `anyhow` will force any specific path.

What difference does it make if `Error::DruidShellOpaqueVariant` is `Arc<anyhow::Error>` or `CustomReimplementationOfAnyhow`? In both cases it implements the `Error` trait and you can print it, wrap it, clone it etc. In neither case can you really get anything more out of it than a string describing the error.

jneem:
Just to elaborate on what @xStrom already said, I agree that we should wrap the native error type (and #989 introduces this for X11), but there are also plenty of internal druid-shell errors for which I don't think we should commit to having a variant in the public API.

cmyr:
/bloat

(curious to know if this works after merge?)

To be fair I haven't read through this carefully, just trying to keep my head above water. I'm certainly happy for any improvements to error stuff in the short term.

jneem:
It doesn't seem to be doing much...

Isn't bloat only for mac, though? Given that I didn't touch the mac backend, I wouldn't be surprised if whatever bloat I introduced just gets optimized out...

cmyr:
yea apparently it won't work if something has been merged. And fair point, that will need reworking at some point. :)

PR #983, opened 2020-05-23
Added a reference to an application using Druid

chris-zen:
This is my work in progress project to build a modular sound synthesizer using Druid for the UI.


chris-zen:
@ForLoveOfCats thanks for your heads up. I already addressed your comments.

PR #989, opened 2020-05-26
X11 present

### LINKED ISSUE #932: X11 shell uses sleep to sync rendering.
From #475:
- Currently we figure out the screen's refresh rate and **sleep** (!) for however long based on the refresh rate, which is probably not what we want to do. Better way to sync draws?
- This might also be what's causing the window to [blink a lot during re-rendering](https://github.com/xi-editor/druid/pull/599#issuecomment-609376693)

The code referred to is [here](https://github.com/xi-editor/druid/blob/1f56a1745dd946a54d05fb2d5e60e7f3d33cbe2f/druid-shell/src/platform/x11/window.rs#L427)

jneem:
Fixes #932 

This adds support for the X11 Present extension, for scheduling drawing, falling back to the old sleep-based method if Present is absent or causes errors.

The main showstopper for now is a bug in [rust-xcb](https://github.com/rtbo/rust-xcb/issues/81) that breaks the present extension. Unfortunately, that project doesn't seem very active -- I see that @xStrom has a PR that's been sitting for a while...

### DIFF
--- a/druid-shell/examples/perftest.rs
+++ b/druid-shell/examples/perftest.rs
@@ -133,7 +133,7 @@ impl WinHandler for PerfTest {
 }
 
 fn main() {
-    env_logger::init();
+    simple_logger::init().expect("Failed to init simple logger");
     let app = Application::new().unwrap();
     let mut builder = WindowBuilder::new(app.clone());
     let perf_test = PerfTest {
--- a/druid-shell/examples/perftest.rs
+++ b/druid-shell/examples/perftest.rs
@@ -133,6 +133,7 @@ impl WinHandler for PerfTest {
 }
 
 fn main() {
+    env_logger::init();
     let app = Application::new().unwrap();
     let mut builder = WindowBuilder::new(app.clone());
     let perf_test = PerfTest {
--- a/druid-shell/src/platform/x11/application.rs
+++ b/druid-shell/src/platform/x11/application.rs
@@ -21,6 +21,8 @@ use std::rc::Rc;
 
 use anyhow::{anyhow, Context, Error};
 use x11rb::connection::Connection;
+use x11rb::protocol::present::ConnectionExt as _;
+use x11rb::protocol::xfixes::ConnectionExt as _;
 use x11rb::protocol::xproto::{ConnectionExt, CreateWindowAux, EventMask, WindowClass};
 use x11rb::protocol::Event;
 use x11rb::xcb_ffi::XCBConnection;
@@ -60,6 +62,8 @@ pub(crate) struct Application {
     window_id: u32,
     /// The mutable `Application` state.
     state: Rc<RefCell<State>>,
+    /// The major opcode of the Present extension, if it is supported.
+    present_opcode: Option<u8>,
 }
 
 /// The mutable `Application` state.
@@ -87,12 +91,67 @@ impl Application {
             quitting: false,
             windows: HashMap::new(),
         }));
+        let present_opcode = match Application::query_present_opcode(&connection) {
+            Ok(p) => p,
+            Err(e) => {
+                log::info!("failed to find Present extension: {}", e);
+                None
+            }
+        };
         Ok(Application {
             connection,
             screen_num: screen_num as i32,
             window_id,
             state,
+            present_opcode,
         })
+    }
+
+    // Check if the Present extension is supported, returning its opcode if it is.
+    fn query_present_opcode(conn: &Rc<XCBConnection>) -> Result<Option<u8>, Error> {
+        let query = conn
+            .query_extension(b"Present")?
+            .reply()
+            .context("query Present extension")?;
+
+        if !query.present {
+            return Ok(None);
+        }
+
+        let opcode = Some(query.major_opcode);
+
+        // If Present is there at all, version 1.0 should be supported. This code
+        // shouldn't have a real effect; it's just a sanity check.
+        let version = conn
+            .present_query_version(1, 0)?
+            .reply()
+            .context("query Present version")?;
+        log::info!(
+            "X server supports Present version {}.{}",
+            version.major_version,
+            version.minor_version,
+        );
+
+        // We need the XFIXES extension to use regions. This code looks like it's just doing a
+        // sanity check but it is *necessary*: XFIXES doesn't work until we've done version
+        // negotiation
+        // (https://www.x.org/releases/X11R7.7/doc/fixesproto/fixesproto.txt)
+        let version = conn
+            .xfixes_query_version(5, 0)?
+            .reply()
+            .context("query XFIXES version")?;
+        log::info!(
+            "X server supports XFIXES version {}.{}",
+            version.major_version,
+            version.minor_version,
+        );
+
+        Ok(opcode)
+    }
+
+    #[inline]
+    pub(crate) fn present_opcode(&self) -> Option<u8> {
+        self.present_opcode
     }
 
     fn create_event_window(conn: &Rc<XCBConnection>, screen_num: i32) -> Result<u32, Error> {
@@ -249,6 +308,27 @@ impl Application {
                     self.finalize_quit();
                 }
             }
+            Event::PresentCompleteNotify(ev) => {
+                let w = self
+                    .window(ev.window)
+                    .context("COMPLETE_NOTIFY - failed to get window")?;
+                w.handle_complete_notify(ev)
+                    .context("COMPLETE_NOTIFY - failed to handle")?;
+            }
+            Event::PresentIdleNotify(ev) => {
+                let w = self
+                    .window(ev.window)
+                    .context("IDLE_NOTIFY - failed to get window")?;
+                w.handle_idle_notify(ev)
+                    .context("IDLE_NOTIFY - failed to handle")?;
+            }
+            // In XCB, errors are reported asynchronously by default, by sending them to the event
+            // loop. You can opt-out of this by using ..._checked variants of a function, and then
+            // getting the error synchronously. We use this in window initialization, but otherwise
+            // we take the async route.
+            Event::Error(e) => {
+                return Err(x11rb::errors::ReplyError::from(e.clone()).into());
+            }
             _ => {}
         }
         Ok(false)
@@ -264,7 +344,7 @@ impl Application {
                         }
                     }
                     Err(e) => {
-                        log::error!("Error handling event: {}", e);
+                        log::error!("Error handling event: {:#}", e);
                     }
                 }
             }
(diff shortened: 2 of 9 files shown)


druid-shell/src/platform/x11/application.rs
```diff
@@ -81,14 +86,65 @@ impl Application {
...
+        log::info!(
+            "X server supports XFIXES version {}.{}",
+            version.major_version(),
+            version.minor_version()
+        );
+
+        Ok(opcode)
+    }
+
+    pub fn present_opcode(&self) -> Option<u8> {
```
xStrom, line 144: Let's add `#[inline]` and a doc comment. Also does this need to be `pub`? Let's make it `pub(crate)`.

druid-shell/src/platform/x11/window.rs
```diff
@@ -615,6 +788,171 @@ impl Window {
...
+        pixmap_id
+    }
+
+    fn create_pixmaps(&mut self, conn: &Rc<Connection>, window_id: u32) {
+        if !self.all_pixmaps.is_empty() {
+            log::error!("BUG: need to free the pixmaps before creating new ones");
+            self.free_pixmaps(conn);
+        }
+
+        // Two buffers is probably enough. We'll create more on demand.
```
xStrom, line 911: Under what circumstances do we need more than two? :thinking:

jneem: As far as I can tell, X doesn't give any guarantees about when it will give us the buffer back. In my testing, I basically always get it back just after vsync; since we present once per vsync, we just alternate between buffers.

One latency-reducing possibility for the future is to render more frequently when we're getting input: if rendering is fast and we get input sometime during the inter-frame time, we could try rendering in the hope that we'll beat the vsync deadline. We might need an extra buffer for this.

xStrom: Okay, then my question would be - are these additional buffers additive or replacements?
So if we have say 10 buffers and we instruct X to render all of these in a row. Does only the last one get rendered, or does this manifest as a 10 frame delay?

If it's extra delay, then we should attempt an optimization solution. If just the last one gets used, then we should add a comment stating that these extra buffers wont cause actual delays.

jneem:
Only the last buffer gets rendered; the others are discarded. I'll add a
comment where the buffers are defined.

jneem:
I'd like to make progress on this, but I'm blocked by some showstoppers in `rust-xcb`, which appears to be inactive...

Is there a good way of making progress there?

Finnerale:
I'm not sure what to do about this...

I suppose we could try to reach out to the maintainer of `xcb` directly and find out if he intends to continue his work on `xcb`, and if not either fork it and maintain it ourselves, or use an alternative ([x11rb](https://github.com/psychon/x11rb) looks interesting).

@xStrom what do you think?

Finnerale:
This might be interesting:
https://github.com/psychon/x11rb/blob/master/doc/comparison.md

jneem:
Hm, yeah. I had a quick look at the docs and they look pretty good (well, the auto-generated code for the extensions is missing docs, but that's not too bad). One downside is that it's fairly new and doesn't have any substantial users on crates.io.

It would also add dependencies on `gethostname` and `nix`, but I guess those aren't a big deal.

psychon:
I never saw anything using the present extension, I think... I kind of expected that one just uses a timer and schedules redraws internally while animations are running.

I did not look at anything in here, but what exactly do you need present for? Things usually use gl when they want vsync'd drawing.

druid-shell/src/platform/x11/window.rs
```diff
@@ -250,6 +260,52 @@ impl WindowBuilder {
...
+            .request_check()
+            {
+                log::info!("Failed to set event mask, disabling Present: {}", e);
+                return None;
+            }
+
+            let region_id = conn.generate_id();
+
+            if let Err(e) =
+                xcb::xfixes::create_region_checked(&conn, region_id, &[]).request_check()
```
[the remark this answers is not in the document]
xStrom: > Creating an empty region should never fail

That's great until it does. Especially given how much diversity there is in Linux-land with implementations. This check isn't really in a hot path, so being more defensive is fine I think.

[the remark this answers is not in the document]
jneem: The general error-handling approach I'm using (which I guess should be written in a comment) is to check everything synchronously when initializing, and then after that to collect errors asynchronously through the event loop.

jneem:
> but what exactly do you need present for? Things usually use gl when they want vsync'd drawing.

The main reason is for vsync, yes. The two options for vsync seem to be present and gl, and present is easier :smile: I found some examples of window managers using present -- they seem to usually have a present path and a gl path, and I just decided to try the present one first.

I was also hoping that present might improve latency (and in particular, give us explicit control over latency). This is more speculative, though, and I don't currently have any way to measure it anyway.

jneem:
Resurrecting this now that #1025 is in.
Items left to do:

- [ ] Be less eager with pixmap resizing
- [ ] Do the pixmaps regardless of whether we're using present (meaning we'll always have double-buffering)
- [ ] Add a way (environment variables?) to disable present for testing

druid-shell/src/platform/x11/application.rs
```diff
@@ -249,6 +308,41 @@ impl Application {
                     self.finalize_quit();
                 }
             }
+            Event::ConfigureNotify(ev) => {
```
xStrom, line 311: Like the above comment says, there needs to be a check against `self.window_id` because this event could also fire for the app `InputOnly` window.

druid-shell/src/platform/x11/window.rs
```diff
@@ -184,24 +186,27 @@ impl WindowBuilder {
...
-            );
+        let cw_values = xproto::CreateWindowAux::new().event_mask(
+            EventMask::Exposure
+                | EventMask::StructureNotify
+                | EventMask::KeyPress
+                | EventMask::KeyRelease
+                | EventMask::ButtonPress
+                | EventMask::ButtonRelease
+                | EventMask::PointerMotion
+                | EventMask::StructureNotify,
```
xStrom, line 205: This is a duplicate `EventMask::StructureNotify` here.

jneem:
Ok, I think this is in better shape now.

In addition to using the present extension, this also improves the non-present path a little, by drawing into a pixmap and then copying out those pixels. (We use the same pixmap-handling code in both paths; the only difference is whether we present or copy.)

I've tested this in several different configurations, and it seems to work without tearing and with very few missed frames. Funny story: I discovered while testing this that my laptop was using generic video drivers, because I never installed the intel ones :neutral_face: . (I did see some tearing with the generic drivers, but they went away with the intel ones.) I also tested on nouveau drivers.

druid-shell/src/platform/x11/window.rs
```diff
@@ -299,6 +398,58 @@ atom_manager! {
...
+    height: u16,
+    /// The depth of the currently allocated pixmaps.
+    depth: u8,
+}
+
+/// The state involved in using X's Present extension.
+#[derive(Debug)]
+struct PresentData {
+    /// A monotonically increasing request counter.
+    serial: u32,
```
Finnerale, line 440: Just to make it slightly more explicit:
```suggestion
    /// A monotonically increasing present request counter.
    serial: u32,
```

druid-shell/examples/perftest.rs
```diff
@@ -133,6 +133,7 @@ impl WinHandler for PerfTest {
 }
 
 fn main() {
+    simple_logger::init().expect("Failed to init simple logger");
```
Finnerale, line 136: I think we can add this to all three examples.

druid-shell/src/platform/x11/application.rs
```diff
@@ -87,14 +91,81 @@ impl Application {
             quitting: false,
             windows: HashMap::new(),
         }));
+
+        let present_opcode = if std::env::var_os("DRUID_SHELL_DISABLE_X11_PRESENT").is_some() {
```
Finnerale, line 95: We should document such toggles somewhere, maybe just in an `# Env` section at the crate root of `druid-shell`.

druid-shell/src/platform/x11/window.rs
```diff
@@ -299,6 +398,58 @@ atom_manager! {
...
+    /// Did we submit a present that hasn't completed yet? If so, this is its serial number.
+    waiting_on: Option<u32>,
+    /// We need to render another frame as soon as the current one is done presenting.
+    needs_present: bool,
+    /// The last MSC that was completed. This can be used to diagnose latency problems, because MSC
+    /// is a frame counter: we should be presenting on every frame, and storing the last completed
+    /// MSC lets us know if we missed one.
+    last_msc: Option<u64>,
+    /// The time at which the last frame was completed.
+    last_ust: Option<u64>,
```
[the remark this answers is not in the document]
jneem: The official docs don't say, but a lwn article suggests "media stream counter" for MSC. UST appears to be in microseconds, at least on my computer, but I still don't know what it stands for.

druid-shell/src/platform/x11/window.rs
```diff
@@ -637,6 +840,214 @@ impl Window {
...
+
+    fn create_pixmaps(
+        &mut self,
+        conn: &Rc<XCBConnection>,
+        window_id: u32,
+        count: usize,
+    ) -> Result<(), Error> {
+        if !self.all_pixmaps.is_empty() {
+            log::error!("BUG: need to free the pixmaps before creating new ones");
+            self.free_pixmaps(conn);
```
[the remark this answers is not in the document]
jneem: Yeah, I guess there's no reason for the error message.

druid-shell/src/platform/x11/window.rs
```diff
@@ -352,45 +502,36 @@ impl Window {
...
-            count: 0,
-            response_type: x11rb::protocol::xproto::EXPOSE_EVENT,
-            sequence: 0,
+    // Ensure that our cairo context is targeting the right drawable, allocating one if necessary.
+    fn update_cairo_surface(&self) -> Result<(), Error> {
+        let mut buffers = borrow_mut!(self.buffers)?;
+        let pixmap = if let Some(p) = buffers.idle_pixmaps.last() {
+            *p
+        } else {
+            buffers.create_pixmap(self.app.connection(), self.id)?
```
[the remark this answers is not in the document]
jneem: Sure, a log makes sense. I'm not sure I understood your second question, though: `buffers.create_pixmap` does add the pixmap to the list of available pixmaps.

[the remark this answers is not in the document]
Finnerale: Oops, I forgot about that :laughing:

jneem:
Seems to be a problem with piet-coregraphics? I'll just wait a bit and see if it gets better...

PR #993, opened 2020-05-26
Statically typed selectors, take two.

Finnerale:
The (hopefully) final part of #908 .

This time it just refactors `Selector` and `Command`.

### DIFF
--- a/druid/src/command.rs
+++ b/druid/src/command.rs
@@ -237,7 +237,7 @@ impl<T> Selector<T> {
         Selector(s, PhantomData)
     }
 
-    pub(crate) const fn symbol(&self) -> SelectorSymbol {
+    pub(crate) const fn symbol(self) -> SelectorSymbol {
         self.0
     }
 }
--- a/druid/src/command.rs
+++ b/druid/src/command.rs
@@ -172,11 +172,10 @@ pub mod sys {
     /// Show the new file dialog.
     pub const NEW_FILE: Selector = Selector::new("druid-builtin.menu-file-new");
 
-    /// System command. A file picker dialog will be shown to the user, and an
-    /// [`OPEN_FILE`] command will be sent if a file is chosen.
+    /// When submitted by the application, a file picker dialog will be shown to the user,
+    /// and an [`OPEN_FILE`] command will be sent if a file is chosen.
     ///
     /// [`OPEN_FILE`]: constant.OPEN_FILE.html
-    /// [`FileDialogOptions`]: ../struct.FileDialogOptions.html
     pub const SHOW_OPEN_PANEL: Selector<FileDialogOptions> =
         Selector::new("druid-builtin.menu-file-open");
 
@@ -185,12 +184,11 @@ pub mod sys {
     /// [`FileInfo`]: ../struct.FileInfo.html
     pub const OPEN_FILE: Selector<FileInfo> = Selector::new("druid-builtin.open-file-path");
 
-    /// Special command. When issued, the system will show the 'save as' panel,
+    /// When submitted by the application, the system will show the 'save as' panel,
     /// and if a path is selected the system will issue a [`SAVE_FILE`] command
     /// with the selected path as the payload.
     ///
     /// [`SAVE_FILE`]: constant.SAVE_FILE.html
-    /// [`FileDialogOptions`]: ../struct.FileDialogOptions.html
     pub const SHOW_SAVE_PANEL: Selector<FileDialogOptions> =
         Selector::new("druid-builtin.menu-file-save-as");
 
--- a/druid/src/command.rs
+++ b/druid/src/command.rs
@@ -260,7 +260,10 @@ impl Command {
 
     /// Used to create a command from the types sent via an `ExtEventSink`.
     pub(crate) fn from_ext(selector: SelectorSymbol, object: Box<dyn Any>) -> Self {
-        Command { selector, object: object.into() }
+        Command {
+            selector,
+            object: object.into(),
+        }
     }
 
     /// Checks if this was created using `selector`.
--- a/druid/src/command.rs
+++ b/druid/src/command.rs
@@ -81,14 +81,14 @@ pub struct Command {
 /// let num = CantClone(42);
 /// let command = Command::new(selector, SingleUse::new(num));
 ///
-/// let object: &SingleUse<CantClone> = command.get_unchecked(selector);
-/// if let Some(num) = object.take() {
+/// let payload: &SingleUse<CantClone> = command.get_unchecked(selector);
+/// if let Some(num) = payload.take() {
 ///     // now you own the data
 ///     assert_eq!(num.0, 42);
 /// }
 ///
 /// // subsequent calls will return `None`
-/// assert!(object.take().is_none());
+/// assert!(payload.take().is_none());
 /// ```
 ///
 /// [`Command`]: struct.Command.html
@@ -390,11 +390,12 @@ impl Into<Option<Target>> for WidgetId {
 #[cfg(test)]
 mod tests {
     use super::*;
+
     #[test]
-    fn get_object() {
+    fn get_payload() {
         let sel = Selector::new("my-selector");
-        let objs = vec![0, 1, 2];
-        let command = Command::new(sel, objs);
+        let payload = vec![0, 1, 2];
+        let command = Command::new(sel, payload);
         assert_eq!(command.get(sel), Some(&vec![0, 1, 2]));
     }
 }
--- a/druid/src/command.rs
+++ b/druid/src/command.rs
@@ -259,8 +259,8 @@ impl Command {
     }
 
     /// Used to create a command from the types sent via an `ExtEventSink`.
-    pub(crate) fn from_ext(selector: SelectorSymbol, object: Arc<dyn Any>) -> Self {
-        Command { selector, object }
+    pub(crate) fn from_ext(selector: SelectorSymbol, object: Box<dyn Any>) -> Self {
+        Command { selector, object: object.into() }
     }
 
     /// Checks if this was created using `selector`.
--- a/druid/src/command.rs
+++ b/druid/src/command.rs
@@ -22,11 +22,14 @@ use std::{
 
 use crate::{WidgetId, WindowId};
 
+/// A [`Selector`]s identity.
+///
+/// [`Selector`]: struct.Selector.html
 pub(crate) type SelectorSymbol = &'static str;
 
 /// An identifier for a particular command.
 ///
-/// The type parameter `T` specifies the commands payload type.
+/// The type parameter `T` specifies the command's payload type.
 ///
 /// This should be a unique string identifier. Certain `Selector`s are defined
 /// by druid, and have special meaning to the framework; these are listed in the
@@ -34,7 +37,7 @@ pub(crate) type SelectorSymbol = &'static str;
 ///
 /// [`druid::commands`]: commands/index.html
 #[derive(Debug, PartialEq, Eq)]
-pub struct Selector<T = ()>(SelectorSymbol, PhantomData<T>);
+pub struct Selector<T = ()>(SelectorSymbol, PhantomData<*const T>);
 
 /// An arbitrary command.
 ///
@@ -42,7 +45,7 @@ pub struct Selector<T = ()>(SelectorSymbol, PhantomData<T>);
 /// and what type of payload it carries, as well as the actual payload.
 ///
 /// If the payload can't or shouldn't be cloned,
-/// wrapping it with [`SingleUse`] allows you to `take` the object.
+/// wrapping it with [`SingleUse`] allows you to `take` the payload.
 ///
 /// # Examples
 /// ```
@@ -55,17 +58,15 @@ pub struct Selector<T = ()>(SelectorSymbol, PhantomData<T>);
 /// assert_eq!(command.get(selector), Some(&vec![1, 3, 10, 12]));
 /// ```
 ///
-/// [`Command::new`]: #method.new
-/// [`Command::get_object`]: #method.get_object
 /// [`SingleUse`]: struct.SingleUse.html
 /// [`Selector`]: struct.Selector.html
 #[derive(Debug, Clone)]
 pub struct Command {
-    selector: SelectorSymbol,
-    object: Arc<dyn Any>,
+    symbol: SelectorSymbol,
+    payload: Arc<dyn Any>,
 }
 
-/// A wrapper type for [`Command`] arguments that should only be used once.
+/// A wrapper type for [`Command`] payloads that should only be used once.
 ///
 /// This is useful if you have some resource that cannot be
 /// cloned, and you wish to send it to another widget.
@@ -111,9 +112,10 @@ pub enum Target {
 ///
 /// [`Command`]: ../struct.Command.html
 pub mod sys {
+    use std::any::Any;
+
     use super::Selector;
     use crate::{FileDialogOptions, FileInfo, SingleUse};
-    use std::any::Any;
 
     /// Quit the running application. This command is handled by the druid library.
     pub const QUIT_APP: Selector = Selector::new("druid-builtin.quit-app");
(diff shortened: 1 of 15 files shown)


druid/src/command.rs
```diff
@@ -172,18 +173,15 @@ pub mod sys {
     /// System command. A file picker dialog will be shown to the user, and an
     /// [`OPEN_FILE`] command will be sent if a file is chosen.
     ///
-    /// The argument should be a [`FileDialogOptions`] struct.
```
xStrom, line 175: I'm not sure that this should be removed. How do developers know that `FileDialogOptions` is the payload type? Should probably reword it from *argument* to *payload* though.

druid/src/command.rs
```diff
@@ -172,18 +173,15 @@ pub mod sys {
...
-    ///
     /// [`OPEN_FILE`]: constant.OPEN_FILE.html
     /// [`FileDialogOptions`]: ../struct.FileDialogOptions.html
-    pub const SHOW_OPEN_PANEL: Selector = Selector::new("druid-builtin.menu-file-open");
+    pub const SHOW_OPEN_PANEL: Selector<FileDialogOptions> =
+        Selector::new("druid-builtin.menu-file-open");
 
-    /// Open a file.
-    ///
-    /// The argument must be a [`FileInfo`] object for the file to be opened.
```
xStrom, line 183: Should probably keep info about `FileInfo` as payload.

druid/src/command.rs
```diff
@@ -228,37 +229,96 @@ pub mod sys {
...
 
+impl<T> Selector<T> {
     /// Create a new `Selector` with the given string.
-    pub const fn new(s: &'static str) -> Selector {
-        Selector(s)
+    pub const fn new(s: &'static str) -> Selector<T> {
+        Selector(s, PhantomData)
+    }
+
+    pub(crate) const fn symbol(self) -> SelectorSymbol {
```
xStrom, line 240: Let's add a short `/// Doc line` and also mark this as `#[inline]`.
Also does this method even need `T`? Can't this live in the regular `impl Selector` block like `NOOP`?

druid/src/command.rs
```diff
@@ -14,25 +14,32 @@
...
-use std::sync::{Arc, Mutex};
+use std::any::{self, Any};
+use std::{
+    marker::PhantomData,
+    sync::{Arc, Mutex},
+};
 
 use crate::{WidgetId, WindowId};
 
+pub(crate) type SelectorSymbol = &'static str;
```
xStrom, line 25: Let's add a short `/// Doc line`.

druid/src/command.rs
```diff
@@ -228,37 +229,96 @@ pub mod sys {
...
+        Selector(s, PhantomData)
+    }
+
+    pub(crate) const fn symbol(self) -> SelectorSymbol {
+        self.0
+    }
+}
+
+impl<T: Any> Selector<T> {
+    pub fn carry(self, object: T) -> Command {
```
xStrom, line 246: We definitely want `/// Doc lines` for all `pub` items. Let's add one here for `carry`.

druid/src/command.rs
```diff
@@ -54,9 +61,8 @@ pub struct Selector(&'static str);
 /// [`Selector`]: struct.Selector.html
 #[derive(Debug, Clone)]
 pub struct Command {
-    /// The command's `Selector`.
-    pub selector: Selector,
-    object: Option<Arc<dyn Any>>,
+    selector: SelectorSymbol,
+    object: Arc<dyn Any>,
```
xStrom, line 65: Can we rename `object` to `payload` to match our docs?

druid/src/command.rs
```diff
@@ -228,37 +229,96 @@ pub mod sys {
...
+
+impl<T: Any> Selector<T> {
+    pub fn carry(self, object: T) -> Command {
+        Command::new(self, object)
     }
 }
 
 impl Command {
     /// Create a new `Command` with an argument. If you do not need
     /// an argument, `Selector` implements `Into<Command>`.
```
xStrom, line 253: The other new docs use the word *payload* which I think is fine. However I think we should be consistent. Let's change all old occurrences of *argument* in relation to commands to *payload*.

druid/src/command.rs
```diff
@@ -54,9 +61,8 @@ pub struct Selector(&'static str);
 /// [`Selector`]: struct.Selector.html
 #[derive(Debug, Clone)]
 pub struct Command {
-    /// The command's `Selector`.
-    pub selector: Selector,
-    object: Option<Arc<dyn Any>>,
+    selector: SelectorSymbol,
```
xStrom, line 64: Can we rename this `selector` to e.g. `selector_symbol`? I think using `selector` to mean two different types can be confusing.

druid/src/command.rs
```diff
@@ -228,37 +229,96 @@ pub mod sys {
...
+    pub fn is<T>(&self, selector: Selector<T>) -> bool {
+        self.selector == selector.symbol()
     }
 
-    /// Return a reference to this `Command`'s object, if it has one.
-    pub fn get_object<T: Any>(&self) -> Result<&T, ArgumentError> {
-        match self.object.as_ref() {
-            Some(o) => o.downcast_ref().ok_or(ArgumentError::IncorrectType),
-            None => Err(ArgumentError::NoArgument),
+    /// Returns `Some(reference)` to this `Command`'s object, if the selector matches.
```
xStrom, line 274: Let's be consistent about wording. *Payload*, *argument*, and *object* all seem to mean the same thing.
Let's pick one (*payload* maybe?) and go with that.

druid/src/command.rs
```diff
@@ -228,37 +229,96 @@ pub mod sys {
...
+    ///
+    /// Panics when the payload has a different type, than what the selector is supposed to carry.
+    /// This can happen when two selectors with different types but the same key are used.
+    ///
+    /// [`get_unchecked`]: #method.get_unchecked
+    pub fn get<T: Any>(&self, selector: Selector<T>) -> Option<&T> {
+        if self.selector == selector.symbol() {
+            Some(self.object.downcast_ref().unwrap_or_else(|| {
+                panic!(
+                    "The selector \"{}\" exists twice with different types. See druid::Command::get_object for more information",
```
xStrom, line 290: `get_object` -> `get`

druid/src/command.rs
```diff
@@ -228,37 +229,96 @@ pub mod sys {
...
+                    selector.symbol()
+                )
+            }))
+        } else {
+            None
         }
     }
+
+    /// Return a reference to this `Command`'s object.
```
xStrom, line 299: Return -> Returns
object -> payload

druid/src/command.rs
```diff
@@ -228,37 +229,96 @@ pub mod sys {
...
+    ///
+    /// If the selector has already been checked, `get_unchecked` can be used safely.
+    /// Otherwise you should either use [`get`] instead, or check the selector using [`is`] first.
```
xStrom, line 302: I would rewrite this as:
```
If the selector has already been checked with [`is`], then `get_unchecked` can be used safely.
Otherwise you should use [`get`] instead.
```

druid/src/command.rs
```diff
@@ -228,37 +229,96 @@ pub mod sys {
...
-    /// Return a reference to this `Command`'s object, if it has one.
-    pub fn get_object<T: Any>(&self) -> Result<&T, ArgumentError> {
-        match self.object.as_ref() {
-            Some(o) => o.downcast_ref().ok_or(ArgumentError::IncorrectType),
-            None => Err(ArgumentError::NoArgument),
+    /// Returns `Some(reference)` to this `Command`'s object, if the selector matches.
+    ///
+    /// Returns `None` when `self.is(selector) == false`.
+    ///
+    /// If the selector has already been checked, [`get_unchecked`] can be used safely.
```
xStrom, line 278: There's not really a scenario where you would `get` and then `get_unchecked` right? So I would write:
```
Alternatively you can check the selector with [`is`] and then use [`get_unchecked`].
```

druid/src/ext_event.rs
```diff
@@ -90,8 +90,7 @@ impl ExtEventSink {
     /// instead you have to pass the [`Selector`] and the (optional) argument
     /// separately, and it will be turned into a `Command` when it is received.
     ///
-    /// The `obj` argument can be any type which implements `Any + Send`, or `None`
-    /// if this command has no argument.
+    /// The `object` argument must implement `Any + Send + Sync`.
```
xStrom, line 93: `object` argument -> `payload`
Also similar change for surrounding docs/variables.

druid/src/command.rs
Finnerale: Sounds good. I went with just `symbol` as its meaning is unique in the `Command` context.

druid/src/command.rs
```diff
@@ -14,25 +14,32 @@
...
+///
 /// This should be a unique string identifier. Certain `Selector`s are defined
 /// by druid, and have special meaning to the framework; these are listed in the
 /// [`druid::commands`] module.
 ///
 /// [`druid::commands`]: commands/index.html
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub struct Selector(&'static str);
+#[derive(Debug, PartialEq, Eq)]
+pub struct Selector<T = ()>(SelectorSymbol, PhantomData<T>);
```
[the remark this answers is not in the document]
Finnerale: Thanks, I forgot about this one.

druid/src/command.rs
Finnerale: The 'regular' `impl Selector` block is actually the `impl Selector<()>` block, just with the default generic. I've made that generic explicit now, to avoid further confusion.

druid/src/command.rs
Finnerale: I think payload is good, I'll update everything to it.

druid/src/command.rs
Finnerale: Adding `#[inline]` to non-pub functions has no effect, to my knowledge.

druid/src/command.rs
Finnerale: The type is visible in the docs, auto completion (RustAnalyzer) and you will get an error if you pass anything else. I think that is enough.

druid/src/command.rs
Finnerale: Same as above.

Finnerale:
@xStrom I think I've addressed all your points.

What I'm not sure about yet: Should I add changelog entries for the methods I've added or renamed, or should it all be compressed to one, less specific entry (as it is now)?
Maybe change it to "*Refactored `Command` and `Selector` to be statically typed, similarly to `Env` and `Key`.*"  to emphasis that there are more changes to it?

I did not rebase it yet, as it will break Github's review history, but if it is ok with you @xStrom , then I'll do the rebase.

druid/src/command.rs
xStrom: Check out [the assembly output here](https://rust.godbolt.org/z/8z2f87). Then remove that `#[inline(always)]` tag and see it grow. So the inline tags definely work on non-pub functions.

The regular `#[inline]` is just a hint, not forced. Did you mean that the hint is ignored for non-pub? That would still be surprising to me though.

druid/src/command.rs
```diff
@@ -14,28 +14,38 @@
...
-use std::sync::{Arc, Mutex};
+use std::any::{self, Any};
+use std::{
+    marker::PhantomData,
+    sync::{Arc, Mutex},
+};
 
 use crate::{WidgetId, WindowId};
 
+/// A [`Selector`]s identity.
```
xStrom, line 25: Let's rewrite this as the following so we don't have the extra **s** outside the code block.
```
The identity of a [`Selector`].
```

druid/src/command.rs
```diff
@@ -225,40 +224,108 @@ pub mod sys {
...
+    pub fn carry(self, payload: T) -> Command {
+        Command::new(self, payload)
     }
 }
 
 impl Command {
-    /// Create a new `Command` with an argument. If you do not need
-    /// an argument, `Selector` implements `Into<Command>`.
-    pub fn new(selector: Selector, arg: impl Any) -> Self {
+    /// Create a new `Command` with payload.
```
xStrom, line 254: with **a** payload

druid/src/command.rs
```diff
@@ -225,40 +224,108 @@ pub mod sys {
...
+    ///
+    /// [`Selector::carry`] can be used to create `Command`s more conveniently.
+    ///
+    /// If you do not need an payload, `Selector` implements `Into<Command>`.
```
xStrom, line 258: ~an~ a
Also lets make the `Selector` a link to the struct.

druid/src/command.rs
Finnerale: Ah yes, with `#[inline(always)]` it does make a difference indeed. But if you try that very same example with `#[inline]` you will see it is the same as without `#[inline]`. `#[inline]` only has an effect when the method is used from another crate.
My assumption is based on this [internals thread](https://internals.rust-lang.org/t/when-should-i-use-inline/598) and [stack overflow question](https://stackoverflow.com/questions/37639276/when-should-inline-be-used-in-rust).

xStrom: It's the same output because optimizations aren't turned on. That example program as fully optimized does nothing, so it's trickier to show that.

The `#[inline]` flag translates into `inlinehint` for LLVM. It's an extra hint for LLVM to consider when deciding whether to inline something. This specific method here will probably be inlined anyway because it's so small.

So I'm fine if you don't want to include the `#[inline]` because it's probably not needed. However the flag definitely works and gets sent to LLVM.

xStrom: Not having `#[inline]` should also improve compile times, because `rustc` does more than just add the LLVM `inlinehint` - [it will codegen into every invocation point](https://github.com/rust-lang/hashbrown/pull/119#issuecomment-537539046).

druid/src/command.rs
```diff
@@ -225,40 +224,109 @@ pub mod sys {
...
+    pub(crate) const fn symbol(self) -> SelectorSymbol {
+        self.0
+    }
+}
+
+impl<T: Any> Selector<T> {
+    /// Convenience method for [`Command::new`] with this selector.
+    ///
+    /// [`Command::new`]: struct.Command.html#method.new
+    pub fn carry(self, payload: T) -> Command {
```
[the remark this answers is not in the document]
xStrom: `command` seems fine but I'm thinking is there any value in using `into_command` given that it does a conversion?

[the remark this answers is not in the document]
cmyr: I think not, because it takes an argument.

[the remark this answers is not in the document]
cmyr: (and it would be `to_`, since self is `Copy`)

druid/src/command.rs
Finnerale: Interesting, thanks! I prefer to leave that inline decision to Rust.

druid/src/command.rs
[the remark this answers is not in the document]
Finnerale: How about just `with(payload)`?
I would like the name to stay rather short and `with` seems to read nicer than `command`.
```rust
submit_command(SHOW_SAVE_PANEL.carry(options));
submit_command(SHOW_SAVE_PANEL.with(options));
submit_command(SHOW_SAVE_PANEL.command(options));
```

druid/src/command.rs
```diff
@@ -225,40 +224,109 @@ pub mod sys {
...
+            payload: Arc::new(payload),
         }
     }
 
     /// Used to create a command from the types sent via an `ExtEventSink`.
-    pub(crate) fn from_ext(selector: Selector, object: Option<Box<dyn Any + Send>>) -> Self {
-        let object: Option<Box<dyn Any>> = object.map(|obj| obj as Box<dyn Any>);
-        let object = object.map(|o| o.into());
-        Command { selector, object }
+    pub(crate) fn from_ext(symbol: SelectorSymbol, payload: Box<dyn Any>) -> Self {
```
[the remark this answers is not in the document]
Finnerale: Yes, that won't allocate anything, so no benefit in adding `Option` here.

druid/src/command.rs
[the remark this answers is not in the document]
xStrom: I like the look of `with`.

Finnerale:
I've renamed `carry` to `with`.

PR #994, opened 2020-05-28
Fix macOS main menu not working before refocus.

### LINKED ISSUE #753: Regression in menu options during initial window launch.
Menu options are not accessible when the window is launched initially. If the window loses and regains focus, the menu items are accessible. This is the case for all the examples and tested on macOS.

An example and discussion can be found in [zulip chat](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/druid-chatter/near/191421810).

As noted by `Colin Rofls` this is a regression.

xStrom:
This PR fixes the issue on macOS where the application menu can't be interacted with until the window loses and regains focus.

There are some pretty wild workarounds [out there](https://ar.al/2018/09/17/workaround-for-unclickable-app-menu-bug-with-window.makekeyandorderfront-and-nsapp.activate-on-macos/) for this, like setting a timer to automatically refocus the window. However as shown in this PR it can be done in an easier fashion. From the surface it seems to be mostly about not letting the app activate before the menu is all set up.

Fixes #753.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -116,6 +116,7 @@ This means that druid no longer requires cairo on macOS and uses Core Graphics i
 - Wheel events now properly update hot state. ([#951] by [@xStrom])
 - X11: Support mouse scrolling. ([#961] by [@jneem])
 - `Painter` now properly repaints on data change in `Container`. ([#991] by [@cmyr])
+- macOS: The application menu is now immediately interactable after launch. ([#994] by [@xStrom])
 
 ### Visual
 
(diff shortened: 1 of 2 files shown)


cmyr:
I'm confused by this; I haven't been able to reproduce the original bug? When the application launches, I'm able to interact with the menus?

xStrom:
Are you macOS Catalina? It's 100% reproduction for me on Catalina.

cmyr:
no I'm on 10.14.6

xStrom:
I'm not fully sure if Catalina is the cause here, but I did see one person mention it [on Stack Overflow](https://stackoverflow.com/a/61006506/138826) that for them the menu issue appeared in Catalina and they needed a new fix.

@totsteps could you check if this PR here fixes the menu issue for you?

totsteps:
I double-checked with the current master, menu if not accessible when running examples!

I can confirm that this PR fixes issue.

PR #995, opened 2020-05-28
Update to kurbo 0.6.1

cmyr:
Suppose this isn't super necessary, but 🤷

### DIFF
--- a/druid-shell/Cargo.toml
+++ b/druid-shell/Cargo.toml
@@ -19,7 +19,7 @@ default-target = "x86_64-pc-windows-msvc"
 # NOTE: When changing the piet or kurbo versions, ensure that
 #       the kurbo version included in piet is compatible with the kurbo version specified here.
 piet-common = "0.1.1"
-kurbo = "0.6.0"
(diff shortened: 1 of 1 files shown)


druid-shell/Cargo.toml
```diff
@@ -19,7 +19,7 @@ default-target = "x86_64-pc-windows-msvc"
 # NOTE: When changing the piet or kurbo versions, ensure that
 #       the kurbo version included in piet is compatible with the kurbo version specified here.
 piet-common = "0.1.1"
-kurbo = "0.6.0"
+kurbo = "0.6.1"
```
raphlinus, line 22: Do we need to directly import kurbo still, or can we use `piet::kurbo`?

xStrom: We don't need to, no. I think the thinking was that keeping the import here will allow for easier bumps if druid wants some new feature, without having to update piet.

cmyr: this is a way to bump our minimum kurbo version without having to do a piet release.

PR #996, opened 2020-05-29
Standardize Text type in Contexts

cmyr:
A long standing pet-peeve of mine: Previously there were a
number of different lifetypes associated with the text factory
depending on where it came from.

This standardizes that, so that all contexts' text() method returns
an owned PietText handle.

This has the added advantage of getting rid of the weird lifetime
in LayoutContext, which will also unblock my macro work.

This is a breaking change in a fairly minor way; `LayoutCtx` returns `Text` now instead of `&mut Text`.


CHANGELOG.md
```diff
@@ -73,6 +73,8 @@ This means that druid no longer requires cairo on macOS and uses Core Graphics i
 - Replaced `Command::one_shot` and `::take_object` with a `SingleUse` payload wrapper type. ([#959] by [@finnerale])
 - Renamed `WidgetPod` methods: `paint` to `paint_raw`, `paint_with_offset` to `paint`, `paint_with_offset_always` to `paint_always`. ([#980] by [@totsteps])
 - `Command` and `Selector` have been reworked and are now statically typed, similarly to `Env` and `Key`. ([#993] by [@finnerale])
+- Standardize the type returned by the contexts' `text()` methods. (#[996] by
```
Finnerale, line 76: The PR link is missing.

cmyr:
as far as I understand the reason we needed the lifetime previously is because we held onto a reference to the type, and so we had to be explicit about its inner lifetime; but now that we just hand one out (not a reference) we can elide it? Still feels weird though

PR #1001, opened 2020-05-30
Refactor WidgetPod::event.

xStrom:
This PR is just the refactoring part of #986 so that we can get the refactoring in first and make the command targeting work more clear.

This new approach is more like `WidgetPod::lifecycle`. It doesn't copy/clone as much and is fewer lines of code allowing for easier understanding.

### DIFF
--- a/druid/src/core.rs
+++ b/druid/src/core.rs
@@ -37,7 +37,7 @@ pub(crate) type CommandQueue = VecDeque<(Target, Command)>;
 /// Generally, container widgets don't contain other widgets directly,
 /// but rather contain a `WidgetPod`, which has additional state needed
 /// for layout and for the widget to participate in event flow.
-/// 
+///
 /// `WidgetPod` will translate internal druid events to regular events,
 /// synthesize additional events of interest, and stop propagation when it makes sense.
 ///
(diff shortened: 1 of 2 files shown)


druid/src/core.rs
```diff
@@ -558,159 +561,168 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
...
-            is_handled: false,
-            is_root: false,
-        };
+        let rect = self.state.layout_rect.unwrap_or_default();
 
-        let rect = child_ctx.widget_state.layout_rect.unwrap_or_default();
+        // If we need to replace either the event or its data.
+        let mut modified_event = None;
+        // Tracking whether we need to merge our state up to the parent.
+        let mut needs_merge = false;
```
cmyr, line 569: is this an optimization (my assumption), or are there times when merging would be incorrect?

if an optimization: my gut feeling is that merging is only a trivial amount of work in the general case, such that this optimization is unlikely to be meaningful even over very large graphs, and I'm not sure it's something it is worth adding more logic to handle, at least without being guided by a concrete performance concern?

xStrom:
I don' think the `needs_merge` adds any significant complexity, but I also don't think it's a significant performance win right now. So I removed it for clarity and added a comment about this situation.

PR #1002, opened 2020-05-31
Increase versions in preparation for 0.6.0.

xStrom:
This PR takes another step towards being ready to publish `0.6.0`.

- The `druid` and `druid-shell` versions were already bumped to `0.6.0`.
- The `druid-derive` version was also already bumped but needlessly far to `0.4.0`. There haven't been any changes or additions since `0.3.0`. Just a few test updates and minor version bumps of dependencies. Thus `0.3.1` is the more logical step according to semver.
- Documentation links/references were updated to `0.6.0`.
- I added a short section to `CONTRIBUTING.md` describing what I did so that the process is more accessible.

- I simplified the two git imports we mention in our documentation.
    * One also specified the version, but according to [The Cargo Book](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations) `version` is ignored when using `git`. It's there so that when you publish your package, that specified version is used instead of the git address.
    * One also specified the branch, but according to [The Cargo Book](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories) cargo will default to `master` anyway, so that is also just noise for beginners.

---

Going forward, I also think we shouldn't bump version numbers until we're close to releasing. As the `druid-derive` `0.4.0` vs `0.3.1` situation shows we can't predict how many changes will be there. Also we might want to do a quick minor release with some bug fixes shortly after a major release, and that would also clash with an immediate major release bump that we've done way ahead of time.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,6 +25,16 @@ then there might be a chance to help out with the preparation of new releases.
 Whether you're new or have prepared druid releases many times already,
 it helps to follow a checklist of what needs to be done. This is that list.
 
+### Increasing the versions
+
+The `druid`, `druid-shell`, and `druid-derive` `Cargo.toml` files need to be updated.
+The `version` field needs to be increased to the next [semver] version that makes sense.
+These packages all also import eachother and those cross-dependency versions need updating too.
+
+You should also search for the previous version number across the whole workspace
+to find any other references that might need updating. There are for example plenty of links
+to specific version documentation that will need updating.
+
 ### Changelog
 
 - Add a new *Unreleased* section by copying the current one.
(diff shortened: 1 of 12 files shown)


cmyr:
since druid-derive depends on druid, I think it needs to go up a a full breaking version anytime druid does?

xStrom:
`druid-derive` doesn't depend on `druid`, or that is to say it is a dev dependency. That doesn't matter for people using `druid-derive`, as it never comes into the picture. This only matters if you want to run the tests of `druid-derive`.

Also, dependency major versions can be bumped without increasing a major version of a library itself. It's the behavior and API of the library itself that matters for semver. In the case of `druid-derive`, `druid` doesn't even come into the picture, because it's only for tests.

PR #1003, opened 2020-06-01
Spinner widget

futurepaul:
This is pretty basic except for the spinning math, which I wish I'd commented while I was writing it!

My layout strategy for this is to only match its container size if that container has a constrained height and width. Otherwise it uses the default widget height key to size itself. Does that sound reasonable?

ezgif-6-18988e9ac782

### DIFF
--- a/druid/src/widget/spinner.rs
+++ b/druid/src/widget/spinner.rs
@@ -1,0 +1,126 @@
+// Copyright 2020 The xi-editor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! An animated spinner widget.
+
+use std::f64::consts::PI;
+
+use druid::kurbo::Line;
+use druid::widget::prelude::*;
+use druid::{theme, Color, Data, KeyOrValue, Point, Vec2};
+
+/// An animated spinner widget for showing a loading state.
+///
+/// To customize the spinner's size, you can place it inside a [`SizedBox`]
+/// that has a fixed width and height.
+///
+/// [`SizedBox`]: struct.SizedBox.html
+pub struct Spinner {
+    t: f64,
+    color: KeyOrValue<Color>,
+}
+
+impl Spinner {
+    /// Create a spinner widget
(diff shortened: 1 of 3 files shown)


druid/src/widget/spinner.rs
```diff
@@ -0,0 +1,132 @@
...
+        self.color = color.into();
+        self
+    }
+
+    /// Set the spinner's color.
+    ///
+    /// The argument can be either a `Color` or a [`Key<Color>`].
+    ///
+    /// [`Key<Color>`]: ../struct.Key.html
+    pub fn set_color(&mut self, color: impl Into<KeyOrValue<Color>>) {
```
xStrom, line 58: These two color methods should mention in their docs that the alpha channel is ignored.

druid/src/widget/spinner.rs
```diff
@@ -0,0 +1,132 @@
...
+
+    fn paint(&mut self, ctx: &mut PaintCtx, _data: &T, env: &Env) {
+        let t = self.t;
+        let (width, height) = (ctx.size().width, ctx.size().height);
+        let center = Point::new(width / 2.0, height / 2.0);
+        let (r, g, b, _) = Color::as_rgba(&self.color.resolve(env));
+        let scale_factor = width.min(height) / 40.0;
+
+        for step in 1..=12 {
+            ctx.paint_with_z_index(1, move |ctx| {
```
xStrom, line 115: Why is the `paint_with_z_index` needed?

druid/src/widget/spinner.rs
```diff
@@ -0,0 +1,132 @@
...
+    ///
+    /// [`Key<Color>`]: ../struct.Key.html
+    pub fn set_color(&mut self, color: impl Into<KeyOrValue<Color>>) {
+        self.color = color.into();
+    }
+}
+
+impl Default for Spinner {
+    fn default() -> Self {
+        Spinner::new()
```
xStrom, line 65: Can't we use `theme::LABEL_COLOR.into(),` in the `Default` implementation?

druid/src/widget/spinner.rs
futurepaul: Ah that was an oversight. Fixed it to use the color's alpha.

druid/src/widget/spinner.rs
futurepaul: Not sure what you're proposing here.

xStrom: So the reason you got that clippy warning was because the `new` function doesn't take any arguments and could be equal to `default`. So what I'm thinking is that we could do the following:

```rust
impl Spinner {
    /// Create a spinner widget
    pub fn new() -> Spinner {
        Spinner::default()
    }
}

impl Default for Spinner {
    fn default() -> Self {
        Spinner {
            t: 0.0,
            color: theme::LABEL_COLOR.into(),
        }
    }
}
```

futurepaul: Ah, that makes sense! Thanks for explaining.

PR #1004, opened 2020-06-01
Don't panic in `Command::get`

jneem:
If a selector's parameter itself has a generic parameter, then prior to
this change there was no panic-free way to get its payload.

### DIFF
--- a/druid/src/command.rs
+++ b/druid/src/command.rs
@@ -282,26 +282,26 @@ impl Command {
         self.symbol == selector.symbol()
     }
 
-    /// Returns `Some(&T)` (this `Command`'s payload) if the selector matches.
+    /// Returns `Some(&T)` (this `Command`'s payload) if the selector matches, and `None` if it
+    /// does not.
     ///
-    /// Returns `None` when `self.is(selector) == false`.
+    /// In most cases, if `self.is(selector)` returns `true` then this method will return
+    /// `Some(&T)`. The exception to this is if there are two selectors with the same symbol, but
+    /// different types.
     ///
     /// Alternatively you can check the selector with [`is`] and then use [`get_unchecked`].
-    ///
-    /// # Panics
-    ///
-    /// Panics when the payload has a different type, than what the selector is supposed to carry.
-    /// This can happen when two selectors with different types but the same key are used.
     ///
     /// [`get_unchecked`]: #method.get_unchecked
     pub fn get<T: Any>(&self, selector: Selector<T>) -> Option<&T> {
         if self.symbol == selector.symbol() {
-            Some(self.payload.downcast_ref().unwrap_or_else(|| {
-                panic!(
+            let ret = self.payload.downcast_ref();
+            if ret.is_none() {
+                log::error!(
                     "The selector \"{}\" exists twice with different types. See druid::Command::get for more information",
                     selector.symbol()
-                )
-            }))
+                );
+            }
+            ret
         } else {
             None
         }


jneem:
I'm very much enjoying the new typed command stuff (thanks @Finnerale !), but this is one place that I think is a regression: there's no panic-free way to get a generic payload. That is, if I have a `Selector<Option<T>>` then there's no way to try to get, say, an `Option<i32>` out of the corresponding command.

This turns up in my attempts to integrate @cmyr's modal stuff, because there we want to be able to send a command with a `Box<dyn Widget<T>>` payload.

Finnerale:
Yes, I'm aware of generic commands, they gave me a bit of a headache as we use them internally.

Considering the example you gave, I think you are running into the exact same problem I did with `NEW_WINDOW` and the menu commands. In druid, I used `Selector<Box<dyn Any>>` to transport the generic payload and utility functions (`EventCtx::new_window`) to construct them. It would be nice if there could be a bit more type information, but I don't think Rust can offer that right now (or in the near future).

I do not want to change the behavior of `Command::get` tough. If this kind of behavior is needed (would be good to see the code in question), then I would rather opt for a new `try_get` or `get_generic` method with a more specific `Result<T, Error>` return type.

xStrom:
Thinking about the naming here and how Rust stdlib methods are named, maybe we should do some renaming?
* `get -> Option<&T>` to `try_get -> Result<&T, Error>`
* `get_unchecked -> &T` to `get -> &T`

Then the `try_get` shouldn't panic, but should return an error instead. This would mirror things like `borrow` / `try_borrow` etc.

---

I think it would be a good idea to add a simple test case here to show the capability that is enabled by not panicking.

Finnerale:
@xStrom Sorry, I can't quite follow. The current naming is based on what [`Vec`](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.get) does in Rust.
- `get` returns `Option<&T>`
- `get_unchecked` returns `&T`

A `try_get` does not exist in `std` to my knowledge, but seems to fit well with similar methods such as `try_from`.

I don't think renaming things makes sens.

xStrom:
I was thinking of [RefCell](https://doc.rust-lang.org/stable/std/cell/struct.RefCell.html#method.try_borrow) and friends. `Command` isn't really a collection, as it contains only one payload.

Even if not renaming things, I think not panicking is a really good property. Logging an error instead doesn't seem that great though. A `Result` seems best.

Finnerale:
Renaming those methods would mean changing the expectations and capabilities of the API as well.
Currently there are two types of failure: expecting a different symbol and having one symbol associated with two types. The current API only cares about the first because the second is caused by incorrect usage of the API and thus causes a panic. This design is the same as what we have already with `Env` and `Key`, if a `Key` is used twice with different types it panics.

And when sticking to these API constraints, then renaming `get` and `get_unchecked` to `get` and `try_get` would mean defaulting to the more likely to panic path. This of corse would not be new in Rust, as you mentioned `RefCell` does that. But I prefer the design of `Vec`, where the safer choice is the default, and the user has to opt-in to the panic.

So yes, not being able to associate generic types with one selector is a 'regression`, but it was a regression by design, after all changing that was the whole point of typed selectors.

xStrom:
Not renaming is fine. I'm more concerned about the panic.

Also, even if we're not going to remove the panic, then at minimum we would do well to address the generic payload issue in documentation. Give developers some solution, with an example of `dyn Any` or whatever.

Finnerale:
Yes, that should definitely be done. That's also why I was curious about the specific usecase @jneem was talking about, right now the only usecase I've had were the druid internal commands, something a user of druid would not really have to deal with. Maybe there is a better solution than `Box<dyn Any>`.

jneem:
To elaborate on the use-case, the envisioned way of creating a modal dialog is to call `ctx.submit_command(SHOW_MODAL.with(modal_widget))`, where `modal_widget` is more-or-less a `Box<dyn Widget<AppState>>` for `AppState` being the same type as the parameter of the containing `Window`. (I say "more-or-less" because the implementation [here](https://github.com/linebender/runebender/blob/master/src/widgets/modal_host.rs) passes a closure instead, while the implementation [here](https://github.com/jneem/druid/blob/modal/druid/src/modal.rs) uses a struct with some additional parameters, but the idea is the same.)

You can see my definition of `SHOW_MODAL` [here](https://github.com/jneem/druid/blob/072ef4904045acdfca1d90f8e0b38c6e3bbed201/druid/src/modal.rs#L58), and although I know that typed commands can't provide full type-safety in the presence of generic parameters, I still think that's the best type signature for it. I realize now that I could better avoid panics by having it be a `Selector<Any>`, but that seems less self-documenting. I'd ideally like to have partial type-safety and freedom from panics at the same time.

xStrom:
I think this `SHOW_MODAL` command should be internal and developers would use an actual method on a context. That method can then under the hood submit the command. Window creation works like that too.

jneem:
I can definitely do that, but I don't think it solves the panic issue. The context method would presumably be something like `EventCtx::show_modal<T>(&mut self, modal: Modal<T>)`, where `T` would need to be unconstrained (or at most `T: Any`) because in the widget that we're calling `show_modal`, the compiler doesn't know the type parameter of the underlying `Window`. Hence, there is still the possibility for the user to make a type error, leading to a panic.

Finnerale:
@jneem Both, `new_window` and `set_menu`, have the same behavior. They panic in debug mode and log a warning when doing a release build.

And I think doing what I did with `new_window` and friends would be the best choice for what you intend to do (it's pretty much the same problem after all). In general I feel like having concrete methods for such APIs is nicer than using commands directly.

jneem:
Oh, I didn't see the check for debug_assertions. Then would a patch to enable the same behavior in `Command::get` be acceptable? i.e, panic only when debug_assertions is enabled, and log otherwise?

Finnerale:
I theory that would be nice, but it won't be possible for `Command::get_unchecked`, and would not help with your problem.

Actually I don't think we can make commands any safer as long as the user has to pick their symbol.
If we want commands to be perfectly safe (and `Env` as well), we would have to find a way to enforce unique symbols.

This could be done using a macro + `TypeId` but it is not super pretty:
```rust
const CMD: Selector = Selector::new(unique_id!());
```
which would become
```rust
const CMD: Selector = Selector::new({
    struct Id;
    TypeId::of::<Id>()
});
```
I think that would work (I've used this approach in the past).

Another option would be a static, mutable counter that is used to initialize selector ids at runtime, but that does not seem like a good solution either.

jneem:
I'm less bothered by panics in `get_unchecked`, because the name is inherently scarier :smile:

And I agree that it wouldn't solve my current problem, which indeed becomes much better with dedicated methods on context (thanks for the suggestion!). But I still think it's worthwhile reducing panics in druid...

Personally, I think auto-generated selector ids would be awesome! I can't think of any reason that I'd want to choose my own...

xStrom:
Regarding auto-generated id's, indeed, what exactly is the reason for manual ids? Also with auto-generated id's we could be more efficient and just use `u64`. I think this id-generation question deserves some attention regardless of whatever else happens here.

Finnerale:
I would guess they have to be chosen manually right now, because that's far easier to implement.

As I've mentioned in my last comment, I've had this issue before, and to my knowledge `TypeId`  is the only reliable option. But once 0.6 is done I'll see if something better came up recently (my experience with this was about a year ago).

I wanted to tackle this issue anyways, guess I'll do it sooner rather than later.

xStrom:
Yeah trying some new more ergonomic automatic strategies sounds good.

Is the issue that it's a `const fn` and so we can't just do the following?
```rust
impl<T> Selector<T> {
    pub const fn new() -> Selector<T> {
        Selector(get_next_id(), PhantomData)
    }
}
```

cmyr:
Feels like there are two things happening here?

The first is we need a general way of supporting payloads that might have generics. I have had to deal with this in runebender, and I have used this solution: https://github.com/linebender/runebender/pull/121/files#diff-2a7182925f2b0adc78b721518cadbe34R38

that is, I've defined the selector in a place where `T` is known, and the selector can always be referenced in that way.

I wonder if something like this is possible as a general solution?

For instance, for menus and window descs, we could require all druid apps to include a 'header' that looked like,

```rust

// in druid:
pub struct AppType<T> {
   pub const SET_MENU: Selector<T> = Selector::new("set-a-typed-menu");
}

// in main.rs
pub const MyApp = AppType<MyData>;

// when creating a menu
ctx.send_command(MyApp::SET_MENU.with(my_menu), None);
```

or something? just spitballing.

The second problem is panicking if a selector has the wrong type. I don't hate this if it would only happen in case of severe error, like reusing a selector id; I would think of it as being like indexing oob in an array.

Finnerale:
@xStrom Yes, with `Selector` being const, we can not just use something like an `AtomicUsize` counter for the id, at least not any time soon (Rust tells me it could be used as `const` on nightly).

@cmyr Generic payloads are tricky, I have a few ideas but nothing satisfying yet. The more I think about it, the less hope I have we can avoid the `Box<Any>` :( But maybe we can find a good pattern to deal with that.

jneem:
I think that `Selector<MenuDesc<T>>` would be preferable to `Selector<Box<Any>>`, even if it can never be completely compiler-checked.

@cmyr, in your example is the idea that there would be some compile-time mechanism to prevent 

```rust
ctx.send_command(AppType::<MyOldDataThatIForgotWhileRefactoring>::SET_MENU.with(my_menu), None)
```
? Or is the idea just that by having a convenient pattern, we make that less likely?

Anyway, I'm going to go ahead and close this PR, and replace it with an issue.

cmyr:
Oops sorry about this: yea, the idea is that you would only use `AppType` through some global alias that would have the same data as whatever you passed to `launch()`.

PR #1006, opened 2020-06-01
Prepare the changelog for 0.6.0.

xStrom:
I cleaned up the changelog in preparation for release.

* Given how much the context method availability changed, I combined all those changes under a single entry.
* I moved all X11 related entries under the highlight, because this is an unreleased feature and so specific fixes aren't that relevant to developers.
* I added a new *Examples* section, because we have a few of those and they don't quite fit under *Docs*.
* I reordered some of the entries for better readibility, grouping entries by their first word and/or theme.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,24 +1,54 @@
 # Changelog
 
+The latest published druid release is [0.6.0](#060---2020-06-01) which was released on 2020-06-01.
+You can find its changes [documented below](#060---2020-06-01).
+
 ## [Unreleased]
 
 ### Highlights
 
-#### Basic X11 backend for druid-shell. ([#599])
+### Added
 
-[@crsaracco] has implemented basic support to run druid on bare-metal X11!
+### Changed
+
+### Deprecated
+
+### Removed
+
+### Fixed
+
+### Visual
+
+### Docs
+
+### Examples
+
+### Maintenance
+
+### Outside News
+
+## [0.6.0] - 2020-06-01
+
+### Highlights
+
+#### X11 backend for druid-shell.
+
+[@crsaracco] got us started and implemented basic support to run druid on bare-metal X11 in [#599].
+Additional features got fleshed out in [#894] and [#900] by [@xStrom]
+and in [#920], [#961], and [#982] by [@jneem].
 
 While still incomplete this lays the foundation for running druid on Linux without relying on GTK.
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -28,96 +77,83 @@ This means that druid no longer requires cairo on macOS and uses Core Graphics i
...
 - Window title can be any `LabelText` (such as a simple `String`). ([#869] by [@cmyr])
 - `Label::with_font` and `set_font`. ([#785] by [@thecodewarrior])
-- `InternalEvent::RouteTimer` to route timer events. ([#831] by [@sjoshid])
 - `MouseEvent` now has a `focus` field which is `true` with window focusing left clicks on macOS. ([#842] by [@xStrom])
 - `MouseButtons` to `MouseEvent` to track which buttons are being held down during an event. ([#843] by [@xStrom])
 - `Env` and `Key` gained methods for inspecting an `Env` at runtime ([#880] by [@Zarenor])
-- `UpdateCtx::request_timer` and `UpdateCtx::request_anim_frame`. ([#898] by [@finnerale])
-- `LifeCycleCtx::request_timer`. ([#954] by [@xStrom])
 - `scale` method to `WinHandler`. ([#904] by [@xStrom])
 - `WinHandler::scale` method to inform of scale changes. ([#904] by [@xStrom])
```
Finnerale, line 96: Those seem to be the same.

xStrom: Good catch. Some sort of rebase failure of the original PR I guess. I can't even see where the shorter line exactly got added.

xStrom:
I improved the above-the-fold experience.

CHANGELOG.md
```diff
@@ -1,24 +1,54 @@
 # Changelog
 
+The latest published druid release is [0.6.0](#060---2020-06-01) which was released on 2020-06-01.
+You can find its changes [documented below](#060---2020-06-01).
```
Finnerale, line 4: Maybe a forced line break would look nicer?

xStrom: I kind of like how it all fits on a single line. However I pushed a line break commit. See if you like it more now.

Finnerale: Oh, I see. On my laptop it is also in one line, which looks nice.

I checked it out on my tablet before, where the last word was on the next line, even though Github had space left on both sides :(.

Considering that most people will probably read it on their PC, having all in on line seems nicer indeed. But thanks for giving this a shot.

PR #1008, opened 2020-06-01
fix broken link in documentation

covercash2:
this is just a quick fix so that `cargo doc` will pass the linter.


Finnerale:
Thanks for catching that one!
The changelog entry should go into the `Docs` section.

PR #1011, opened 2020-06-02
Export `Image` and `ImageData` by default

covercash2:
this change exposes all of the `Image` functions and `ImageData::empty` without exposing the `image` crate. it also shows hints about the "image" feature in the API docs.

### next
this PR doesn't add any functionality, really. still without the "image" feature enabled you can only draw an empty `Image` widget, but this is just an incremental step toward a more robust API for this feature.
i already have a bunch of documentation changes for `image.rs`, but i wanted to push this change first (i'm also working on my git habits).


druid/src/widget/mod.rs
```diff
@@ -52,8 +50,6 @@ mod view_switcher;
 mod widget;
 mod widget_ext;
 
-#[cfg(feature = "image")]
-#[cfg_attr(docsrs, doc(cfg(feature = "image")))]
 pub use self::image::{Image, ImageData};
```
xStrom, line 53: I don't think this `self` is needed in front of this import. Can you check if it compiles without it?

druid/src/widget/image.rs
```diff
@@ -127,6 +144,35 @@ impl ImageData {
...
+                    self.format,
+                )
+                .unwrap();
+            ctx.draw_image(&im, size.to_rect(), interpolation);
+        })
+    }
+}
+
+#[cfg(feature = "image")]
+use std::{convert::AsRef, error::Error, path::Path};
```
xStrom, line 171: Let's not move imports to the middle of the file. Keep it at the top.

druid/src/widget/image.rs
```diff
@@ -27,6 +23,17 @@ use crate::{
...
+///
+/// `druid` does not support loading image data by default.
+/// However, there is an optional dependency on the `image` crate
+/// that makes loading simple. This feature can be enabled with cargo.
+/// For example, in `Cargo.toml`:
+/// ```no_compile
+/// [dependencies.druid]
+/// version = "0.6.0"
+/// features = ["image"]
+/// ```
```
xStrom, line 35: I think the documentation for `Image` / `ImageData` should mention the `image` feature just briefly and link to a more detailed description how to use features elsewhere.

I think adding a section to the root of the crate (`lib.rs`) is a fine place. Just before the `Examples` section for example. The new section could list all the features we have and a brief description of what they do. Then below this list we could have a simple `Cargo.toml` example how to enable a feature.

druid/src/widget/image.rs
```diff
@@ -173,30 +219,10 @@ impl ImageData {
...
-                )
-                .unwrap();
-            ctx.draw_image(&im, size.to_rect(), interpolation);
-        })
-    }
 }
 
+#[cfg(feature = "image")]
+#[cfg_attr(docsrs, doc(cfg(feature = "image")))]
 fn has_alpha_channel(image: &image::DynamicImage) -> bool {
```
xStrom, line 226: This being a private function, it won't need the doc config. Also perhaps this makes sense to move under the `ImageData` impl block as an assocaited function.

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ You can find its changes [documented below](#060---2020-06-01).
 
 ### Changed
 
+- `Image` and `ImageData` exported by default. ([#1011] by [@covercash2])
```
xStrom, line 14: Add the PR link definition to the bottom of the file.

druid/src/widget/image.rs
covercash2: yeah, just got a little trigger happy with the copy paste without thinking. i thought to put this as an associated fn, but decided against it at first since it wasn't necessary (trying to be more atomic with my git commits). easy enough, though

covercash2: looking at it now, i'm thinking i may just inline this function. it's only used in once place, and doesn't totally make sense as an associated fn (doesn't need `self` or `Self`).

xStrom: Associated functions by definition never need `self`. Those are methods.

PR #1012, opened 2020-06-02
Fix nightly feature docs.

xStrom:
This PR fixes the nightly doc generation feature where individual item docs will have a notice about being guarded by a feature flag.

None of the `Cargo.toml` files were specifying the custom cfg flag that we use to guard the nightly docs feature with.

I also removed two unneeded uses of this doc cfg usage. It's only needed next to the `mod foo;` line and doesn't need to be duplicated on `use` lines.

### DIFF
--- a/druid-derive/Cargo.toml
+++ b/druid-derive/Cargo.toml
@@ -10,6 +10,10 @@ edition = "2018"
 [lib]
 proc-macro = true
 
+[package.metadata.docs.rs]
+rustdoc-args = ["--cfg", "docsrs"]
+default-target = "x86_64-pc-windows-msvc"
+
 [dependencies]
 syn = "1.0.29"
 quote = "1.0.6"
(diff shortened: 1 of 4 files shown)


Finnerale:
This does not seem to break anything, but I could not find any difference either.
Where exactly does this have an impact?

xStrom:
It has an impact on docs.rs when the `svg` feature is enabled. Which it is not right now.

If you wish to see it in action, then do these modifications:
- In `druid/src/lib.rs` replace `#![cfg_attr(docsrs, feature(doc_cfg))]` with `#![feature(doc_cfg)]`
- In `druid/src/widget/mod.rs` replace `#[cfg_attr(docsrs, doc(cfg(feature = "svg")))]` with `#[doc(cfg(feature = "svg"))]`

Then run: `cargo +nightly doc --manifest-path=druid/Cargo.toml --features svg --open`

(Alternatively you can skip the modifications and invoke `rustdoc` directly with `--cfg docsrs`)

Then when you check out the `Svg` widget you'll see that the docs have blue notices saying that it's feature gated.

PR #1013, opened 2020-06-02
Standardize web terminology.

### LINKED ISSUE #999: Towards less confusion over web support.
## Background
Druid supports targeting the web, however the naming we use in documentation is all over the place.

There are **web** and **Wasm**. Currently both are used pretty much as synonyms. I think we should be more precise with our terms to avoid confusing people.

### Web
Web is a really well known term and definitely delivers the message. Druid can run on the web. However the web supports multiple languages other than Wasm. Druid mostly runs via Wasm, although indirectly also via others. The distinction is that Wasm is involved at all, because it would also be possible to just transpile straight into JavaScript.

### WebAssembly
Wasm is less known but in some circles where it is known it can have a higher *cool factor* than plain *web*. Those who do know Wasm would probably guess that this means that druid runs on the web. However WebAssembly also [supports non-web targets](https://webassembly.org/docs/non-web/) which druid certainly does not. (*Not via Wasm anyway.*)

## Standardizing the terms
I think **web** is the better term to go with. Orders of magnitude of more people know what it means and I think it's also closer to the practical reality we have here - druid runs on the web and not on any other Wasm targets.

So whenever in documentation we need just a single word to describe this backend/capability we would use **web**. However when we're talking about it in more detail, it would be fine to say **web via Wasm** or describe that the **web** capability is achieved thanks to **Wasm**. That is to say, I'm not proposing on removing all mentions of **Wasm** - I'm just trying to limit it to contexts talking about **web** - so that when people see it used, they know that it has something to do with the web.

## Open for feedback
Did I miss something? Are there good reasons to go with a different route? Please comment.

xStrom:
This PR standardized the terminology when talking about targeting the web. The rationale for this change was discussed in #999.

Generally the term to use is `web`. It's the web backend, and these apps are targeting the web. When going into more detail, WebAssembly (Wasm) can also come up as an implementation detail.

I also renamed the `switch` example to `switches` so that it won't conflict with the reserved keyword in JavaScript.

I also defined the default members of the workspace. This means that when you write `cargo clippy` in the root workspace, it won't start also building all the web examples. You can still get that behavior with `cargo clippy --workspace` and the CI will still test these.

Fixes #999.


druid/examples/web/build.rs
```diff
@@ -122,21 +122,15 @@ mod examples {
...
             examples_in.push_str(&format!("    pub mod {};\n", example_str));
 
-            // The "switch" example name would conflict with JavaScript's switch statement. So we
-            // rename it here to switch_demo.
-            let js_entry_fn_name = if &example_str == "switch" {
-                "switch_demo".to_string()
-            } else {
-                example_str.to_string()
-            };
+            let js_entry_fn_name = example_str.to_string();
```
Finnerale, line 133: Without the special case we could just pass `example_str` to the template directly.

PR #1016, opened 2020-06-03
Remove `KeyCode.is_printable`

ForLoveOfCats:
Instead of checking if the KeyCode is printable in TextBox the more sound logic is to check if the key event's text option is Some. This allows the numpad on GTK to still be used for text insertion even when the keycodes are yet to be fixed. By removing the method altogether this will force applications to check the option as well.

The Windows shell relies on determining the printablity of a keydown to avoid sending double keydown events to the application (with one of them containing no text). While that is very much a hack it is entirely out of scope for this PR to fix and would likely require a deep dive to find a suitable solution. As a result I've moved the original `is_printable`'s match over in order to continue to use as a stopgap and retain current behaviour. I've additionally documented that thoroughly so that anyone looking at it in the future won't have to do quite so much git blame finagling to figure out what the heck it is doing.

### DIFF
--- a/druid-shell/src/keycodes.rs
+++ b/druid-shell/src/keycodes.rs
@@ -148,20 +148,3 @@ pub enum KeyCode {
 
     Unknown(platform::RawKeyCode),
 }
-
-impl KeyCode {
-    pub fn is_printable(self) -> bool {
-        use KeyCode::*;
-        match self {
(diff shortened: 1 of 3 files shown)


ForLoveOfCats:
Windows CI is failing. I'll take a look in a moment, probably something silly like not having the correct namespace in scope or something along those lines

cmyr:
My inclination is to close this and hold out for the larger-scale refactor described in #1040. Does that sound reasonable @ForLoveOfCats?

Issue #891, opened 2020-04-29
Roadmap for 3D


raphlinus:
This issue captures my thinking on 3D, but I'm open to discussion, especially if people bring a lot of energy and motivation.

There's a lot of interest in allowing access to 3D graphics from druid, but at the same time it's not in any way blocking Runebender, so it's hard to prioritize.

That said, one reason I'm very interested in 3D API's is to support better 2D rendering. There are at least two paths to that right now - [Pathfinder](https://github.com/servo/pathfinder) and [piet-gpu](https://github.com/linebender/piet-gpu). These are fairly different in their approaches, as Pathfinder is designed for compatibility with installed GPU hardware and drivers, while piet-gpu explores cutting-edge compute capabilities. As of recently Pathfinder exposes enough of the 2D imaging model that we could consider it, and there is already a [Pathfinder piet backend](https://github.com/linebender/piet/pull/177) in progress.

Thus, we need to consider the approach to 3D in layers. One is: what should druid-shell expose? That layer can be consumed by druid to provide the best piet experience possible, even without exposing 3D. The other is: what should druid expose?

There is also the question of complexity. There are many, many 3D API's out there, with a complex matrix of compatibility and capabilities. Any approach to 3D must involve runtime detection, with some sort of fallback. Adding to the complexity, using 3D codepaths creates integration problems unique to desktop apps, not shared by the more typical game use cases: incremental present capability, low latency present modes (based on [wait objects](https://docs.microsoft.com/en-us/windows/uwp/gaming/reduce-latency-with-dxgi-1-3-swap-chains)), smooth resize, etc.

One very appealing approach is to adopt [wgpu](https://github.com/gfx-rs/wgpu) as the primary integration point. The runtime detection would be wgpu or no wgpu, plus of course finer grained feature detection as provided by wgpu. Not all platforms can support wgpu, but compatibility work is envisioned (from the wgpu web page, OpenGL is currently unsupported but in progress).

[...]

PR #1025, opened 2020-06-09
Port X11 backend from xcb to x11rb.

jneem:
This ports the X11 backend from the `xcb` crate to the `x11rb` crate, as discussed in #989. Overall, I have a positive impression of `x11rb`, as compared to `xcb`: it's actively maintained, it uses actual rust enums in many places, and it doesn't need unsafe for event handling. Also, it supports the present extension (a blocker for #989) and the xinput extension (which presumably we'll need at some point), unlike `xcb`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ You can find its changes [documented below](#060---2020-06-01).
 ### Maintenance
 
 - Standardized web targeting terminology. ([#1013] by [@xStrom])
+- X11: Ported the X11 backend to `x11rb`. ([#1025] by [@jneem])
 
 ### Outside News
 
@@ -327,6 +328,7 @@ Last release without a changelog :(
 [#1008]: https://github.com/xi-editor/druid/pull/1008
 [#1011]: https://github.com/xi-editor/druid/pull/1011
 [#1013]: https://github.com/xi-editor/druid/pull/1013
+[#1025]: https://github.com/xi-editor/druid/pull/1025
 [#1028]: https://github.com/xi-editor/druid/pull/1028
 [#1042]: https://github.com/xi-editor/druid/pull/1042
 [#1043]: https://github.com/xi-editor/druid/pull/1043
--- a/druid-shell/src/platform/x11/application.rs
+++ b/druid-shell/src/platform/x11/application.rs
@@ -123,6 +123,8 @@ impl Application {
             )?
             .check()?
         {
+            // TODO: https://github.com/psychon/x11rb/pull/469 will make error handling easier with
+            // the next x11rb release.
             return Err(x11rb::errors::ReplyError::X11Error(err).into());
         }
 
--- a/druid-shell/src/platform/x11/application.rs
+++ b/druid-shell/src/platform/x11/application.rs
@@ -20,12 +20,10 @@ use std::convert::TryInto;
 use std::rc::Rc;
 
 use anyhow::{anyhow, Context, Error};
-use xcb::{
-    ButtonPressEvent, ButtonReleaseEvent, ClientMessageEvent, Connection, DestroyNotifyEvent,
-    ExposeEvent, KeyPressEvent, MotionNotifyEvent, BUTTON_PRESS, BUTTON_RELEASE, CLIENT_MESSAGE,
-    COPY_FROM_PARENT, CW_EVENT_MASK, DESTROY_NOTIFY, EVENT_MASK_STRUCTURE_NOTIFY, EXPOSE,
-    KEY_PRESS, MOTION_NOTIFY, WINDOW_CLASS_INPUT_ONLY,
-};
+use x11rb::connection::Connection;
+use x11rb::protocol::xproto::{ConnectionExt, CreateWindowAux, EventMask, WindowClass};
+use x11rb::protocol::Event;
+use x11rb::xcb_ffi::XCBConnection;
 
 use crate::application::AppHandler;
 
@@ -40,7 +38,7 @@ pub(crate) struct Application {
     /// The X server might also host other displays.
     ///
     /// A display is a collection of screens.
-    connection: Rc<Connection>,
+    connection: Rc<XCBConnection>,
     /// The default screen of the connected display.
     ///
     /// The connected display may also have additional screens.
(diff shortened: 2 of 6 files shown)


druid-shell/src/platform/x11/application.rs
```diff
@@ -20,12 +20,10 @@ use std::convert::TryInto;
...
-use xcb::{
-    ButtonPressEvent, ButtonReleaseEvent, ClientMessageEvent, Connection, DestroyNotifyEvent,
-    ExposeEvent, KeyPressEvent, MotionNotifyEvent, BUTTON_PRESS, BUTTON_RELEASE, CLIENT_MESSAGE,
-    COPY_FROM_PARENT, CW_EVENT_MASK, DESTROY_NOTIFY, EVENT_MASK_STRUCTURE_NOTIFY, EXPOSE,
-    KEY_PRESS, MOTION_NOTIFY, WINDOW_CLASS_INPUT_ONLY,
-};
+use x11rb::connection::Connection;
+use x11rb::protocol::xproto::{ConnectionExt, CreateWindowAux, EventMask, WindowClass};
+use x11rb::protocol::Event;
+use x11rb::xcb_ffi::XCBConnection;
```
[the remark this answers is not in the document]
Finnerale: Currently, we are using a mix, sometimes grouped, sometimes separated, which is why I would like to choose one and go with that everywhere.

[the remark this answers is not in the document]
Finnerale: Considering that the related `rustfmt` option `merge_imports` is still unstable / only partially implemented, it is probably best to  just stick to separated imports and reevaluate that once it is properly supported by `rustfmt`.

cmyr:
it should be possible to adapt the `bloat` stuff to take a backend as an argument or something 🤔

jneem:
In this case, we would also need it to take a feature flag...

psychon:
> But 100 KB larger binaries is quite a bit of bloat, I wonder where that comes from :thinking:

> xcb avoids this by leaning on the parsing code in libxcb and then just using FFI to access the structs.

libxcb is basically a big `std::mem::transmute` (well, it is C code... it's a pointer cast). The whole idea of libxcb is to describe the X11 on-the-wire protocol as C structs so that "no actual parsing" happens / the C code directly reads the on-the-wire-representation. (This is also the origin of most of the API weirdness.) (This also means that there is not much length checking and a malicious X11 server could send a packet of length 42 containing a list which claims to have a million entries.)

Thus, x11rb really does a lot more than libxcb. If you want to call that "bloat" or not is up for debate. :-)

(By the way, x11rb doing actual parsing is why it supports more extension than the XCB crate; some of the on-the-wire stuff does not really map to C `struct`s, but since x11rb "invents" its own representation, it can work with these more easily.)

TL;DR: Even when also considering the size of `libxcb.so` and all the `libxcb-foo.so`, x11rb will likely result in larger code.

druid-shell/src/platform/x11/window.rs
```diff
@@ -171,62 +204,64 @@ impl WindowBuilder {
...
-            // Window class type
-            WINDOW_CLASS_INPUT_OUTPUT as u16,
-            // Visual ID
-            visual_id,
-            // Window properties mask
-            &cw_values,
-        )
-        .request_check()?;
+        // TODO: https://github.com/psychon/x11rb/pull/469 will make error handling easier with the
+        // next x11rb release.
```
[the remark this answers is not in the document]
psychon: According to https://github.com/psychon/x11rb/issues/484, the timeline for a new release is "about one week".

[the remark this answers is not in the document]
jneem: Nice, then I'll probably just wait until then...

jneem:
It's not that I think 100kb is excessive, but I do think it might be possible (if not a high priority) to improve. According to `cargo bloat`, `Event::parse` is more than 9kb of code (this is only with randr). Then there's also a bunch of things like `KeyPressEvent::try_parse` at around the 3kb mark. I bet that rustc is doing a lot of inlining, and that the error-handling code could be de-duplicated.

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ You can find its changes [documented below](#060---2020-06-01).
 ### Maintenance
 
 - Standardized web targeting terminology. ([#1013] by [@xStrom])
+- Ported the X11 backend to `x11rb`. ([#1025] by [@jneem])
```
xStrom, line 37: Let's continue the tradition from the previous release and prefix backend specific changes.
Something like:
```md
- X11: Ported the backend from `xcb` to `x11rb`. ([#1025] by [@jneem])
```

druid-shell/src/platform/x11/application.rs
```diff
@@ -74,59 +72,61 @@ struct State {
 
 impl Application {
     pub fn new() -> Result<Application, Error> {
-        let (conn, screen_num) = Connection::connect_with_xlib_display()?;
+        let (conn, screen_num) = XCBConnection::connect(None)?;
```
xStrom, line 75: I haven't really investigated this stuff with either library, but is this an equivalent port? The `connect_with_xlib_display` docs say that going through XLib is needed for OpenGL support. I'm not sure if cairo uses OpenGL but if it does - does the new connection also provide that?

Finnerale: To my knowledge Cairo does not use OpenGL by default. There is an [experimental feature](https://www.cairographics.org/OpenGL/) for it but otherwise it uses the platform library (X11 on Linux). So it should be fine if there is no OpenGL available, unless we want to use it ourselves.

jneem: I suspect that we will want OpenGL support eventually. If not for cairo, then for #891-related stuff. But it looks like it will need some work on the x11rb side.

psychon: I guess you will want to use the `x11` crate for this. (The following "talks `C`", I haven't looked at the Rust bindings of any of this.)

- create you `Display*` via `x11`.
- Get the underlying XCB Connection via `XGetXCBConnection`. This gives you an `xcb_connection_t*`.
- This pointer can be given to `x11rb::XCBConnection::from_raw_xcb_connection(dpy, false)`.
- If you want to handle events through x11rb and libxcb, you should call `XSetEventQueueOwner(dpy, XCBOwnsEventQueue)`. Otherwise, libX11 might randomly eat your events / move them to its own event queue.
- One you drop / `XCloseDisplay` the `Display*`, any further use of the `XCBConnection` might cause demons to fly through your nose, so be careful (by the way: Similar fun can result with cairo, but cairo makes it much easier to accidentally continue using the connection after it was closed / freed.).

Congratulations, you now have a single X11 connection which you can access through all of libX11, libxcb and x11rb.

API-wise, I do not see anything that could be added to x11rb. The two functions I mentioned above live in `libX11.so` and `libX11-xcb.so` and x11rb certainly has no business depending on them.

druid-shell/src/platform/x11/window.rs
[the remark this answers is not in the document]
psychon: Done

druid-shell/src/platform/x11/application.rs
jneem: Thanks for the detailed information! I'd prefer not to block this PR on OpenGL support, so I've just dropped in a comment for now.

druid-shell/src/platform/x11/application.rs
```diff
@@ -72,6 +72,14 @@ struct State {
...
 impl Application {
     pub fn new() -> Result<Application, Error> {
+        // If we want to support OpenGL, we will need to open a connection with Xlib support (see
+        // https://xcb.freedesktop.org/opengl/ for background).  There is some sample code for this
+        // in the `rust-xcb` crate (see `connect_with_xlib_display`), although it may be missing
+        // something: according to the link below, If you want to handle events through x11rb and
+        // libxcb, you should call XSetEventQueueOwner(dpy, XCBOwnsEventQueue). Otherwise, libX11
+        // might randomly eat your events / move them to its own event queue.
+        //
+        // https://github.com/xi-editor/druid/pull/1025/files/76b923417183bd103f61e56b56a56474b7417cec#r442777892
```
Finnerale, line 82: This link is broken for me.

jneem: Thanks, fixed, and I'll merge when CI lets me. Definitely nice to be using a library that is both rusty and actively maintained to the point where the maintainer drops by and tells us what we should be doing better :smile:

PR #1031, opened 2020-06-13
Added multi-line option to textbox and changed painting to match any text size

rhzk:
This pull request adds a option to allow multi-line instance of textbox.
It only checks for '\n' line-breaks so '\r' alone might be a issue.

I also changed the painting of different elements so it should fit text of any size. (Old painting code had hard numbers that just happened to work with the default text size of 15.0)

I would like to have text_size as a member but I am unsure how to give it a default value as I cant access env.get() in the TextBox::new()

Shift+Enter does not trigger 'KeyCode::Return' should it be add it to that or to a new KeyCode?

### DIFF
--- a/druid/Cargo.toml
+++ b/druid/Cargo.toml
@@ -37,6 +37,7 @@ unicode-segmentation = "1.6.0"
 xi-unicode = "0.2.0"
 fnv = "1.0.7"
 instant = { version = "0.1.4", features = ["wasm-bindgen"] }
+bytecount = "0.6.0"
 
 # Optional dependencies
 im = { version = "15.0.0", optional = true }
--- a/druid/src/widget/textbox.rs
+++ b/druid/src/widget/textbox.rs
@@ -16,12 +16,14 @@
 
 use std::time::Duration;
 
+extern crate bytecount;
+
 use crate::{
     Application, BoxConstraints, Cursor, Env, Event, EventCtx, HotKey, KeyCode, LayoutCtx,
     LifeCycle, LifeCycleCtx, PaintCtx, Selector, SysMods, TimerToken, UpdateCtx, Widget,
 };
 
-use crate::kurbo::{Affine, Line, Point, RoundedRect, Size, Vec2};
+use crate::kurbo::{Affine, Line, Point, Rect, RoundedRect, Size, Vec2};
 use crate::piet::{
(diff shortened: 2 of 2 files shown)


Finnerale:
For the text size you can probably set it to zero in the constructor and then initialize it to the actual value during `LifeCycle::WidgetAdded`.

rhzk:
Thanks for the tip.

Another issue is how to check if there is a scroll widget attached and communicate with it.

Currently the focus border is drawn around the textbox when a scroll widget is attached, I need to either get the size of the scroll and draw the focus according to that size. or disable it and draw it in the scroll instead. This picture illustrates the problem (textbox grows infinite, scroll fixed size):

scroll

There is also a need to update the scroll position when people navigate downwards in the text using arrow keys

HoNile:
I think the correct way to count lines and handle all possible line-breaks is to use `text.lines().count()`

ForLoveOfCats:
On the topic of widgets handling being in a `scroll` see https://github.com/xi-editor/druid/issues/799

rhzk:
On the documentation for text.lines().count() it says it calls next() over and over. I don't know if that gets optimized but it sounds like a lot of jump calls. And could maybe be a performance problem with large amounts of text.

HoNile:
I have searched on the web and it look like counting only '\n' should work, specially I found this https://llogiq.github.io/2016/09/24/newline.html so if xi work by counting '\n' it's ok in all cases I guess.

Notes:
- Iterator should perform well https://doc.rust-lang.org/book/ch13-04-performance.html
- '\r' is not used anymore by any major OS, source Wikipédia, that's probably why `lines` work on '\n' or '\r\n'.

rhzk:
That is interesting thanks, I might also set up some performance tests to see how they compare to each other.

cmyr:
My next in-the-code project is going to be a large reworking of our text API, that will include multi-line support; I'm not sure it's worth merging temporary fixes in the meantime, since there's some fairly significant work to do under the hood.

rhzk:
This was mostly fixing the painting code so that it can draw the cursor / selection over multiple lines,
and making sure the widget draws the lines with the correct height. It might still be useful after you have rewritten the text API. Can put this on hold for now and then I can update it if/when needed to reflect the changes

cmyr:
going to close this for now, a solution for this is in progress.

PR #1033, opened 2020-06-14
impl ListIter for im::HashMap

futurepaul:
This is motivated by my own need for a stable "key" of a list item so I can use that specific list item in another part of my app.

I'm guessing this is possible to do without im but it looks a lot more difficult so I will kindly ask not to be selected for that task :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ You can find its changes [documented below](#060---2020-06-01).
 
 ### Added
 
+- `im::HashMap` support for the `List` widget. ([#1033] by [@futurepaul])
+
 ### Changed
 
 - `Image` and `ImageData` exported by default. ([#1011] by [@covercash2])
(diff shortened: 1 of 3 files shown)


futurepaul:
After going through the trouble of getting this compile with a new example, I'm realizing this example should probably just be an extension or modification of the existing `list.rs` example. thoughts?

xStrom:
This was briefly mention [on zulip](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/Getting.20index.20of.20item.20in.20a.20List) and I tend to agree - `HashMap` doesn't seem like a natural fit for a `List` because the [order is undefined](https://docs.rs/im/15.0.0/im/index.html#maps) - although stable in the current implementation. `im::OrdMap` would be a better fit for sure and would also help with your key requirement.

Example wise, yeah I'd say a strong no for another example just for the map. Every new example introduces maintenance burden. If there's a decent way to add it to the existing `list` example - that's cool. Otherwise I think just not having an example is fine too.

cmyr:
Agree that `OrdMap` is a better fit for this, otherwise this seems reasonable (outside of the fact that I want a better abstraction than `ListIter` for collections, but one thing at a time)

Also don't think this really needs a new example.

futurepaul:
This makes a lot of sense @xStrom / @cmyr. Unfortunately I don't see the `iter_mut` method on im's OrdMap... might just be an oversight on their part?

xStrom:
That's interesting indeed. I asked it in [im#138](https://github.com/bodil/im-rs/issues/138). Hopefully we'll know if this is something we could add.

Issue #883, opened 2020-04-28
text work planning & discussion


cmyr:
# Druid text planning

This is intended as a medium-level sketch for how we will implement multi-line text, text editing, and (eventually) rich text in druid.

This covers a number of things: the storage of text itself, the storage of style and color information, the storage of selection state, and the storage of line breaks and width measurement, as well as the relationships between these, as is  required for things like interactive editing.

## Representation of text

We are going to adapt the [xi-rope] crate as our main way of representing text buffers. This is a not-insignificant depedency, but ultimately it feels worthwhile; the design of the rope (using smart pointers under the hood) makes it very well suited to druid's `Data` model, and xi-rope includes a large API for things like representing and applying edits, diffing, text search, and rich
text spans.

We will use xi-rope's `Rope` to store our text, and the `Spans` type to (later) represent rich text information.

These two types will be used together, and will be wrapped up in some druid type, with a name like `TextBuffer`.

## Line breaking and width measurement

In order to display text to the user in a meaningful way, we need to be able to do line breaking and width measurement. A version of this is provided by piet; we will wrap this API in druid, providing our own `TextLayout` type.

This type will own a copy of the `TextBuffer`, and will also have a width; it will be responsible for both painting text as well as for converting from points on the screen (e.g. from a mouse click) to offsets in the buffer.

You will create a `TextLayout` from a `TextBuffer` and a `PietText` object (discussion: can we make `PietText` not be bound to a lifetime?); you will then be able to paint it, using a convenience method looking something like, `TextLayout::draw(&self, ctx: &mut PaintCtx, point: Point, env: &Env)`. You will also be able to query the layout for information like the offset in the buffer that corresponds to a given point in the layout.

When the buffer changes elsewhere, you will update the `TextLayout`, by calling a method like (e.g.) `TextLayout::update_buffer(&mut self, buffer: &TextBuffer)` and then the `TextLayout` will update its internal state. An important part of this design is that the layout will be able to update itself incrementally, by comparing the old and the new buffers; however for our first pass we may just recompute the entire layout from scratch.

This `TextLayout` type will be the basis for all text that is displayed in druid; we will discourage the use of piet directly.

## Selection state and editing

[...]

PR #1036, opened 2020-06-15
Support multiple lines in labels

### LINKED ISSUE #428: Multiline labels
The label widget is currently a single line, which is a serious limitation. #391 was a draft PR to add this, but a more robust solution is to have support for multiline text layouts at the piet level. https://github.com/linebender/piet/pull/103 is a work-in-progress PR to add that. This issue tracks support in druid for when that lands in piet.

jpochyla:
First try at multiline support in Label. Supports automatic wrapping if there's not enough horizontal space. I'm not really sure about the padding calculations, caching the text layout, or whether this is a good time to add support for vertical-clipping as well.

Thanks!

Fixes #428

### DIFF
--- a/druid/src/widget/label.rs
+++ b/druid/src/widget/label.rs
@@ -23,10 +23,6 @@ use crate::{
     LifeCycleCtx, LocalizedString, PaintCtx, Point, Size, UpdateCtx, Widget,
 };
 
-// a fudgey way to get an approximate line height from a font size
-const LINE_HEIGHT_FACTOR: f64 = 1.2;
-// a fudgey way of figuring out where to put the baseline, relative to line height
-const BASELINE_GUESS_FACTOR: f64 = 0.8;
 // added padding between the edges of the widget and the text.
 const LABEL_X_PADDING: f64 = 2.0;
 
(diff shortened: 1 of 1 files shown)


druid/src/widget/label.rs
```diff
@@ -195,14 +191,14 @@ impl<T: Data> Label<T> {
         self.font = font.into();
     }
 
-    fn get_layout(&mut self, t: &mut PietText, env: &Env) -> PietTextLayout {
+    fn get_layout(&mut self, t: &mut PietText, env: &Env, bounds: &Size) -> PietTextLayout {
         let font_name = self.font.resolve(env);
         let font_size = self.size.resolve(env);
 
         // TODO: caching of both the format and the layout
```
jpochyla, line 198: We should cache the text layout only between consecutive `Widget::layout` and `Widget::paint` calls, not in further scope, is that correct?

jpochyla: Related, but since I can't comment on non-modified lines: in `Label::update`, condition `!old_data.same(data)` is wrong, is it not? `data` can stay the same, but we need to request layout even if the environment changes.

jpochyla:
Closing, text layout is getting improved properly in #883. Thank you!

Issue #778, opened 2020-03-30
Implement 'set_windows_position'


rhzk:
Function to set/move the window position on the screen.

Theres also the question about multiple displays, and if the window should be restrained within bounds.

Other GUIs I have used in the past usually did not have a restrain option, so it was up to the programmers using the GUI to implement their own if they wanted it.
We could potentionally have a set_windows_position_restrained(), that would keep the window within the display/displays if values would exceed them. In other word prevent the window from being drawn outside any visible space.

Multiple displays are a bigger problem tho as it might differ between operating system and display modes.

If lets say I have 2 fhd monitors(1920*1080)*2 and I call the function with 1921, Should the window then be drawn on the second monitor? This might be problematic if the resolution is different tho.

Maybe the best option here would be functions 'get_displays' 'set_window_display' and display.get_resolution'

cmyr:
as you mention, this is tied to having some way of querying the connected displays.

We could preliminarily do a simple version that only uses the main display. Even for that to be useful, though, we do want some way of querying the platform for the size of the current display.

this would require new API in druid-shell.

This doesn't need to be crazy; I would add a new `screen` module in druid-shell, and I would initially just have a single top-level function, `fn main_screen_size() -> Size`.

Then we would add methods to `WindowHandle` in druid-shell for controling the window frame. I would expect, at the very least, a 'set_frame' method; we could definitely also have 'set_frame_constrained'.

On your last point; in general windows have to be on one monitor or another, for the reasons you describe. If you drag a window halfway onto another monitor, I don't know of a platform where it will be displayed half and half.

In any case: bottom line is yes, we should have this, and that if you're interested in working on it you should feel free!

xStrom:
> If you drag a window halfway onto another monitor, I don't know of a platform where it will be displayed half and half.

What do you mean by this? I'm confused because on Windows you can definitely have half of a window on one monitor and the other half on the other monitor.

ForLoveOfCats:
I will concur that X11 (I'm 90% sure Wayland is the same) will also display half the window on one monitor and half on the other

cmyr:
hmm, noted.

JAicewizard:
I just saw a PR related to this, and I would like to add that Wayland does not allow you to set window position. This may changer in the future since this completely blocks wine from working under wayland, but from what I have read wayland maintainers are stubborn about it.

[...]

Issue #634, opened 2020-03-11
implement 'show titlebar' window option


cmyr:
This was implemented on gtk in #587.  This is at least somewhat complicated on mac; it isn't 100% clear what it should correspond to. If there is no window border on mac that means the window also is not resizable or moveable by default. Is that the intended behaviour?  this needs to be clarified.

cc @Psykopear: exactly what is the intended behaviour of a window that has `show_titlebar` false?  can that window be resized or moved? We may need to slightly rethink this..

futurepaul:
In electron a "frameless" window is non-draggable by default and it's up to the user to implement a draggable region: https://www.electronjs.org/docs/api/frameless-window

Electron also has more granular settings for mac, including hiding the title bar but showing the traffic light buttons.

futurepaul:
Some Apple documentation here: here's [how to hide the titlebar](https://developer.apple.com/documentation/appkit/nswindow/1419167-titlebarappearstransparent), but it warns that this only makes sense when `NSFullSizeContentViewWindowMask` is true. But `NSFullSizeContentViewWindowMask` is "deprecated." So... \o/

futurepaul:
Some [relevant discussion for winit](https://github.com/rust-windowing/winit/pull/408)

Psykopear:
So, from a Linux perspective, I'd expect that setting `show_titlebar` as false starts the application with no border and no titlebar (equivalent to the `decorated` gtk option).
Then it's up to the application to implement a draggable region, like futurepaul said for electron.
And that's it, I think that other properties like the ability to resize and move the window should remain unchanged when setting `show_titlebar`.

The `resizable` should only give a fixed size to the window (on tiling window managers that support floating windows, the window might also implicitly become floating, but that's up to the WM I think).

Non movable windows also could be required (I'm thinking of applications like conky), and always as a separate setting. That *could* also imply the non resizable property, but not the other way around.

I don't know if those patterns makes sense on other OS, but at least on Linux they are used independently. Some examples could be:

- Bitwig studio as mentioned in another issue, draws its own borders and titlebars, but it's moveable and resizable. I think the Spotify client does the same?
- Conky or Ulauncher are non resizable, non moveable and borderless/titlebarless

My point is that having separate options for moveability, resizability and borderless drawing would be an expected pattern in Linux applications. Can't say anything about Mac, but for example bitwig works the same on Windows, so it could work similarly there.

[...]

Issue #626, opened 2020-03-10
ERROR [druid_shell::platform::windows::window] ResizeBuffers failed: 0x80070057 when minimize a window


usagi:
I got the error log in the terminal when I minimize the window( 100% repro ):

```sh
ERROR [druid_shell::platform::windows::window] ResizeBuffers failed: 0x80070057
```

It is not a critical error, but I want to run an app without an errors.

- OS: Windows 10
- toolchain: x86_64-pc-windows-msvc; 1.43.0-nightly (d3c79346a 2020-02-29)
- druid: `druid = { git = "https://github.com/xi-editor/druid.git", version = "0.5" }`
- source ( It is same to the example in the README.md ):

```rs
use druid::{
  widget::{
    Align,
    Button,
    Flex,
    Label,
    Padding,
    WidgetExt
  },
  AppLauncher,
  LocalizedString,
  Widget,
  WindowDesc
};

fn main()
{
  let main_window = WindowDesc::new(ui_builder);
  let data = 0_u32;
  AppLauncher::with_window(main_window).use_simple_logger().launch(data).expect("launch failed");
}

fn ui_builder() -> impl Widget<u32>
{
  // The label text will be computed dynamically based on the current locale and count
  let text = LocalizedString::new("hello-counter").with_arg("count", |data: &u32, _env| (*data).into());
  let label = Label::new(text).padding(5.0).center();
  let button = Button::new("increment", |_ctx, data, _env| *data += 1).padding(5.0);

  Flex::column().with_child(label, 1.0).with_child(button, 1.0)
}
```

PR #1037, opened 2020-06-15
Screen Module, window titlebar, position, size, DPI

rhzk:
Changes:

All changes are only implemented on Windows (but have skeletons for other platforms)

* Added Screen module, where you can get infomation about the screen and monitors.

* Added setting window size, position, maximized and minimized state.
* Added option to disable the default titlebar.
* Added option to let windows handle a custom titlebar like the default one (move, double click to maximize, etc (based on win-settings)).
 * Changed window creation to use window's default window size if the user does not specify a window size.
* Implemented a solution for re-entrancy on Windows. #1068
* Improved DPI on windows (10), should now work correctly across monitors with different dpi settings

This should close the Window's side of #1044,  #778, #634, #427. And #626

Some use-cases I could think of for this:
* Restore window size/position on next launch, as they where when it was closed.
* Create custom look for titlebar.
* Layout multiple windows (I think GIMP does/used to do this: [picture](https://upload.wikimedia.org/wikipedia/commons/b/bc/Gimp-gnome-2.2.8.png)).
* Create window(s) on a particular monitor.
* Change size to fit content, or make content fit window size (This might require additional API).
* Create window at a spesific point (For example an app that shows information about files when users hovers over them or pop up windows, like confirm dialogs/error messages etc)


xStrom:
This is interesting and I'll take a closer look in the coming days.
However for now I wanted to mention one thing.

> There also needs to be a way to exit the GUI from a widget (Borderless window has no 'X' button)

There's the [`CLOSE_WINDOW`](https://docs.rs/druid/0.6.0/druid/commands/constant.CLOSE_WINDOW.html) command that can achieve this already.

rhzk:
Oh I missed that, is there also a way to access the window inside a widget?

rhzk:
Windows does not allow changing the style (title bar) after it has been created, In that case the window has to be recreated. I don't know if there is much use-case for that tho.

druid-shell/src/platform/windows/window.rs
```diff
@@ -1369,6 +1378,89 @@ impl WindowHandle {
...
+                    warn!(
+                        "failed to resize window: {}",
+                        Error::Hr(HRESULT_FROM_WIN32(GetLastError()))
+                    );
+                };
+                // The GUI needs to redraw / layout components again here
+                // Unsure why it does not do this, MoveWindow sends the same messages as when you grab corner of window.
+            }
+        }
+    }
```
rhzk, line 1440: Anyone know how I could tell the GUI that the window size has changed here?
The GUI needs to layout and paint components again.

xStrom: Look into the `WM_SIZE` handling code.

rhzk:
Demo of custom title-bar :)

titlebar_demo

druid-shell/src/platform/windows/window.rs
rhzk: But shouldn't that run when it receives WM_SIZE? I also tried sending it this way:
let lparam = MAKELONG(size.width as u16, size.height as u16);
w.wndproc.window_proc(hwnd, WM_SIZE, 0, lparam as isize);

rhzk: Oh I see the problem, to change the size I need the hwnd and to get that I have to call self.state.upgrade() to borrow the state. When the window then receives WM_SIZE it needs to borrow the state as mutable self.state.try_borrow_mut(). But since I am already borrowing it this fails, and instead of redrawing the GUI it does nothing

cmyr:
I definitely think we are going to want something like a screen module, will give this a scan soon!

rhzk:
I might be wrong but from my understanding of the current windows code it seems to be a circular reference between
WindowHandle -> WindowState -> WndProc -> WindowHandle.

This was solved by putting WindowHandle inside a RefCell. The issue now is that whenever a function inside WindowHandle is called for example:
```
pub fn invalidate(&self) {
        if let Some(w) = self.state.upgrade() {
            let hwnd = w.hwnd.get();
            unsafe {
                if InvalidateRect(hwnd, null(), FALSE) == FALSE {
                    log::warn!(
                        "InvalidateRect failed: {}",
                        Error::Hr(HRESULT_FROM_WIN32(GetLastError()))
                    );
                }
            }
        }
    }

```
&self is borrowed, meaning there is a borrow on WindowHandle.
The windows function "InvalidateRect" creates a WM_PAINT message and does a context-switch,
calling "fn window_proc", that then tries to handle the WM_PAINT:
```
 WM_PAINT => unsafe {
                if let Ok(mut s) = self.state.try_borrow_mut() {
```
Here is tries to borrow WindowState, but fails because WindowState contains WindowHandle, and WindowHandle is already borrowed, inside its own &self function.

From my understanding this causes all Windows messages created inside WindowHandle to be dropped, and the current windows code might have to be rewritten / reorganized.

jneem:
Thanks for bringing this up, because it's reminded me that I intended two write down some thoughts on the interior mutability/borrowing/re-entrancy issue (which comes up in all platforms). However, I think that the *specific* example that you've given isn't really a problem. Unless I misunderstood, I believe you're conflating shared borrows like `&self` (which can be taken an unlimited number of times simultaneously) and the "interior-mutability" kind of borrows used in `RefCell` (which panic at runtime if taken re-entrantly). The `WindowHandle::invalidate` doesn't take any of these, so there's no issue if `WindowHandle::invalidate` causes another re-entrant call into `WindowHandle`.

rhzk:
Well i added a new function to WindowsHandle:
```
pub fn set_size(&self, size: Size) {
        if let Some(w) = self.state.upgrade() { <-- Tried without this and instead coded hwnd into WindowHandle so i dont think
                                                                          it is the problem.
            let hwnd = w.hwnd.get();
            unsafe {
                let position = self.get_position();
                if MoveWindow(hwnd, position.x as i32, position.y as i32, size.width as i32, size.height as i32, 1) == 0 {
                    warn!(
                        "failed to resize window: {}",
                        Error::Hr(HRESULT_FROM_WIN32(GetLastError()))
                    );
                };
            }
        }
    }
```
MoveWindow creates a WM_SIZE message and in the window_proc this always happens:
```
WM_SIZE => unsafe {
    if let Ok(mut s) = self.state.try_borrow_mut() { <------ Fails
      // Code ommited...
    } else {
        self.log_dropped_msg(hwnd, msg, wparam, lparam); <---- This is always hit, meaning the window does not get repainted 
                                                                    with the new size
    }
}
```

Since it fails on try_borrow_mut() It seems to be a problem between borrow checker and the system call.
For the previous functionality in WindowHandle this haven't really been a problem as dropped messages can be ignored but for 
resizing the window we end up with this problem:

The "mimizie" button sets the window to a specific size, but the UI is not redrawn to reflect that size (because message is dropped) until i resize it with the borders (in that case WM_SIZE is not dropped as there is not a borrow lock)

resize_fail

jneem:
Ah, ok, so I'd be willing to bet that the previous borrow came from handling the mouse click that triggered the minimize -- see #1068 for a more detailed description. Is there a way on windows to trigger a `WM_SIZE` asynchronously? Because that would fix it; otherwise, you might need a solution like the one described in #1068.

rhzk:
Well this is basically the code for the button
```
impl<T: Data> Widget<T> for MinimizeButton {
    fn event(&mut self, ctx: &mut EventCtx, event: &Event, _data: &mut T, _env: &Env) {
        match event {
            Event::MouseDown(_) => {
                let window = ctx.window().clone(); // &WindowHandle, holds a borrow?
                                                  // clone() does not fix it either.
                window.set_size(Size::new(200.0,200.0)); // WM_SIZE fails
            }
```
Attemted to set a bool in MouseDown() instead and have the window code handled somewhere else later but same problem.
Could just leave this as-is until somebody has a proper fix for #1068

jneem:
Yeah, this might be easier once #1068 is fixed. Alternatively, I did a bit of quick googling, and it looks like maybe [SetWindowPos](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwindowpos?redirectedfrom=MSDN) with the `SWP_ASYNCWINDOWPOS` flag might do what you want? The docs aren't really clear on what happens when you do it in the same thread as the window...

druid-shell/src/platform/windows/application.rs
```diff
@@ -105,6 +105,11 @@ impl Application {
         unsafe {
             // Handle windows messages
             loop {
+                if let Ok(state) = self.state.try_borrow() {
+                    for hwnd in &state.windows {
+                        SendMessageW(*hwnd, DS_HANDLE_DROPPED, 0,0);
+                    }
+                }
```
rhzk, line 112: Sends a custom message each loop to handle dropped WM_ messages

druid-shell/src/platform/windows/window.rs
```diff
@@ -160,6 +161,13 @@ trait WndProc {
         -> Option<LRESULT>;
 }
 
+struct WindowMessage {
```
rhzk, line 164: Holds dropped messages

druid-shell/src/platform/windows/window.rs
```diff
@@ -329,6 +344,15 @@ impl MyWndProc {
         );
     }
 
+    fn put_msg_to_queue(&self, hwnd: HWND, msg: UINT, wparam: WPARAM, lparam: LPARAM) {
+        if let Ok(mut dropped) = self.dropped.try_borrow_mut() {
+            let m = WindowMessage{hwnd, msg, wparam, lparam};
+            dropped.push_back(m);
+        } else {
+            self.log_dropped_msg(hwnd, msg, wparam, lparam);
+        }
```
rhzk, line 353: Add dropped msgs to queue or if that fails log them

druid-shell/src/platform/windows/window.rs
```diff
@@ -875,6 +899,19 @@ impl WndProc for MyWndProc {
                     None
                 }
             }
+            DS_HANDLE_DROPPED => {
+                if let Ok(mut d) = self.dropped.try_borrow_mut() {
+                    let m = d.pop_front();
+                    if let Some(m) = m {
+                        self.window_proc(m.hwnd, m.msg, m.wparam, m.lparam)
```
rhzk, line 906: Rerun dropped messages

rhzk:
It might not be the best solution but it is one for now

borderless

Slight problem with Event::MouseMove as it does not get triggered when the button suddenly reposition itself

druid-shell/src/platform/windows/window.rs
```diff
@@ -1280,6 +1319,87 @@ impl WindowHandle {
...
+                        "failed to get window rect: {}",
+                        Error::Hr(HRESULT_FROM_WIN32(GetLastError()))
+                    );
+                };
+                return Point::new(rect.left as f64, rect.top as f64)
+            }
+        }
+        warn!(
+            "failed to get window hwnd"
+        );
```
jneem, line 1362: I don't think this needs a warning: it just means that the window was dropped.

druid-shell/src/platform/windows/window.rs
```diff
@@ -1280,6 +1319,87 @@ impl WindowHandle {
...
+                    );
+                };
+                let width = rect.right - rect.left;
+                let height = rect.bottom - rect.top;
+                return Size::new(width as f64, height as f64);
+            }
+        }
+        warn!(
+            "failed to get window size"
+        );
```
jneem, line 1399: I don't think this needs a warning either.

druid-shell/src/screen.rs
```diff
@@ -0,0 +1,64 @@
...
+
+use crate::kurbo::{Rect, Size};
+use crate::platform;
+use std::fmt;
+use std::fmt::Display;
+
+#[derive(Clone)]
+pub struct Monitor {
+    primary: bool,
+    rect: Rect, // The working Rectangle of the monitor in pixels related to Screen.size
```
jneem, line 25: For reasons unknown to me, our convention is to put the comment above the field.

druid-shell/src/screen.rs
```diff
@@ -0,0 +1,64 @@
...
+    work_rect: Rect, // The working Rectangle of the monitor in pixels related to Screen.size excluding reserved space (the taskbar)
```
jneem, line 26: How cross-platform is this notion? Does it exclude the menu bar on mac? I vaguely know that there is a notion of "struts" in X11, but I don't think that the remaining region is necessarily a rectangle.

druid-shell/src/screen.rs
jneem: Also, my inclination would be to use display points instead of pixels, because that way the user doesn't need to do as many conversions.

druid-shell/src/screen.rs
```diff
@@ -0,0 +1,64 @@
...
+}
+
+impl Monitor {
+    pub fn new(primary: bool, rect: Rect, work_rect: Rect) -> Self {
```
jneem, line 30: Why do we want a public constructor for `Monitor`? (but if we do, it should have API docs)

druid-shell/src/screen.rs
```diff
@@ -0,0 +1,64 @@
...
+//! Module to get information about monitors
+
+use crate::kurbo::{Rect, Size};
+use crate::platform;
+use std::fmt;
+use std::fmt::Display;
+
+#[derive(Clone)]
+pub struct Monitor {
+    primary: bool,
```
jneem, line 24: What's the meaning of `primary`? It's private, and doesn't seem to be used anywhere.

druid-shell/src/screen.rs
```diff
@@ -0,0 +1,64 @@
...
+            "({}, {})({}, {})",
+            self.rect.x0, self.rect.x1, self.rect.y0, self.rect.y1
+        )?;
+        Ok(())
+    }
+}
+
+pub struct Screen {}
+impl Screen {
+    pub fn get_monitors() -> Vec<Monitor> {
```
jneem, line 57: Missing API docs.

druid-shell/src/screen.rs
```diff
@@ -0,0 +1,64 @@
...
+        platform::screen::get_monitors()
+    }
+
+    pub fn get_display_size() -> Size {
```
jneem, line 61: Missing API docs. (in particular, dp or pixels?)

druid-shell/src/window.rs
```diff
@@ -123,6 +123,26 @@ impl WindowHandle {
         self.0.show_titlebar(show_titlebar)
     }
 
+    /// Set the position of the window
```
jneem, line 126: What are the units? And I'm guessing this will set the position of the top-left of the window?

druid-shell/src/window.rs
```diff
@@ -123,6 +123,26 @@ impl WindowHandle {
...
+    pub fn set_position(&self, position: Point) {
+        self.0.set_position(position)
+    }
+
+    /// Get the position of the window
```
jneem, line 131: Units? And meaning of position?

druid-shell/src/window.rs
```diff
@@ -123,6 +123,26 @@ impl WindowHandle {
...
+    pub fn get_position(&self) -> Point {
+        self.0.get_position()
+    }
+
+    /// Set the size of the window
```
jneem, line 136: Units?

druid-shell/src/window.rs
```diff
@@ -123,6 +123,26 @@ impl WindowHandle {
...
+    pub fn set_size(&self, size: Size) {
+        self.0.set_size(size)
+    }
+
+    /// Gets the size of the window
```
jneem, line 141: Units?

druid-shell/src/window.rs
```diff
@@ -269,6 +289,11 @@ impl WindowBuilder {
         self.0.show_titlebar(show_titlebar)
     }
 
+    /// Set the window's initial position.
+    pub fn set_position(&mut self, position: Point) {
```
jneem, line 293: Units and meaning of position?

druid/src/app.rs
```diff
@@ -234,6 +236,12 @@ impl<T: Data> WindowDesc<T> {
         self
     }
 
+    /// Set the initial position for this window.
+    pub fn set_position(mut self, position: Point) -> Self {
```
jneem, line 240: Units and meaning of position?

druid-shell/src/platform/gtk/screen.rs
```diff
@@ -0,0 +1,28 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! GTK Monitors and Screen information.
+
+use crate::screen::Monitor;
+use crate::kurbo::Size;
+
+pub fn get_display_size() -> Size {
```
jneem, line 20: Maybe this should be `pub(crate)`? I think it's inaccessible outside `druid-shell` anyway, because we don't export the `platform` module, but `pub(crate)` would be clearer anyway.

druid-shell/src/platform/windows/application.rs
jneem: I'm not convinced that it's worth putting in a temporary solution for this, given that we want a more permanent one soonish anyway.

Actually, this looks like a good place (and mechanism) for handling the permanent solution: the only difference is that we queue up blocking operations instead of messages: so there'd be an
```
enum BlockingOp {
   SaveDialog(...),
   LoadDialog(...),
   SetSize(...),
   // anything else?
}
```

Each window would queue up these operations and run them on `DS_HANDLE_DROPPED`

druid-shell/src/platform/windows/window.rs
```diff
@@ -874,6 +891,19 @@ impl WndProc for MyWndProc {
                     None
                 }
             }
+            DS_HANDLE_DROPPED => {
+                if let Ok(mut d) = self.dropped.try_borrow_mut() {
+                    let m = d.pop_front();
```
jneem, line 896: This looks like it's just taking one thing from the queue. Why not all of them?

druid-shell/src/platform/web/screen.rs
```diff
@@ -0,0 +1,28 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Monitor and Screen information ignored for web.
+
+use crate::screen::Monitor;
+use crate::kurbo::Size;
+
+pub fn get_display_size() -> Size {
+    //Ignore?
```
jneem, line 21: Maybe it's worth a warning?

druid-shell/src/screen.rs
Zarenor: I think it should be accessible somewhere? On windows, the primary is the one whose top-left corner is the origin, and the user can change it when they reconfigure the desktop. It's not necessarily the 'first' monitor.

druid-shell/src/platform/windows/application.rs
rhzk: How would you detect when the handle is dropped?

druid-shell/src/screen.rs
rhzk: Sure, could implements conversion to dp instead

druid-shell/src/screen.rs
rhzk: Yeah it should be private, will update this

druid-shell/src/platform/gtk/screen.rs
rhzk: I re-exported screen in druid so it would be called by Screen::get_display_size();

druid-shell/src/platform/windows/window.rs
rhzk: This was because it would then return the same return message as the original message, if we run a loop over all of them what should the return be? Some messages return None, some Some(0) and Some(1). Maybe it does not matter?

rhzk:
Do you think it would be better to use display points? I feel like that could be less accurate and maybe problematic across monitors

druid-shell/src/platform/windows/application.rs
jneem: At this point in the event loop, it should always be dropped. That is, we'll ensure that we never return to this loop while holding a `RefCell` borrow.

druid-shell/src/platform/windows/window.rs
jneem: Ok, good point, but then maybe there should be some way in the main event loop to ensure that *all* the queued events get handled? Otherwise you'll get events re-ordered: if two commands get queued (say cmd1 and cmd2) while the borrow is held and one (cmd3) arrives right after it's dropped, then they'll get processed in the order cmd1, cmd3, cmd2.

jneem:
I'm not sure that display points is the right thing (actually, I'm not even sure how we plan to handle two monitors with different DPI in druid, generally speaking). What I'm sure of is that the units should be documented on all the public API, and also we need to think about (and document) what the position means in the presence of multiple monitors (like, is it per-monitor or is it a coordinate on some larger "virtual" screen).

It might also be worth taking a step back and documenting what we expect the screen module to be used for, and why. I'm probably a bad person for understanding the big picture here, because I usually use a tiling window manager and so my windows get very little say on how they get positioned...

rhzk:
Currently the position are more of an abstract/virtual type, since you might want to place a window between 2 monitors. I am thinking we could also add a function that takes a specific monitor and position within that monitor.

For example in the example above i used .set_position(Point::new(-760.0,-1080.0)) to place it in the top-left screen of my 2nd monitor. But we could also have a: 
.set_position_on_monitor(2nd_monitor, 0.0,0.0) that would also place it on top-left corner
(or in some other unit / other coordinate system if that is better) and maybe a better function name

druid-shell/src/platform/windows/window.rs
rhzk: Yeah I thought about this and its not ideal. I am implementing BlockingOp now and I think that should solve this problem

druid-shell/src/platform/gtk/screen.rs
rhzk: Ah yes. I missunderstood you, they should be pub(crate). Will update this

druid-shell/src/platform/windows/window.rs
```diff
@@ -357,6 +370,57 @@ impl MyWndProc {
     fn has_menu(&self) -> bool {
         self.with_window_state(|state| state.has_menu.get())
     }
+
+    // Here we handle messages generated by WindowHandle
+    // that needs to run after borrow is released
+    fn handle_blocked(&self) {
```
rhzk, line 376: This is problematic for `open_file_sync()` and `save_as_sync()` as they need to return `Option<FileInfo>`

Any idea how we could handle `get_file_dialog_path()` here but get the return value in WindowHandle?

jneem: Rather than add a callback, I'd suggest adding methods to the `WinHandler` trait, one for every blocking operation that needs a "reply".

rhzk: Should it then trigger a Event? That would work for widgets, but not for button closure e.g:
```
let button = Button::new("Select File").on_click(|ctx, _data: &mut String, _env| {
        let i = ctx
            .window()
            .clone()
            .open_file_sync(FileDialogOptions::new());
    });
```

jneem: Hm, I might have changed my mind: a callback has certain advantages. But on the other hand, the two-call protocol is being introduced in #1039, and I think it's important to be consistent. @xStrom?

The way the two-call protocol would work, is that druid would call `WindowHandle::open_file_dialog` (or some name like that) and then when the file dialog is closed, druid-shell will call `WinHandler::file_dialog_finished` (or some name like that). Druid's implementation of `WinHandler` will take care of sending the `OPEN_FILE` (or `OPEN_PANEL_CANCELLED`, once #1061 lands) command down the widget tree.

Then we would deprecate `open_file_sync`. Users of druid should open files by sending a `SHOW_OPEN_PANEL` command and then listening for a `OPEN_FILE` command.

jneem: Sorry for the spam; I just realized the problem with a callback: it's hard to do anything useful in a callback, because most of the useful things (like modifying the data, or sending a command) won't pass the borrow checker.

rhzk: Agreed, We can hold of on changing `open_file_sync()` and `save_as_sync()` for later, does not have to be in this PR.

rhzk:
This should close the Window's side of #1044,  #778, #634, #427.
And contains a solution for #1068, but the `open_file_sync()` and `save_as_sync()` functions need to be changed to use it.

rhzk:
Demo of custom titlebar:

custom_titlebar

Demo of DPI:

dpi

rjwittams:
@rhzk  Are you still looking at this? I did a half arsed version of some of this for popups/combos/tooltips on the mac. I will branch off your branch - so if you do look again, probably best not to force push

druid-shell/src/screen.rs
```diff
@@ -0,0 +1,94 @@
...
+impl Screen {
+    /// Returns a vector of all the [`monitors`] on the system.
+    ///
+    /// [`monitors`]: struct.Monitor.html
+    pub fn get_monitors() -> Vec<Monitor> {
+        platform::screen::get_monitors()
+    }
+
+    /// Returns the total virtual screen space on the system, in pixels.
+    pub fn get_display_size() -> Size {
```
rjwittams, line 91: I think it would make sense for this to be a Rect - the bounding box of all monitors on virtual space. 
On the mac at least you can have screens with origins at negative coordinates vs the 'main screen' which is always at 0, 0 ( bottom left corner as mac coordinates are upside down .. )

rjwittams:
I have the mac parts kind of working for the screen and window module at https://github.com/rjwittams/druid/tree/screen-rjw. 

I think we need to specify upfront that screen coordinates, like all other coordinates, are with y increasing downwards, and Mac needs to translate on the way in and out - otherwise its not possible to have a consistent relative positioning.  

Also to make use of the virtual screen dimensions to do this, its necessary to have not just the height of the virtual screen but the max y coordinate. (Because druid_y = max_screen_y - mac_y ). So I think get_display_size should be get_display_rect(). If there are platforms where screen coords are guaranteed positive, then they just use 0, 0 as the origin

druid-shell/src/screen.rs
rhzk: Just to make sure I understand you correctly, lets say we have a virtual screen where the rect would be
top = -100, left = -100, bottom = 0, right = 0. currently this function would return a Size(100,100) but instead you suggest it should return the Rect(-100, -100, 0, 0) ?

rjwittams: I think so - that way you can still get the size (off the rect) but also know the extents of virtual screen coordinates

rhzk:
Interesting that mac works with inverted y-axis. I agree that the best solution is proably to specify that we have y going downwards and flip it on mac, so that the api is consistant.

If the origin (0, 0) on mac is in the bottom-left corner of the primary screen I suppose we also need to move it to the top left corner.

Given 2 monitors (blue rectangles) with 100x100 in size each the transformation would be something like this for y-coordinates:

img

Thank you for doing the Mac side, I dont own a Mac so doing anything mac related would be like guessing.

rjwittams:
I think that picture is correct. I haven't done the flipping about yet, will do on the macbook with a tv at various orientations!

druid-shell/src/platform/windows/application.rs
```diff
@@ -105,6 +104,11 @@ impl Application {
         unsafe {
             // Handle windows messages
             loop {
+                if let Ok(state) = self.state.try_borrow() {
```
raphlinus, line 107: My experience is that putting *any* additional functionality in the runloop is problematic. The logic doesn't get invoked when the system is using its own runloop, which happens on live resize and also while a file dialog is open. Also, if and when we ever run as a guest (for example as a VST plugin), then we'll also be at the mercy of the host's runloop. But if we use the function sparingly, maybe we're ok.

Also, I think the name "DS_HANDLE_DROPPED" is confusing and should be changed to "BLOCKED" to be consistent with other functions.

See below where I suggest a comment to add.

druid-shell/src/platform/windows/screen.rs
```diff
@@ -0,0 +1,61 @@
...
+    let primary = info.dwFlags == MONITORINFOF_PRIMARY;
+    let rect = Rect::new(info.rcMonitor.left as f64, info.rcMonitor.top as f64, info.rcMonitor.right as f64, info.rcMonitor.bottom as f64);
+    let work_rect = Rect::new(info.rcWork.left as f64, info.rcWork.top as f64, info.rcWork.right as f64, info.rcWork.bottom as f64);
+    let m = Monitor::new(primary, rect, work_rect);
+    MONITORS.push(m);
+    TRUE
+}
+
+
+pub(crate) fn get_monitors() -> Vec<Monitor> {
```
raphlinus, line 50: I believe there is a safety issue here due to the use of static mutable global state, and the fact that this function is exposed up to the user through the `Screen` interface, which is `Send` and `Sync`. One solution is to be more threadsafe by using the dwData mechanism to pass in a pointer to the Vec being mutated, another thing to consider is ensuring that `Screen` can't be used across thread boundaries.

Another possibility for the safety concern is to use a Mutex. But that in turn requires once_cell (or lazy_static, we have deps on both), so it's a little long-winded.

druid-shell/src/platform/windows/util.rs
```diff
@@ -144,6 +146,8 @@ type CreateDXGIFactory2 =
 #[allow(non_snake_case)] // For member fields
 pub struct OptionalFunctions {
     pub GetDpiForSystem: Option<GetDpiForSystem>,
+    pub GetDpiForWindow: Option<GetDpiForWindow>,
+    pub SetProcessDpiAwarenessContext: Option<SetProcessDpiAwarenessContext>,
```
raphlinus, line 150: I was going to suggest cleaning this up a bit, but now I'm not sure. We don't currently use `GetDpiForMonitor`, but I think it is the function we want to use for compatibility on Windows 8.1. Similar for SetProcessDpiAwareness.

druid-shell/src/platform/windows/window.rs
```diff
@@ -212,6 +229,12 @@ const DS_RUN_IDLE: UINT = WM_USER;
 /// time it is handled, we can successfully borrow the handler.
 pub(crate) const DS_REQUEST_DESTROY: UINT = WM_USER + 1;
 
+/// Message relaying a request to handle dropped messages.
```
raphlinus, line 232: I'd rework the name and comment. First, messages aren't dropped if they're placed in the queue. We use the word "blocked" and that's ok, but I think "deferred" would actually be best. Here's my suggestion for the comment:

/// A message which must be delivered deferred due to reentrancy.
///
/// Due to the architecture of Windows, it sometimes delivers messages
/// reentrantly. This is problematic when the mutable state for the window
/// is borrowed, as it's not possible to safely dispatch the message. The two
/// choices are to drop the message, or put it in a queue for deferred
/// processing.
///
/// This mechanism should be used very sparingly, as delivery of these
/// deferred messages can not be guaranteed in a timely and reliable
/// fashion. In particular, we do not run our own runloop when handling
/// live resize, when a modal dialog is open, or when the application is a
/// a guest (such as a VST plugin).

druid-shell/src/platform/windows/window.rs
```diff
@@ -454,11 +596,97 @@ impl WndProc for MyWndProc {
...
+                    warn!(
+                        "failed to get window style: {}",
+                        Error::Hr(HRESULT_FROM_WIN32(GetLastError()))
+                    );
+                    return Some(0);
+                }
+                if !self.has_titlebar() && (style & WS_CAPTION) != 0 {
+                    let s: *mut NCCALCSIZE_PARAMS = lparam as *mut NCCALCSIZE_PARAMS;
+                    if let Some(mut s) = s.as_mut() {
+                        s.rgrc[0].top -= (31.0 * self.scale().x()) as i32;
```
raphlinus, line 647: These hard-coded numbers seem scary to me. I would be more comfortable if they were at least moved to `const` with docstring explaining where they came from.

druid-shell/src/platform/windows/window.rs
```diff
@@ -454,11 +596,97 @@ impl WndProc for MyWndProc {
...
+                        if (style & WS_MAXIMIZE) != 0 {
+                            s.rgrc[0].top += (7.0 * self.scale().x()) as i32;
+                        }
+                    }
+                }
+                // Let the default WinProc handle the message.
```
raphlinus, line 653: If you return `None` here, it invokes the default.

druid-shell/src/platform/windows/window.rs
```diff
@@ -1277,35 +1532,140 @@ impl WindowHandle {
...
+                warn!(
+                    "failed to borrow blocked queue"
+                );
+            }
+        }
+    }
+
+    // Gets the window state.
+    pub fn get_window_state(&self) -> WindowSizeState {
+        // We cant store state internally because it could be modified externaly.
```
raphlinus, line 1640: "can't" and "externally"

druid-shell/src/platform/windows/window.rs
```diff
@@ -121,6 +125,14 @@ pub struct WindowHandle {
     state: Weak<WindowState>,
 }
 
+enum BlockingOp {
```
raphlinus, line 128: I would prefer the name "DeferredOp" but won't block on it (see several other comments).

druid-shell/src/platform/windows/window.rs
```diff
@@ -1277,35 +1532,140 @@ impl WindowHandle {
         }
     }
 
-    // TODO: Implement this
-    pub fn show_titlebar(&self, _show_titlebar: bool) {}
+    pub fn show_titlebar(&self, show_titlebar: bool) {
+        if let Some(w) = self.state.upgrade() {
+            w.has_titlebar.set(show_titlebar);
+            if let Ok(mut q) = w.blocked_queue.try_borrow_mut() {
```
raphlinus, line 1538: Maybe consider a newtype for `BlockedQueue` (or "DeferredQueue" to use my preferred name), and have a method to handle all this boilerplate.

raphlinus:
Actually I'm going to put my comment re reentrancy into 1068.

druid-shell/src/platform/windows/application.rs
rhzk: I wasnt to happy about this aswell. Given that `GetMessage` will block until a message is received its not guaranteed that the deferred messages are run immediately, and like you said on things like live resize the message is handled internally and our loop dont run. But what would be a good solution here? If we where to run a loop in a separate thread or use PeekMessage druid would waste/use a lot more cpu time. Maybe we could send a message whenever something is added to DeferredOps as to trigger the runloop?

druid-shell/src/platform/windows/window.rs
```diff
@@ -641,17 +710,45 @@ impl WndProc for MyWndProc {
...
                 if !self.has_titlebar() && (style & WS_CAPTION) != 0 {
                     let s: *mut NCCALCSIZE_PARAMS = lparam as *mut NCCALCSIZE_PARAMS;
                     if let Some(mut s) = s.as_mut() {
-                        s.rgrc[0].top -= (31.0 * self.scale().x()) as i32;
-                        if (style & WS_MAXIMIZE) != 0 {
-                            s.rgrc[0].top += (7.0 * self.scale().x()) as i32;
+                        if let Some(func) = OPTIONAL_FUNCTIONS.GetSystemMetricsForDpi {
+                            // This function is only supported on windows 10
+                            let dpi = self.scale().x() * 96.;
+                            // Height of the different parts that make the titlebar
```
rhzk, line 734: Note to self: change from 96. to the constant we have

raphlinus (approved):
These changes look good, though I haven't done any testing yet; I'm planning to do that while merging for 1191. Thanks!

raphlinus (changes requested):
I ran this on Windows 10 and unfortunately the dpi scaling on the calc example is broken there. It seems like created window size is not being scaled up 2x.

raphlinus:
Here's a screenshot of calc on Windows 10 with a 2x hi-dpi monitor:

calc

rhzk:
Yeah I found the issue, it was related to dpi changes i did to support 8.1. Will fix it asap

rhzk:
Should be resolved now, thanks for the report :)

rhzk:
Should be resolved now

raphlinus (changes requested):
Unfortunately the "parse" example now seems to have a very tiny window. The "flex" example seems mostly fine though. I say "mostly" because at 2x dpi it starts up with some kind of incorrect rendering (I didn't analyze it) then flash to correct rendering quickly. I'm ok with that being addressed in a followup though.

The flashing doesn't happen in master. I didn't do any serious bisection to figure out which patch caused the regression.

rhzk:
So many edge cases, a solution to one causes a problem for another..

Thanks for the report, should be fixed

rhzk:
Yeah the flashing occurs because I need the hwnd to apply everything correctly, so some window changes happens right after the window is created. I tried some solutions for this earlier that did not work, so I will need some time to find a good solution.

raphlinus (approved):
This is still not perfect. For windows created with default size (as in the "parse") example, it is initially created with a reasonable size, then a resize to a size that's 2x a reasonable size. For windows with given size (as in "flex") it's initially the 1x, ie half the intended size, then it resizes to 2x.

I'm going to click "approved" because this is a usable state and I really want to unblock other work, and think the fix probably won't cause conflicts. I remember researching this months ago and realizing it's a complicated problem - you don't really know what the dpi is until you know what monitor the window will appear on. This is one reason I didn't take  on dynamic dpi switching back then :) If I were fixing this myself, I'd probably use the system dpi as a first guess, but that might create flashing and other issues in mixed-dpi configurations.

I'll start the simpler_d2d merge as soon as this is merged.

Thanks!

raphlinus:
I've done some digging and see the problem pretty clearly: the `handle.set_size(handle.get_size())` at the end of `build()` is taking a pixel size and interpreting it as scaled pixels. The fundamental problem here is that `self.size` is not a real `Size` value, but rather may have `CW_USEDEFAULT` values in it. A workable fix is to call `handle.set_size(self.size)` but only when it's a valid value.

That leaves a couple problems, one is the flashing, the other is that there's functionality for setting only one of width and height that we're not surfacing, but those are much smaller potatoes.

rhzk:
Yeah that is the problem, currently the window is created with the specified size without any dpi-scaling, then we get the DPI for the window and fix the scaling. There is currently 2 problems with this, one being the "flashing" and the other being the size windows suggest as a default size, as this ends up way to big when scaled.

The first issue is complicated to solve and I dont yet have a solution for it, the second issue should be easy to solve

raphlinus:
If you call SetWindowPos synchronously at the end of the `build` function I'm not seeing any flashing, but calling WM_CREATE with a wrong value still doesn't feel entirely right.

rhzk:
I think there is still flashing, it just happens very fast. Ideally I want to avoid it all together

raphlinus:
FWIW I looked at the winit code, and they seem to create the window using CW_USEDEFAULT for all geometry, then use SetWindowPos after window creation to fix up the size.

raphlinus:
My read of the code is that we'll still get a good size for the window on dpi change even if we start with CW_USEDEFAULT and don't ever set it explicitly, because it's based on the "suggested size" from the WM_DPICHANGE message. But we can't test that while Direct2D is all crashy.

raphlinus:
As discussed on Zulip, I'm going to merge this now even though it has some rough ends, then merge #1191 on top of it, and propose a PR to fix the size behavior, all in the interest of getting master to a reasonably non-broken state quickly. Thanks for all the work on this PR, and apologies if that steps on your toes at all.

PR #1039, opened 2020-06-15
Add alerts and their macOS implementation.

### LINKED ISSUE #383: add support for showing error/alert dialogs
initially this can just be dialogs that display an error message; at some point we will also want to have dialogs that present multiple options.

xStrom:
## API

This PR adds a cross-platform alert API that will enable a good user experience on Windows/macOS/Linux.

The API is designed in a way that a single `AlertOptions` configuration will result in a native feeling alert on all these platforms. The buttons are ordered automatically, so none of this *windows-app-on-mac* or vice versa feeling when you encounter an alert with a weird layout.

The API also favors a good user experience with contextualized (and translated) button labels. This is something that all the platform guideline documents agree on. Yes you can find plenty of legacy apps with basic labels like *Confirm* or *Yes* but neither of those are actually recommended.

## macOS implementation

This PR also contains the macOS implementation of this API to get things started and show that the API works.

I am aware of at least two macOS `druid-shell` issues that show up in relation to the alerts work. These issues aren't directly related to the alerts API though so I am tackling them separately. The first one is timers not firing when modal dialogs are up, which is resolved in #1028. When that fix is applied another issue surfaces - `Application::quit` doesn't properly end the runloop when an app modal dialog is up - it just ends the modal runloop. This is an issue I'm still investigating but will also solve separately as it's more of a bug in the `quit` method.

## Alert example

I also added a new `alert` example. It shows a few different ways to use alerts, although some of it can feel needlessly complex for an example. I built it more as a testing tool to see how edge cases are handled when multiple alerts show up in various ways. This made me think that perhaps we should have an `examples` directory but also some sort of `testing` directory of example-like apps that help us confirm whether various implementations are working properly. For now, I think the `alert` example can live as an example, but this distinction is something to think about for the future.

## Future work

I actually started this alerts work with the Windows implementation, which I've had 80% done for months now. This helped me really understand the cross-platform requirements. Moving forward I'll finish up the Windows implementation and post that as a separate PR.

---

Fixes #383.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ You can find its changes [documented below](#060---2020-06-01).
 
 ### Added
 
+- Alert dialogs for time sensetive notifications. ([#1039] by [@xStrom])
+
 ### Changed
 
 - `Image` and `ImageData` exported by default. ([#1011] by [@covercash2])
@@ -334,6 +336,7 @@ Last release without a changelog :(
 [#1018]: https://github.com/xi-editor/druid/pull/1018
 [#1025]: https://github.com/xi-editor/druid/pull/1025
 [#1028]: https://github.com/xi-editor/druid/pull/1028
+[#1039]: https://github.com/xi-editor/druid/pull/1039
 [#1042]: https://github.com/xi-editor/druid/pull/1042
 [#1043]: https://github.com/xi-editor/druid/pull/1043
 [#1050]: https://github.com/xi-editor/druid/pull/1050
--- a/druid-shell/src/alert.rs
+++ b/druid-shell/src/alert.rs
@@ -1,0 +1,386 @@
+// Copyright 2020 The xi-editor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Alert dialogs.
+
+use crate::util::ConstString;
+
+/// A token that uniquely identifies an alert dialog.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Hash)]
+pub struct AlertToken(usize);
+
+/// Contains the result of the alert after it was closed.
+#[derive(Debug, Clone)]
+pub struct AlertResponse {
+    token: AlertToken,
+    button: AlertButton,
+}
+
+/// Options for alert dialogs.
+///
+/// An alert dialog is a modal dialog that provides the user with some information
+/// and optionally asks the user to make some decision.
+///
+/// # User experience
+///
+/// You should try to minimize the use of alerts. Alerts interrupt whatever task
+/// the user is currently doing. Only use alerts for time sensitive information.
+/// If there is some piece of information that the user must know immediately,
+/// e.g. that storage space is running very low and will soon cause a critical error,
+/// or if the storage space has already depleted and the critcial error has happened.
+/// Alerts are also useful when there is some decision that the user must make
+/// and the application can't continue without making that decision. Even so
+/// you should think hard when designing your application whether you can have
+/// reasonable default behavior which can happen without alerts. It's better to
+/// offer an *undo* feature than to ask for confirmation on deletion.
+///
+/// The fewer alerts the user encounters, the better their experience.
+///
+/// # Modality
+///
+/// While an alert dialog is open it takes interaction priority and the user won't be
+/// able to continue interacting with the parent window while the alert is open.
+///
+/// Multiple alert dialogs can be open at once, but only the most recent one will be interactable.
+/// When that alert dialog is closed, the next most recent will become interactive.
+#[derive(Debug, Clone, Default)]
+pub struct AlertOptions {
+    /// The context of the alert.
+    pub(crate) context: String,
+    /// The primary message of the alert.
+    pub(crate) message: String,
+    /// The supplemental text of the alert.
+    pub(crate) description: String,
+    /// The icon of the alert.
+    pub(crate) icon: Option<AlertIcon>,
+    /// The buttons to be shown on the alert.
+    pub(crate) buttons: Vec<AlertButton>,
+}
+
+/// Alert dialog icon.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum AlertIcon {
+    /// The alert dialog is providing important time sensitive information.
+    Information,
+    /// The alert dialog is warning about a potential future problem.
+    Warning,
+    /// The alert dialog is informing of a critical problem that has already happened.
+    Error,
+}
+
+/// Alert dialog button type.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum AlertButtonType {
+    /// Positive buttons are used for choices that agree with the primary alert message.
+    Positive,
+    /// Negative buttons are used for choices that disagree with the primary alert message.
+    Negative,
+    /// Cancel buttons reflect the user's desire to not answer at all
+    /// and instead cancel the task which initiated the alert.
+    Cancel,
+}
+
+/// A specific button for the alert dialog.
+///
+/// There are three types of buttons:
+/// - **Positive** - for choices that agree with the primary alert message.
+/// - **Negative** - for choices that disagree with the primary alert message.
+/// - **Cancel** - used by the user to show a desire to not answer at all
+///   and instead cancel the task which initiated the alert.
+///
+/// If the alert is just reporting information and it only needs a single button,  
+/// then you should use a positive button with the label **OK** or a translation of it.
+///
+/// The cancel button should always be labeled **Cancel** or a translation of it.
+///
+/// In all other cases you should give a short but descriptive label that includes a verb.
+/// The label should let the user know what will happen when they click the button.
+/// Assume that the user didn't actually read the alert message. This assumption ends up
+/// true more often than not, because alerts by design appear unexpectedly while the user
+/// is trying to accomplish some other task.
+///
+/// Labels like **Send message**, **View all**, **Reconnect**, **Remind me tomorrow**, and
+/// **Delete file** make it much more clear what will happen compared to a generic **OK**,
+/// **Confirm**, or **Yes**.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct AlertButton {
+    pub(crate) button_type: AlertButtonType,
+    pub(crate) label: ConstString,
+}
+
+impl AlertToken {
+    /// A token that does not correspond to any alert.
+    pub const INVALID: AlertToken = AlertToken(0);
+
+    /// Create a new token with the specified `id`.
+    pub(crate) const fn new(id: usize) -> AlertToken {
+        AlertToken(id)
+    }
+}
+
+impl AlertResponse {
+    /// Create a new alert response.
+    pub(crate) fn new(token: AlertToken, button: AlertButton) -> AlertResponse {
+        AlertResponse { token, button }
+    }
+
+    /// Returns the [`AlertToken`] that identifies the alert.
+    ///
+    /// [`AlertToken`]: struct.AlertToken.html
+    #[inline]
+    pub fn token(&self) -> AlertToken {
+        self.token
+    }
+
+    /// Returns the [`AlertButton`] that closed the alert.
+    ///
+    /// # Canceled by the platform
+    ///
+    /// If the platform cancels the alert, e.g. when the parent window gets closed,
+    /// then this will return a cancel button that was defined in [`AlertOptions`].
+    /// If no cancel button was provided then this will be [`AlertButton::CANCEL`].
+    ///
+    /// [`AlertButton`]: struct.AlertButton.html
+    /// [`AlertOptions`]: struct.AlertOptions.html
+    /// [`AlertButton::CANCEL`]: struct.AlertButton.html#associatedconstant.CANCEL
+    #[inline]
+    pub fn button(&self) -> &AlertButton {
+        &self.button
+    }
+}
+
+impl AlertButton {
+    /// A positive button with the English label **OK**.
+    pub const OK: AlertButton = AlertButton::const_positive("OK");
+    /// A cancel button with the English label **Cancel**.
+    pub const CANCEL: AlertButton = AlertButton::const_cancel("Cancel");
+
+    /// Create a new positive button with the specified `label`.
+    ///
+    /// If this is the only button of the alert, use **OK** or a translation of it as the `label`.
+    /// Otherwise you should give a short but descriptive label like **Save the draft** that lets
+    /// the user know what will happen when they click the button.
+    ///
+    /// See [`AlertButton`] for more information.
+    ///
+    /// [`AlertButton`]: struct.AlertButton.html
+    pub fn positive(label: impl Into<String>) -> AlertButton {
+        AlertButton {
+            button_type: AlertButtonType::Positive,
+            label: ConstString::new(label),
+        }
+    }
+
+    /// Create a new negative button with the specified `label`.
+    ///
+    /// You should give a short but descriptive label like **Delete the draft** that lets the user
+    /// know what will happen when they click the button.
+    ///
+    /// See [`AlertButton`] for more information.
+    ///
+    /// [`AlertButton`]: struct.AlertButton.html
+    pub fn negative(label: impl Into<String>) -> AlertButton {
+        AlertButton {
+            button_type: AlertButtonType::Negative,
+            label: ConstString::new(label),
+        }
+    }
+
+    /// Create a new cancel button with the specified `label`.
+    ///
+    /// The cancel button `label` should always be **Cancel** or a translation of it.
+    ///
+    /// See [`AlertButton`] for more information.
+    ///
+    /// [`AlertButton`]: struct.AlertButton.html
+    pub fn cancel(label: impl Into<String>) -> AlertButton {
+        AlertButton {
+            button_type: AlertButtonType::Cancel,
+            label: ConstString::new(label),
+        }
+    }
+
+    /// Create a new `const` positive button with the specified `label`.
+    ///
+    /// This is a specialized `const` version of the [`positive`] function.
+    ///
+    /// See [`AlertButton`] for more information.
+    ///
+    /// [`positive`]: #method.positive
+    /// [`AlertButton`]: struct.AlertButton.html
+    pub const fn const_positive(label: &'static str) -> AlertButton {
+        AlertButton {
+            button_type: AlertButtonType::Positive,
+            label: ConstString::from_static(label),
+        }
+    }
+
+    /// Create a new `const` negative button with the specified `label`.
+    ///
+    /// This is a specialized `const` version of the [`negative`] function.
+    ///
+    /// See [`AlertButton`] for more information.
+    ///
+    /// [`negative`]: #method.negative
+    /// [`AlertButton`]: struct.AlertButton.html
+    pub const fn const_negative(label: &'static str) -> AlertButton {
+        AlertButton {
+            button_type: AlertButtonType::Negative,
+            label: ConstString::from_static(label),
+        }
+    }
+
+    /// Create a new `const` cancel button with the specified `label`.
+    ///
+    /// This is a specialized `const` version of the [`cancel`] function.
+    ///
+    /// See [`AlertButton`] for more information.
+    ///
+    /// [`cancel`]: #method.cancel
+    /// [`AlertButton`]: struct.AlertButton.html
+    pub const fn const_cancel(label: &'static str) -> AlertButton {
+        AlertButton {
+            button_type: AlertButtonType::Cancel,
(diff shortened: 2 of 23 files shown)


justinmoon:
Tested the example on MacOS and everything worked very nicely. Looked very "native".

Tested the example on Manjaro Linux with i3 desktop environment and:
- "New window" worked.
- "Quit in style" quits the app after ~3 second without showing any alerts.
- Nothing else worked.

ratmice:
>When that fix is applied another issue surfaces - Application::quit doesn't properly end the runloop when an app modal dialog is up - it just ends the modal runloop. This is an issue I'm still investigating but will also solve separately as it's more of a bug in the quit method.

Interesting... It looks like this is because the stop method only applies to the current runloop https://github.com/xi-editor/druid/blob/6a429144309cef53c5d8c772b9e9c25416c50cee/druid-shell/src/platform/mac/application.rs#L96 

It's not clear to me why this isn't just using `terminate:`, so perhaps that or calling stop via `performSelectorOnMainThread:withObject:waitUntilDone:`

xStrom:
> It's not clear to me why this isn't just using `terminate:`,

Because terminate kills the process and cleanup code won't run.

> so perhaps that or calling stop via `performSelectorOnMainThread:withObject:waitUntilDone:`

That's not enough either. I've found a way to stop two levels of runloops with `abortModal` + `stop`, however not three levels yet. I'll try some sort of handle based targeting soon probably.

druid-shell/src/platform/mac/alert.rs
```diff
@@ -0,0 +1,120 @@
...
+    }
+}
+
+unsafe fn handle_result(view: &mut Object, request: &AlertRequest, tag: NSInteger) {
+    let button = match tag {
+        BUTTON_ID_PRIMARY => Some(request.options.primary.clone()),
+        BUTTON_ID_CANCEL | NSModalResponseAbort | NSModalResponseStop | NSModalResponseCancel => {
+            None
+        }
+        id => {
```
cmyr, line 102: I like to include `if` conditions as part of the branch, if possible. I see why that's annoying here, because the conditional is so verbose; one option would be to add a simple function, `fn is_valid_id(id: NSInteger, request: &AlertRequest) -> bool`, so that we can just do the match as `id if is_valid_id(id, request) => { .. }`.

not a blocker or anything, just one of those little refactors that might make this slightly easier to read, not that it's particularly tricky or anything. :)

druid-shell/src/platform/mac/alert.rs
```diff
@@ -0,0 +1,120 @@
...
+            }
+        }
+    };
+    let view_state = {
+        let view_state: *mut c_void = *view.get_ivar("viewState");
+        &mut *(view_state as *mut ViewState)
+    };
+    (*view_state)
+        .handler
+        .alert_response(AlertResponse::new(request.token, button));
```
cmyr, line 119: From a casual look it feels like this is the only unsafe in this function? In that case I might either split it into a separate `unsafe` function or else make this function `safe` and just use an `unsafe` block here.

Edit: my thinking here is that we always want to make unsafe blocks as small as possible, because they demand much more scrutiny during review or even just reading the code; in particular future changes to this function have to be carefully reviewed for soundness.

druid-shell/src/platform/mac/appkit.rs
```diff
@@ -73,3 +75,97 @@ impl NSView for id {
         msg_send![self, addTrackingArea: trackingArea]
     }
 }
+
+pub trait NSButton: Sized {
```
cmyr, line 79: I don't mind this, but I've also found the trait-based patterns used in the `Cocoa` crate tend not to not feel quite worth it, API-wise, and I tend to just just `msg_send` directly. ymmv.

druid-shell/src/platform/mac/appkit.rs
```diff
@@ -73,3 +75,97 @@ impl NSView for id {
...
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum NSAlertStyle {
+    Warning,
+    Informational,
+    Critical,
+}
+
+pub const NSModalResponseContinue: NSInteger = -1002;
```
cmyr, line 106: I often like to include a link to my source when I'm adding constants like this.

druid-shell/src/alert.rs
```diff
@@ -0,0 +1,451 @@
...
+///
+/// On macOS the information and warning icons are the same.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum AlertIcon {
+    /// The alert dialog is providing important time sensitive information.
+    Information,
+    /// The alert dialog is warning about a potential future problem.
+    Warning,
+    /// The alert dialog is informing of a critical problem that has already happened.
+    Error,
```
cmyr, line 115: I would mark this as non-exhaustive. In particular we will likely want at some point the option of supplying a custom icon?

druid-shell/src/alert.rs
```diff
@@ -0,0 +1,451 @@
...
+///
+/// The cancel button should always be labeled **Cancel** or a translation of it.
+///
+/// In all other cases you should give a short but descriptive label that includes a verb.
+/// The label should let the user know what will happen when they click the button.
+/// Assume that the user didn't actually read the alert message. This assumption ends up
+/// true more often than not, because alerts by design appear unexpectedly while the user
+/// is trying to accomplish some other task.
+///
+/// Labels like **Send message**, **View all**, **Reconnect**, **Remind me tomorrow**, and
```
cmyr, line 131: I might also include links to things like the macOS human interface guidelines.

druid-shell/src/alert.rs
```diff
@@ -0,0 +1,451 @@
...
+
+    /// Create a new `const` button with the specified `label`.
+    ///
+    /// You should give a short but descriptive label like **Send message** that lets
+    /// the user know what will happen when they click the button.
+    ///
+    /// See [`AlertButton`] for more information.
+    ///
+    /// [`AlertButton`]: struct.AlertButton.html
+    pub const fn new(label: &'static str) -> AlertButton {
```
cmyr, line 193: Thinking out loud: at least from `druid` we're going to want to be (eventually) pushing localized text wherever possible, and I wonder if we should have like `static` and `new` instead of `new` and `dynamic`? That is, having the method folks are most likely to use be the flexible one.

I'm also not especially concerned about the allocation cost, given the use case.

druid-shell/src/alert.rs
```diff
@@ -0,0 +1,451 @@
...
+pub(crate) struct AlertRequest {
+    pub(crate) token: AlertToken,
+    pub(crate) options: AlertOptions,
+}
+
+/// Contains the result of the alert after it was closed.
+#[derive(Debug, Clone)]
+pub struct AlertResponse {
+    token: AlertToken,
+    button: Option<AlertButton>,
```
cmyr, line 34: I presume that the button is used to resolve what action was taken?

My inclination would be to not use `AlertButton` for that, and instead have the user pass in some other token for each option? Although now that I write this out I see how annoying it might be.

I guess this comes down to my concern about using string equality to check for intent, I could imagine cases (maybe?) where a locale changes while a modal is up and the user checks against a changed string, or similar?

Perhaps a more reasonable concern is that this doesn't work as well when the buttons are localized, because you wan't have a `const` `AlertButton` sitting around.

So I'm not convinced either way, but would be curious to hear your thoughts.

druid-shell/src/platform/mac/appkit.rs
xStrom: I've found the discoverability of `msg_send` rather tricky, because there's no direct documentation of using the macOS API through it. It requires a bunch of understanding how Objective-C works etc. With a Rust API you can know just Rust.

xStrom: As an example, [here's the Apple docs](https://developer.apple.com/documentation/appkit/nsalert/1524296-beginsheetmodal) for `beginSheetModal(for:completionHandler:)`.

Took me a while to figure out that the declaration given:

```objc
func beginSheetModal(for sheetWindow: NSWindow, 
   completionHandler handler: ((NSApplication.ModalResponse) -> Void)? = nil)
```

Translates into:

```rust
// completionHandler: RcBlock<(NSInteger,), ()>
msg_send![self, beginSheetModalForWindow: window completionHandler: completionHandler]
```

druid-shell/src/platform/mac/appkit.rs
xStrom: I think that's a good idea in general, however I found these specific constants through a combination of testing+println and digging through random Objective-C/Swift apps on the internet. So no authoritative source to link.

I haven't found Apple docs that define constants like they are available from Microsoft for example. If there are any that exist, that would be great to know. A [few constants here and there are defined](https://developer.apple.com/documentation/appkit/nsalert/style/informational) on developer.apple.com but most aren't.

druid-shell/src/alert.rs
xStrom: I am definitely open to this, in an earlier iteration I had `const_new` and `new`. My reasoning for changing it was that most apps aren't going to localize anything.

I think `static` is a reserved keyword just like `const`. So the method name would have to be something different. Any ideas?

druid-shell/src/platform/mac/appkit.rs
cmyr: a tip about the docs; always switch to the objective-c version, which will match much more closely the code that you end up writing with `msg_send`.

druid-shell/src/platform/mac/appkit.rs
cmyr: so what I like to do here is grep in `/Library/Developer` and you'll often find the header file where something is declared, and these header files often include helpful documentation.

druid-shell/src/alert.rs
cmyr: I think `const_new` is good enough.

druid-shell/src/alert.rs
xStrom: This is a pain point, I wrestled with various designs and I agree this is not perfect. Right now the idea is that you either have a `const AlertButton` or you store the localized `AlertButton` and check against that. This stored one wouldn't change when the locale changes, but it can be a bit of a pain to plumb around because it does not implement `Copy`.

I was thinking of also having some sort of `AlertButtonToken` but that runs into the same issue that commands/selectors do, in that this token would have to be a manually specified string to support `const`. Also having a token would make things less streamlined for const buttons.

Thinking about this some more now, perhaps there is a way to make this somewhat better in a way that wouldn't decrease the ergonomics of const buttons.

We could define that a const button's token is equal to its label. Then `AlertResult` could have `Option<AlertButtonToken>` and we could also support `AlertButton == AlertButtonToken` by using the button's underlying token for equality. To create localized buttons you would have to supply both a token to identify the button, and a label. The `AlertButtonToken` would implement `Copy` and be able to be `const`. The ergonomics win from having such an easily usable token might be greater than the ergonomics loss of having to define the tokens.

druid-shell/src/alert.rs
xStrom: Yeah custom icons could come later. However I'm not sure why anyone would be matching on this enum, as we only take it as input and never output it. So right now it would just create an additional useless match arm for `druid-shell` itself.

cmyr: I guess this is more a philosophical point; if this is exposed publicly then any addition would be a semvar major change. Maybe not a practical concern right now, but something I like to keep in mind.

druid-shell/src/alert.rs
cmyr: Here's another consideration:

there are special cases, like 'ok' and 'cancel', where in general we want to let the platform provide the strings and handle the localization, (If possible?) and so we'd want to have some nice way of handling this.

some thoughts: we could maybe stick with the current mechanism, but have the "action identifier" or something be different from the display text? And maybe we could also have convenient `is_confirm()` and `is_cancel()` methods?

honestly... I defer to you here, you've been thinking about it more, I don't think that from a distance I'm going to come up with some sort of better solution.

xStrom: I'll play around a bit more with this API and see if anything seems more ergonomical. The idea of `is_cancel` is there already in form of `None` but I guess an additional method to `AlertResponse` wouldn't hurt much. A similar method for the primary button could also be worth it.

---

I don't think that letting the platform localize some buttons is a good idea. I've used apps like that and it's always a horrible user experience because the platform and app won't have a matching set of translated languages. Solving this is one of the key goals I personally have regarding localization.

If the app doesn't have the language but the platform has, then the alert message (and some buttons!) will be in English while OK/Cancel will be translated. The other way around you could have the message and some buttons translated and then suddenly some English OK/Cancel thrown in to the mix because the platform doesn't have that language. All of this gets even more insane once you mix RTL languages in.

The app should have full localization control. This will allow the user experience in the app to be consistent. If the app doesn't support a locale, then at least everything will be in a locale it supports. On the other hand if it does support a specific locale, then the platform won't be holding it back.

The user should also be able to override the language in the app. Nothing pisses me off more than when an app translates everything into my locale (at Google translate level no less) and has no way to change it. Also in today's globalized world many people have organization provided computers where the language is stuck at whatever the organization has chosen. For example my girlfriend works for a Danish organization and has a Danish language Windows - but she doesn't speak Danish. Not being able to override this choice in apps (or having some buttons suddenly be in Danish) is a very bad experience.

cmyr: That makes sense, the only annoyance is that we then want to ship localization files that have the same strings as the platform localization files for every supported locale, which isn't the end of the world.

cmyr:
An additional thought I missed in my original review: I'll try and use this in runebender shortly and see how it feels, that will hopefully give me better perspective.

druid-shell/src/alert.rs
```diff
@@ -0,0 +1,451 @@
...
+    /// [`AlertButton`]: struct.AlertButton.html
+    pub fn dynamic(label: impl Into<String>) -> AlertButton {
+        AlertButton {
+            label: Cow::Owned(label.into()),
+        }
+    }
+}
+
+impl Default for AlertOptions {
+    /// Create a default set of alert options.
```
Finnerale, line 219: I think a short overview of what default means for `AlertOptions` would be good. I.e. no icon and only an OK button.

druid-shell/src/lib.rs
```diff
@@ -41,6 +42,7 @@ mod platform;
 mod scale;
 mod window;
 
+pub use alert::{AlertButton, AlertIcon, AlertOptions, AlertResponse, AlertToken};
```
Finnerale, line 45: Our `crate` module is getting quite convoluted, maybe we should stop reexporting everything?

druid-shell/src/alert.rs
```diff
@@ -0,0 +1,451 @@
...
+/// The label should let the user know what will happen when they click the button.
+/// Assume that the user didn't actually read the alert message. This assumption ends up
+/// true more often than not, because alerts by design appear unexpectedly while the user
+/// is trying to accomplish some other task.
+///
+/// Labels like **Send message**, **View all**, **Reconnect**, **Remind me tomorrow**, and
+/// **Delete file** make it much more clear what will happen compared to a generic **OK**,
+/// **Confirm**, or **Yes**.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct AlertButton {
```
Finnerale, line 135: According to the  [style guide](https://doc.rust-lang.org/1.0.0/style/style/naming/README.html#avoid-redundant-prefixes-[rfc-356]) we should be calling these just `Button` or `Options` and then use them like `alert::Button` or `alert::Options`.

This will of course mean that the naming will collide with `Button` if you were to import everything into the same scope, but I don't think typing `alert::Button` is worse than `AlertButton`.

This would also make sens if we would not reexport everything in the crate root, because then you would have to use `alert::AlertButton` which is pretty redundant.

So if we want to continue reexporting everything, `AlertButton` makes sense, otherwise we should remove the prefix.

druid-shell/src/alert.rs
xStrom: These links tend to 404 annoyingly fast and we also don't want to leave the impression that following any one platform's guidelines will lead to good results everywhere.

That said, I'm not trying to *hide* this information so I'll add a section linking to Apple/Microsoft/Gnome guidelines with a note about cross-platform considerations.

druid-shell/src/alert.rs
```diff
@@ -0,0 +1,451 @@
...
+    pub fn button(&self) -> Option<&AlertButton> {
+        self.button.as_ref()
+    }
+}
+
+impl AlertButton {
+    /// A primary button with the English label **OK**.
+    pub const OK: AlertButton = AlertButton::new("OK");
+    /// A cancel button with the English label **Cancel**.
+    pub const CANCEL: AlertButton = AlertButton::new("Cancel");
```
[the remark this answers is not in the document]
xStrom: Having these increases the ergonomics of prototyping and English-only apps. The generic **OK** (or a translation) is also the recommended label for a single button alert.

I fully support proper localization and making that as easy as possible. I thought a lot about it while designing this API. However I'm also a realist. We're not going to make everyone create localized apps overnight. There will be plenty of developers who want their English-only super easy alerts. Having these two buttons is the compromise. They're extremely common and simplify creating alerts for simple toy apps or prototypes.

Showing a simple error message with an OK is as easy as:
```rust
ctx.alert(AlertOptions::error().message("Bad things happened!"));
```
Then when these developers want more buttons they will run into the whole localization / proper button label business.

Also `AlertOptions::cancelable` isn't enough for a near-future API addition to allow configuring the default button. The idea for that is something like `opts.default(button)`.

druid-shell/src/alert.rs
```diff
@@ -0,0 +1,451 @@
...
+/// In window scope this means the user can't interact with the parent window
+/// and in application scope this means the user can't interact with any other window.
+///
+/// # Multiple alerts
+///
+/// There is no guaranteed behavior when multiple alert dialogs are requested with
+/// overlapping modality scope. Depending on platform specifics, all the alerts
+/// might be visible or just some of them. Regardless of visibility, only one of
+/// the alert dialogs will be interactable. When that alert dialog is closed,
+/// another one will become interactable.
```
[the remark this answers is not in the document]
xStrom: So I think alerts should be rarely used to begin with. There's already a statement to that effect saying *You should try to minimize the use of alerts.*

Now when there is good reason to use alerts, they can and should be used. If multiple of them show up at once with overlapping modality, then it's not ideal but it works.

I'm not sure that an extra note about this is worth it. Having the developer do some sort of reference counting on whether a previous alert of a window has closed or not to postpone a new alert that they want to eventually send anyway -- it just doesn't sound like a good idea. Right now we just send all these requests to the platform and the platform chooses whether to delay showing the newest one or not. This note here is about the platform logic not being consistent - sometimes the alerts are delayed but other times not.

druid-shell/src/lib.rs
xStrom: It's definitely a worthwhile question. For now I'm just following what we've been doing.

If we decide to organize them differently then that would also mean a ton of broken docs links though. So I think even if we want to change that and move things into modules, we should wait until better doc linking is stabilized which [looks like](https://github.com/rust-lang/rust/issues/43466) it might happen this year.

druid-shell/src/alert.rs
xStrom: I think for this PR they should remain like `AlertButton` as we have `WidgetId`, `FileInfo`, `FlexParams` etc.

However this is worth having a discussion over whether we should rename everything and also stop re-exporting everything to the root namespace and organize things more nicely into modules.

We should open a separate issue for this as a general problem+discussion. Feel free to get that started.

druid-shell/src/alert.rs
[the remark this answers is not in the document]
Finnerale: Sounds reasonable, somehow I did not think about simple notice dialogs, but yeah for those `OK` makes perfect sense.

druid-shell/src/alert.rs
Finnerale: I've opened an issue for it (#1051).

cmyr:
Just had a thought here: what if you passed a `Command` along with each alert option? then we could not worry about tokens etc, and just handle the command as usual?

cmyr:
An update: I implemented a "you have unsaved changes" dialog when trying to quit in runebender, and I'm fairly happy with the result. I do think this idea of just associating a command with each button could be helpful. If we wanted to be very rigorous, we might also want to use a token to identify a specific modal; I'm not sure how necessary this is in practice.

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ You can find its changes [documented below](#060---2020-06-01).
 
 ### Added
 
+- Alert dialogs for time sensetive notifications. ([#1039] by [@xStrom])
```
bjorn3, line 12: nit:
```suggestion
- Alert dialogs for time sensitive notifications. ([#1039] by [@xStrom])
```

jneem:
What's the status of this? Are there any major changes to the API that need to be made? If not, I'd like to resolve the conflicts and get this merged. (My main limitation is that I can't test on mac, so I don't want to make non-trivial changes to the mac part)

raphlinus:
Given that the original author has gone radio silent (I hope you're ok, @xStrom, and just taking a break; in any case we appreciate your work on this), I'd be very grateful for you to champion getting this merged, especially as it's a very often requested feature. I'm willing to take on mac testing if nobody else steps up.

cmyr:
closing this in favor of #1256

xStrom:
I am now back and will update this PR shortly. I have some API improvement ideas that weren't pushed that I'll also review - and I'll read through the feedback and address that as well.

arifd:
How is progress on this coming along? Is there anything I can do to help? Basically alerts and `PasswordInput` are the only two things blocking me from use Druid in real world applications.

cmyr:
Yea, I need to pick this up and get it finished.

PR #1042, opened 2020-06-18
Simplify Scale by removing DPI.

xStrom:
This PR removes any notion of DPI from `Scale` and makes it strictly scale factor based.

Previously `Scale` always considered 96 as the target DPI but with this change that decision is in the platform code's hands. This will allow for easier implementation of future platforms that have a different target DPI like Android with 160.

This also means that all the DPI-related stuff can be an implementation detail and it doesn't need to be surfaced to developers using druid. Especially as DPI doesn't really match any actual inches etc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,12 +12,13 @@ You can find its changes [documented below](#060---2020-06-01).
 ### Changed
 
 - `Image` and `ImageData` exported by default. ([#1011] by [@covercash2])
+- `Scale::from_scale` to `Scale::new`, and `Scale` methods `scale_x` / `scale_y` to `x` / `y`. ([#1042] by [@xStrom])
 
 ### Deprecated
 
 ### Removed
 
-- DPI related functions/methods from `Scale`. ([#1042] by [@xStrom])
+- `Scale::from_dpi`, `Scale::dpi_x`, and `Scale::dpi_y`. ([#1042] by [@xStrom])
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ You can find its changes [documented below](#060---2020-06-01).
 
 ### Removed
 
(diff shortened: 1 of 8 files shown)


druid-shell/src/platform/gtk/window.rs
```diff
@@ -169,12 +172,13 @@ impl WindowBuilder {
         window.set_resizable(self.resizable);
         window.set_decorated(self.show_titlebar);
 
-        // Get the GTK reported DPI
-        let dpi = window
+        // Get the scale factor based on the GTK reported DPI
+        let scale_factor = window
             .get_display()
             .map(|c| c.get_default_screen().get_resolution() as f64)
```
jneem, line 178: Doesn't `get_resolution` already given an `f64`?

druid-shell/src/platform/gtk/window.rs
```diff
@@ -47,6 +47,9 @@ use super::dialog;
 use super::menu::Menu;
 use super::util;
 
+/// When the GTK reported DPI differs from this we will scale coordinates to achieve it.
+const SCALE_TARGET_DPI: f64 = 96.0;
```
Finnerale, line 51: I think it would be good to add a not that this is the value reported by GTK (or Windows below) when the scale is 1x in both directions. At least that's how I interpret this constant.

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ You can find its changes [documented below](#060---2020-06-01).
 
 ### Removed
 
+- DPI related functions/methods from `Scale`. ([#1042] by [@xStrom])
+
```
Finnerale, line 21: And renamed `from_scale` to `new`, `Scale::scale_x` to `Scale::x` , same for `y`.

druid-shell/src/platform/gtk/window.rs
xStrom: Indeed it does, good catch!

xStrom:
Thanks, this is ready for another look. :eyes:

PR #1046, opened 2020-06-19
Cache piet::Image in Image widget

### LINKED ISSUE #1035: Cache piet::Image in Image widget
Using the [`image`](https://github.com/xi-editor/druid/blob/master/druid/examples/image.rs) example with a large image is insanely slow.

A first read may suggest that the problem is in repeating `ctx.make_image` at each `paint` event, is there a way to cache this and reuse the image from the previous frame? Since it depends on the `PaintCtx` I'm not sure how to do it.

jpochyla:
Adds a simple, never-invalidated cache for Piet image data. I'm not sure about the code style, please comment (or edit) in case you feel it could be better.

Fixes #1035

### DIFF
--- a/druid/src/widget/image.rs
+++ b/druid/src/widget/image.rs
@@ -19,15 +19,16 @@
 use std::{convert::AsRef, error::Error, path::Path};
 
 use crate::{
-    piet::{ImageFormat, InterpolationMode},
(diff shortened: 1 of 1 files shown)


jpochyla:
> This looks good! The only place we should invalidate is in the two set_ methods.

Thanks! Added.

PR #1048, opened 2020-06-19
Enable trivially_copy_pass_by_ref clippy lint.

xStrom:
Rust 1.44 moved the `trivially_copy_pass_by_ref` clippy lint to pedantic. As discussed on [zulip](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/Scale.20clippy.20fail) let's turn it back on, at least for now.

This PR also fixes the corresponding error in relation to the recent `Scale` rework.

### DIFF
--- a/druid-derive/src/lib.rs
+++ b/druid-derive/src/lib.rs
@@ -14,6 +14,8 @@
 
 //! derive macros for druid.
 
+#![deny(clippy::trivially_copy_pass_by_ref)]
+
 extern crate proc_macro;
 
 mod attr;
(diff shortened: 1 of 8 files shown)


xStrom:
Unfortunately for some bizarre reason they have decided that `clippy.toml` should not have the power to enable/disable lints. There is [some movement](https://github.com/rust-lang/rust-clippy/issues/1313) to have this feature in `Cargo.toml` but it doesn't look like it'll arrive any time soon.

Issue #1040, opened 2020-06-15
KeyEvent rework


raphlinus:
This is a tracking issue for redoing `KeyEvent`. Currently there are a number of things wrong with the implementation, which we're seeing as Control-X etc not working except as menu shortcuts (#1030), issues with numpad arrow keys (#318 and a number of related issues and PR's). Issue #357 rolls up a number of the numpad related issues.

I propose to rework `KeyEvent` to fix these. The following goals are in scope:

* Key events should be as consistent as possible across platforms. This especially extends to the text field in keys with modifiers, scan codes, etc.

* Export a reliable way for text widgets to determine whether the key event represents a potential text insertion. This will involve rethinking `is_printable` (see #119 and #1016).

* Detailed information should be available and accurate. This includes tracking both left and right shift, as well as indicating whether the key is a repeat from being held down.

* The Windows implementation should be UTF-16 aware, though key text in the astral plane is probably unlikely (see #174).

* Expected behavior should be documented, including guidance about how and when to interpret keys as text insertion, and recognition of command modifiers.

I propose that we converge as much as possible toward the web [KeyboardEvent](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent) mechanism. Note that the [KeyboardEvent.keyCode](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) field has extensive documentation and is likely a useful reference, though it is officially deprecated.

Also note, the issue of key events intersects in a very profound way with IME (input method editor), which is on our medium to long term roadmap but is unlikely to see work soon. Unfortunately, IME is the correct way to get important features such as "dead key" for Latin accents. In general, it's the goal of this work to not interfere with future development of IME, though I will take a look at whether there might be stopgap measures to get some of that functionality if well supported by the platform.

ratmice:
Given the convergence with the web KeyboardEvent mechanism,
I think it is worth looking at the [keyboard-types](https://docs.rs/keyboard-types) crate (I believe is the basis for the KeyboardEvent implementation in servo).

raphlinus:
Very interesting. Poking around it, I see some things that are fairly well developed, but others that are strangely not. For example, there is no left shift / right shift distinction in modifiers, although there is a `location` that can be used to detect the press of the shift key itself (it feels to me that edge-triggered tracking of these is something of an antipattern). I suspect that there were long arguments about this and there is a reason, but if it's suboptimal I'm not sure I want to adopt it wholesale.

[...]

PR #1049, opened 2020-06-20
Keyboard event rework

raphlinus:
As of this checkpoint, we generate an keyboard-types KeyboardEvent,
but don't plumb it through as the actual druid-shell keyboard event.

Also, there is no dead-key handling at this point.

Work towards #1040


druid/src/widget/textbox.rs
```diff
@@ -134,7 +135,7 @@ impl TextBox {
             EditAction::ModifySelection(movement) => self.move_selection(movement, text, true),
             EditAction::SelectAll => self.selection.all(text),
             EditAction::Click(action) => {
-                if action.mods.shift {
+                if action.mods.contains(Modifiers::SHIFT) {
```
cmyr, line 138: I don't love the ergonomics here, would be nice if we could just do `action.has_shift()` or something, or at least `action.modifiers.includes_shift()`.

This is maybe related to the idea of still having our own `KeyboardEvent` type.

druid/src/event.rs
```diff
@@ -85,18 +86,12 @@ pub enum Event {
...
     Wheel(MouseEvent),
     /// Called when a key is pressed.
-    ///
-    /// Note: the intent is for each physical key press to correspond to
-    /// a single `KeyDown` event. This is sometimes different than the
-    /// raw events provided by the platform. In particular, Windows sends
-    /// one or both of WM_KEYDOWN (a raw key code) and WM_CHAR (the
-    /// Unicode value), depending on the actual key.
-    KeyDown(KeyEvent),
+    KeyDown(KeyboardEvent),
```
cmyr, line 89: Do we think we will ever want to augment `KeyboardEvent`, such as to add a timestamp, or to expose the platform event? In that case it might make sense to still keep our own `KeyboardEvent` type, that will be a cheap conversion from `keyboard_types::KeyboardEvent`, and that will also make it easier for us to add convenience methods to that API?

It might also be nice to have our own `Modifiers` type, so that we could have more control over the API.

druid-shell/src/hotkey.rs
```diff
@@ -260,14 +265,30 @@ impl From<SysMods> for RawMods {
...
-impl From<KeyCode> for KeyCompare {
-    fn from(src: KeyCode) -> KeyCompare {
-        KeyCompare::Code(src)
+impl IntoKey for Key {
+    fn into_key(self) -> Key {
+        self
+    }
+}
+
+impl<'a> IntoKey for &'a str {
```
cmyr, line 274:
```suggestion
impl IntoKey for &str {
```

I believe you can now omit this lifetime.

druid-shell/src/platform/gtk/menu.rs
```diff
@@ -142,9 +143,16 @@ impl Menu {
 }
 
 fn register_accelerator(item: &GtkMenuItem, accel_group: &AccelGroup, menu_key: HotKey) {
-    let gdk_keyval = match menu_key.key {
-        KeyCompare::Code(key_code) => key_code.into(),
-        KeyCompare::Text(text) => text.chars().next().unwrap() as u32,
+    let gdk_keyval = match &menu_key.key {
+        Key::Character(text) => text.chars().next().unwrap() as u32,
```
cmyr, line 147: Appreciate this was existing code, but this unwrap feels bad, do we have a way of ensuring `Character` is never the empty string?

druid-shell/src/platform/gtk/window.rs
```diff
@@ -832,24 +834,62 @@ fn get_mouse_click_count(event_type: gdk::EventType) -> u8 {
...
     let keyval = key.get_keyval();
     let hardware_keycode = key.get_hardware_keycode();
 
     let keycode = hardware_keycode_to_keyval(hardware_keycode).unwrap_or(keyval);
 
     let text = gdk::keyval_to_unicode(keyval);
-
-    keyboard::KeyEvent::new(keycode, repeat, get_modifiers(key.get_state()), text, text)
+    /*
+    println!(
```
cmyr, line 862: ✂️

druid-shell/src/platform/gtk/window.rs
```diff
@@ -832,24 +834,62 @@ fn get_mouse_click_count(event_type: gdk::EventType) -> u8 {
...
+                Key::Unidentified
+            }
+        } else {
+            Key::Unidentified
+        }
+    });
+    let code = keycodes::hardware_keycode_to_code(hardware_keycode);
+    let location = keycodes::raw_key_to_location(keycode);
+    let is_composing = false;
+    // TODO: set this to "up" for key up events
```
cmyr, line 882: this is future work?

druid-shell/src/platform/mac/keyboard.rs
```diff
@@ -0,0 +1,381 @@
...
+        | Code::MetaLeft
+        | Code::MetaRight
+        | Code::CapsLock
+        | Code::Help => true,
+        _ => false,
+    }
+}
+
+impl KeyboardState {
+    pub fn new() -> KeyboardState {
```
cmyr, line 291: `pub(crate)` maybe?

druid-shell/src/platform/mac/keyboard.rs
```diff
@@ -0,0 +1,381 @@
...
+        let last_mods = NSEventModifierFlags::empty();
+        KeyboardState { last_mods }
+    }
+
+    pub fn process_native_event(&mut self, event: id) -> Option<KeyboardEvent> {
```
cmyr, line 296: ditto

druid-shell/src/platform/mac/keyboard.rs
```diff
@@ -0,0 +1,381 @@
...
+                        // that case, the best thing to do is synthesize the event
+                        // from the modifiers. But a challenge there is that we
+                        // might get multiple events.
+                        return None;
+                    }
+                }
+                _ => unreachable!(),
+            };
+            let is_composing = false;
+            let repeat: bool = event_type == NSEventType::NSKeyDown && msg_send!(event, isARepeat);
```
cmyr, line 331: total nit but convention is braces for `msg_send!`, which I like because it looks like objc.
```suggestion
            let repeat: bool = event_type == NSEventType::NSKeyDown && msg_send![event, isARepeat];
```

druid-shell/src/platform/mac/keyboard.rs
```diff
@@ -0,0 +1,381 @@
...
+            let key = if let Some(key) = code_to_key(code) {
```
cmyr, line 332: temping to write this as,

```rust
let key = code_to_key(code).unwrap_or_else(|| { ... })
```

🤷

druid-shell/src/platform/mac/keyboard.rs
```diff
@@ -0,0 +1,381 @@
...
+                key
+            } else {
+                let characters = from_nsstring(event.characters());
```
cmyr, line 335: maybe even tempting to just move this into a `chars_to_key` fn.

druid-shell/src/platform/web/keycodes.rs
```diff
@@ -14,17 +14,228 @@
...
+        KeyboardEvent::DOM_KEY_LOCATION_LEFT => Location::Left,
+        KeyboardEvent::DOM_KEY_LOCATION_RIGHT => Location::Right,
+        KeyboardEvent::DOM_KEY_LOCATION_NUMPAD => Location::Numpad,
+        // Should be exhaustive but in case not, use reasonable default
+        _ => Location::Standard,
+    }
+}
+
+/*
+// TODO: use macro in window to avoid code dupl
```
cmyr, line 223: 👀

druid-shell/src/platform/windows/keyboard.rs
```diff
@@ -0,0 +1,781 @@
...
+use winapi::shared::minwindef::{HKL, INT, LPARAM, UINT, WPARAM};
+use winapi::shared::ntdef::SHORT;
+use winapi::shared::windef::HWND;
+use winapi::um::winuser::{
+    GetKeyState, GetKeyboardLayout, MapVirtualKeyExW, PeekMessageW, ToUnicodeEx, VkKeyScanW,
+    MAPVK_VK_TO_CHAR, MAPVK_VSC_TO_VK_EX, PM_NOREMOVE, VK_CAPITAL, WM_CHAR, WM_INPUTLANGCHANGE,
+    WM_KEYDOWN, WM_KEYUP, WM_SYSCHAR, WM_SYSKEYDOWN, WM_SYSKEYUP,
+};
+
+use winapi::um::winuser::{
```
cmyr, line 34: this can be combined with the block above?

druid-shell/src/platform/windows/keyboard.rs
```diff
@@ -0,0 +1,781 @@
...
+        _ => Location::Standard,
+    }
+}
+
+impl KeyboardState {
+    /// Create a new keyboard state.
+    ///
+    /// There should be one of these per window. It loads the current keyboard
+    /// layout and retains some mapping information from it.
+    pub fn new() -> KeyboardState {
```
cmyr, line 497: can this be `pub(crate)`? here and below

druid-shell/src/platform/windows/keyboard.rs
```diff
@@ -0,0 +1,781 @@
...
+    /// As a general theory, many keyboard events generate a sequence of platform
+    /// messages. In these cases, we stash information from all messages but the
+    /// last, and generate the event from the last (using `PeekMessage` to detect
+    /// that case). Mozilla handling is slightly different; it actually tries to
+    /// do the processing on the first message, fetching the subsequent messages
+    /// from the queue. We believe our handling is simpler and more robust.
+    ///
+    /// # Safety
+    ///
+    /// The `hwnd` argument must be a valid `HWND`. Similarly, the `lparam` must
```
cmyr, line 534:
```suggestion
    /// The `hwnd` argument must be a valid `HWND`. Similarly, the `lparam` must be
```

druid-shell/src/platform/windows/keyboard.rs
```diff
@@ -0,0 +1,781 @@
...
+    /// from raw platform messages. It should be called for each relevant message,
+    /// which comprises: `WM_KEYDOWN`, `WM_KEYUP`, `WM_CHAR`, `WM_SYSKEYDOWN`,
+    /// `WM_SYSKEYUP`, `WM_SYSCHAR`, and `WM_INPUTLANGCHANGE`.
+    ///
+    /// As a general theory, many keyboard events generate a sequence of platform
+    /// messages. In these cases, we stash information from all messages but the
+    /// last, and generate the event from the last (using `PeekMessage` to detect
+    /// that case). Mozilla handling is slightly different; it actually tries to
+    /// do the processing on the first message, fetching the subsequent messages
+    /// from the queue. We believe our handling is simpler and more robust.
```
cmyr, line 530: I think it's great to write stuff down here. Something that would be particularly helpful is an example of a keyboard event that generates multiple messages.

druid-shell/src/platform/x11/keycodes.rs
```diff
@@ -14,321 +14,373 @@
...
-
-            _ => {
-                log::warn!("Warning: unknown keyval {}", raw);
-                KeyCode::Unknown(raw)
-            }
-        }
+use crate::keyboard_types::{Code, Key, Location, Modifiers};
+use x11rb::protocol::xproto::Keycode;
+
+// TODO: there's duplication here with gtk, as hardware keycodes
```
cmyr, line 20: maybe worth doing now? I would have a `shared` module in `platform` that is `cfg'd` for linux.

druid-shell/src/lib.rs
```diff
@@ -47,12 +45,13 @@ pub use clipboard::{Clipboard, ClipboardFormat, FormatId};
...
+pub use hotkey::{key_event_for_test, HotKey, IntoKey, RawMods, SysMods};
 pub use menu::Menu;
 pub use mouse::{Cursor, MouseButton, MouseButtons, MouseEvent};
 pub use scale::{Scalable, Scale, ScaledArea};
 pub use window::{
     IdleHandle, IdleToken, Text, TimerToken, WinHandler, WindowBuilder, WindowHandle,
 };
+
+// Discussion question: export types to top-level? Just KeyboardEvent?
+pub use keyboard_types;
```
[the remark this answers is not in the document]
raphlinus: I agree with you that re-exporting would be following current practices. However, `Key` in particular collides with the item of the same name in `env`. So I believe doing this would force the submodule discussion.

druid-shell/src/platform/gtk/keycodes.rs
```diff
@@ -12,264 +12,315 @@
...
-            _5 => KeyCode::Key5,
-            _6 => KeyCode::Key6,
-            _7 => KeyCode::Key7,
-            _8 => KeyCode::Key8,
-            _9 => KeyCode::Key9,
-            minus => KeyCode::Minus,
-            equal => KeyCode::Equals,
-            BackSpace => KeyCode::Backspace,
+#[allow(clippy::just_underscores_and_digits, non_upper_case_globals)]
+pub fn raw_key_to_key(raw: RawKey) -> Option<Key> {
```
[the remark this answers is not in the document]
raphlinus: This feels like a dialog. Is this something that there would be interest in? So far my toes dipped in these waters have met a chilly reception. The *major* client would be servo, but they're on winit, for which adopting any of this stuff would be a fairly major change.

I think the best way to go about this is a blog and then discussion on reddit or whatnot.

druid/src/widget/textbox.rs
raphlinus: I agree with you that the bitflags ergonomics are a downgrade, but one reason I went with it is that it's very much standard Rust practice. I'm also worried about too much timtowtdi, especially if we start having methods on both `Modifiers` and types that contain it.

druid/src/event.rs
raphlinus: Deep question with several facets.

Yes, we *absolutely* want to add timestamp at some point. But there are deep questions how to do that. I think we may want Event to become a product of timestamp and the sum of individual event types, rather than having timestamp become a new field in each of those variants.

Our own newtype over KeyboardEvent has some appeal, including those convenience methods, but I'm not convinced it overcomes the activation energy for making a new type. Certainly if we went the latter approach for timestamps that would be compelling. If we did a newtype, I would be open to having it deref to `keyboard_types::KeyboardEvent` (and probably an into).

druid-shell/src/platform/gtk/menu.rs
raphlinus: Right, if I were defining the type I'd have nonempty as an invariant. But in this particular code path we're basically consuming client code that can construct the value directly. I'm comfortable making this an `expect` with a helpful message, as I think `Key::Character("")` is something that is (a) unlikely and (b) reasonable to expect to fail.

druid-shell/src/platform/gtk/window.rs
raphlinus: Nah, I did it and forgot to remove the TODO.

druid-shell/src/platform/mac/keyboard.rs
raphlinus: It's a style call but I'm comfortable not putting it in a function, especially because future evolution of this (dead keys) will involve lots more of the keyboard event other than chars.

druid-shell/src/platform/windows/keyboard.rs
raphlinus: It can be, but I had the vk's in their own block so that it wouldn't be just one mass of symbols. I'll make the change though, because it can create confusion.

druid-shell/src/platform/windows/keyboard.rs
raphlinus: The reason *this* is not pub(crate) is that I want to keep the same code in win-win, where it's publicly exported. But maybe that should be a diff.

druid-shell/src/platform/windows/keyboard.rs
raphlinus: I added a couple examples.

druid-shell/src/platform/x11/keycodes.rs
raphlinus: There's also sharing between mac and x11. I'm worried the cfg stuff will get tangly, but it's probably worth it because more than one source of truth is a real problem.

druid-shell/src/hotkey.rs
raphlinus: Yes. I wonder why this isn't a clippy lint. Maybe would be too noisy on existing codebases.

druid-shell/src/lib.rs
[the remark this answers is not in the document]
cmyr: I might just reexport `Key` as `KeyCode` for now?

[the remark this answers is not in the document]
raphlinus: It's possible but I have a pretty strong aversion to having two names for the same type.

druid-shell/src/keyboard.rs
```diff
@@ -0,0 +1,220 @@
...
+///
+/// Note that this type is similar to [`KeyboardEvent`] in keyboard-types,
+/// but has a few small differences for convenience. It is missing the `state`
+/// field because that is already implicit in the event.
+///
+/// [`KeyboardEvent`]: keyboard_types::KeyboardEvent
+#[derive(Clone, Debug, Default, Eq, Hash, PartialEq)]
+pub struct KeyEvent {
+    /// Whether the key is pressed or released.
+    pub state: KeyState,
```
cmyr, line 37: meant to remove this?

druid-shell/src/keyboard.rs
```diff
@@ -0,0 +1,220 @@
...
+    /// Location for keys with multiple instances on common keyboards.
+    pub location: Location,
+    /// Flags for pressed modifier keys.
+    pub modifiers: Modifiers,
+    /// True if the key is currently auto-repeated.
+    pub repeat: bool,
+    /// Events with this flag should be ignored in a text editor
+    /// and instead composition events should be used.
+    pub is_composing: bool,
+}
```
cmyr, line 51: in case you were curious we could save 8 bytes everywhere this is used by jamming `is_composing`, `repeat`, and `location` into spare bits on modifiers 🥴 🤷

druid-shell/src/keyboard.rs
raphlinus: Originally I did, but then I found that it was quite convenient to have in the Windows implementation in particular, as the `process_event` method could just return `Option<KeyEvent>` in response to a fairly large set of platform messages it processes. Also, leaving it in decreases the friction for interconverting with upstream keyboard-types.

cmyr:
Okay, I played around with moving runebender over to this, and I do have a few notes.

- **direct field access**: I wonder if we should be using this less? It does cause headaches when we make changes. This includes a bunch of (perhaps all reasonable) name changes; `mods` is now `modifiers`, `is_repeat` is now `repeat`, etcetera. This does make migration more of a hassle. If we were using methods for most of these things it would be easier to just deprecate them.

- **reexporting typealiases with deprecation warnings**: I think it might make sense to have something like,

    ```rust
    #[deprecated(since = "0.7.0", note = "Use KbKey instead")]
    pub type KeyCode = KbKey;
    #[deprecated(since = "0.7.0", note = "Use Modifiers instead")]
    pub type KeyModifiers = Modifiers;
    ```
    in `druid/src/lib.rs`. Again, this just means there's a bunch less red squiggle when updating.

- **naming, more generally**: `MouseEvent` uses `mods`, and `KeyEvent` uses `modifiers`. It's `is_composing` but not `is_repeat`. A few of these feel inconsistent.

- **`IntoKey`**: since `KbKey` now uses a `String` internally, I think it might still make sense to have a `KeyCompare` trait? In particular we could implement this for &str, without allocating, so you could do `event.matches(" ")`, instead of `event.key == KbKey::Character(" ".to_string())` (or something).

If we're moving away from direct field access, it might make sense to use the same names we currently use, but to mark them as deprecated. Playing around, I ended up with something like,


```rust
#[derive(Clone, Debug, Default, Eq, Hash, PartialEq)]
pub struct KeyEvent {
    /// Whether the key is pressed or released.
    pub state: KeyState,
    /// Logical key value.
    #[deprecated(since = "0.7.0", note = "use key() method instead")]
    //FIXME: this should be named `key`; we use key_code to ease migration from 0.6
    pub key_code: KbKey,
    /// Physical key position.
    code: Code,
    /// Location for keys with multiple instances on common keyboards.
    location: Location,
    /// Flags for pressed modifier keys.
    #[deprecated(since = "0.7.0", note = "use modifiers() method instead")]
    pub mods: Modifiers,
    /// True if the key is currently auto-repeated.
    #[deprecated(since = "0.7.0", note = "use is_repeat() method instead")]
    pub is_repeat: bool,
    /// Events with this flag should be ignored in a text editor
    /// and instead composition events should be used.
    is_composing: bool,
}
#[allow(deprecated)]
impl KeyEvent {
    /// Flags for pressed modifier keys.
    pub fn modifiers(&self) -> Modifiers {
        self.mods
    }

    /// Return the `KbKey` for this event.
    #[inline]
    pub fn key(&self) -> &KbKey {
        &self.key_code
    }

    /// Physical key position.
    pub fn code(&self) -> Code {
        self.code
    }

    /// Location for keys with multiple instances on common keyboards.
    pub fn location(&self) -> Location {
        self.location
    }
  // etc
}
```

This doesn't feel great, and may not be worth it, but maybe it is? I'm not sure. The idea would be that after the next release we would make these fields private.


In any case, offering this up for discussion. If some of this is compelling, I can put up a patch.

raphlinus:
I'm not sure how much a `KeyCompare` trait will help. It doesn't in and of itself reduce allocations when constructing a `HotKey`, which is the main place `IntoKey` is used now. (We could of course use a Cow or some TinyStr variant if we're making our own type). I do very much appreciate the desire to be able to match a KbKey without allocation - that might be another question for upstream keyboard-types.

raphlinus:
Yes, it is I believe possible to improve the implementation of HotKey to allocate less without much changing the external interface.

PR #1053, opened 2020-06-23
Implement std::fmt::Debug for ImageData

theduke:
Stumbled over this today, bit annoying to not have it.

Not 100 percent sure about the format. 

I chose default struct formatting for convenience, which might be a tiny bit misleading because the fields are private and I used more descriptive names.


theduke:
Deriving would include the binary data, which would render stdout pretty much unusable.

cmyr:
right, makes sense. :)

PR #1054, opened 2020-06-23
Container::rounded takes KeyOrValue<f64> instead of f64

### LINKED ISSUE #1052: Container::rounded takes f64 directly
The method [Container::rounded](https://docs.rs/druid/0.6.0/druid/widget/struct.Container.html#method.rounded) takes a `f64` directly, instead of an `impl Into<KeyOrValue<f64>>`. This seems to be an oversight. Fixing it would allow customisation through `Env` keys, like with the other visual properties of `Container`.

binomial0:
Fixes #1052.
This is my first PR, and I'm not completely sure how things are done. However, the change is really small, so I don't think there's much that can go wrong.
I checked that all the tests pass, and I also tried this out in an unpublished application that I had lying around.

I think this change should be semver-safe, but I'm not sure. Some downstream crate's type inference might break, because the `rounded` function becomes more general. However, in the most common case you simply pass a float literal, and that case doesn't break. Still, I don't have any experience in Rust API design to completely rule out this problem.

Performance is probably irrelevant. At the moment, `paint()` unconditionally loads the border radius, even if it doesn't need it. I could improve that with some way-too-clever caching code that would be annoying to write and, more importantly, horrible to read, so I decided not to do that.


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ You can find its changes [documented below](#060---2020-06-01).
 
 - `Image` and `ImageData` exported by default. ([#1011] by [@covercash2])
 - `Scale::from_scale` to `Scale::new`, and `Scale` methods `scale_x` / `scale_y` to `x` / `y`. ([#1042] by [@xStrom])
+- `Container::rounded` takes `KeyOrValue<f64>` instead of `f64`. ([1054] by [@binomial0])
```
jneem, line 16: #1054, not 1054

druid/src/widget/container.rs
```diff
@@ -41,7 +41,7 @@ impl<T: Data> Container<T> {
         Self {
             background: None,
             border: None,
-            corner_radius: 0.0,
+            corner_radius: (0.0).into(),
```
jneem, line 44: Mostly just out of curiosity, but is there a reason you went with the parentheses?

CHANGELOG.md
binomial0: Oh, thanks.

druid/src/widget/container.rs
binomial0: Not really. I subconsciously thought it wouldn't parse without them, but it does. I'll remove them.

jneem:
Thanks!

PR #1057, opened 2020-06-24
Revisit the druid <-> druid-shell animation and invalidation interface.

jneem:
This implements (only for GTK so far) the invalidation scheme discussed [on zulip](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/purpose.20of.20request_anim), with a few small differences. I'm posting it in the hope of getting a little feedback before I dig into the other backends. (I went with GTK first because I think it'll be the hardest one, because there's a mismatch between what we want to do and what GTK wants us to do.)

The main issue with the current setup is that animation frames are propagated in the paint callback. This means that they can't invalidate any regions, which in turn means that when druid requests an animation frame, it has to invalidate the whole window just to be sure. So this patch introduces a two-phase paint protocol: first, `druid-shell` calls `prepare_paint`, which is allowed to invalidate whatever regions that it wants. Immediately after that, `druid-shell` calls `paint` and provides an invalidation region that must be filled. In the discussion on zulip, I proposed having `prepare_paint` return the region that it wants to be invalidated, but in fact it was simpler to just ask `prepare_paint` to use the existing invalidation methods.

Note that this is a breaking change, in that widgets that previously relied on `request_anim_frame` doing the invalidation will now need to do their own.

Here's a little gif showing (with debug_invalidation enabled) a scroll bar that doesn't repaint the whole window when it fades out.
scroll


druid-shell/src/platform/windows/util.rs
```diff
@@ -127,6 +129,15 @@ pub(crate) fn recti_to_rect(rect: RECT) -> Rect {
     )
 }
 
+/// Converts a `Region` into a vec of winapi `RECT`, with a scaling applied.
+pub(crate) fn region_to_rectis(region: &Region, scale: Scale) -> Vec<RECT> {
```
raphlinus, line 133: "rectis" seems an odd name.

jneem: I'm not in love with it either, but I copied it from `rect_to_rectis`...

jneem:
I think this is ready for another look. I've tested it on all platforms but mac, and I've been dogfooding it for a couple of weeks in my own druid project.

jneem:
I checked the conflict with #1037: merging this PR into #1037 results in just two conflicts. There's a trivial one in `druid/src/lib.rs` and one 8ish-line conflict in `druid-shell/src/platform/window.rs`. Anyway, much less than I expected, and so if I don't hear otherwise then I'll try to get this one in first.

PR #1061, opened 2020-06-25
Add OPEN_PANEL_CANCELLED and SAVE_PANEL_CANCELLED commands

cmyr:
This lets the user take special action if needed in either of
these cases.

I needed this in order to implement a "you have unsaved changes" prompt in Runebender, while testing out #1039. I think that we may want to further modify the general alerts API, and I think that we should try and make the open and save panels use a similar API to whatever we come up with for alerts, more generally, in which case this might be a stop-gap solution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ You can find its changes [documented below](#060---2020-06-01).
 ### Highlights
 
 ### Added
+- `OPEN_PANEL_CANCELLED` and `SAVE_PANEL_CANCELLED` commands. ([#1061] by @cmyr)
 
 - Added ctrl/shift key support to textbox. ([#1063] by [@vkahl])
 
@@ -351,6 +352,7 @@ Last release without a changelog :(
 [#1050]: https://github.com/linebender/druid/pull/1050
 [#1054]: https://github.com/linebender/druid/pull/1054
 [#1058]: https://github.com/linebender/druid/pull/1058
+[#1061]: https://github.com/linebender/druid/pull/1061
 [#1062]: https://github.com/linebender/druid/pull/1062
 [#1081]: https://github.com/linebender/druid/pull/1081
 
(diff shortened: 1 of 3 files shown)


Finnerale:
Yeah, I've been thinking about adding context methods for file dialogs similarly to those for windows, menus and future alerts.
I guess we will probably want to use the same pattern for this as with alerts.

It's probably okish to add those as long as they go into no release. Did you leave of the changelog intentionally? I think it should be included temporarily, even if they are going to be removed again.

cmyr:
oops changelog was an oversight, updated.

CHANGELOG.md
```diff
@@ -340,6 +341,7 @@ Last release without a changelog :(
 [#1043]: https://github.com/linebender/druid/pull/1043
 [#1050]: https://github.com/linebender/druid/pull/1050
 [#1054]: https://github.com/linebender/druid/pull/1054
+[#1061]: https://github.com/xi-editor/druid/pull/1062
```
Finnerale, line 344:
```suggestion
[#1061]: https://github.com/linebender/druid/pull/1061
```

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ You can find its changes [documented below](#060---2020-06-01).
 ### Highlights
 
 ### Added
+- Add `OPEN_PANEL_CANCELLED` and `SAVE_PANEL_CANCELLED` commands ([#1061] by @cmyr)
```
Finnerale, line 11:
```suggestion
- `OPEN_PANEL_CANCELLED` and `SAVE_PANEL_CANCELLED` commands. ([#1061] by @cmyr)
```

druid/src/command.rs
```diff
@@ -206,6 +209,9 @@ pub mod sys {
     pub const SHOW_SAVE_PANEL: Selector<FileDialogOptions> =
         Selector::new("druid-builtin.menu-file-save-as");
 
+    /// Sent when the user cancels an open file panel.
```
Finnerale, line 212:
```suggestion
    /// Sent when the user cancels an save file panel.
```

cmyr, line 212:
```suggestion
    /// Sent when the user cancels a save file panel.
```

PR #1066, opened 2020-06-26
Restrict on_click() to left clicks only.

konkers:
Previously any click would trigger the on_click() handler.

### DIFF
--- a/druid/src/widget/click.rs
+++ b/druid/src/widget/click.rs
@@ -17,7 +17,7 @@
 //! [`Controller`]: struct.Controller.html
 
 use crate::widget::Controller;
-use crate::{Data, Env, Event, EventCtx, LifeCycle, LifeCycleCtx, Widget};
+use crate::{Data, Env, Event, EventCtx, LifeCycle, LifeCycleCtx, MouseButton, Widget};
 
 /// A clickable [`Controller`] widget. Pass this and a child widget to a
 /// [`ControllerHost`] to make the child interactive. More conveniently, this is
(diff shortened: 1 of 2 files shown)


druid/src/widget/click.rs
```diff
@@ -52,12 +52,14 @@ impl<T: Data> Click<T> {
...
-                ctx.set_active(true);
-                ctx.request_paint();
+            Event::MouseDown(mouse_event) => {
+                if mouse_event.button == MouseButton::Left {
+                    ctx.set_active(true);
+                    ctx.request_paint();
+                }
             }
-            Event::MouseUp(_) => {
-                if ctx.is_active() {
```
Finnerale, line 60: The `ctx.is_active()` check must still be there. Otherwise, you can press left click, drag the cursor off the button, release and it will be considered a click on the button. You can try this in the `calc` example.

druid/src/widget/click.rs
```diff
@@ -52,12 +52,14 @@ impl<T: Data> Click<T> {
...
+            Event::MouseUp(mouse_event) => {
+                if mouse_event.button == MouseButton::Left {
```
cmyr, line 62:
```suggestion
                if ctx.is_active() && mouse_event.button == MouseButton::Left {
```

cmyr (approved):
This is a reasonable patch, I'm going to address @Finnerale's main concern and merge. I can't modify the changelog easily, so I'm just going to skip it. 🤷

Finnerale (approved):
Ok, I'll put it on my to-do list for the changelog cleanup :+1:

PR #1071, opened 2020-06-30
Replace all references of druid with Druid.

Finnerale:
This should replace all references of 'druid' with 'Druid', following the discussion on Zulip.

I think this is the best choice (and [most voted for](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/name.20style/near/202073103)).

If I remember correctly @xStrom made the argument to use `druid` when referring to the crate, and while I like the idea it turned out to be impossible to define (at least for me) when exactly one is talking about 'the crate', so I preferred 'Druid' in most cases.

I also changed 'Mdbook' and 'mdbook' to the correct spelling 'mdBook' :p

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,22 +100,22 @@ doesn't suit your use case, perhaps one of the others will!
 
 ### druid-shell
 
-The Druid toolkit uses druid-shell for a platform-abstracting application shell.
-druid-shell is responsible for starting a native platform runloop, listening to
+The Druid toolkit uses `druid-shell` for a platform-abstracting application shell.
+`druid-shell` is responsible for starting a native platform runloop, listening to
 events, converting them into a platform-agnostic representation, and calling a
 user-provided handler with them.
 
-While druid-shell is being developed with the Druid toolkit in mind, it is
+While `druid-shell` is being developed with the Druid toolkit in mind, it is
 intended to be general enough that it could be reused by other projects
-interested in experimenting with Rust GUI. The druid-shell crate includes a
+interested in experimenting with Rust GUI. The `druid-shell` crate includes a
 couple of [non-druid examples].
 
 ### piet
 
-Druid relies on the [piet library] for drawing and text layout. Piet is a 2D graphics
+Druid relies on the [Piet library] for drawing and text layout. Piet is a 2D graphics
 abstraction with multiple backends: `piet-direct2d`, `piet-coregraphics`, `piet-cairo`,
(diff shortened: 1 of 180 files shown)


docs/README.md
```diff
@@ -1,29 +1,29 @@
 # Druid mdbook documentation
```
chris-morgan, line 1: s/mdbook/mdBook/

xStrom:
The `druid` case is clear when you want to differentiate from `druid-shell` - I think a few places have this context.

xStrom:
There's also the `AUTHORS` file and all the copyright headers in all source files.

Finnerale:
> The druid case is clear when you want to differentiate from druid-shell - I think a few places have this context.

Hmm, I found even those places difficult because we are still talking about Druid as a "product" not just the entity on `crates.io` which is why I just went with Druid, considering that the main argument for the change was consistency this seems like the best choice.

> There's also the AUTHORS file and all the copyright headers in all source files.

Woops. After doing the replacement I noticed that `widget/parse.rs` was unmodified and in fact did not have a copyright header, so I've added one.

README.md
```diff
@@ -100,12 +100,12 @@ doesn't suit your use case, perhaps one of the others will!
 
 ### druid-shell
```
xStrom, line 101: Perhaps we shoud change all mentions of *druid-shell* in (at least) this section to `druid-shell`.

README.md
```diff
@@ -100,12 +100,12 @@ doesn't suit your use case, perhaps one of the others will!
...
+The Druid toolkit uses druid-shell for a platform-abstracting application shell.
+druid-shell is responsible for starting a native platform runloop, listening to
 events, converting them into a platform-agnostic representation, and calling a
 user-provided handler with them.
 
-While druid-shell is being developed with the druid toolkit in mind, it is
+While druid-shell is being developed with the Druid toolkit in mind, it is
 intended to be general enough that it could be reused by other projects
 interested in experimenting with Rust GUI. The druid-shell crate includes a
 couple of [non-druid examples].
```
xStrom, line 111: *non-druid* -> *non-Druid* or non-`druid` ?

README.md
Finnerale: I changed this for the content of the section but I left the heading as-is because it looked weird with the `background`.

README.md
Finnerale: I went with non-`druid`.

Issue #935, opened 2020-05-15
X11 shell does not support idle handles.


Finnerale:
The X11 shell should implement `WindowHandle::get_idle_handle`.

jneem:
Closed by #1072

PR #1072, opened 2020-06-30
Adds an idle loop to the x11 backend, and uses it for repainting.

### LINKED ISSUE #932: X11 shell uses sleep to sync rendering.

jneem:
Fixes #932 and #935.

### DIFF
--- a/druid-shell/src/platform/x11/application.rs
+++ b/druid-shell/src/platform/x11/application.rs
@@ -44,6 +44,10 @@ pub(crate) struct Application {
     ///
     /// A display is a collection of screens.
     connection: Rc<XCBConnection>,
+    /// An `XCBConnection` is *technically* safe to use from other threads, but there are lots of
+    /// subtleties; let's just avoid the issue altogether. As far as public API is concerned, this
+    /// causes `druid_shell::WindowHandle` to be `!Send` and `!Sync`.
+    marker: std::marker::PhantomData<*mut XCBConnection>,
     /// The default screen of the connected display.
     ///
     /// The connected display may also have additional screens.
@@ -122,6 +126,7 @@ impl Application {
             idle_read,
             idle_write,
             present_opcode,
+            marker: std::marker::PhantomData,
         })
     }
 
@@ -369,10 +374,19 @@ impl Application {
         let mut last_idle_time = Instant::now();
         loop {
             let next_idle_time = last_idle_time + timeout;
-            poll_with_timeout(&self.connection, self.idle_read, next_idle_time)
-                .context("Error while waiting for X11 connection")?;
+            self.connection.flush()?;
 
-            while let Some(ev) = self.connection.poll_for_event()? {
+            // Before we poll on the connection's file descriptor, check whether there are any
+            // events ready. It could be that XCB has some events in its internal buffers because
+            // of something that happened during the idle loop.
+            let mut event = self.connection.poll_for_event()?;
+
+            if event.is_none() {
+                poll_with_timeout(&self.connection, self.idle_read, next_idle_time)
+                    .context("Error while waiting for X11 connection")?;
+            }
+
+            while let Some(ev) = event {
                 match self.handle_event(&ev) {
                     Ok(quit) => {
                         if quit {
@@ -383,34 +397,13 @@ impl Application {
                         log::error!("Error handling event: {:#}", e);
                     }
                 }
+                event = self.connection.poll_for_event()?;
             }
 
             let now = Instant::now();
             if now >= next_idle_time {
                 last_idle_time = now;
-                // Clear out our idle pipe. Each write to it adds one byte; it's unlikely that there
-                // will be much in it, but read it 16 bytes at a time just in case.
-                let mut read_buf = [0u8; 16];
-                loop {
-                    match nix::unistd::read(self.idle_read, &mut read_buf[..]) {
-                        Err(nix::Error::Sys(nix::errno::Errno::EINTR)) => {}
-                        // According to write(2), this is the outcome of reading an empty, O_NONBLOCK
-                        // pipe.
-                        Err(nix::Error::Sys(nix::errno::Errno::EAGAIN)) => {
-                            break;
-                        }
-                        Err(e) => {
-                            return Err(e).context("Failed to read from idle pipe");
-                        }
-                        // According to write(2), this is the outcome of reading an O_NONBLOCK pipe
-                        // when the other end has been closed. This shouldn't happen to us because we
-                        // own both ends, but just in case.
-                        Ok(0) => {
-                            break;
-                        }
-                        Ok(_) => {}
-                    }
-                }
+                drain_idle_pipe(self.idle_read)?;
 
                 if let Ok(state) = self.state.try_borrow() {
                     for w in state.windows.values() {
@@ -442,7 +435,6 @@ impl Application {
                     for window in state.windows.values() {
                         window.destroy();
                     }
-                    log_x11!(self.connection.flush());
                 }
             }
         } else {
@@ -452,7 +444,6 @@ impl Application {
 
     fn finalize_quit(&self) {
         log_x11!(self.connection.destroy_window(self.window_id));
-        log_x11!(self.connection.flush());
     }
 
     pub fn clipboard(&self) -> Clipboard {
@@ -472,8 +463,37 @@ impl Application {
     }
 }
 
-/// Returns when there is an event ready to read from `conn`, or there is data ready to read from
-/// `idle` and the `timeout` has passed.
+/// Clears out our idle pipe; `idle_read` should be the reading end of a pipe that was opened with
+/// O_NONBLOCK.
+fn drain_idle_pipe(idle_read: RawFd) -> Result<(), Error> {
+    // Each write to the idle pipe adds one byte; it's unlikely that there will be much in it, but
+    // read it 16 bytes at a time just in case.
+    let mut read_buf = [0u8; 16];
+    loop {
+        match nix::unistd::read(idle_read, &mut read_buf[..]) {
+            Err(nix::Error::Sys(nix::errno::Errno::EINTR)) => {}
+            // According to write(2), this is the outcome of reading an empty, O_NONBLOCK
+            // pipe.
+            Err(nix::Error::Sys(nix::errno::Errno::EAGAIN)) => {
+                break;
+            }
+            Err(e) => {
+                return Err(e).context("Failed to read from idle pipe");
+            }
+            // According to write(2), this is the outcome of reading an O_NONBLOCK pipe
+            // when the other end has been closed. This shouldn't happen to us because we
+            // own both ends, but just in case.
+            Ok(0) => {
+                break;
+            }
+            Ok(_) => {}
+        }
+    }
+    Ok(())
+}
+
+/// Returns when there is an event ready to read from `conn`, or we got signalled by another thread
+/// writing into our idle pipe and the `timeout` has passed.
 // This was taken, with minor modifications, from the xclock_utc example in the x11rb crate.
 // https://raw.githubusercontent.com/psychon/x11rb/a6bd1453fd8e931394b9b1f2185fad48b7cca5fe/examples/xclock_utc.rs
 fn poll_with_timeout(conn: &Rc<XCBConnection>, idle: RawFd, timeout: Instant) -> Result<(), Error> {
@@ -482,10 +502,12 @@ fn poll_with_timeout(conn: &Rc<XCBConnection>, idle: RawFd, timeout: Instant) ->
     use std::os::unix::io::AsRawFd;
 
     let fd = conn.as_raw_fd();
-    let mut poll_fds = [
+    let mut both_poll_fds = [
         PollFd::new(fd, PollFlags::POLLIN),
         PollFd::new(idle, PollFlags::POLLIN),
     ];
+    let mut just_connection = [PollFd::new(fd, PollFlags::POLLIN)];
+    let mut poll_fds = &mut both_poll_fds[..];
 
     // We start with no timeout in the poll call. If we get something from the idle handler, we'll
     // start setting one.
@@ -497,12 +519,16 @@ fn poll_with_timeout(conn: &Rc<XCBConnection>, idle: RawFd, timeout: Instant) ->
                 .contains(PollFlags::POLLIN)
         };
 
-        match poll(&mut poll_fds, poll_timeout) {
+        match poll(poll_fds, poll_timeout) {
             Ok(_) => {
                 if readable(&poll_fds[0]) {
                     break;
                 }
-                if readable(&poll_fds[1]) {
+                if poll_fds.len() == 1 || readable(&poll_fds[1]) {
+                    // Now that we got signalled, stop polling from the idle pipe and use a timeout
+                    // instead.
+                    poll_fds = &mut just_connection;
+
                     let now = Instant::now();
                     if now >= timeout {
                         break;
--- a/druid-shell/src/platform/x11/application.rs
+++ b/druid-shell/src/platform/x11/application.rs
@@ -16,8 +16,10 @@
 
 use std::cell::RefCell;
 use std::collections::HashMap;
-use std::convert::TryInto;
+use std::convert::{TryFrom, TryInto};
+use std::os::unix::io::RawFd;
 use std::rc::Rc;
+use std::time::{Duration, Instant};
 
 use anyhow::{anyhow, Context, Error};
 use x11rb::connection::Connection;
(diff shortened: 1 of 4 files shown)


druid-shell/src/platform/x11/application.rs
```diff
@@ -350,20 +361,71 @@ impl Application {
...
+        // Try to figure out the refresh rate of the current screen. We run the idle loop at that
+        // rate.
+        let refresh_rate = util::refresh_rate(self.connection(), self.window_id).unwrap_or(60.0);
+        let timeout = Duration::from_millis((1000.0 / refresh_rate) as u64);
+        let mut last_idle_time = Instant::now();
         loop {
-            if let Ok(ev) = self.connection.wait_for_event() {
+            let next_idle_time = last_idle_time + timeout;
+            poll_with_timeout(&self.connection, self.idle_read, next_idle_time)
+                .context("Error while waiting for X11 connection")?;
```
psychon, line 373: Right now the order of the code here is the following (all wrapped in a big `loop { ... }`):

1. `poll` for readability
2. get events from x11rb until none are left in the queue
3. run idle handlers if there are some

If anything in step (3) causes x11rb to read an event from the X11 connection and enqueue it in its internal state, you could end up blocking in `poll`.

I propose to do something like this: https://github.com/psychon/x11rb/blob/87567eafa964efc6acd6bb7b76b69c1f1ff71215/examples/xeyes.rs#L308
Well... not really...

Anyway, what I mean:
Currently step (1) is `poll()` (pseudo-code). I propose to replace this with `let mut event = self.connection.poll_for_event(); if event.is_none() { poll(); } `.
Somehow you would then have to change step (2) so that this extra event is not lost.

Does this make some sense?

druid-shell/src/platform/x11/application.rs
```diff
@@ -350,20 +361,71 @@ impl Application {
...
     }
 
-    pub fn run(self, _handler: Option<Box<dyn AppHandler>>) {
+    fn run_inner(self) -> Result<(), Error> {
+        // Try to figure out the refresh rate of the current screen. We run the idle loop at that
+        // rate.
+        let refresh_rate = util::refresh_rate(self.connection(), self.window_id).unwrap_or(60.0);
+        let timeout = Duration::from_millis((1000.0 / refresh_rate) as u64);
+        let mut last_idle_time = Instant::now();
         loop {
```
psychon, line 370: I'd propose to add a call to `self.connection.flush()?` either at the end of the beginning of the loop. That way, you always flush before blocking in `poll` and thus all other calls to `flush` become unnecessary.

druid-shell/src/platform/x11/application.rs
```diff
@@ -404,4 +466,57 @@ impl Application {
...
+        PollFd::new(idle, PollFlags::POLLIN),
+    ];
+
+    // We start with no timeout in the poll call. If we get something from the idle handler, we'll
+    // start setting one.
+    let mut poll_timeout = -1;
+    loop {
+        let readable = |p: &PollFd| {
+            p.revents()
+                .unwrap_or_else(PollFlags::empty)
```
psychon, line 496: By the way, I have no idea why this `unwrap_or_else` is necessary, but I know that the x11rb example also has it...

druid-shell/src/platform/x11/application.rs
```diff
@@ -404,4 +466,57 @@ impl Application {
...
+            Ok(_) => {
+                if readable(&poll_fds[0]) {
+                    break;
+                }
+                if readable(&poll_fds[1]) {
+                    let now = Instant::now();
+                    if now >= timeout {
+                        break;
+                    } else {
+                        poll_timeout = c_int::try_from(timeout.duration_since(now).as_millis())
```
psychon, line 510: Why would the `idle` fd become readable, but no timeout occurred yet? As written, you use 100% CPU as soon as this case occurs until the timeout elapsed.

druid-shell/src/platform/x11/application.rs
```diff
@@ -404,4 +466,57 @@ impl Application {
...
+    use std::os::unix::io::AsRawFd;
+
+    let fd = conn.as_raw_fd();
+    let mut poll_fds = [
+        PollFd::new(fd, PollFlags::POLLIN),
+        PollFd::new(idle, PollFlags::POLLIN),
+    ];
+
+    // We start with no timeout in the poll call. If we get something from the idle handler, we'll
+    // start setting one.
```
psychon, line 491: This comment makes no sense to me and seems to be related to my comment below about the 100% CPU usage.

No new idle handler can appear while this code is running. Or are you working with threads? If so, this whole approach would break down (because another thread could cause x11rb to read in events, so events would end up in the internal queue, but the socket would not be readable, so `poll` would not return).

Later edit: Actually... this whole self-pipe thread looks like something that is needed only when threads are involved. Without threads, you cannot be blocking in `poll()` while at the same time deciding to need to wake up. With threads, your use of x11rb seems racy (I am open to suggestions on how to improve x11rb here, but I do not see any and especially `XCBConnection` is based on libxcb and cannot be modified here anyway).

druid-shell/src/platform/x11/window.rs
```diff
@@ -1117,21 +1151,56 @@ fn key_mods(mods: u16) -> Modifiers {
     ret
 }
 
+/// A handle that can get used to schedule an idle handler. Note that
+/// this handle is thread safe.
```
psychon, line 1155: Why does the handle have to be thread safe? (See my other comment about threads)

jneem:
Thanks for the comments! Regarding the threading: the intention (although it probably isn't correct right now) is to do the interaction with X11 only the main thread, but to support threads for other things. For example, `IdleHandle` is `Send + Sync`, so you can hand it to another thread. That thread can use the handle to queue a callback, but then that callback will be run on the main thread. Hopefully this explains the use of pipes.

Regarding the interaction with `x11rb`, could you expand on what the thread-safety issues are? I noticed that `XCBConnection` is [`Send + Sync`](https://github.com/psychon/x11rb/blob/87567eafa964efc6acd6bb7b76b69c1f1ff71215/src/xcb_ffi/raw_ffi.rs#L33), but maybe it shouldn't be? Or maybe a better question is: if I wrap our `XCBConnection` in a struct that is `!Sync`, will that avoid all potential threading issues. (Should I make it `!Send` also?) Actually, on windows our `WindowHandle` type is already `!Send` and `!Sync`, so this should probably be replicated on other platforms to avoid portability issues.

druid-shell/src/platform/x11/application.rs
jneem: That makes perfect sense, thanks!

druid-shell/src/platform/x11/application.rs
jneem: I'm not sure I understand this comment... we have to do *something* in the unlikely event that `poll (2)` sets `revents` to something invalid.

druid-shell/src/platform/x11/application.rs
jneem: I hope the comment about threads below explained why `idle` can become readable. The 100% CPU should be fixed.

druid-shell/src/platform/x11/application.rs
psychon: Yeah, this is not really meant as a comment to you. What I mean is: Why can't `nix` do this "something" for us? Is there any other sensible option than `unwrap_or_else(PollFlags::empty)`?

psychon:
> Regarding the interaction with x11rb, could you expand on what the thread-safety issues are?

The issue is that libxcb and XCBConnection have an internal queue of pending events that were read from the socket, but not yet passed to the caller. Thus "an event is (potentially) pending" is not the same as "the socket is readable".

Let's say that the main thread is "almost done" with a main loop iteration and is about to call `poll()`. We now have a race condition and another thread gets to run some code.
This other thread now does `conn.get_input_focus()?.reply()`. Thus, it sends a request and waits for its reply. To wait for the reply, it has to read X11 packages from the connection until it "finds" the reply to the request. Any events that it reads ends up in the internal queue.
Now the main thread continues with `poll()` and blocks even though there are pending events.

Hopefully this makes the issue clearer.

> I noticed that XCBConnection is Send + Sync, but maybe it shouldn't be?

Well, you could do what Qt5 is doing: You spawn a thread that does `loop { let event = conn.wait_for_event(); send_to_main_loop(event); }`. This `wait_for_event()` is made to work correctly by x11rb (via a condition variable, `std::sync::Condvar`). Thus, in this scenario, everything works correctly and the connection really is `Sync`.

The issue is that "`poll()` on the socket to know when to call `poll_for_event()`" is not thread-safe, but all the APIs themselves are thread-safe.

Does that make sense? Do you have any ideas how x11rb could work differently?

jneem:
Thanks, that makes a lot of sense. The only thing I would suggest to change in x11rb would be to have an explanation like the one you gave in the API docs. I've been kind of spoiled by rust over the last couple years, to the extent that if something `Send + Sync`, I just assume that there are no threading issues...

As far as druid is concerned, it sounds to me like everything will be fine as long as we restrict the xcb connection to one thread. As far as I can tell, access to the connection is restricted to the `WindowHandle` and the cairo context that we create; I've marked the `WindowHandle` as `!Send` and `!Sync`, and the cairo context is exposed as `piet::CairoRenderContext`, which is already `!Send` and `!Sync`. So I think the threading issue should be ok now.

druid-shell/src/platform/x11/application.rs
```diff
@@ -404,4 +466,92 @@ impl Application {
...
+                    if now >= timeout {
+                        break;
+                    } else {
+                        poll_timeout = c_int::try_from(timeout.duration_since(now).as_millis())
+                            .unwrap_or(c_int::max_value())
+                    }
+                }
+            }
+
+            Err(nix::Error::Sys(nix::errno::Errno::EINTR)) => {}
```
Finnerale, line 551: What case does this handle?

druid-shell/src/platform/x11/application.rs
```diff
@@ -350,23 +369,62 @@ impl Application {
         Ok(false)
     }
 
-    pub fn run(self, _handler: Option<Box<dyn AppHandler>>) {
+    fn run_inner(self) -> Result<(), Error> {
+        // Try to figure out the refresh rate of the current screen. We run the idle loop at that
+        // rate.
+        let refresh_rate = util::refresh_rate(self.connection(), self.window_id).unwrap_or(60.0);
+        let timeout = Duration::from_millis((1000.0 / refresh_rate) as u64);
+        let mut last_idle_time = Instant::now();
```
Finnerale, line 377: I'm not sure why we need this when using the present extension.
I suppose that's the reason why `poll_with_timeout` ignores its timeout when there is no idle action?

druid-shell/src/platform/x11/application.rs
```diff
@@ -404,4 +466,92 @@ impl Application {
...
+    let mut both_poll_fds = [
+        PollFd::new(fd, PollFlags::POLLIN),
+        PollFd::new(idle, PollFlags::POLLIN),
+    ];
+    let mut just_connection = [PollFd::new(fd, PollFlags::POLLIN)];
+    let mut poll_fds = &mut both_poll_fds[..];
+
+    // We start with no timeout in the poll call. If we get something from the idle handler, we'll
+    // start setting one.
+    let mut poll_timeout = -1;
```
Finnerale, line 523: Are we doing this just to "sync" with the refresh rate, or does this also have a purpose when using the present extension?

druid-shell/src/platform/x11/application.rs
```diff
@@ -404,4 +466,92 @@ impl Application {
...
+            p.revents()
+                .unwrap_or_else(PollFlags::empty)
+                .contains(PollFlags::POLLIN)
+        };
+
+        match poll(poll_fds, poll_timeout) {
+            Ok(_) => {
+                if readable(&poll_fds[0]) {
+                    break;
+                }
```
Finnerale, line 535: I assume this means we got an X11 event? Would be good to as a comment.

druid-shell/src/platform/x11/application.rs
```diff
@@ -404,4 +466,92 @@ impl Application {
...
+                if poll_fds.len() == 1 || readable(&poll_fds[1]) {
+                    // Now that we got signalled, stop polling from the idle pipe and use a timeout
+                    // instead.
+                    poll_fds = &mut just_connection;
```
Finnerale, line 539: What is the reason to keep waiting when using present?

druid-shell/src/platform/x11/application.rs
```diff
@@ -404,4 +466,92 @@ impl Application {
...
+        PollFd::new(idle, PollFlags::POLLIN),
+    ];
+    let mut just_connection = [PollFd::new(fd, PollFlags::POLLIN)];
+    let mut poll_fds = &mut both_poll_fds[..];
+
+    // We start with no timeout in the poll call. If we get something from the idle handler, we'll
+    // start setting one.
+    let mut poll_timeout = -1;
+    loop {
+        let readable = |p: &PollFd| {
```
Finnerale, line 525: We could just use a normal `fn` here:
```suggestion
        fn readable(p: &PollFd) -> bool {
```

druid-shell/src/platform/x11/application.rs
psychon: The C function `poll` returns `-EINTR` when a unix signal is received. This is necessary so that the signal can be handled (kernel-side reasons), but is quite useless for userspace. So, this case here just restarts the call.

After writing this: @jneem What about the `poll_timeout` argument? Some time could have passed before the signal came in and so `poll_timeout` might need to be recalculated.

jneem: @psychon There's a good chance that I'm missing something obvious, but I think that once timeout gets set, then I recalculate it for every poll. Because if we ever set the timeout then we also set `poll_fds` to `just_connection`. Then every subsequent time through the loop the `poll_fds.len() == 1` will succeed and so we will either `break` or recalculate `poll_timeout`. Did I miss something?

druid-shell/src/platform/x11/application.rs
jneem: The idea behind rate-limiting the idle loop is just to [avoid excessive wake-ups](https://wiki.ubuntu.com/Kernel/PowerManagement/IdentifyingIssues) by consolidating the idle commands and dong them all at once. It probably isn't that important, but I needed to support it anyway when present isn't being used. I've expanded on the comment a little.

druid-shell/src/platform/x11/application.rs
jneem: I hope I answered this above.

druid-shell/src/platform/x11/application.rs
psychon: Ah, sorry for not looking carefully enough.

jneem: Not at all! I'm very happy to have another pair of eyes on it!

jneem:
Ok, I think the review comments are addressed now.

druid-shell/src/platform/x11/application.rs
Finnerale: I see, but `poll_with_timeout` is already doing the waiting for external events, so why do we have to rate limit `run_idle` in this function separately? When there is an X11 event, why not just handle the idle queue as well?

jneem: Hm, I guess that would make the most sense when using present, but when using the fallback it would mean redrawing on every event. Thinking more about it, it would make more sense for the fallback to use timers instead of the idle loop for drawing but I haven't got around to implementing timers yet... would it be ok if I left in a TODO comment?

I'm thinking that the ideal behavior for the idle loop would be to run it after every collection of events, or if there are no events then wait at least 1/refresh_rate before running the idle loop.

Finnerale: > but when using the fallback it would mean redrawing on every event.
> Thinking more about it, it would make more sense for the fallback to use timers instead of the idle loop for drawing

Yeah, I think that would make more sense, delaying idle events just because we could be using a fallback render loop seems bad to me.

> would it be ok if I left in a TODO comment?

I would prefer an issue (or adding it to the existing timer issue), but yes.

> I'm thinking that the ideal behavior for the idle loop would be to run it after every collection of events, or if there are no events then wait at least 1/refresh_rate before running the idle loop.

You are referring to X11 events I suppose? In that case, I agree with this idea.

PR #1075, opened 2020-07-02
Re-export `druid_shell::Scalable` & remove to_px/to_dp helpers on Scale

ForLoveOfCats:
This allows `use`-ing `Scalable` and thus various type's implementation of `to_px` ect

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,12 +15,14 @@ You can find its changes [documented below](#060---2020-06-01).
 - `Scale::from_scale` to `Scale::new`, and `Scale` methods `scale_x` / `scale_y` to `x` / `y`. ([#1042] by [@xStrom])
 - Major rework of keyboard event handling ([#1049] by [@raphlinus])
 - `Container::rounded` takes `KeyOrValue<f64>` instead of `f64`. ([#1054] by [@binomial0])
+- Re-export `druid_shell::Scalable` under `druid` namespace. ([#1075] by [@ForLoveOfCats])
 
 ### Deprecated
 
 ### Removed
 
 - `Scale::from_dpi`, `Scale::dpi_x`, and `Scale::dpi_y`. ([#1042] by [@xStrom])
+- `Scale::to_px` and `Scale::to_dp`. ([#1075] by [@ForLoveOfCats])
 
 ### Fixed
 
@@ -239,6 +241,7 @@ Last release without a changelog :(
 [@covercash2]: https://github.com/covercash2
 [@raphlinus]: https://github.com/raphlinus
 [@binomial0]: https://github.com/binomial0
+[@ForLoveOfCats]: https://github.com/ForLoveOfCats
 [@chris-zen]: https://github.com/chris-zen
 
 [#599]: https://github.com/linebender/druid/pull/599
@@ -347,6 +350,7 @@ Last release without a changelog :(
 [#1050]: https://github.com/linebender/druid/pull/1050
 [#1054]: https://github.com/linebender/druid/pull/1054
 [#1058]: https://github.com/linebender/druid/pull/1058
+[#1075]: https://github.com/linebender/druid/pull/1075
 [#1062]: https://github.com/linebender/druid/pull/1062
 
 [Unreleased]: https://github.com/linebender/druid/compare/v0.6.0...master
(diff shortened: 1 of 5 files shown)


cmyr:
@xStrom does this make sense as public API to you?

ForLoveOfCats:
It kinda already is. `Scale` has the public and very confusingly named `to_px` which accepts a generic arg which impls `Scalable` before calling the argument's own `to_px` passing in the instance of `Scale`.  I say it is confusingly named because for the types which implement `Scalable` their `to_px` functions convert themselves using the `Scale` passed in but this apparent helper function on `Scale` does the opposite by converting the argument using itself as the scale.

When I initially drafted this PR I hadn't looked into these functions to much but now after using this branch for my project and poking around I think that the helper function on `Scale` should be removed and simply use the methods on these `Scalable` types.

My specific use case of having this functionality provided to applications is that I'm using it to handle font size in a DPI aware way by converting to absolute pixel values.

ForLoveOfCats:
I've removed the helper functions and corrected any callsites I could find with a quick grep. I decided to leave the `px_to_dp*` functions as despite not finding any places which call any of them they seem reasonable enough to exist on `Scale` as is. I expect the CI to probably fail on some non-GTK shell

ForLoveOfCats:
After fixing the CI and fiddling with merge conflicts I think this is ready for a review. Thankfully removing those functions turned out to be a fairly small change

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ You can find its changes [documented below](#060---2020-06-01).
 - `Scale::from_scale` to `Scale::new`, and `Scale` methods `scale_x` / `scale_y` to `x` / `y`. ([#1042] by [@xStrom])
 - Major rework of keyboard event handling ([#1049] by [@raphlinus])
 - `Container::rounded` takes `KeyOrValue<f64>` instead of `f64`. ([#1054] by [@binomial0])
+- Re-export `druid_shell::Scalable` under `druid` namespace and remove `to_px`/`to_dp` helper fns on `Scale` ([#1075] by [@ForLoveOfCats])
```
Finnerale, line 18: This should be split into two entries, one for the addition and one for the removal. Something like this:
### Added
- Re-export `druid_shell::Scalable` under `druid` namespace. ([#1075] by [@ForLoveOfCats])
### Removed
- `to_px`/`to_dp` helper functions on `Scale` in favor of those on `Scalable`. ([#1075] by [@ForLoveOfCats])

ForLoveOfCats:
So none of the CI seems to be running and the changelog still has a merge conflict after manually fixing it so I'm just going to let this sit for a while and hope Github sorts itself out

CHANGELOG.md
```diff
@@ -15,12 +15,14 @@ You can find its changes [documented below](#060---2020-06-01).
...
 - Major rework of keyboard event handling ([#1049] by [@raphlinus])
 - `Container::rounded` takes `KeyOrValue<f64>` instead of `f64`. ([#1054] by [@binomial0])
+- Re-export `druid_shell::Scalable` under `druid` namespace ([#1075] by [@ForLoveOfCats])
 
 ### Deprecated
 
 ### Removed
 
 - `Scale::from_dpi`, `Scale::dpi_x`, and `Scale::dpi_y`. ([#1042] by [@xStrom])
+- `Scale::to_px` and `Scale::to_dp` ([#1075] by [@ForLoveOfCats])
```
Finnerale, line 25:
```suggestion
- `Scale::to_px` and `Scale::to_dp`. ([#1075] by [@ForLoveOfCats])
```

CHANGELOG.md
```diff
@@ -15,12 +15,14 @@ You can find its changes [documented below](#060---2020-06-01).
 - `Scale::from_scale` to `Scale::new`, and `Scale` methods `scale_x` / `scale_y` to `x` / `y`. ([#1042] by [@xStrom])
 - Major rework of keyboard event handling ([#1049] by [@raphlinus])
 - `Container::rounded` takes `KeyOrValue<f64>` instead of `f64`. ([#1054] by [@binomial0])
+- Re-export `druid_shell::Scalable` under `druid` namespace ([#1075] by [@ForLoveOfCats])
```
Finnerale, line 18:
```suggestion
- Re-export `druid_shell::Scalable` under `druid` namespace. ([#1075] by [@ForLoveOfCats])
```

ForLoveOfCats:
GH claims there is only one conflict and it is the following
image
Combined with the CI bugging out I think that Github is being silly so I'm going to let this sit for a bit and see if it figures itself out

ForLoveOfCats:
I managed to fix the merge conflict by just moving the conflicting line up a line, not ideal but better than screwing up the PR's git log. The CI also seems to be functioning again, not sure if the conflict was blocking the CI from running as I'm 90% sure that I've had PRs run CI with merge conflicts but I don't know if the CI runs on just the PR's branch or on the master branch w/ the PR merged which would cause it to be blocked :shrug:

xStrom:
Sorry for not taking a look sooner, but yeah I think this change is fine.

PR #1077, opened 2020-07-03
Fix comparison between Modifier with NUM_LOCK and RawMods.

HoNile:
This is the most basic way I found to solve an issue where arrow key (and some others) doesn't work on texbox, I tracked the issue to the comparison between a Modifier and a RawMods because of NUM_LOCK.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -353,6 +353,7 @@ Last release without a changelog :(
 [#1054]: https://github.com/linebender/druid/pull/1054
 [#1058]: https://github.com/linebender/druid/pull/1058
 [#1062]: https://github.com/linebender/druid/pull/1062
+[#1077]: https://github.com/linebender/druid/pull/1077
 
 [Unreleased]: https://github.com/linebender/druid/compare/v0.6.0...master
 [0.6.0]: https://github.com/linebender/druid/compare/v0.5.0...v0.6.0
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ You can find its changes [documented below](#060---2020-06-01).
 - GTK: Don't crash when receiving an external command while a file dialog is visible. ([#1043] by [@jneem])
 - Fix derive `Data` when type param bounds are defined ([#1058] by [@chris-zen])
 - Ensure that `update` is called after all commands. ([#1062] by [@jneem])
+- Fix comparison between Modifier with NUM_LOCK and RawMods. ([#1077] by [@HoNile])
```
Finnerale, line 34: The PR link should be added to the others, below the profile links.

HoNile: Done !

druid-shell/src/hotkey.rs
```diff
@@ -112,7 +112,7 @@ impl HotKey {
     /// [`KeyboardEvent`]: keyboard_types::KeyEvent
     pub fn matches(&self, event: impl Borrow<KeyEvent>) -> bool {
         let event = event.borrow();
-        self.mods == event.mods && self.key == event.key
+        self.mods == (event.mods & (!Modifiers::NUM_LOCK)) && self.key == event.key
```
raphlinus, line 115: This doesn't cover the case of CAPS_LOCK or other fancy modifiers either. What I recommend is doing `|` over the four basic modifiers: shift, ctrl, alt, meta.

I had this on my radar to address as part of the rework, but missed it. Thanks for finding it, and sorry.

HoNile: Indeed I miss some other cases, I have fixed it by now, but I cannot add shift and ctrl modifier as they are used to manage the behavior from text selection/movement in text_input.rs line 67 in handle_event.

HoNile: Hum now I failed some doctest, I'll check it later. Maybe it's a too simplistic fix.

raphlinus:
#1079 is the fix I had in mind. Can you confirm it works for you?

HoNile:
Yeah #1079 work fine.

PR #1079, opened 2020-07-03
Compare only base modifiers for hotkey match

raphlinus:
Fixes the issue raised in #1077


raphlinus:
Good catch, I didn't add the other lock modifiers in MODIFIER_MAP in platform/gtk/window.rs. I could add that to this PR.

Finnerale:
We also could merge this now and do a separate PR for GTK mods, either way would be fine

PR #1081, opened 2020-07-04
Fix gtk key repeat logic

### LINKED ISSUE #1080: Event::KeyDown reports incorrect `repeat` value after holding multiple keys at specific order
Steps to reproduce:
1. Press & hold any  key A (except modifiers)
2. Press & hold any other key B (except modifiers)
3. Release key A
4. Wait for repeat timeout while holding key B

It will report `Event::KeyDown` for key B with `KeyEvent::repeat = false` for one time.

I am using GTK backend, haven't tested on others.

Druid version: commit 64eb47c6e74d71425d26be9259ef2fd131709d23 (latest master when reporting)

raphlinus:
Keep track of which keys are pressed and use that to determine whether
a key is a repeat. The previous logic of comparing the keyval to the
last value fails in the case of multiple keys.

Also add lock modifiers, a followup from #1079.

Fixes #1080

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,7 @@ You can find its changes [documented below](#060---2020-06-01).
 - GTK: Don't crash when receiving an external command while a file dialog is visible. ([#1043] by [@jneem])
 - Fix derive `Data` when type param bounds are defined ([#1058] by [@chris-zen])
 - Ensure that `update` is called after all commands. ([#1062] by [@jneem])
-- Fix Gtk key repeat logic ([#1081] by [@raphlinus])
+- GTK: Don't interrupt `KeyEvent.repeat` when releasing another key. ([#1081] by [@raphlinus])
 
 ### Visual
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,6 @@ You can find its changes [documented below](#060---2020-06-01).
 - `Scale::from_scale` to `Scale::new`, and `Scale` methods `scale_x` / `scale_y` to `x` / `y`. ([#1042] by [@xStrom])
 - Major rework of keyboard event handling ([#1049] by [@raphlinus])
 - `Container::rounded` takes `KeyOrValue<f64>` instead of `f64`. ([#1054] by [@binomial0])
-- Fix Gtk key repeat logic ([#1081] by [@raphlinus])
 
 ### Deprecated
 
@@ -30,6 +29,7 @@ You can find its changes [documented below](#060---2020-06-01).
 - GTK: Don't crash when receiving an external command while a file dialog is visible. ([#1043] by [@jneem])
 - Fix derive `Data` when type param bounds are defined ([#1058] by [@chris-zen])
 - Ensure that `update` is called after all commands. ([#1062] by [@jneem])
+- Fix Gtk key repeat logic ([#1081] by [@raphlinus])
 
 ### Visual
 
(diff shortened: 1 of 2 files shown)


raphlinus:
For more context on the logic for modifiers: Mozilla doesn't use gdk for this, but reaches through and calls xkb_keymap_mod_get_index to map the modifiers (see `KeymapWrapper::SetModifierMasks` in [nsGtkKeyUtils.cpp](https://searchfox.org/mozilla-central/source/widget/gtk/nsGtkKeyUtils.cpp)). The name it looks up for num-lock is XKB_MOD_NAME_NUM, which I believe is "Mod2". Calling this function directly is not appealing, as ideally we'd like to preserve the ability to build and test the gtk port on non-X11 systems.

I also have a comment in there about using [Keymap](https://docs.rs/gdk/0.12.0/gdk/struct.Keymap.html) as a more principled way of getting the numlock state. Oddly, Mozilla does this for nsBidiKeyboard.cpp, but not for numlock. Maybe somebody who knows this better than me can shed some light. In the meantime, I'm pretty satisfied that the logic in this PR is good enough, and doing something else might cause more problems.

raphlinus:
Also, Mozilla's logic is a bit different. Instead of having a hashmap, they have a "last key pressed" (which is a hardware keycode), but *don't* clear it on keyup if the value doesn't match. I also confirm that Mozilla works here. I think maybe that logic might be a touch more robust, as it's less likely that a key will be stuck in the down position on something like focus change.

azymohliad:
I can confirm that it fixes [#1080](https://github.com/linebender/druid/issues/1080).

druid-shell/src/platform/gtk/window.rs
```diff
@@ -839,6 +843,10 @@ const MODIFIER_MAP: &[(gdk::ModifierType, Modifiers)] = &[
     (ModifierType::MOD1_MASK, Modifiers::ALT),
     (ModifierType::CONTROL_MASK, Modifiers::CONTROL),
     (ModifierType::META_MASK, Modifiers::META),
+    (ModifierType::LOCK_MASK, Modifiers::CAPS_LOCK),
+    // Note: this is the usual value on X11, not sure how consistent it is.
+    // Possibly we should use `Keymap::get_num_lock_state()` instead.
+    (ModifierType::MOD2_MASK, Modifiers::NUM_LOCK),
```
Finnerale, line 849: I tested this on Wayland and I'm getting `MOD2_MASK` as well.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ You can find its changes [documented below](#060---2020-06-01).
 - `Scale::from_scale` to `Scale::new`, and `Scale` methods `scale_x` / `scale_y` to `x` / `y`. ([#1042] by [@xStrom])
 - Major rework of keyboard event handling ([#1049] by [@raphlinus])
 - `Container::rounded` takes `KeyOrValue<f64>` instead of `f64`. ([#1054] by [@binomial0])
+- Fix Gtk key repeat logic ([#1081] by [@raphlinus])
```
Finnerale, line 18: This should go into the `Fixed` section.
```markdown
- GTK: Don't interrupt `KeyEvent.repeat` when releasing another key. ([#1081] by [@raphlinus])
```

Finnerale: @raphlinus Did you miss my suggestion for the changlog entry or did you not like it?

raphlinus: Sorry, I missed it, I only caught the part about it being in the wrong section. That's what I get for trying to do this stuff before I've had my coffee.

PR #1082, opened 2020-07-05
Constrain aspect ratio

derekdreery:
This PR adds a method that works kinda like `BoxConstraints::constrain`, but prioritises matching the aspect ratio rather than the L1 distance between the requested size and the returned size. It's useful when you want to render a widget that looks better when its aspect ratio is preserved.

This function only expresses a preference: if the parent widget gives tight constraints then they will be used regardless of aspect ratio requested. If the requested aspect ratio is not available, this method chooses the size that minimizes the difference between aspect ratios.

It's not quite ready for merge - I wanted to start a conversation. I'll write some more on zulip.

### DIFF
--- a/druid/src/box_constraints.rs
+++ b/druid/src/box_constraints.rs
@@ -164,7 +164,7 @@ impl BoxConstraints {
     }
 
     /// Find the `Size` within these `BoxConstraint`s that minimises the difference between the
-    /// returned `Size`'s aspect ratio and `aspect_ratio` An aspect ratio is defined as
+    /// returned `Size`'s aspect ratio and `aspect_ratio`, where *aspect ratio* is defined as
     /// `height / width`.
     ///
     /// If multiple `Size`s give the optimal `aspect_ratio`, then the one with the `width` nearest
@@ -174,9 +174,6 @@ impl BoxConstraints {
     ///
     /// Use this function when maintaining an aspect ratio is more important than minimizing the
     /// distance between input and output size width and height.
-    ///
-    /// TODO should aspect_ratio be restricted to 0 < ar < infinity and panic otherwise? Same for
-    /// `width` and `self`.
     pub fn constrain_aspect_ratio(&self, aspect_ratio: f64, width: f64) -> Size {
         // Minimizing/maximizing based on aspect ratio seems complicated, but in reality everything
         // is linear, so the amount of work to do is low.
@@ -185,9 +182,12 @@ impl BoxConstraints {
             height: width * aspect_ratio,
         };
 
+        // It may be possible to remove these in the future if the invariant is checked elsewhere.
+        let aspect_ratio = aspect_ratio.abs();
+        let width = width.abs();
+
         // Firstly check if we can simply return the exact requested
         if self.contains(ideal_size) {
-            //println!("ideal");
             return ideal_size;
         }
 
@@ -208,14 +208,12 @@ impl BoxConstraints {
         // Check each possible intersection (or not) of the aspect ratio line with the constraints
         if aspect_ratio > min_w_max_h {
             // outside max height min width
-            //println!("outside max height min width");
             Size {
                 width: self.min.width,
                 height: self.max.height,
             }
         } else if aspect_ratio < max_w_min_h {
             // outside min height max width
-            //println!("outside min height max width");
             Size {
                 width: self.max.width,
                 height: self.min.height,
@@ -223,7 +221,6 @@ impl BoxConstraints {
         } else if aspect_ratio > min_w_min_h {
             // hits the constraints on the min width line
             if width < self.min.width {
-                //println!("min width");
                 // we take the point on the min width
                 Size {
                     width: self.min.width,
@@ -231,14 +228,12 @@ impl BoxConstraints {
                 }
             } else if aspect_ratio < max_w_max_h {
                 // exits through max.width
-                //println!("max width");
                 Size {
                     width: self.max.width,
                     height: self.max.width * aspect_ratio,
                 }
             } else {
                 // exits through max.height
-                //println!("max height");
                 Size {
                     width: self.max.height * aspect_ratio.recip(),
                     height: self.max.height,
@@ -248,21 +243,18 @@ impl BoxConstraints {
             // final case is where we hit constraints on the min height line
             if width < self.min.width {
                 // take the point on the min height
-                //println!("min height");
                 Size {
                     width: self.min.height * aspect_ratio.recip(),
                     height: self.min.height,
                 }
             } else if aspect_ratio > max_w_max_h {
                 // exit thru max height
-                //println!("max height");
                 Size {
                     width: self.max.height * aspect_ratio.recip(),
                     height: self.max.height,
                 }
             } else {
                 // exit thru max width
-                //println!("max width");
                 Size {
                     width: self.max.width,
                     height: self.max.width * aspect_ratio,
--- a/druid/src/box_constraints.rs
+++ b/druid/src/box_constraints.rs
@@ -187,7 +187,7 @@ impl BoxConstraints {
 
         // Firstly check if we can simply return the exact requested
         if self.contains(ideal_size) {
-            println!("ideal");
+            //println!("ideal");
             return ideal_size;
         }
 
--- a/druid/src/box_constraints.rs
+++ b/druid/src/box_constraints.rs
@@ -285,7 +285,7 @@ mod tests {
 
     #[test]
     fn constrain_aspect_ratio() {
-        for (bc, aspect_ratio, width, output) in vec![
+        for (bc, aspect_ratio, width, output) in [
             // The ideal size lies within the constraints
             (bc(0.0, 0.0, 100.0, 100.0), 1.0, 50.0, Size::new(50.0, 50.0)),
             (bc(0.0, 10.0, 90.0, 100.0), 1.0, 50.0, Size::new(50.0, 50.0)),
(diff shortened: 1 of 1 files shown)


cmyr:
quoting from @Zarenor on zulip:

>Catching up on the chat here - Did we ever get any other discussion on this? I think we've several times brought up possibly changing the expected behavior of widgets with regard to layout constraints and clipping, etc. (which we then expect built-ins to follow)

I agree with the premise, and I was thinking about this a lot when I was redoing the flex widget. 

### unexpected sizes should be rare

my thinking is maybe in a slightly different direction, though. Basically: if you end up with a size that is not the size you would like, how do you draw your content?

An initial comment here: for our own widgets, this should be hard to do. If they are used in 'standard' ways (that is, using our built-in layout widgets in proscribed ways, where exactly what is 'proscribed' is definitely fuzzy right now, but just imagine we have a good book chapters about this) then widgets that have expected sizes should always end up being an appropriate size. A label might be narrower than its text, certainly, but a checkbox should have the correct height, etcetera.

The only way you should be able to get really weird constraints for a default widget should be by forcing it to some size, say with an expanding `SizedBox` or similar.

### how do you lay yourself out when you have an unexpected size?

 this is more the crux of the matter, and is ultimately about how you implement a widget's `paint` method. At the start of this method, you call `PaintCtx::size()`, which tells you the size of the region you are painting in. How do you use this space? For instance, with a checkbox, we might decide that the checkbox should always draw itself square, and that it should draw itself centered. Centered is an arbitrary choice; we might decide that checkboxes should always be right aligned, for instance. With a label, we might decide to always draw the text at the top of the available space, aligned to the appropriate leading edge for the text direction.

The take away, for me, is that these decisions are **policy** decisions. We need to figure out what our expectations are, and we need to write them down, likely as part of the book that details how to implement custom widgets.

From this, my current opinion is that I'm not sure this PR is the correct direction, since we can achieve most of the same outcomes by implementing the policy in `paint`. I might be wrong though: one thought that jumps out at me right now is that, by pushing this on to the constraints, we might give the parent widget an opportunity to use extra space better than we do?

Please let me know what obvious things I have missed. 😁

Zarenor:
I certainly agree that these are *policy* decisions, but I'd like to get them written somewhere, and then treat violations of them in the built in widgets as bugs (or as noted exceptions to policy.

Secondly, I disagree about handling most of this in `paint` - I think, at least for the standard layout widgets, a violated constraint might should trigger a relayout. This makes us vulnerable to malicious widgets, though.
The problem is I'm not sure if we want to make layout more stateful. The conundrum I keep finding myself in when thinking through designing new druid layout widgets is what state might be needed, and how minimal we've thus far made that state (in the current set of layout widgets). 
And I don't see a way to alter future relayout behavior without added statefulness.

derekdreery:
So I would definitely not claim that this method is the correct one to choose in all circumstances, but that it provides an option to developers of custom widgets to use. I think I would argue that the correct way to handle unexpected sizes depends on the widget - there will not be a 1 size fits all choice. 

The new method in this PR simply provides a different way of choosing size based on constraints.

dhardy:
BTW, I'm kind of butting in here, but I already thought about sizing quite a lot for KAS, culminating in the current design where widgets specify a minimum size, a preferred size, a stretch policy, explicit margins, and can make their height-requirements dependent on width. As @cmyr says, it results in widgets usually getting what they ask for, but potentially having to put up with less given insufficient space. I put in some extra work to ensure that space is distributed fairly when insufficient to meet requirements (to avoid abuse from oversized widgets) and that resizes are idempotent (respecting the current layout as much as possible). Take a look [here](https://github.com/kas-gui/kas/blob/master/src/layout/size_rules.rs#L85) (and at the rather crazy `solve_seq` method). I also have a bunch of rough notes on the topic if you're interested.

derekdreery:
@dhardy thanks for butting in. I hadn't even considered deviating from the [flutter layout algorithm](https://medium.com/snapp-mobile/flutter-anatomy-layout-internals-part-1-de99b772ab99) - this method is simply providing a method to choose a point in the space of possible sizes (given by the flutter algo) that is as close as possible to the requested size, but that favours preserving aspect ratio over L1 distance.

dhardy:
I just started simple and added bits until I got what seemed right. For example this is a grid with a few spans and cell-sizing based on contents (the reason the empty cell above the edit-box is so big is because the layout system considers nothing else needs the space anyway — probably I should make the edit-box more greedy in this case).

I kept things simple as much as possible: first in a bottom-to-top pass each parent collects a list of child size-requirements, adds them together and passes up to the top-level solver, then going back top-to-bottom, the allocated extents are shared out by each parent. This means there is not a large list of constraints gathered and solved-at-once (making this square with width-for-height trade-offs seemed like a major scaling problem with anything like text where there thousands of possible widths).

But my system isn't perfect; e.g. it can calculate height-requirements based on width but not vice-versa. I also haven't implemented any type of flow-box, though I don't see any technical difficulties to doing so.

layout

druid/src/box_constraints.rs
```diff
@@ -155,4 +155,218 @@ impl BoxConstraints {
...
+    /// Test whether these constraints contain the given `Size`.
+    pub fn contains(&self, size: impl Into<Size>) -> bool {
+        let size = size.into();
+        (self.min.width <= size.width && size.width <= self.max.width)
+            && (self.min.height <= size.height && size.height <= self.max.height)
+    }
+
+    /// Find the `Size` within these `BoxConstraint`s that minimises the difference between the
+    /// returned `Size`'s aspect ratio and `aspect_ratio` An aspect ratio is defined as
+    /// `height / width`.
```
cmyr, line 168:
```suggestion
    /// Find the `Size` within these `BoxConstraint`s that minimises the difference between the
    /// returned `Size`'s aspect ratio and `aspect_ratio`, where *aspect ratio* is defined as
    /// `height / width`.
```

druid/src/box_constraints.rs
```diff
@@ -155,4 +155,218 @@ impl BoxConstraints {
...
+    ///
+    /// If multiple `Size`s give the optimal `aspect_ratio`, then the one with the `width` nearest
+    /// the supplied width will be used. Specifically, if `width == 0.0` then the smallest possible
+    /// `Size` will be chosen, and likewise if `width == f64::INFINITY`, then the largest `Size`
+    /// will be chosen.
+    ///
+    /// Use this function when maintaining an aspect ratio is more important than minimizing the
+    /// distance between input and output size width and height.
+    ///
+    /// TODO should aspect_ratio be restricted to 0 < ar < infinity and panic otherwise? Same for
```
cmyr, line 178: I think it's reasonable to at least `abs()` aspect ratio / width. I think it should also be an invariant that constraints are never negative, and we should perhaps be enforcing this on creation, but that would be a separate PR.

druid/src/box_constraints.rs
```diff
@@ -155,4 +155,218 @@ impl BoxConstraints {
...
+        // Minimizing/maximizing based on aspect ratio seems complicated, but in reality everything
+        // is linear, so the amount of work to do is low.
+        let ideal_size = Size {
+            width,
+            height: width * aspect_ratio,
+        };
+
+        // Firstly check if we can simply return the exact requested
+        if self.contains(ideal_size) {
+            //println!("ideal");
```
cmyr, line 190: lets clean this up :)

cmyr:
@derekdreery I definitely agree that the correct way to handle unexpected sizes depends on the widget, although I think there can be general guidelines.

@dhardy that's fascinating! I definitely *think* in terms of a "preferred content size" when doing flutter-style layout, at least with some widgets, but that logic all exists in each widget's layout method. The beauty of the box algorithm is how incredibly simple it is, and I would be scared to start trying to augment it too much, but I'm very glad you've come up with a mechanism that works for you, although I'm curious if there were concrete limitations of box layout that inspired you to augment it?

dhardy:
@cmyr I am not really familiar with the Flutter model, but what happens when e.g. you have a "label" child and an "edit box" child, and the window has more space than required for the sum of the children? Ideally you want the edit box to expand. Or what if you have a row of children with various sizes, and want to squash all of these into an area not quite large enough to meet their size requirements? Ideally you want "fair distribution" of available space, meaning that the largest children lose the most space. Is the Flutter model powerful enoguh to handle this? (The article linked above is far too sparse on details.)

cmyr:
In that instance you would have those two children in a 'flex' container, with the label marked as non-flexible and the edit box marked as flexible; after laying out the inflexible children, the rest of the available space would be used by the flexible child.

In the second instance, if you do not have enough space to lay out the children you would clip or overflow (your choice) or you would embed in a scroll widget.

what a "fair" distribution would be is, I think, totally a matter of policy and of the particular widgets; it would be very hard to come up with a generalization of that problem; and I don't think it's unreasonable to (for instance) expect layouts that might reasonable overflow to take some explicit decision around how to handle that, such as allowing items to wrap, or requiring scrolling etc.

dhardy:
Then I guess in my model, all parents are flex containers. (There is also the option to embed in a scroll region if desired.) My idea of a "fair" distribution is influenced by the idea that a widget requesting a huge amount of space should not break everything else, while space should be used to maximum utility. This constrains things somewhat.

cmyr:
@derekdreery I'm happy to have you merge this if you like!

derekdreery:
Brill, I think these kinds of utility functions are useful, mostly because they are a pain to implement and handling edge cases is fiddly. I will mae sure it's tidy before merging.

druid/src/box_constraints.rs
derekdreery: good spot :)

derekdreery:
I've updated to address your review. Would you like to review again, or shall I merge?

cmyr:
@derekdreery looks good to me, feel free to squash :)

cmyr:
@derekdreery I'd like to get this off the PR list so I'm going to merge, if you decide you don't like it we can always revert down the road.

PR #1086, opened 2020-07-06
WIP: Attempt to speed up CI by caching cargo dependencies

justinmoon:
I'm attempting to adapt an ["official example"](https://github.com/actions/cache/blob/master/examples.md#rust---cargo). This same technique worked for my project which uses Druid (7 minute build -> 2 minute build), but it doesn't speed things up here.

I suspect the problem might be that [Cargo.lock isn't checked in](https://github.com/linebender/druid/blob/master/.gitignore#L3) -- [as discussed in this thread](https://github.com/actions-rs/meta/issues/21#issuecomment-629712798).

Perhaps druid being a "workspace" also interferes somehow?

[Note about CI in cargo book](https://doc.rust-lang.org/cargo/guide/cargo-home.html#caching-the-cargo-home-in-ci).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
             ~/.cargo/git
             target
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-        if: contains(matrix.os, 'ubuntu')
+        if: contains(matrix.os, 'ubuntu') || contains(matrix.os, 'MacOS')
 
       - name: install libgtk-3-dev libx11-dev
         run: |
@@ -154,7 +154,7 @@ jobs:
             ~/.cargo/git
             target
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-        if: contains(matrix.os, 'ubuntu')
+        if: contains(matrix.os, 'ubuntu') || contains(matrix.os, 'MacOS')
 
       # libgtk-dev seems to be needed by e.g. druid-derive
       - name: install libgtk-dev
@@ -264,7 +264,7 @@ jobs:
             ~/.cargo/git
             target
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-        if: contains(matrix.os, 'ubuntu')
+        if: contains(matrix.os, 'ubuntu') || contains(matrix.os, 'MacOS')
 
       - name: install libgtk-dev libx11-dev
         run: |
@@ -336,7 +336,7 @@ jobs:
             ~/.cargo/git
             target
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-        if: contains(matrix.os, 'ubuntu')
+        if: contains(matrix.os, 'ubuntu') || contains(matrix.os, 'MacOS')
 
       - name: install libgtk-dev
         run: |
@@ -397,7 +397,7 @@ jobs:
             ~/.cargo/git
             target
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-        if: contains(matrix.os, 'ubuntu')
+        if: contains(matrix.os, 'ubuntu') || contains(matrix.os, 'MacOS')
 
       - name: install stable toolchain
         uses: actions-rs/toolchain@v1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - name: Generate Cargo.lock
+        run: cargo generate-lockfile
+
       - uses: actions/cache@v2
         with:
           path: |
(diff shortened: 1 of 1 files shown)


cmyr:
cool, work on this stuff is definitely appreciated!

justinmoon:
I'm attempting to adapt an ["official example"](https://github.com/actions/cache/blob/master/examples.md#rust---cargo). This same technique worked for my project which uses Druid (7 minute build -> 2 minute build), but it doesn't speed things up here.

I suspect the problem might be that [Cargo.lock isn't checked in](https://github.com/linebender/druid/blob/master/.gitignore#L3) -- [as discussed in this thread](https://github.com/actions-rs/meta/issues/21#issuecomment-629712798).

Blisto91:
This [Pull request](https://github.com/actions/cache/pull/325) at the "official example" repository might have some additional info and help.

Blisto91:
I am pretty new to rust and github. So i hope my writing makes sense and that i'm doing it in the correct format. :)

I looked into the examples of the pull request linked above and read a bit more into how it worked.
The Testing a library example seems very useful for Druids use case.

We can lower the size of the cache by only caching ~/.cargo/registry/index and ~/.cargo/registry/cache instead of the whole ~/.cargo/registry directory.
The dependency crates are stored twice in the folder as unpacked files in /registry/cache and raw source code in /registry/src.
We can omit /registry/src to avoid redundant data. 

The /target directory can also be omitted since it isn't included in the Druid repository unless i'm mistaken?

cmyr:
@justinmoon where did this work end up?

Blisto91:
If you are having trouble getting it to work then i could try and take a swing at it in a separate PR. If you wouldn't mind :).

Been doing a lot of reading and some testing with the CI and i think i have something that should work.

justinmoon:
@cmyr: I haven't had any time to play with it, and probably won't :(

@Blisto91: Go for it

Blisto91:
So i pulled down my own PR again that only handled the downloaded dependency crates and crates.io index, because over time i wasn't too happy with the modest improvement. Sometimes it actually took a bit longer because restoring a Github cache seems to fluctuate between a few seconds to several minutes, even for the same small cache.

Tho i do think there is proper gains to be had from caching the build target.
The biggest problem here is we have to satisfy the 5GB Github cache storage limit. Especially the Ubuntu build gets very large just on it's own (around 5GB on my own linux laptop) and most of the size is in the examples folder.
With the impressive compression the cache does this is actually reduced to a bit below 2GB when packed and stored.
The MacOS and Windows files are in the 300 - 500MB range when compressed.
There probably have to be separate cache for stable and nightly and maybe also the wasm files so we can blow the budget quite quickly.

My first attempt was to keep incremental compilation off and then clean only the druid artifacts with Cargo clean -p and or Powershell.
But the CI kept failing on macOS after restoring the files saying it couldn't find the rental-impl dependency for fluent-bundle, even tho it was present. I tried to investigate the problem and found a Rust github issue that linked to [this nightly flag](https://github.com/rust-lang/cargo/issues/7916) and was able to get it working on nightly only with it.
It also seemed work by reversing the order the druid crates are tested in. But i think it seemed like a bad solution.

The next attempt was to enable incremental and then just optimize the build target for size.
These build settings got the size down substantially on my linux laptop.
opt-level = 'z'
lto = 'fat' or 'thin'
codegen-units = 1
But i again ran into build trouble on macOS where it on nightly failed with lto enabled. But lto stood for most of the size savings on it's own.
Seems there are issues with lto on atleast nightly currently and people are still complaining on a closed github issue that highlighted the problem. But can't seem to find it again currently.

Sorry for the long ramble! Just wanted to share some findings.

Blisto91:
Just read through [this pinned issue](https://github.com/actions/cache/issues/267) over on the GitHub cache repository.

I think we should maybe put this on hold a bit until at least the cache restore times are more consistent.

CleanCut:
@Blisto91 [that pinned issue](https://github.com/actions/cache/issues/267) has been updated - v2.1.0 of the cache action has been released, which should resolve the inconsistent cache restore times.

Blisto91:
@CleanCut ye I saw and I've seen much more consistant restore times in my tests since then.
I did have one new cache where it took 15-20 minutes to save it. I think the new improvement might only apply to restore times?

Now I just want windows to use Zstd and the new rust feature resolver to get stabilized and then I'm a satisfied man 😁
A Github cache storage limit increase to 10 or 15GB wouldn't be unwelcome either hehe.

CleanCut:
> Edit: [Saw this](https://github.com/github/roadmap/issues/66) on the new github roadmap.
> With a proper increased cache size limit we could get away with caching the whole build target without cleaning or needing external cloud storage. That would probably show as a nice speed boost and avoid dependency failures with the current feature resolver.

That would be really cool. 👍

cmyr:
@justinmoon What's the status here? Do you think it's worth merging?

Blisto91:
@cmyr I think there is a good chance the CI will fail on subsequent runs after the cache have been saved.
Especially had problems on mac during my testing. 
I think saving the whole target folder is risky until the new rust feature resolver.

I have toyed a bit with sccache and i think that might work fine.
If @justinmoon doesn't mind then i could try to take another shot at it with a new PR.

Issue #1087, opened 2020-07-07
Focusing widgets


sysint64:
Hello there, at the moment API for focusing widgets quite unfinished, I have been playing around with it.

I would like to suggest such an API: add two widgets FocusScope - serves as a scope for its descendants, restricting focus traversal to the scoped controls, it can be useful if I would like to add some overlay UI like alert and strict scope for this area. We could create FocusSope in the root by default.

Focus widget - is a control widget that can capture the focus and spread focus state to its descendants; in that case, all descendants can receive keyboard events. It's nice to have such a widget because we could handle all focus events like request next and previous focus only there. And we can wrap up focusable widgets, which could handle keyboard events to some other decorator widgets that also could be rendered differently depends on focus state.

I have already implemented the Focus widget and now working on the FocusScope widget; it's much trickier.

**Example:**

```rust
let input = Focus::new(
    AccessorDecorator::new(
        FocusDecorator::new(),
        AccessorDecorator::new(
            TextboxDecorator::new(),
            TextBox::new()
                .with_placeholder("Enter the command")
                .fix_width(TEXT_BOX_WIDTH)
                .lens(HelloState::name),
        )
        .padding(2.0),
    ),
);
```

PR #1089, opened 2020-07-11
WIP: Focusing widgets

sysint64:
Hello there, this is an implementation of #1087.


druid/src/tests/mod.rs
```diff
@@ -332,7 +334,7 @@ fn simple_lifecyle() {
         harness.send_initial_events();
         assert_matches!(record.next(), Record::L(LifeCycle::WidgetAdded));
         assert_matches!(record.next(), Record::E(Event::WindowConnected));
-        assert_matches!(record.next(), Record::E(Event::WindowSize(_)));
+        assert_matches!(record.next(), Record::None);
```
sysint64, line 337: Not sure if this is important to record `WindowSize` here. It can't be recorder due to `WindowDesc::new` now implicitly wrap root widget with `FocusScope`

sysint64:
Finally, I have managed to get the first working example:

ezgif-4-cfdd6d3332c9

sysint64:
Inspired by flutter [Focus](https://api.flutter.dev/flutter/widgets/Focus-class.html) and [FocusNode](https://api.flutter.dev/flutter/widgets/FocusScope-class.html) widgets.

sysint64:
Also, I would like to add some tests a little bit later.

sysint64:
@cmyr, @raphlinus can someone have a look? Can this patch be useful?

cmyr:
@sysint64 thanks for the ping, will check it out today.

druid/examples/focus.rs
```diff
@@ -0,0 +1,135 @@
...
+
+//! Demonstrates focus and focus scope widgets
+
+pub use druid::theme;
+use druid::widget::prelude::*;
+use druid::widget::{Button, Container, Flex, Focus, FocusScope, Label, WidgetExt};
+
+use druid::{AppLauncher, Color, Data, LocalizedString, Point, Rect, WidgetPod, WindowDesc};
+
+pub struct FocusDecorator<T> {
```
cmyr, line 23: In examples I would be generous with documentation; for instance here I would say something like, "A container that draws a border around its child when focused.".

druid/src/command.rs
```diff
@@ -242,6 +242,13 @@ pub mod sys {
 
     /// Redo.
     pub const REDO: Selector = Selector::new("druid-builtin.menu-redo");
+
+    /// Notify descendens that focus node has changed focus
+    pub const FOCUS_NODE_FOCUS_CHANGED: Selector<bool> =
```
cmyr, line 247: What does the `bool` indicate?

druid/src/core.rs
```diff
@@ -30,6 +30,52 @@ use crate::{
 /// Our queue type
 pub(crate) type CommandQueue = VecDeque<(Target, Command)>;
 
+/// An object that can be used by a widget to obtain the focus and to handle keyboard events.
+#[derive(Debug, Clone, Copy)]
+pub struct FocusNode {
```
cmyr, line 35: I have very little information from these types as to what they represent or how they are used; we'll want to really figure out how to document this stuff before merging.

druid/src/widget/focus.rs
```diff
@@ -0,0 +1,143 @@
...
+    pub fn new(child: impl Widget<T> + 'static) -> Self {
+        Focus {
+            child: WidgetPod::new(child).boxed(),
+            focus_node: FocusNode::empty(),
+            auto_focus: false,
+            focus_requested: false,
+        }
+    }
+
+    /// Builder-style method to set the `Focus`'s auto focus.
```
cmyr, line 42: What does 'auto focus' mean? (I presume it means "has focus when the program starts", but we should be explicit about this.) What happens when multiple widgets are auto focused?

druid/src/widget/focus.rs
```diff
@@ -0,0 +1,143 @@
...
+            Event::KeyDown(key_event) => {
+                match key_event {
+                    // Tab and shift+tab
+                    k_e if HotKey::new(None, KbKey::Tab).matches(k_e) => {
+                        ctx.focus_next();
+                    }
+                    k_e if HotKey::new(SysMods::Shift, KbKey::Tab).matches(k_e) => {
+                        ctx.focus_prev();
+                    }
+                    _ => (),
```
cmyr, line 73: I think we should send the key event to our child first, and then only handle it ourselves if `is_handled() == false`; this would let a text field decide whether or not it wanted to insert a tab character or not.

druid/src/widget/focus.rs
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    fn lifecycle(&mut self, ctx: &mut LifeCycleCtx, event: &LifeCycle, data: &T, env: &Env) {
+        let previous_focus_node = ctx.focus_node();
+
+        match event {
+            LifeCycle::WidgetAdded => {
+                self.focus_node.widget_id = Some(ctx.widget_id());
+                ctx.set_focus_node(self.focus_node);
+                ctx.register_for_focus();
```
cmyr, line 100: I'm a little concerned about how complicated this seems to be, although I supposed it's fine if druid users never need to do this themselves, and they just use the `Focus` widget if they want to participate in focus?

druid/src/widget/textbox.rs
```diff
@@ -365,13 +371,30 @@ impl Widget<String> for TextBox {
...
 
     fn lifecycle(&mut self, ctx: &mut LifeCycleCtx, event: &LifeCycle, _data: &String, _env: &Env) {
         match event {
-            LifeCycle::WidgetAdded => ctx.register_for_focus(),
+            LifeCycle::WidgetAdded => {
+                let previous_focus_node = ctx.focus_node();
+                self.focus_node.widget_id = Some(ctx.widget_id());
+                ctx.set_focus_node(self.focus_node);
+                ctx.register_for_focus();
+                ctx.set_focus_node(previous_focus_node);
```
cmyr, line 385: it's a bit concerning that I would have to do this much bookkeeping in order to create a custom widget that participated in focus without using a `Focus` wrapper? It might make sense to only allow focus to work via `Focus`, if it's a significant simplification.

cmyr (commented):
Overall I like the idea of this patch, but I find some of the details a little confusing, and would want to think about how we document and explain this functionality.

I think a good place to start would be just a simple explanation of the types added in the patch and how they work together, and what I would need to know (if I am writing a widget) in order to participate in focus? What invariants do I need to uphold etc.

Also: should we remove or change `register_for_focus`, or make it private?

druid/src/command.rs
sysint64: Nothing, I just don't know how to represent an empty type.

druid/examples/focus.rs
sysint64: ok

druid/src/widget/focus.rs
sysint64: Ok

druid/src/widget/focus.rs
sysint64: I think druid users should never do this themselves; they should use the `Focus` widget if they would like to manage focus. After this PR, I wanted to remove focus management from the `TextBox` widget. I've done this already in my pet project: https://github.com/tech-paws/druid-frontend/blob/master/src/ui/widgets/textbox.rs But I think in druid it should be a little bit different.

druid/src/widget/textbox.rs
sysint64: Yes, definitely, only the `Focus` widget should manage focus; I want to do this in the next iteration. It will already work fine if delete all stuff with focus management in the `TextBox`, but you should always wrap this widget with `Focus,` and this will break backward compatibility.

druid/src/core.rs
sysint64: I use this as an additional object to get information about the currently focused widget and about the current `FocusScope`.

sysint64:
> Also: should we remove or change register_for_focus, or make it private?

@cmyr I think we could make it private.

druid/src/widget/focus.rs
sysint64: > has focus when the program starts

this. I'll try to make more understandable the doc comment.

druid/src/command.rs
sysint64: No, my mistake, this `bool` indicates whether `focus_node` is focused or not.

sysint64: I'll fix comment

cmyr: And for the record if you don't need a type you would use the 'unit type', `()`.

druid/src/widget/focus.rs
```diff
@@ -56,12 +58,14 @@ impl<T: Data> Widget<T> for Focus<T> {
...
 
+        self.child.event(ctx, event, data, env);
+
         match event {
             Event::MouseDown(_) => {
                 ctx.request_focus();
                 ctx.request_paint();
             }
-            Event::KeyDown(key_event) => {
+            Event::KeyDown(key_event) if !ctx.is_handled => {
```
sysint64, line 68: Is it correct?

cmyr: yes, this looks correct to me.

sysint64:
@cmyr I've added two tests - for multiple autofocus and focus scopes. Also fixed doc comments. `FocusNode` is used only as a read-only structure for druid users, so I've changed doc comment to:
```rust
/// Read-only structure to get information about the currently focused widget and current focus scope.
```

sysint64:
I'm trying implicitly wrap `TextBox` with `Focus` widget, but I have no idea how to do that in rust :( I would like to have something like this after calling `TextBox::new()`

```rust
Focus::new(TextBox::new())
```

but then I will not be able to set placeholder or other parameters, because the result type is `Focus`.

sysint64:
I could add `FocusController` to manage focus inside a widget without wrapping with the `Focus` widget.

sysint64:
It would be nice to have something like a `build` method in Flutter.

sysint64:
Well, I have an idea - add something like `build` in Flutter. 
Here how I've done this:
https://github.com/sysint64/druid/blob/5988f2aab721f0f73a21409ace68f3ba72149c53/druid/src/widget/textbox.rs#L41-L113

We could have an additional widget to create composable widgets like this.

sysint64:
@cmyr I have removed focus handling from the `TextBox` widget and removed the `register_for_focus` method. Also, I have changed the doc comment for the `FocusNode` struct to this:
```
/// An object that can be used by a widget to obtain the focus and to handle keyboard events.
/// This is a read-only structure to get information about the currently focused widget
/// and current focus scope.
```

For the `FocusScopeNode` as well:
```
/// An object that can be used by a widget to create a focus scope.
/// This is a read-only structure to get information about the current focus scope.
```

cmyr:
Thanks for this, I've been wanting to take a deeper look, and having just been in the textbox myself I should be in a good position for a review.

I want do add that I really appreciate a lot of the experimenting you've been doing; I think you are working on tricky problems, and sometimes review will be slow just because you're sort of out working on niche stuff and it can be hard to find the mental space to really consider it properly.

sysint64:
> I want do add that I really appreciate a lot of the experimenting you've been doing; I think you are working on tricky problems, and sometimes review will be slow just because you're sort of out working on niche stuff and it can be hard to find the mental space to really consider it properly.

Thanks, it's totally okay, I'm not in a hurry

cmyr:
Want to just ping this and say that I'm intending to get around to it, it's just that focus is a complicated area with lots of moving parts and I need to set aside at least half a day to give this proper consideration. Thank you for your patience!

sysint64:
Hi, I'm still here. Just have some work to be done. Soon I'll fix conflicts.

sysint64:
@cmyr I have added `FocusController` to manage focus, so we can easily use it inside the `TextBox` widget.

druid/examples/focus.rs
```diff
@@ -0,0 +1,136 @@
...
+        let rect = Rect::from_origin_size(Point::ORIGIN, size);
+        self.child.set_layout_rect(ctx, data, env, rect);
+
+        size
+    }
+
+    fn paint(&mut self, ctx: &mut PaintCtx, data: &T, env: &Env) {
+        self.child.paint(ctx, data, &env);
+
+        if ctx.focus_node().is_focused {
```
cmyr, line 63: couldn't this still just be `if ctx.is_focused()`? like we have an `is_focused()` method that internally checks the `FocusNode`?

druid/src/widget/focus_scope.rs
```diff
@@ -0,0 +1,81 @@
...
+            child: WidgetPod::new(child).boxed(),
+            focus_scope_node: FocusScopeNode { widget_id: None },
+        }
+    }
+}
+
+impl<T: Data> Widget<T> for FocusScope<T> {
+    fn event(&mut self, ctx: &mut EventCtx, event: &Event, data: &mut T, env: &Env) {
+        let previous_focus_scope = ctx.focus_scope();
+        ctx.set_focus_scope_node(self.focus_scope_node);
```
cmyr, line 40: Just an observation: an idiomatic way to get around writing these dual methods is to have a method on the `ctx`, like `with_focus_scope`, that takes a closure, similar to how `PaintCtx::with_save` works; then you could write,

```rust
ctx.with_focus_scope(self.focus_scope_node, |ctx| {
    self.child_event(ctx, event, data, env)
});

File: druid/src/widget/focus.rs [hunk 12]
```diff
@@ -0,0 +1,232 @@
...
+// limitations under the License.
+
+//! A focus widget.
+
+use druid::widget::prelude::*;
+
+use druid::{commands, Data, FocusNode, HotKey, KbKey, Point, Rect, SysMods, Widget, WidgetPod};
+
+#[derive(Debug, Clone)]
+pub(crate) struct FocusController {
```
cmyr, line 22: A rule of thumb: anything that druid's built-in widgets use in their implementations should be public; so if this is used in `TextBox` we should expose it, so that other people can write their own textboxes.

druid/src/widget/focus.rs
```diff
@@ -0,0 +1,232 @@
...
+                ctx.request_paint();
+            }
+            Event::KeyDown(key_event) if !ctx.is_handled => {
+                match key_event {
+                    // Tab and shift+tab
+                    k_e if HotKey::new(None, KbKey::Tab).matches(k_e) => {
+                        ctx.focus_next();
+                    }
+                    k_e if HotKey::new(SysMods::Shift, KbKey::Tab).matches(k_e) => {
+                        ctx.focus_prev();
```
cmyr, line 57: we'll want some way to make this behaviour configurable at some point, I think? but no worries for now.

druid/src/widget/focus.rs
```diff
@@ -0,0 +1,232 @@
...
+            focus_node: FocusNode::empty(),
+            previous_focus_node: FocusNode::empty(),
+            requested_focus: false,
+            auto_focus: false,
+        }
+    }
+
+    pub(crate) fn event_start(&mut self, ctx: &mut EventCtx) {
+        self.previous_focus_node = ctx.focus_node();
+        ctx.set_focus_node(self.focus_node);
```
cmyr, line 41: I'm still confused about this, and I'm giving up on trying to figure it out from the code. What happens if we don't set this? What things will break?

cmyr (commented):
So overall I think this is definitely an improvement, but I'm curious about some aspects of the implementation.

When I imagine how I expect this to work, it mostly involves taking the logic in `window.rs` around handling `focus_next` and `focus_prev`, and moving that to something (a widget, realistically) that can be inserted at various points in the tree.

There may be reasons why my idea doesn't work; it's just the thing that occurs to me.

I think something that I've struggled with a bit in reviewing this work is that I haven't had a good understanding of the design and the rationale for the design? Like I'm sure it exists, it just hasn't been obvious to me.


In any case, I do think this is definitely better than what we have now, and I'm inclined to merge it, but I would love to have a better understanding of the design, if there are any concerns you have with it, etc. 😎

I've left a bunch of other varied comments inline.

Thanks again for your patience!

druid/src/command.rs
```diff
@@ -266,6 +266,20 @@ pub mod sys {
...
+
+    /// Notify descendens that focus node has changed focus.
+    /// The payload indicates wheter the focus_node in the context is focused or not.
+    pub const FOCUS_NODE_FOCUS_CHANGED: Selector<bool> =
+        Selector::new("druid-builtin.focus_node_focus_changed");
+
+    /// Request focus for the widget
+    pub const REQUEST_FOCUS: Selector<WidgetId> = Selector::new("druid-builtin.request_focus");
+
+    /// Request next focus
```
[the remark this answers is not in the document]
sysint64: Agree.

druid/src/widget/focus.rs
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    pub(crate) fn event_end(&mut self, ctx: &mut EventCtx, event: &Event) {
+        match event {
+            Event::MouseDown(_) => {
+                ctx.request_focus();
+                ctx.request_paint();
```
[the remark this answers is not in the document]
sysint64: Yeah, that makes sense.

druid/src/contexts.rs
```diff
@@ -147,6 +149,16 @@ impl_context_method!(
             self.widget_state.id
         }
 
+        /// get the `FocusNode` of the current widget.
+        pub fn focus_node(&self) -> FocusNode {
```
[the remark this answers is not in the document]
sysint64: So this method is to get information about the current `FocusNode` and check if it is focused. In that case, descendants of the `Focus` widget can read that property and maybe draw some decorations, like in the example where we can draw a border for the Button widget.

druid/examples/focus.rs
sysint64: Yes, it can be, but I kept `is_focused` unchangeable because `focus_node()` is the current focus widget information. On the other hand, `is_focused` shows if the current widget is focused. I just was not sure if we could change the behavior of the `is_focused` method.

druid/src/widget/focus_scope.rs
sysint64: Oh yeah, thanks a lot! I actually don't like this solution with dual methods, but my attempt using `Drop` was failed because of the borrow checker, and I haven't come across any different solution. I'll fix this.

druid/src/widget/focus.rs
sysint64: Okay, got it.

druid/src/widget/focus.rs
sysint64: I agree, I think it would be great if we could configure it.

druid/src/widget/focus.rs
sysint64: Firstly lets me explain what `FocusNode` actually is and how I use it. The main idea is to pass information about the `Focus` widget state to its descendants. Right now, `is_focused` shows if the current widget has a focus and hence `is_focused = false` for its descendants. In the example, we can draw decoration over the `Button` just because we have `focus_node` information that depicts the `Focus` widget state. Also, descendants of the `Focus` widget should also be able to receive keyboard events, that's why we should set the current `focus_node` to the context, to check if descendants can receive events. In code I have added this: `ctx.state.focus_node.is_focused || self.state.has_focus` to check if widget can receive keyboard events.

And here is a little diagram:

```
"Focus" <- Owner of `focus_node` put it to the context; `is_focused = true`
|--"Widget 1" <- Can read `focus_node` by calling `ctx.focus_node()`; `is_focused = false` but `ctx.focus_node().is_focused = true`
|--|--"Widget 2" <- The same properties as Widget 1
|--"Widget 3" <- The same properties as Widget 1
```

cmyr: okay, that's a good explanation, thanks. I think it would make sense to write something like that down in the docs for `FocusNode`?

sysint64: Okay, I'll try )

sysint64: @cmyr what about this:
```
/// A type that depicts the current [`Focus`] widget state for its descendants.
///
/// This type can be retrieved from any descendants of [`Focus`].
/// For example, this type can be used to draw focus decoration by reading the current is_focused
/// value that shows if [`Focus`] widget is_focused. Unlike ctx.is_focused, ctx.focus_node().is_focused
/// will be true even if the current widget is not focused and if the parent [`Focus`] has focus.
///
/// This type also allows descendants to receive keyboard events if the parent [`Focus`] widget has focus.
```

druid/src/widget/textbox.rs
```diff
@@ -170,6 +173,14 @@ impl<T> TextBox<T> {
     pub fn editor(&self) -> &Editor<T> {
         &self.editor
     }
+
+    /// Builder-style method to set the `TextBox`'s auto focus.
+    /// Has focus when the widget is created.
+    /// When multiple widgets are auto-focused, the last created widget will gain the focus.
+    pub fn with_auto_focus(mut self, auto_focus: bool) -> Self {
```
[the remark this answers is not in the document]
sysint64: I think this makes sense. I also was thinking about situations where we could open modal dialog without opening a new window or maybe show some panel, and when this panel is shown, autofocus is triggered.

fxdave:
@sysint64 Thanks for this feature. If you have time, could you please update the PR?
I've just tried this branch, because I'm missing this functionality, but I got dependency errors:
```cargo
error: failed to select a version for `piet`.
    ... required by package `piet-web v0.3.1`
    ... which satisfies dependency `piet-web = "^0.3.1"` (locked to 0.3.1) of package `piet-common v0.3.2`
    ... which satisfies dependency `piet-common = "=0.3.2"` of package `druid-shell v0.7.0 (https://github.com/sysint64/druid?branch=focus#3918b24e)`
    ... which satisfies git dependency `druid-shell` of package `druid v0.7.0 (https://github.com/sysint64/druid?branch=focus#3918b24e)`
    ... which satisfies git dependency `druid` of package `vonal-rust v0.1.0 (/home/dbiro/Projects/All/Programming/vonal-rust)`
versions that meet the requirements `^0.3.0` (locked to 0.3.0) are: 0.3.0

all possible versions conflict with previously selected packages.

  previously selected package `piet v0.3.1`
    ... which satisfies dependency `piet = "^0.3.1"` of package `piet-common v0.3.2`
    ... which satisfies dependency `piet-common = "=0.3.2"` of package `druid-shell v0.7.0 (https://github.com/sysint64/druid?branch=focus#3918b24e)`
    ... which satisfies git dependency `druid-shell` of package `druid v0.7.0 (https://github.com/sysint64/druid?branch=focus#3918b24e)`
    ... which satisfies git dependency `druid` of package `vonal-rust v0.1.0 (/home/dbiro/Projects/All/Programming/vonal-rust)`

failed to select a version for `piet` which could resolve this conflict
```

If @cmyr wouldn't at least I would use this branch :D

PR #1092, opened 2020-07-12
added EditAction::SelectNone

sysint64:
this PR adds a new edit action SelectNone as the opposite of SelectAll.

### DIFF
--- a/druid/src/text/editable_text.rs
+++ b/druid/src/text/editable_text.rs
@@ -58,11 +58,11 @@ pub trait EditableText: Sized {
     /// Get the next codepoint offset from the given offset, if it exists.
     fn next_codepoint_offset(&self, offset: usize) -> Option<usize>;
 
-    /// Get the line beginning offset from the given offset
-    fn left_end_of_line(&self, offset: usize) -> usize;
+    /// Get the preceding line break offset from the given offset
+    fn preceding_line_break(&self, offset: usize) -> usize;
 
-    /// Get the line ending offset from the given offset
-    fn right_end_of_line(&self, offset: usize) -> usize;
+    /// Get the next line break offset from the given offset
+    fn next_line_break(&self, offset: usize) -> usize;
 
     /// Returns `true` if this text has 0 length.
     fn is_empty(&self) -> bool;
@@ -163,7 +163,7 @@ impl EditableText for String {
         s.to_string()
     }
 
-    fn left_end_of_line(&self, from: usize) -> usize {
+    fn preceding_line_break(&self, from: usize) -> usize {
         let mut offset = from;
 
         for byte in self.get(0..from).unwrap_or("").bytes().rev() {
@@ -176,7 +176,7 @@ impl EditableText for String {
         0
     }
 
-    fn right_end_of_line(&self, from: usize) -> usize {
+    fn next_line_break(&self, from: usize) -> usize {
         let mut offset = from;
 
         for char in self.get(from..).unwrap_or("").bytes() {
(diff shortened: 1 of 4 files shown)


sysint64:
@cmyr okay, I got it.

cmyr:
@sysint64 ping on this?

sysint64:
@cmyr I'm still here, just a lot of work, I'll work on my PR's on the following weekends.

sysint64:
@cmyr does `Movement::LeftOfLine` do the same thing as should `BeginningOfDocument`? at least the current implementation seems to me what I have tried to do.

cmyr:
It may be the same now, but it shouldn't be the same when we are handling multi-line text correctly; if it is that's likely a bug in the current implementation.

sysint64:
@cmyr Sorry for the delay, I have implemented two methods - `line_beginning` and `line_ending` for `EditableText`, and added `Movement::BeginningOfDocument` and `Movement::EndingOfDocument`.

druid/src/text/movement.rs
```diff
@@ -31,6 +31,10 @@ pub enum Movement {
     LeftOfLine,
     /// Move to right end of visible line.
     RightOfLine,
+    /// Move to the beginning of the document
+    BeginningOfDocument,
+    /// Move to the end of the document
+    EndingOfDocument,
```
cmyr, line 37: I would call these `StartOfDocument` and `EndOfDocument`.

druid/src/text/editable_text.rs
```diff
@@ -58,6 +58,12 @@ pub trait EditableText: Sized {
     /// Get the next codepoint offset from the given offset, if it exists.
     fn next_codepoint_offset(&self, offset: usize) -> Option<usize>;
 
+    /// Get the line beginning offset from the given offset, if it exists.
+    fn line_beginning(&self, offset: usize) -> Option<usize>;
+
+    /// Get the line ending offset from the given offset, if it exists.
+    fn line_ending(&self, offset: usize) -> Option<usize>;
+
```
cmyr, line 66: Let's call these `left_end_of_line` and `right_end_of_line`. 'beginning' and 'end' are ambiguous, since in right-to-left languages the beginning of the line is on the right. :)

druid/src/text/editable_text.rs
```diff
@@ -156,6 +162,32 @@ impl EditableText for String {
...
+    fn line_beginning(&self, from: usize) -> Option<usize> {
+        let mut offset = from;
+
+        for grapheme in self.get(0..from)?.graphemes(true).rev() {
+            if grapheme.chars().next()? == '\n' {
+                return Some(offset);
+            }
+            offset -= grapheme.len();
+        }
```
cmyr, line 174: We don't need to be fancy with graphemes here; we're only looking for the 0x80 (`\n`) byte, and we can must iterate the bytes directly, which is much cheaper.

Additionally I think this doesn't need to return an `Option`; it can return `0` if there's no preceding newline.

druid/src/text/editable_text.rs
```diff
@@ -156,6 +162,32 @@ impl EditableText for String {
...
+        let mut offset = from;
+
+        for grapheme in self.get(from..)?.graphemes(true) {
+            if grapheme.chars().next()? == '\n' {
+                return Some(offset);
+            }
+            offset += grapheme.len();
+        }
+
+        Some(self.len())
```
cmyr, line 189: Same as above, except that we can just return `self.len()` if we don't find a hit.

druid/src/text/editable_text.rs
sysint64: `\n` is 0x0a

sysint64:
@cmyr fixed.

druid/src/text/editable_text.rs
```diff
@@ -58,6 +58,12 @@ pub trait EditableText: Sized {
     /// Get the next codepoint offset from the given offset, if it exists.
     fn next_codepoint_offset(&self, offset: usize) -> Option<usize>;
 
+    /// Get the line beginning offset from the given offset
+    fn left_end_of_line(&self, offset: usize) -> usize;
+
+    /// Get the line ending offset from the given offset
+    fn right_end_of_line(&self, offset: usize) -> usize;
```
cmyr, line 65: Ugh I'm sorry, I screwed up here, was thinking somehow of an API a level higher in the stack.

These should really be `preceding_line_break` and `next_line_break`. Whether or not they are on the left or the right will depend on  the writing direction.

druid/src/text/movement.rs
```diff
@@ -51,8 +55,11 @@ pub fn movement(m: Movement, s: Selection, text: &impl EditableText, modify: boo
             }
         }
 
-        Movement::LeftOfLine => 0,
-        Movement::RightOfLine => text.len(),
+        Movement::LeftOfLine => text.left_end_of_line(s.end),
+        Movement::RightOfLine => text.right_end_of_line(s.end),
```
cmyr, line 59: These are also the wrong names, given the implementation. Ideally these *would* be the names and we would check writing direction on the paragraph, but we don't have that ability right now.

I think it's fine to leave this as is in this PR, and I'll open an issue describing the problem.

cmyr (commented):
Okay apologies, I was a little brain dead during my previous review. There's still a naming issue here, and a larger issue with the naming of the movement types themselves. I'd like to at least get the naming correct here, and we can address the larger issue separately. Sorry for the run-around!

sysint64:
@cmyr it looks like there is an error on the master branch located in `druid/src/box_constraints.rs`

sysint64:
this is what I get after rebase:
```
error[E0428]: the name `tests` is defined multiple times
   --> druid/src/box_constraints.rs:375:1
    |
276 | mod tests {
    | --------- previous definition of the module `tests` here
...
375 | mod tests {
    | ^^^^^^^^^ `tests` redefined here
    |
    = note: `tests` must be defined only once in the type namespace of this module
```

sysint64:
fixed here: #1187

cmyr (approved):
Thanks!

Issue #934, opened 2020-05-15
X11 shell does not support timers.


Finnerale:

xStrom:
I've been thinking of tackling this in the near future. A bunch of examples are broken because of this, so it would be nice to have working.

jneem:
As part of this, the fallback painting code should be changed to use timers. Right now it uses the idle loop, but this isn't ideal: it forces us to either write a bunch of special-case code for painting or else rate-limit the idle loop to avoid painting to often (this is the current state).

Finnerale:
I've opened #1186 as a separate issue for the fallback painting, because this issue has basically been resolved.

PR #1096, opened 2020-07-18
Implement timers for the X11 platform

psychon:
This fixes half of #934, but does *not* deal with https://github.com/linebender/druid/issues/934#issuecomment-660412889

Timers are collected in a `BinaryHeap` per window, sorted by deadline. The main loop gets the first deadline of each window and uses the minimum of that for the timeout used in the call to `poll`. I am least comfortable with my changes to `poll_with_timeout()`. Hopefully, this function can be simplified a lot when someone (else) handles https://github.com/linebender/druid/issues/934#issuecomment-660412889.

This was tested with `cargo run --example invalidate --no-default-features --features x11` (and with nothing else). The result looks similar to the version without `--features x11`, which should be GTK. Before this PR, `--features x11` lead to an empty window instead.

### DIFF
--- a/druid-shell/src/platform/x11/application.rs
+++ b/druid-shell/src/platform/x11/application.rs
@@ -565,8 +565,8 @@ fn poll_with_timeout(conn: &Rc<XCBConnection>, idle: RawFd, timer_timeout: Optio
         };
         // ...and convert the deadline into an argument for poll()
         let poll_timeout = if let Some(deadline) = deadline {
-            if deadline < now {
-                0
+            if deadline <= now {
+                break;
             } else {
                 let millis = c_int::try_from(deadline.duration_since(now).as_millis())
                     .unwrap_or(c_int::max_value() - 1);
--- a/druid-shell/src/platform/x11/application.rs
+++ b/druid-shell/src/platform/x11/application.rs
@@ -575,7 +575,7 @@ fn poll_with_timeout(conn: &Rc<XCBConnection>, idle: RawFd, timer_timeout: Optio
                     break;
                 }
                 let now = Instant::now();
-                if now >= idle_timeout.min(timer_timeout.unwrap_or(idle_timeout)) {
+                if timer_timeout.is_some() && now >= timer_timeout.unwrap() {
                     break;
                 }
                 if poll_fds.len() == 1 || readable(poll_fds[1]) {
@@ -583,6 +583,9 @@ fn poll_with_timeout(conn: &Rc<XCBConnection>, idle: RawFd, timer_timeout: Optio
                     // instead.
                     poll_fds = &mut just_connection;
                     poll_timeout = to_timeout(idle_timeout, now);
+                    if now >= idle_timeout {
+                        break;
+                    }
                 }
             }
 
--- a/druid-shell/src/platform/x11/application.rs
+++ b/druid-shell/src/platform/x11/application.rs
@@ -568,8 +568,11 @@ fn poll_with_timeout(conn: &Rc<XCBConnection>, idle: RawFd, timer_timeout: Optio
             if deadline < now {
                 0
             } else {
-                c_int::try_from(deadline.duration_since(now).as_millis())
-                    .unwrap_or(c_int::max_value())
+                let millis = c_int::try_from(deadline.duration_since(now).as_millis())
+                    .unwrap_or(c_int::max_value() - 1);
+                // The above .as_millis() rounds down. This means we would wake up before the
+                // deadline is reached. Add one to 'simulate' rounding up instead.
+                millis + 1
             }
         } else {
             // No timeout
--- a/druid-shell/src/platform/x11/application.rs
+++ b/druid-shell/src/platform/x11/application.rs
@@ -381,7 +381,18 @@ impl Application {
         let timeout = Duration::from_millis((1000.0 / refresh_rate) as u64);
         let mut last_idle_time = Instant::now();
         loop {
+            // Figure out when the next wakeup needs to happen
+            let next_timeout = if let Ok(state) = self.state.try_borrow() {
+                state.windows
+                    .values()
+                    .filter_map(|w| w.next_timeout())
+                    .min()
+            } else {
+                log::error!("Getting next timeout, application state already borrowed");
+                None
+            };
             let next_idle_time = last_idle_time + timeout;
+
             self.connection.flush()?;
 
             // Before we poll on the connection's file descriptor, check whether there are any
@@ -390,7 +401,7 @@ impl Application {
             let mut event = self.connection.poll_for_event()?;
 
             if event.is_none() {
-                poll_with_timeout(&self.connection, self.idle_read, next_idle_time)
+                poll_with_timeout(&self.connection, self.idle_read, next_timeout, next_idle_time)
                     .context("Error while waiting for X11 connection")?;
             }
 
@@ -409,6 +420,17 @@ impl Application {
             }
 
             let now = Instant::now();
+            if let Some(timeout) = next_timeout {
+                if timeout <= now {
+                    if let Ok(state) = self.state.try_borrow() {
+                        for w in state.windows.values() {
+                            w.run_timers(now);
+                        }
+                    } else {
+                        log::error!("In timer loop, application state already borrowed");
+                    }
+                }
+            }
             if now >= next_idle_time {
                 last_idle_time = now;
                 drain_idle_pipe(self.idle_read)?;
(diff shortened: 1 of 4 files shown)


druid-shell/src/platform/x11/application.rs
```diff
@@ -539,18 +574,15 @@ fn poll_with_timeout(conn: &Rc<XCBConnection>, idle: RawFd, timeout: Instant) ->
                     // There is an X11 event ready to be handled.
                     break;
                 }
+                let now = Instant::now();
+                if now >= idle_timeout.min(timer_timeout.unwrap_or(idle_timeout)) {
+                    break;
+                }
```
jneem, line 580: I think we want to break if `(past the timeout) || (past the idle timeout && there is an idle event)`. Maybe something like:
```suggestion
let now = Instant::now();
if timer_timeout.is_some() && now >= timer_timeout.unwrap() {
    break;
}
```

druid-shell/src/platform/x11/application.rs
```diff
@@ -539,18 +574,15 @@ fn poll_with_timeout(conn: &Rc<XCBConnection>, idle: RawFd, timeout: Instant) ->
...
                     poll_fds = &mut just_connection;
-
-                    let now = Instant::now();
-                    if now >= timeout {
-                        break;
-                    } else {
-                        poll_timeout = c_int::try_from(timeout.duration_since(now).as_millis())
-                            .unwrap_or(c_int::max_value())
-                    }
+                    poll_timeout = to_timeout(idle_timeout, now);
```
jneem, line 585: ... and then here you can keep
```suggestion
                    poll_timeout = to_timeout(idle_timeout, now);
                    if now >= timeout {
                        break;
                    }
```

Also, I think when resetting poll_timeout here you're forgetting the timer timeout. So maybe define `first_timeout` above the loop to be `idle_timeout.min(timer_timeout.unwrap_or(idle_timeout))`, and use `first_timeout` here.

druid-shell/src/platform/x11/util.rs
```diff
@@ -87,3 +91,38 @@ macro_rules! log_x11 {
...
+    }
+
+    pub(crate) fn token(&self) -> TimerToken {
+        self.token
+    }
+}
+
+impl Ord for Timer {
+    fn cmp(&self, other: &Self) -> Ordering {
+        // Ordering is so that earliest deadline sorts first
```
jneem, line 119: Maybe worth pointing out here that we're putting it in a max-heap (because when I think "sorting", the first time that comes to mind is sorting vecs, and then the comment is confusing).

Finnerale:
I did not yet have a look at the code, but I tried it with the `timer` example and noticed two problems:
- The square seems to move slower than in the GTK example
- The square accelerates when I move the cursor to roughly GTK speed

druid-shell/src/platform/x11/application.rs
psychon: > if (past the timeout) || (past the idle timeout && there is an idle event)

Let me rewrite that a bit:
* `now >= timer_timeout || now >= idle_timeout`
Next, if `now` is larger than one of them, then `now` is also larger than the smaller one of the two, so the above is equivalent to:
*  `now >= min(timer_timeout, idle_timeout)`
However, `idle_timeout` is an `Option`, so could be `None`. Let's handle this via `timer_timeout.unwrap_or(idle_timeout)`. There are two cases here:
* If there is a `timer_timeout`, this is equivalent to the above (take the smaller one of the two)
* If there is no `timer_timeout`, this does `now >= min(idle_timeout, idle_timeout)`, and obviously `min(a, a) == a`

I agree that this code is too clever, but I think it does what you want it to do. Hopefully, changing the `idle_timeout` logic to use a timer should simplify all of this, but... dunno, somehow I didn't want to do that.

----

For your suggested change: What happened to `idle_timeout`? It is no longer mentioned.

psychon:
> The square seems to move slower than in the GTK example

Yup. No idea yet what happens. I'll investigate.

To make this a bit less subjective, I did:
```diff
diff --git a/druid/examples/timer.rs b/druid/examples/timer.rs
index f31637a..5949fed 100644
--- a/druid/examples/timer.rs
+++ b/druid/examples/timer.rs
@@ -15,6 +15,7 @@
 //! An example of a timer.
 
 use std::time::Duration;
+use std::time::Instant;
 
 use druid::widget::prelude::*;
 use druid::widget::BackgroundBrush;
@@ -24,6 +25,7 @@ static TIMER_INTERVAL: Duration = Duration::from_millis(10);
 
 struct TimerWidget {
     timer_id: TimerToken,
+    timeout_requested: Instant,
     simple_box: WidgetPod<u32, SimpleBox>,
     pos: Point,
 }
@@ -49,12 +51,20 @@ impl Widget<u32> for TimerWidget {
         match event {
             Event::WindowConnected => {
                 // Start the timer when the application launches
+                self.timeout_requested = Instant::now();
                 self.timer_id = ctx.request_timer(TIMER_INTERVAL);
             }
             Event::Timer(id) => {
                 if *id == self.timer_id {
+                    let elapsed = self.timeout_requested.elapsed();
+                    if elapsed < TIMER_INTERVAL {
+                        println!("Timeout is {:?} early", TIMER_INTERVAL - elapsed);
+                    } else {
+                        println!("Timeout is {:?} late", elapsed - TIMER_INTERVAL);
+                    }
                     self.adjust_box_pos(ctx.size());
                     ctx.request_layout();
+                    self.timeout_requested = Instant::now();
                     self.timer_id = ctx.request_timer(TIMER_INTERVAL);
                 }
             }
@@ -119,6 +129,7 @@ impl Widget<u32> for SimpleBox {
 pub fn main() {
     let window = WindowDesc::new(|| TimerWidget {
         timer_id: TimerToken::INVALID,
+        timeout_requested: Instant::now(),
         simple_box: WidgetPod::new(SimpleBox),
         pos: Point::ZERO,
     })
```
Output is:
Gtk:
```
Timeout is 89.263µs early
Timeout is 563.821µs early
Timeout is 360.599µs early
Timeout is 338.149µs early
Timeout is 587.387µs early
Timeout is 77.686µs early
Timeout is 621.249µs early
Timeout is 234.995µs early
Timeout is 408.812µs early
Timeout is 41.96µs late
Timeout is 414.332µs early
Timeout is 748.723µs early
Timeout is 528.43µs early
Timeout is 780.295µs early
```

X11:
```
Timeout is 129.242µs late
Timeout is 1.397298ms late
Timeout is 1.619458ms late
Timeout is 6.540825ms late
Timeout is 6.784348ms late
Timeout is 164.414µs late
Timeout is 1.38442ms late
Timeout is 1.589043ms late
Timeout is 6.516359ms late
Timeout is 6.808551ms late
Timeout is 132.782µs late
Timeout is 1.398832ms late
Timeout is 1.599091ms late
Timeout is 6.582464ms late
Timeout is 6.730138ms late
Timeout is 131.859µs late
```

psychon:
Something is being slow and I do not know what. I also added
```diff
diff --git a/druid-shell/src/platform/x11/application.rs b/druid-shell/src/platform/x11/application.rs
index 8d879eb..a1cc61f 100644
--- a/druid-shell/src/platform/x11/application.rs
+++ b/druid-shell/src/platform/x11/application.rs
@@ -569,6 +569,9 @@ fn poll_with_timeout(conn: &Rc<XCBConnection>, idle: RawFd, timer_timeout: Optio
                 .contains(PollFlags::POLLIN)
         };
 
+        let before = Instant::now();
+        let p = poll(poll_fds, poll_timeout);
+        println!("Poll for {} returned {:?} in time {:?}", poll_timeout, p, before.elapsed());
         match poll(poll_fds, poll_timeout) {
             Ok(_) => {
                 if readable(poll_fds[0]) {
```
Output is:
```
Poll for 9 returned Ok(0) in time 9.121401ms
Timeout is 6.617062ms late
Poll for 9 returned Ok(0) in time 9.117681ms
Timeout is 6.450287ms late
Poll for 9 returned Ok(0) in time 9.123917ms
Timeout is 6.895794ms late
Poll for 9 returned Ok(0) in time 9.124663ms
Timeout is 6.599921ms late
Poll for 9 returned Ok(0) in time 9.125751ms
```
The `timer` example uses a timeout of 10 ms. This ends up as a 9 ms timeout for `poll()` and `poll()` seems to be obey this request quite well. No idea where the other 6 ms are spent. The output above also suggests that there is exactly one call to `poll()` before the timer is run, so this is also not due to something "forgetting" to run the timeouts.

No idea what to make of this.

psychon:
Found it. It's something like a rounding issue.

When the `timer_timeout` is 9.5 ms in the future, `poll()` is called with a timeout of 9 ms. It then wakes up before the timeout, thus `poll()` is called again. However, this new call still uses the old timeout.

Recomputing the timeout in each loop iteration "does the trick".

druid-shell/src/platform/x11/application.rs
```diff
@@ -525,36 +549,57 @@ fn poll_with_timeout(conn: &Rc<XCBConnection>, idle: RawFd, timeout: Instant) ->
...
+            Some(earliest_timeout)
+        } else {
+            timer_timeout
+        };
+        // ...and convert the deadline into an argument for poll()
+        let poll_timeout = if let Some(deadline) = deadline {
+            if deadline < now {
+                0
+            } else {
+                c_int::try_from(deadline.duration_since(now).as_millis())
```
psychon, line 571: I guess it would be best if this would round up instead of round down (in `as_millis()`). Since that does not seem to be easily possible: How about adding a `+1` to the result? That way, `poll` should sleep until "something interesting" happens. Otherwise, it is possible that `Instant::now()` is less than 1 ms before "something interesting" and the code here would busy loop with `poll()` with a timeout of zero...

jneem: I think adding 1 ms is a good approach.

druid-shell/src/platform/x11/application.rs
```diff
@@ -525,36 +549,57 @@ fn poll_with_timeout(conn: &Rc<XCBConnection>, idle: RawFd, timeout: Instant) ->
...
+        // Compute the deadline for when poll() has to wakeup
+        let deadline = if honor_idle_timeout {
+            Some(earliest_timeout)
+        } else {
+            timer_timeout
+        };
+        // ...and convert the deadline into an argument for poll()
+        let poll_timeout = if let Some(deadline) = deadline {
+            if deadline < now {
+                0
```
jneem, line 569: Why not just break here?

psychon: A combination of "historic reasons" and "I do not follow understand the idle drawing code". The previous code definitely did not return early and might have went into a short busy loop. I think. I am not sure.

I'll just change it to a break (and also to `<=` instead of `<`, but that shouldn't make much of a difference).

PR #1099, opened 2020-07-19
Don't fail fast in CI

Blisto91:
Currently when a job fails the other jobs in the os matrix gets cancelled
I think it would make sense to try and finish the tests on the other operating systems to see if anything else pops up.


Blisto91:
Second self close!

When looking more into i can see why it might make sense get a quick CI response by failing fast

PR #1102, opened 2020-07-25
Give menu entries with "shift" shortcuts upper-case letters.

### LINKED ISSUE #1085: Multiwin example filling terminal with hotkey warning
Whenever I run the multiwin example by `cargo run --example multiwin`, the terminal will fill up with this error:

```
WARN  [druid_shell::hotkey] warning: HotKey HotKey { mods: CtrlShift, key: Character("s") } includes shift, but text is lowercase. Text is matched literally; this may cause problems.
```

When using the hotkey, it does appear to work (at least, it shows a file save prompt), so I don't know why it may cause problems. The error should probably say what issues may arise rather than just say that they could. Or, as a comment next to the error suggests as a possibility, to normalize the casing so there would be no issues.

If the warning is intentional, I don't think it's appropriate to appear in an example about multiple windows. Instead there should be an example made specifically for the purpose of generating warnings.

jneem:
Fixes #1085.

This doesn't actually cause any behavior change that I can see (on gtk backend), but according to the warning message it's the right thing to do.

### DIFF
--- a/druid/src/menu.rs
+++ b/druid/src/menu.rs
@@ -493,7 +493,7 @@ pub mod sys {
             }
             #[cfg(not(target_os = "windows"))]
             {
-                item.hotkey(SysMods::CmdShift, "z")
+                item.hotkey(SysMods::CmdShift, "Z")
             }
         }
     }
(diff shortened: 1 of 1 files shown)


Finnerale:
This should be tested on Windows as well, before it gets merged.

jneem:
Finally got around to spinning up the windows vm, and it works fine there.

Issue #295, opened 2019-11-11
missing impls for bring_to_front_and_focus 


cmyr:
This is a non-critical method on `WindowHandler` that was added in druid#293.

Issue #1045, opened 2020-06-19
Some comments on the X11 platform


psychon:
Hi,

I cam across Druid via https://www.reddit.com/r/rust/comments/h9z66n/changelog_for_druid_06_updated_from_05_two_weeks/.  The changelog then links to a couple of things which then lead to more issues.  I feel like I provide some useful input and hence I am opening this issue. I will try to add things to the already existing issues where applicable.

----

https://github.com/xi-editor/druid/pull/599#issuecomment-609376693

> Resizing the window on the left site makes it blink temporarily towards the
 direction I drag on the right.

Random guess:

https://github.com/xi-editor/druid/blob/d810d30b95da52b3ca89efe63bd3b88b15b5a42b/druid-shell/src/platform/x11/window.rs#L182

You are telling the X11 server to fill any exposed area with white before sending an expose event. I'd recommend just removing this line. The [default background is None](https://www.x.org/releases/X11R7.6/doc/xproto/x11protocol.html#requests:CreateWindow), which means that the server just sends an Expose event without drawing anything beforehand.

> This goes completely crazy when sizing it < 0 width, the window just flies
> away to the right joy. Does not happen when resizing on the right side

This makes no sense to me. X11 windows must have at least a size of 1x1. Sounds like a weird interaction with your WM, but I am not really sure what is going on.

> Application does not exit when closing the window

You seem to have figured out `WM_PROTOCOLS` already. Anyway, I recommend reading EWMH and scrolling through the table of contents of ICCCM (I have never read all of ICCCM myself, I think).

----

https://github.com/xi-editor/druid/pull/599#issuecomment-609459142

> Window redraw flashes: [...]

How are you redrawing? I see that cairo is involved. Are you trying directly to the cairo surface that represents your window? In that case, redrawing is "live" and intermediate steps are already visible (welcome to X11).

I recommend looking at double-buffering. With cairo, this is as simple as `cairo_surface_create_similar` (this is the C function) on the surface.  Internally, this creates an X11 pixmap (if called on an X11 surface) and sets this up as a cairo surface. You can then draw to this surface and only copy your finished drawing to the X11 window.

If you are already doing double-buffering: Sorry for missing that part of the code.

By the way, you are calling `cairo_surface_flush()` where necessary, right?  (Necessary = the drawing should actually be sent to the X11 server now.)

----

https://github.com/xi-editor/druid/issues/475

> Currently figure out the screen's refresh rate and sleep (!) for however long
[...]

PR #1104, opened 2020-07-27
bring_to_front_and_focus implementation on gtk

Kartoffelsaft:
This should be a working implementation of the bring_to_front_and_focus function for the WindowHandle mentioned in #295.

Also, I noticed that this focus function has a documentation comment above it but the function below just has a regular comment. I didn't bother to change it as it seemed unrelated to what I was doing, and I didn't know if that was intentional. The documentation seems to have gotten it anyway.

### DIFF
--- a/druid-shell/src/platform/gtk/window.rs
+++ b/druid-shell/src/platform/gtk/window.rs
@@ -562,8 +562,12 @@ impl WindowHandle {
 
     /// Bring this window to the front of the window stack and give it focus.
     pub fn bring_to_front_and_focus(&self) {
-        //FIXME: implementation goes here
-        log::warn!("bring_to_front_and_focus not yet implemented for gtk");
+        if let Some(state) = self.state.upgrade() {
+            // The GTK docs say not to use this and instead use present_with_timestamp.
+            // This works though, and the docs also don't say how the timestamp
+            // is reperesented (seconds maybe?)
+            state.window.present();
+        }
     }
 
     // Request invalidation of the entire window contents.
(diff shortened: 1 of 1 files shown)


jneem:
This gtk API is probably derived from the x11 API, where when you request input focus you're supposed to supply the timestamp of the event that made you want to request input focus. As I recently learned in #1045, you aren't allowed to use the current time; it has something to do with focus stealing prevention. The relevant snippet from ICCCM is

> Clients that use a SetInputFocus request must set the time field to the timestamp of the event that caused them to make the attempt. [...] Note that clients must not use CurrentTime in the time field.

The underlying issue here is that the druid-shell API doesn't allow for a meaningful timestamp. But given that constraint, there's actually no difference between `present` and `present_with_time(GDK_CURRENT_TIME)`. I just checked the gtk source:

```c
void
gtk_window_present (GtkWindow *window)
{
  gtk_window_present_with_time (window, GDK_CURRENT_TIME);
} 
```
```

**Comment by Kartoffelsaft** [comment]
Given what @jneem said, I think it should be left at `present`, as using `present_with_time` implies that we know what timestamp to use, which we don't. That said, I do think that a `// TODO: ` should be added to indicate that `present_with_time` should be used if and when the timestamp could be known.

And having looked at the GTK source myself, it does do some conversion so that it doesn't break the X11 side of things:
```
gtk_window_present_with_time (GtkWindow *window,
			      guint32    timestamp)
{
// ...
      if (timestamp == GDK_CURRENT_TIME)
        {
// #ifdef ...
	      timestamp = gdk_x11_display_get_user_time (display);
// #endif ...
	    timestamp = gtk_get_current_event_time ();
        }
```

PR #1108, opened 2020-07-30
Scrolling changes and small other fixes to support druid_table

rjwittams:
Druid_table is at https://github.com/rjwittams/druid_table/  for now and has been discussed on Zulip. 

I had to make some small changes to druid to get things working (synchronising the sibling widgets).

- Disabling scroll bars on a Scroll
- Adding a callback to Scroll to emit scroll offsets
- Using a command to set scroll position (including from within layout so it always happens from within event - the callback passes along the evt context to allow sending commands)
- Making some things in environments (Key, KeyOrValue) Clone as I need to copy them around between sibling widgets.
- Added an example I was using to demonstrate synchronising scroll positions

### DIFF
--- a/druid/src/contexts.rs
+++ b/druid/src/contexts.rs
@@ -328,20 +328,30 @@ impl_context_method!(EventCtx<'_, '_>, UpdateCtx<'_, '_>, LifeCycleCtx<'_, '_>, 
 });
 
 // methods on event, update, and lifecycle
-impl_context_method!(EventCtx<'_, '_>, UpdateCtx<'_, '_>, LifeCycleCtx<'_, '_>, LayoutCtx<'_,'_>, {
-    /// Submit a [`Command`] to be run after this event is handled.
-    ///
-    /// Commands are run in the order they are submitted; all commands
-    /// submitted during the handling of an event are executed before
-    /// the [`update`] method is called; events submitted during [`update`]
-    /// are handled after painting.
-    ///
-    /// [`Command`]: struct.Command.html
-    /// [`update`]: trait.Widget.html#tymethod.update
-    pub fn submit_command(&mut self, cmd: impl Into<Command>, target: impl Into<Option<Target>>) {
-        self.state.submit_command(cmd.into(), target.into())
+impl_context_method!(
+    EventCtx<'_, '_>,
+    UpdateCtx<'_, '_>,
+    LifeCycleCtx<'_, '_>,
+    LayoutCtx<'_, '_>,
+    {
+        /// Submit a [`Command`] to be run after this event is handled.
+        ///
+        /// Commands are run in the order they are submitted; all commands
+        /// submitted during the handling of an event are executed before
+        /// the [`update`] method is called; events submitted during [`update`]
+        /// are handled after painting.
+        ///
+        /// [`Command`]: struct.Command.html
+        /// [`update`]: trait.Widget.html#tymethod.update
+        pub fn submit_command(
+            &mut self,
+            cmd: impl Into<Command>,
+            target: impl Into<Option<Target>>,
+        ) {
+            self.state.submit_command(cmd.into(), target.into())
+        }
     }
-});
+);
 
 impl EventCtx<'_, '_> {
     /// Set the cursor icon.
@@ -618,7 +628,6 @@ impl PaintCtx<'_, '_, '_> {
             transform: current_transform,
         })
     }
-
 }
 
 impl<'a> ContextState<'a> {
--- a/druid/src/contexts.rs
+++ b/druid/src/contexts.rs
@@ -318,6 +318,17 @@ impl_context_method!(EventCtx<'_, '_>, UpdateCtx<'_, '_>, LifeCycleCtx<'_, '_>, 
         self.request_layout();
     }
 
+    /// Set the menu of the window containing the current widget.
+    /// `T` must be the application's root `Data` type (the type provided to [`AppLauncher::launch`]).
+    ///
+    /// [`AppLauncher::launch`]: struct.AppLauncher.html#method.launch
+    pub fn set_menu<T: Any>(&mut self, menu: MenuDesc<T>) {
+        self.state.set_menu(menu);
+    }
+});
+
+// methods on event, update, and lifecycle
+impl_context_method!(EventCtx<'_, '_>, UpdateCtx<'_, '_>, LifeCycleCtx<'_, '_>, LayoutCtx<'_,'_>, {
     /// Submit a [`Command`] to be run after this event is handled.
     ///
     /// Commands are run in the order they are submitted; all commands
@@ -329,14 +340,6 @@ impl_context_method!(EventCtx<'_, '_>, UpdateCtx<'_, '_>, LifeCycleCtx<'_, '_>, 
     /// [`update`]: trait.Widget.html#tymethod.update
     pub fn submit_command(&mut self, cmd: impl Into<Command>, target: impl Into<Option<Target>>) {
         self.state.submit_command(cmd.into(), target.into())
-    }
-
-    /// Set the menu of the window containing the current widget.
-    /// `T` must be the application's root `Data` type (the type provided to [`AppLauncher::launch`]).
-    ///
-    /// [`AppLauncher::launch`]: struct.AppLauncher.html#method.launch
-    pub fn set_menu<T: Any>(&mut self, menu: MenuDesc<T>) {
-        self.state.set_menu(menu);
     }
 });
 
@@ -523,10 +526,6 @@ impl LayoutCtx<'_, '_> {
     pub fn set_paint_insets(&mut self, insets: impl Into<Insets>) {
         self.widget_state.paint_insets = insets.into().nonnegative();
     }
-
-    pub fn submit_command(&mut self, cmd: impl Into<Command>, target: impl Into<Option<Target>>) {
-        self.state.submit_command(cmd.into(), target.into())
-    }
 }
 
 impl PaintCtx<'_, '_, '_> {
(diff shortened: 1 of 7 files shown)


ForLoveOfCats:
I like this apart from the fact that of course I'm going to want https://github.com/linebender/druid/pull/1107 merged in and this rebased on that :) The ability to disable scrollbars is definitely needed and a command to ask a scroll to scroll to a position is very helpful for the 95% of time where composition suits the situation.

Adding clone to `Key` and friends looks harmless to me but I'm not an expert in that area. I do know that there was the worry recently about having to many fringe examples but I'll defer to someone else on that.

The CI failing is identical to what is on the aforementioned PR so it doesn't seem to be Github Actions bugging out like we've had in the past so we should probably figure out how we are depending on `harfbuzz_rs` as `cargo tree` shows nothing on my Linux system which suggests that perhaps someone on macOS should take a look as that is the platform it is definitely failing on which surprises me as the macOS piet backend uses CoreGraphics to handle text.

rjwittams:
I still think 1107 needs splitting up into multiple stages (list changes, behaviour split inside the scroll file, and then gratuitous moves/renames), and this shouldn’t be blocked by it. 
I’m definitely not convinced the current scroll control is ideal.

druid/src/widget/scroll.rs
```diff
@@ -410,9 +515,21 @@ impl<T: Data, W: Widget<T>> Widget<T> for Scroll<T, W> {
             }
         }
 
+        if let Event::Command(ref cmd) = event {
+            if let Some(scroll_to) = cmd.get(SCROLL_TO) {
```
jneem, line 519: Just a remark that this is possible without changing `Scroll`, by using `WidgetExt::controller` to wrap the scroll in a controller that intercepts the command. (That kind of composition is my "ideal" for how druid should work, but as you know I don't have a concrete solution for the command-submission/handler/callback half of that ideal; therefore, this is just a remark, not a request for a change.)

druid/src/widget/scroll.rs
```diff
@@ -456,7 +573,16 @@ impl<T: Data, W: Widget<T>> Widget<T> for Scroll<T, W> {
...
+
+        let scroll_to = ScrollTo::new(
+            Some(f64::max(self_size.width - size.width, self.scroll_offset.x)),
+            Some(f64::max(
+                self_size.height - size.height,
+                self.scroll_offset.y,
+            )),
+        );
+
+        ctx.submit_command(SCROLL_TO.with(scroll_to), ctx.widget_state.id);
```
jneem, line 585: When you submit a command in `layout` or `lifecycle`, it won't be processed until (at least) the next frame. I think scrolling immediately is better.

druid/src/widget/view_switcher.rs
```diff
@@ -79,9 +79,7 @@ impl<T: Data, U: PartialEq> Widget<T> for ViewSwitcher<T, U> {
             self.active_child = Some(WidgetPod::new((self.child_builder)(&child_id, data, env)));
             self.active_child_id = Some(child_id);
             ctx.children_changed();
-        }
-
-        if let Some(child) = self.active_child.as_mut() {
+        } else if let Some(child) = self.active_child.as_mut() {
```
jneem, line 82: If I remember correctly from zulip, there's some slightly tricky interaction with lifecycle, right? Maybe add a comment.

druid/src/contexts.rs
```diff
@@ -523,6 +523,10 @@ impl LayoutCtx<'_, '_> {
     pub fn set_paint_insets(&mut self, insets: impl Into<Insets>) {
         self.widget_state.paint_insets = insets.into().nonnegative();
     }
+
+    pub fn submit_command(&mut self, cmd: impl Into<Command>, target: impl Into<Option<Target>>) {
+        self.state.submit_command(cmd.into(), target.into())
+    }
```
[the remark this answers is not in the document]
rjwittams: I noticed that too late! Can look into it

druid/src/contexts.rs
```diff
@@ -633,7 +641,7 @@ impl<'a> ContextState<'a> {
         }
     }
 
-    fn submit_command(&mut self, command: Command, target: Option<Target>) {
+    pub(crate) fn submit_command(&mut self, command: Command, target: Option<Target>) {
```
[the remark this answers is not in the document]
rjwittams: I think I meant to revert this - can't remember why I changed it!

druid/src/widget/scroll.rs
```diff
@@ -162,13 +244,21 @@ impl<T, W: Widget<T>> Scroll<T, W> {
     /// Update the scroll.
     ///
     /// Returns `true` if the scroll has been updated.
-    pub fn scroll(&mut self, delta: Vec2, size: Size) -> bool {
-        let mut offset = self.scroll_offset + delta;
-        offset.x = offset.x.min(self.child_size.width - size.width).max(0.0);
-        offset.y = offset.y.min(self.child_size.height - size.height).max(0.0);
+    pub fn scroll(&mut self, ctx: &mut EventCtx, delta: Vec2, size: Size) -> bool {
```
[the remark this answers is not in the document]
rjwittams: I think you are right, I just didn't want to break anyone

druid/src/contexts.rs
```diff
@@ -615,6 +619,10 @@ impl PaintCtx<'_, '_, '_> {
             transform: current_transform,
         })
     }
+
+    pub fn viewport_offset(&self) -> Vec2 {
```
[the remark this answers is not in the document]
rjwittams: Can do. I will work out if I actually needed this in the end

druid/src/widget/scroll.rs
rjwittams: Hm, thats different than I would expect. Surely there would be some expectations of commands that widgets support out of the box? Given that it is the main way of interacting with them. It does feel like you might end up having to replicate all of your 'interactive' methods as commands...

druid/src/widget/scroll.rs
rjwittams: This was because the scroll callback takes an EventContext (so they can submit commands). I think I may have experimented with giving them the ContextState (and thats why I changed the visibility on that method)? Or something wrapped around it. Maybe need their own context that is minimal and just lets them do what we expect, that can be derived from any other context...

druid/src/widget/view_switcher.rs
rjwittams: Can do

druid/src/widget/scroll.rs
jneem: I did a quick grep. Currently, `TextBox` takes several commands, but not so much for the other built-in widgets. Probably this is partly because we don't have lots of complex built-in widgets and partly because some things that might be slots in (e.g.) qt instead get their behavior from Data (e.g. the text for a `Label`).

rjwittams: I do think it is possible that later down the line we will come up with a sane way somehow put all of this "possibly reactive" stuff  in data, and when its not needed seamlessly do the wrapping and unwrapping hygienically. But I haven't quite worked it out yet..

rjwittams: I've thought about it and I don't think this holds water. I fully understand the benefits of composition, but being able to tell a scroll area to scroll is not at all specialised. Having every app come up with their own series of adhoc commands for widgets is not a good idea. Commands literally are the way you communicate with widgets after they have been boxed up in another widget and really every post construction api method should have a command counterpart.

druid/src/widget/view_switcher.rs
rjwittams: This has ended up in someone elses PR anyway

druid/src/contexts.rs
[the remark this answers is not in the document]
rjwittams: I didn't need it so its gone

druid/src/widget/scroll.rs
jneem: > really every post construction api method should have a command counterpart

I realize you're probably sick of seeing me write this, but I remain very much unconvinced about this part. In my admittedly limited experience with druid, post-construction APIs are only used locally (like in the example here, where you use it to communicate between siblings). If we had a better story for this kind of local communication, I think the non-local case would be rare enough that ad-hoc controllers would be good enough.

Either way, I'd be interested in revisiting this conversation when we all have more experience with composite widgets. As far as this PR goes, I'm not going to complain about it...

druid/src/widget/scroll.rs
```diff
@@ -162,13 +240,21 @@ impl<T, W: Widget<T>> Scroll<T, W> {
     /// Update the scroll.
     ///
     /// Returns `true` if the scroll has been updated.
-    pub fn scroll(&mut self, delta: Vec2, size: Size) -> bool {
-        let mut offset = self.scroll_offset + delta;
-        offset.x = offset.x.min(self.child_size.width - size.width).max(0.0);
-        offset.y = offset.y.min(self.child_size.height - size.height).max(0.0);
+    pub fn scroll_by(&mut self, submit: &mut SubmitCommand, delta: Vec2, size: Size) -> bool {
```
jneem, line 243: I see you've decided to go for the breaking change. Which is fine, but it (along with the other publicly-visible changes in this PR) will need a CHANGELOG entry...

rjwittams: Ok. My thoughts were that the signature had changed so much, may as well break the name.
It already was slightly odd, due to the size, which was obtained from a context from the scrolls own widget pod - calling it from outside meant doing tricks to obtain that. So it was never great as public api - what would be needed for that is “scroll here when you can”, ie a pending scroll - but this is recreating the command system afaict. I’m going to make an alternate PR, based on the binding/scope stuff I mentioned on Zulip. It’s much closer to the composition you are talking about above, and there aren’t any callbacks. Then people can pick between approaches.

rjwittams:
Not much point in this if I'm going to drip feed in bits from binding scroll.

PR #1112, opened 2020-08-04
ViewSwitcher uses Data instead of PartialEq

justinmoon:
As I was playing around with @Finnerale's wonderful [druid-enums](https://github.com/Finnerale/druid-enums) package, I attempted to create a `ViewSwitcher` based on an enum `MyEnum`. I would get an error:

```
no implementation for `MyEnum == MyEnum`
```

It seemed like I could fix this by recursively deriving or implementing `PartialEq` for every type inside `MyEnum`. But perhaps it makes much more sense to use `Data` for such comparisons, right? So that's what I tried to implement here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -136,6 +136,7 @@ This means that druid no longer requires cairo on macOS and uses Core Graphics i
 - The `WindowHandle::get_dpi` method got replaced by `WindowHandle::get_scale`. ([#904] by [@xStrom])
 - The `WinHandler::size` method now gets a `Size` in display points. ([#904] by [@xStrom])
 - Standardized the type returned by the contexts' `text` methods. ([#996] by [@cmyr])
+- `ViewSwitcher` uses `Data` type constraint instead of `PartialEq`. ([#1112] by [@justinmoon])
 
 ### Removed
 
@@ -257,6 +258,7 @@ Last release without a changelog :(
 [@vkahl]: https://github.com/vkahl
 [@psychon]: https://github.com/psychon
 [@sysint64]: https://github.com/sysint64
+[@justinmoon]: https://github.com/justinmoon
 
 [#599]: https://github.com/linebender/druid/pull/599
 [#611]: https://github.com/linebender/druid/pull/611
--- a/druid/src/widget/view_switcher.rs
+++ b/druid/src/widget/view_switcher.rs
@@ -32,7 +32,7 @@ pub struct ViewSwitcher<T, U> {
     active_child_id: Option<U>,
 }
 
-impl<T: Data, U: PartialEq> ViewSwitcher<T, U> {
+impl<T: Data, U: Data> ViewSwitcher<T, U> {
     /// Create a new view switcher.
     ///
     /// The `child_picker` closure is called every time the application data changes.
@@ -55,7 +55,7 @@ impl<T: Data, U: PartialEq> ViewSwitcher<T, U> {
     }
 }
 
-impl<T: Data, U: PartialEq> Widget<T> for ViewSwitcher<T, U> {
+impl<T: Data, U: Data> Widget<T> for ViewSwitcher<T, U> {
     fn event(&mut self, ctx: &mut EventCtx, event: &Event, data: &mut T, env: &Env) {
         if let Some(child) = self.active_child.as_mut() {
             child.event(ctx, event, data, env);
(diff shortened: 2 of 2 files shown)


druid/src/widget/view_switcher.rs
```diff
@@ -75,10 +75,13 @@ impl<T: Data, U: PartialEq> Widget<T> for ViewSwitcher<T, U> {
...
-            self.active_child = Some(WidgetPod::new((self.child_builder)(&child_id, data, env)));
-            self.active_child_id = Some(child_id);
-            ctx.children_changed();
+        if let Some(active_child_id) = &self.active_child_id {
+            if child_id.same(&active_child_id) {
+                self.active_child =
+                    Some(WidgetPod::new((self.child_builder)(&child_id, data, env)));
+                self.active_child_id = Some(child_id);
+                ctx.children_changed();
+            }
```
justinmoon, line 84: There must be a more elegant a more elegant way to implement this control flow ...

Zarenor: I think you're missing a negation in there - we're creating a new child if `child_id` is different from `active_child_id` (the new check here checks for sameness without negation)

Also, what happens here in the case where `active_child_id` is `None`? I think in that case we should be creating the appropriate child? This was handled by some-wrapping the non-optional `child_id` (returned from `child_picker`) in the old version. If the active one was none, we enter the change block.

justinmoon: You were right about the negation. Fixed that.

As for the second question, I don't know. I only have a hazy understanding of how this component actually works. But it doesn't seem the existing code does this so perhaps that could be a separate PR?

Finnerale:
This will also need a changelog entry in the `Changed` section, see [CONTRIBUTING.md](https://github.com/linebender/druid/blob/master/CONTRIBUTING.md#changelog).

druid/src/widget/view_switcher.rs
Finnerale: > Also, what happens here in the case where `active_child_id` is `None`? 

In that case, we would have a bug somewhere else. Because `Widget::lifecycle(LifeCycle::WidgetAdded)` must be the first event a widget receives and thus `active_child_id` can only be `None` if that rule gets violated somewhere.

Just doing
```rust
        if !child_id.same(self.active_child_id.as_ref().unwrap()) {
```
would be fine in my opinion.

justinmoon: Then why is it `Option` at all?

Finnerale: Because we have to initialize it lazily during `WidgetAdded`, and thus we have to assign some value (`None`) to it until we get there.

justinmoon: Went with @Finnerale's suggestion here.

justinmoon:
> This will also need a changelog entry in the `Changed` section, see [CONTRIBUTING.md](https://github.com/linebender/druid/blob/master/CONTRIBUTING.md#changelog).

Done.

CHANGELOG.md
```diff
@@ -136,6 +136,7 @@ This means that druid no longer requires cairo on macOS and uses Core Graphics i
 - The `WindowHandle::get_dpi` method got replaced by `WindowHandle::get_scale`. ([#904] by [@xStrom])
 - The `WinHandler::size` method now gets a `Size` in display points. ([#904] by [@xStrom])
 - Standardized the type returned by the contexts' `text` methods. ([#996] by [@cmyr])
+- `ViewSwitcher` uses `Data` type constraint instead of `PartialEq` ([#1112] by [@justinmoon])
```
Finnerale, line 139: You will also have to add a link for `[#1112]` to the others (below the names), and the dot is missing :stuck_out_tongue: 
```suggestion
- `ViewSwitcher` uses `Data` type constraint instead of `PartialEq`. ([#1112] by [@justinmoon])
```
Those aside, this looks good to me. @Zarenor what do you think?

justinmoon: Fixed

Zarenor:
This looks alright to me. 

I'm a little worried about this making a failure of a widget to get `WidgetAdded` turn into a panic (it's a situation that happens often-enough when building new child-having widgets), but I don't think that should be a blocker unless it becomes a more serious issue than I think it is.

Thanks for working this out, @justinmoon !

PR #1113, opened 2020-08-05
Skip update in ViewSwitcher after switching.

Finnerale:
As @justinmoon reported in an issue on `druid-enums` [here](https://github.com/Finnerale/druid-enums/issues/2), there was a problem when wrapping a `Matcher` into `ViewSwitcher`. This was caused, because `Matcher` relies on receiving `LifeCycle::WidgetAdded` as first event, and `ViewSwitcher` was violating that rule by calling update directly on a newly created widget.

I'm not perfectly sure yet, how those things are interacting with each other and what other problems we might have here, but this change makes sense to me regardless of other problems we might have.

`ViewSwitcher` caused the following warning before (can be seen in the `view_switcher` example):
```
old_data missing in WidgetId(27), skipping update.
```


rjwittams:
That was included in my #1108 as I also found it. The fix seems to work ok!

PR #1114, opened 2020-08-05
Add password option to text box.

derekdreery:
This is a little bit of functionality that I needed. I'm not sure if it belongs in druid, but I thought I'd make the PR to stimulate discussion.

### DIFF
--- a/druid/src/widget/textbox.rs
+++ b/druid/src/widget/textbox.rs
@@ -109,7 +109,7 @@ impl TextBox {
             text.to_string()
         };
         piet_text
-            .new_text_layout(&font, &text.to_string(), std::f64::INFINITY)
+            .new_text_layout(&font, &text, std::f64::INFINITY)
             .build()
             .unwrap()
     }
--- a/druid/src/widget/textbox.rs
+++ b/druid/src/widget/textbox.rs
@@ -15,6 +15,7 @@
 //! A textbox widget.
 
 use std::time::Duration;
+use unicode_segmentation::UnicodeSegmentation;
 
 use crate::{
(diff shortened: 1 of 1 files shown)


derekdreery:
Just found that the caret position uses the wrong text - I'll fix then re-enable the PR.

druid/src/widget/textbox.rs
```diff
@@ -90,8 +99,17 @@ impl TextBox {
             .build()
             .unwrap();
 
+        let text = if self.password_mode {
+            let mut t = String::new();
+            for _ in text.grapheme_indices(true) {
+                t.push('*');
```
jneem, line 105: I think a [black circle](http://www.fileformat.info/info/unicode/char/25CF/index.htm) looks nicer, personally :shrug:

druid/src/widget/textbox.rs
```diff
@@ -90,8 +99,17 @@ impl TextBox {
...
+            }
+            t
```
cmyr, line 107: not a blocker but I would write this as,

```rust
let text = if self.password_mode {
    text.grapheme_indices(true).map(|_| '*').collect()
} else {
    text.to_string()
};
```

or even (because I think it's a bit annoying that we're allocating a `String` from a `&str` and then immediately borrowing it again):

```rust
static PASSWORD_CHARS: &str = "*****************************************";

let text = if self.password_mode {
    let num_chars = text.grapheme_indices(true).count().min(PASSWORD_CHARS.len());
    &PASSWORD_CHARS[..num_chars]
} else {
    text
};
```

cmyr:
@derekdreery status of this? textbox stuff is all in flux right now, so should probably close and redo as needed once the other stuff lands?

derekdreery:
Yeah lets close - there's no point adding something that's going to need changing every time the API changes. I will reopen once text has settled down.

PR #1128, opened 2020-08-13
First cut at request_update

### LINKED ISSUE #1073: Updating child widgets?
I have a custom widget that has a tree of child widgets, where the data of the child widgets is made up of both the widget's data and some of the widget's internal state. The `event`, `lifecycle`, `update`, `layout`, and `paint` methods all handle assembling these together.

In particular, the `event` method looks like this:

```rust
fn event(&mut self, ctx: &mut EventCtx, event: &Event, tools: &mut ToolTable, env: &Env) {
    let mut data = (self.state.to_owned(), tools.to_owned());

    self.widgets.event(ctx, event, &mut data, env);

    let (new_state, new_tools) = data;

    if !new_state.same(&self.state) {
        self.state = new_state;
    }

    if !new_tools.same(&new_tools) {
        *tools = new_tools;
    }
}
```

This works great, except that when the widget's internal state changes, the children do not get updated.

Does it make sense to add a `request_update` method on `EventCtx`? Or is there a better way to do this?

raphlinus:
Fixes #1073. This version is largely for discussion, as it hasn't been
tested at all.

### DIFF
--- a/druid/src/contexts.rs
+++ b/druid/src/contexts.rs
@@ -25,7 +25,7 @@ use crate::piet::Piet;
 use crate::piet::RenderContext;
 use crate::{
     commands, Affine, Command, ContextMenu, Cursor, Insets, MenuDesc, Point, Rect, SingleUse, Size,
-    Target, Text, TimerToken, Vec2, WidgetId, WindowDesc, WindowHandle, WindowId,
+    Target, Text, TimerToken, Vec2, WidgetId, WidgetPod, WindowDesc, WindowHandle, WindowId,
 };
 
 /// A macro for implementing methods on multiple contexts.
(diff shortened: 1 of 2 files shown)


druid/src/contexts.rs
```diff
@@ -495,6 +495,23 @@ impl EventCtx<'_, '_> {
             );
         }
     }
+
+    /// Request an update cycle.
+    ///
+    /// After this, `update` will be called on the widget in the next update cycle, even
+    /// if there's not a data change.
+    pub fn request_update<T, W>(&mut self) {
```
rjwittams, line 503: Generic parameters aren't needed here

rjwittams:
This does now work for e.g Tabs in my binding-scroll branch

druid/src/contexts.rs
```diff
@@ -495,6 +495,14 @@ impl EventCtx<'_, '_> {
             );
         }
     }
+
+    /// Request an update cycle.
+    ///
+    /// After this, `update` will be called on the widget in the next update cycle, even
+    /// if there's not a data change.
```
cmyr, line 502: I think it's worth highlighting that this method is only exposed for special cases, such as when a container widget is synthesizing data for its children, and that in general it should be unnecessary/avoided.

druid/src/tests/mod.rs
```diff
@@ -463,3 +463,29 @@ fn register_after_adding_child() {
...
+        })
+        .update_fn(move |_, _ctx, _old_data, _data, _env| {
+            updated_clone.set(true);
+        });
+    Harness::create_simple((), widget, |harness| {
+        harness.send_initial_events();
+        assert!(!updated.get());
+        harness.submit_command(REQUEST_UPDATE, None);
+        assert!(updated.get());
+    })
```
cmyr, line 490: nice to see how straight forward and readable this is.

Issue #1135, opened 2020-08-17
Add Prism as a co-Lens


swfsql:
TL;DR This is an exploration for using enums with Prisms, similarly on how structs are used with Lenses.  
Next step is try having them working comfortably with/as Widgets.

## Widget, Data, and Lenses

A Lens is, as [documented](https://docs.rs/druid/0.6.0/druid/trait.Lens.html), "a datatype that gives access to a part of a larger data structure".

This is a useful concept/tool that often facilitates the handling and manipulation of data, of which *reusable* Widgets benefit from. A Widget may be reusable because it's defined to work with, let's say, a `bool`, and there are a lot of places and different libraries that could want to also use that Widget.

One detail of Widgets is that they need to have their instance defined inside some function, inside some scope - and note that such functions usually return `-> impl Widget<MyWholeData>` or something along these lines. Assume `MyWholeData` to be a struct that contains some stuff, including a `bool`, which we want our Widget to work it.  
But we can't just use/return a `Widget<bool>` - a widget that works with / manipulates a `bool` - in a function that is expecting to return a Widget that work with / manipulate a `MyWholeData`. And we also *cannot* access the data by ourselves (to then hand it over to the widget), because let's say we might try having something like: 
```rust
my_widget_that_works_with_bool()
    .actually_here_is_the_data_that_you_work_with(my_whole_data.my_bool)
```
But that is not possible, because here in our scope we are only defining (the instance of) the Widget, and only that. We don't actually have / hold an instance of the data (the `my_whole_data`, which would be an instance of `MyWholeData`) and therefore we can't really "access" the `bool` and "hand it over" to the `Widget<bool>`.

What we really do is to *instruct* (that instance of) the `Widget<bool>` *on how to* handle a `MyWholeData` data when given such data. And what it will do, is to access the inner `bool` contained in the `MyWholeData` by itself! So that's where we use lenses: it's that instruction itself.  
(and then, after the correct lensing is applied, that `Widget<bool>` can now be treated as a `Widget<MyWholeData>`, which was the intention, which is the return type of the function)

[...]

PR #1136, opened 2020-08-17
Initial impl for adding Prism for enums

swfsql:
Related to https://github.com/linebender/druid/issues/1135  
(and other links in the issue).

# Testing

```sh
git clone https://github.com/swfsql/druid.git
cd druid
git checkout prism-for-enums
cargo test --test with_prism
cargo test --test with_traversal
cargo run --example either_prism
cargo run --example login
```

Notes:  
The login example is based on the `druid-enums` login example.  
I copied most of the interface, but adapted into the prisms, while also trying to be as non-`dyn` as possible.

I currently don't trust that the implementation is correct, it really should be checked by someone more intimate with lenses from Haskell/Scala.  
Not to mention that the current example logs warning/errors regarding widget events/lifecycle/layout. -- edit: I believe newer versions of druid minimized those warnings/errors.

There are many TODOs, and almost no documentation.  
I'm not sure if this whole prism part should be simply moved into a complementary crate (and then be imported separately), or if both lenses and prisms should be made into a third-party crate.

Finally, the current "widget" usage of prisms should still be improved.. I guess!


druid-derive/src/lens.rs
```diff
@@ -104,12 +104,30 @@ fn derive_struct(input: &syn::DeriveInput) -> Result<proc_macro2::TokenStream, s
         let field_ty = &f.ty;
 
         quote! {
-            impl #impl_generics druid::Lens<#ty#ty_generics, #field_ty> for #twizzled_name::#field_name #where_clause {
-                fn with<#val_ty_par, #func_ty_par: FnOnce(&#field_ty) -> #val_ty_par>(&self, data: &#ty#ty_generics, f: #func_ty_par) -> #val_ty_par {
+            impl #impl_generics druid::Lens<
```
rjwittams, line 107: Are these just formatting changes? Did rustfmt do it? I wonder why

druid/examples/either_prism.rs
```diff
@@ -0,0 +1,126 @@
...
+    }
+}
+
+impl Default for SliderOrLabel {
+    fn default() -> Self {
+        Self::Label("Click to reveal slider".into())
+    }
+}
+
+impl Widget<SliderOrLabel> for Checkbox {
```
rjwittams, line 49: I think there should be some kind of widget wrapper here doing the adaptation of checkbox to the enum.
This pattern of implementing From on boolean then reimplementing Widget seems quite verbose and hard to explain.

druid/src/optics/lens/lens.rs
```diff
@@ -35,13 +37,13 @@ use crate::Data;
 /// sophistication, for example combinators to combine lenses.
 ///
 /// [Haskell lens]: http://hackage.haskell.org/package/lens
-pub trait Lens<T: ?Sized, U: ?Sized> {
+pub trait Lens<S: ?Sized, A: ?Sized> {
```
rjwittams, line 40: What is the goal of these renames?

druid-derive/src/lens.rs
swfsql: Yes, they are only formatting changes; and rustfmt did not do it; I did it so I could read more easily as more variables were used

druid/examples/either_prism.rs
swfsql: Yes, I agree. I'm not comfortable with the current example implementation, and I plan to have / believe there are better ways to implement it. I'd say the purpose of this example is to be temporary, and to show how/why it should be improved.

druid/src/optics/lens/lens.rs
swfsql: This is to match more closely how they are named in external resources, as in [haskell](https://hackage.haskell.org/package/lens-4.19.2/docs/Control-Lens-Lens.html#t:Lens-39-), [Scala](https://www.optics.dev/Monocle/api/monocle/Lens$.html) and in some blog posts/videos.. I'm not sure what they mean: I found something like "Source" and "Target", which doesn't makes sense to me. But overall, I think this terminology is very adopted in external resources.

But I didn't make it entirely consistent, as I ended up preserving the A -> B -> C terminology in some cases, where multiple lenses are used.

I applied the renames to the prisms as well (as I initially copied the original lens terminology).

cmyr:
I'm very curious about this idea, but I think maybe the PR is the wrong place to start; I would like to see a clearer description of what the problem is, how this proposes solving the problem, and how that compares to other explorations like the `druid-enums` work. This doesn't need to be a massive document or anything, but it should state as simply as possible what problem we're solving and why we're taking this approach. I think it probably makes the most sense to have this discussion in #1135. Does that sound reasonable?

cmyr:
marking this as 'blocked' until we've come to some consensus in #1135. 👌

druid/examples/either_prism.rs
swfsql: I'm not sure if this example is ever getting better, but I copied/adapted an example from `druid-enums`' `login.rs` example, which is more likely to work.

druid/src/optics/lens/lens.rs
swfsql: Recently I dropped all `T`-`U`, `S`-`A` and `A`-`B`-`C` terminology and just stuck with `T1`-`T2` and `T1`-`T2`-`T3` as I think it becomes easier to read all of the related functions. The exception is where only `T` is used.

swfsql:
Hi, I happened to revisit this and I'm opting to close it. Although I haven't used it, Xilem design appears to be easier and also very good.

PR #1137, opened 2020-08-18
Rework ExtEventSink.

jneem:
This ties the external event handles to the windows rather than to the global app.

I'm not sure if this is a good idea (hence the draft status). It makes my usages of `ExtEventSink` rather nicer, because I can just get a sink from a context; see the changes to `examples/blocking_function.rs` to see why this is convenient.


ForLoveOfCats:
I don't have the time right now to read through the required code to be able to perform any sort of review but I would like to say that I am *greatly* in favor of making it easy to get the sink from a context. That solves a major pain point in my usage of it.

Finnerale:
I think I'll find some time for this today, so for now I'll just complain about the missing changlog entry.

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ You can find its changes [documented below](#060---2020-06-01).
 - `Scale::from_scale` to `Scale::new`, and `Scale` methods `scale_x` / `scale_y` to `x` / `y`. ([#1042] by [@xStrom])
 - Major rework of keyboard event handling. ([#1049] by [@raphlinus])
 - `Container::rounded` takes `KeyOrValue<f64>` instead of `f64`. ([#1054] by [@binomial0])
+- `ExtEventSink` was redone to work per-window instead of globally. ([#1137] by [@jneem])
```
Finnerale, line 26: I like this and it is likely much better for most use cases, but I wonder if there is a valid need for an event sink at the delegate level? Because with this it seems no longer possible to have some application wide background thread / event sink.

jneem: Yeah, this was my main concern too, and I was kind of hoping that someone who's using it that way would stop by and explain their use case.

I thought about trying to support both global and per-window ExtEventSinks; not only is it a bit tricky, the existing global one doesn't work that predictably/reliably for a reason that I don't know how to fix: it needs an idle loop to work, and the idle loop is tied to a window. The existing `ExtEventSink` works around this by jumping to a new window when the existing one is closed, but it makes command-targeting work strangely and I think it will also fail if all windows are closed but the application is still running.

Finnerale:
Maybe we could have something like `get_external_handle(Target)`?
And then you could use it like this
```rust
ctx.get_external_handle(ctx.widget_id());
ctx.get_external_handle(ctx.window_id());
ctx.get_external_handle(Target::Global);
```
I'm not sure though, if the latter would make sens at all.

jneem:
I'm not sure that I understand your objection, @rjwittams. In each widget, you could do
```rust
let ev = ctx.get_external_handle();
let id = ctx.window_id();
std::thread::spawn(move || {
    //...
    handle.submit_command(.., Target::Widget(id));
});
```

And then you can have one thread per widget. They don't need to cooperate at all -- the window will get commands from both sources and will handle dispatching them to their intended targets.

PR #1151, opened 2020-08-22
Scope widget cherry-picked from binding-scroll branch.

rjwittams:
As discussed many times on Zulip. 

Scope allows you to encapsulate data so local reactivity is possible without polluting your global app state.


jneem:
I understand that it will be convenient to do the example(s) later, but what about docs for the public API functions?

druid/src/widget/scope.rs
```diff
@@ -0,0 +1,278 @@
...
+/// b) Provide two way synchronisation between the input and the state via a ScopeTransfer
+///
+/// Convenience methods are provided to make a policy from a function and a lens.
+/// It may sometimes be advisable to implement ScopePolicy directly if you need to
+/// mention the type of a Scope.
+///
+/// # Examples
+/// ```
+/// #[derive(Data, Lens)]
+/// struct AppState{
```
raphlinus, line 153: Minor formatting nits, there seem to be spaces missing; rustfmt apparently doesn't catch issues inside doctests, which seems to me an oversight.

rjwittams: I ran rustfmt on the example so it looks standardised now

druid/src/widget/scope.rs
```diff
@@ -0,0 +1,282 @@
...
+        state: SP::State,
+        transfer: SP::Transfer,
+    },
+}
+
+/// A widget that allows encapsulation of application state.
+///
+/// This is useful in circumstances where
+/// * A (potentially reusable) widget is composed of a tree of multiple cooperating child widgets
+/// * Those widgets communicate amongst themselves using Druids reactive data mechanisms
```
jneem, line 124: Druid's

Issue #1029, opened 2020-06-12
Stepper does not clamp against rounding errors


thePalindrome:
The Stepper, despite being (currently, see #762 ) exclusively for floats, does not seem to implement clamping against rounding errors.

Given the following compilable example (with warnings):

```rs
use druid::{AppLauncher, WindowDesc, Widget, PlatformError, LensWrap, Lens, Data, LensExt};
use druid::widget::{
    Padding, Stepper, Switch, WidgetExt, Flex, Label, TextBox, Parse
};
fn main() {
AppLauncher::with_window(WindowDesc::new(build_ui)).launch(State {
     n_value: 20.0
     });
}
fn build_ui() -> impl Widget<State> {
let mut row = Flex::row();
let n_textbox = LensWrap::new(
        Parse::new(TextBox::new()),
        State::n_value.map(|x| Some(*x), |x, y| *x = y.unwrap_or(0.0)),
    );
    let stepper = LensWrap::new(
        Stepper::new()
            .with_range(0.0, u32::MAX.into())
            .with_step(1.0)
            .with_wraparound(false),
        State::n_value,
    );
row.add_child(Padding::new(5.0, n_textbox));
row.add_child(Padding::new(5.0, stepper));
row.center()
}

#[derive(Lens, Data, Clone)]
struct State {
    n_value: f64,
}
```

Use the buttons to go up and down a few times, and you'll note that some values will end up being skipped.

As a note, I'm building off of 0.6.0, and would state that building off of master is something that should be discouraged (so if this is fixed in master, I request this not be closed until the release is pushed onto crates.io)

Finnerale:
This has nothing to do with rounding errors, in fact I've had this issue 'fixed' at one point and that fix caused it to no longer work on macOS and thus was reverted.

I'll take a look if we can solve that properly by now.

The original fix (for GTK): #713 
And the revert: #716

---

Also, I don't think building druid from master is discouraged at all, we make sure that master builds and runs at any time and using it allows catching bugs earlier during development. Keeping issues open until their solution is on crates.io sounds interesting, but also annoying (we'll have to close them all once we make a release), and I don't want to spent time on that yet, considering how fast druid is progressing. But once we reach stable 1.0 that might be something to consider.

Finnerale:
Implementing the ideas from #859 would solve this.
> - No extra double or triple click events, it's merely an extra event property for those who care.

xStrom:
[...]

Issue #859, opened 2020-04-19
Designing a unified multi-click API


xStrom:
While working on #843 I also started thinking about double-clicks and how to best do them. I think that the behavior should match across platforms so that druid users can just read the druid docs and not worry about platform specifics.

Double-clicks are good, triple-clicks possibly worth it, but what about quadruple-clicks or even more? I'm not sure there's really much value in going beyond three. It's definitely more work.

* Windows natively supports only up to double, so will need a custom implementation anyway.
* GTK seems to support triple clicks, so we can probably piggyback on that for up to triple.
* macOS supports up to triple on most buttons, but seems to be practically unbounded for left and right buttons. We can piggyback on the platform for up to triple, including `unbounded_count % 3`.
* Web has some support via [UIEvent.detail](https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/detail) but I haven't tested how well it works.

Custom implementations (which Windows will need anyway) will need to be careful to follow the user configured OS setting for double click timeout.

Thus I propose the following unified API:
* Support for up to triple-clicks.
* Both `MouseDown` and `MouseUp` contain the click count.
* No extra double or triple click events, it's merely an extra event property for those who care.

cmyr:
I think this design makes sense.

If we're using a simple timer to determine whether a click is part of a 'click sequence' I'm not if it makes much difference whether or not we stop at three, but I also don't think quadruple clicks are a problem we need to worry about in this century? Famous last words?

ratmice:
Question about click count, If i recall, mac OS would send sequentially higher click counts, so you would get an event with `MouseDown` and click count 1, then one with 2, etc.
Is this what you had in mind, or would their be a single event with click count 2.

Double clicks and event propagation can sometimes be tricky, presume it's possible that the first click gets put on some system platform event queue, same for the second click +incr click count,
the first event causes the window to close, the target for the 2nd event now no longer exists.

Anyhow its just an awkward case to keep in mind

cmyr:
It's a timer. playing around, here are three events with different positions that count as a click sequence, and then two events in the same location that don't:

```
(155.7109375, 201.76953125): 1
(155.41796875, 201.26171875): 2
(154.8828125, 200.51171875): 3

(147.55859375, 248.6640625): 1
(147.55859375, 248.6640625): 1
```

xStrom:
Yes my earlier macOS findings match this, indeed it's a timer.

[...]

PR #716, opened 2020-03-20
Revert "Fix Stepper wraparound and double step"

cmyr:
This reverts commit 6e9a1edb5b25a302bad1fbd3ea9b2abf788fa9a6.

as mentioned in this zulip thread (https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/druid.23683.20Clickable/near/191191656) this had a bug that caused clicks to not register; something to do with the timer.

PR #1154, opened 2020-08-23
Temp fix for stepper triggering multiple times with double/triple click

ForLoveOfCats:
Closes https://github.com/linebender/druid/issues/1029

A better solution would be to implement https://github.com/linebender/druid/issues/859 as then multi-click events would only fire one mouse down event but until someone does that (which will take some work with the different platforms which will be the tricky part) it made sense to handle the current situation a bit more gracefully. I'm on Linux so this is only tested on the GTK and x11 shells so it would be great if someone could test this on Windows and macOS. Something to note is that the x11 behavior appeared to already behave correctly but that seems to be because it doesn't implement multi-click at all [see here](https://github.com/linebender/druid/blob/master/druid-shell/src/platform/x11/window.rs#L825).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@ You can find its changes [documented below](#060---2020-06-01).
 - Allow submit_command from the layout method in Widgets ([#1119] by [@rjwittams])
 - Allow derivation of lenses for generic types ([#1120]) by [@rjwittams])
 - Switch widget: Toggle animation being window refresh rate dependent ([#1145] by [@ForLoveOfCats])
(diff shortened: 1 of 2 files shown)


Finnerale (changes requested):
This will break macOS. I've tried it before in #713 and we had to revert it #716 because on macOS normal clicks will count as double or triple clicks, or rather GTK considers double clicks also as single clicks.

Anyway, this will probably have to wait for #859 or require GTK specific patching.

ForLoveOfCats:
Damn, that's frustrating. Thanks for linking the related context

PR #1155, opened 2020-08-24
Move AnimFrame from lifecycle to event.

### LINKED ISSUE #1060: Changing the data on every frame
I'm working on an animation program, and I'm keeping the current time of the animation in my `Data`. While playing back the animation, I want to update the data on every frame, but I can't find a good way to do this right now.

Originally, I was firing a timer every 16ms, but the [druid-shell docs](https://docs.rs/druid-shell/0.6.0/druid_shell/struct.WindowHandle.html#method.request_timer) (the druid docs is missing this warning, but I'll submit a PR) suggest that it isn't a good idea. Empirically, it works well on GTK but poorly on windows.

I switched to using `AnimFrame`, but since `lifecycle` can't modify the data directly I had to (as suggested in the docs) change the data indirectly through a command. Unfortunately, the change to the data doesn't happen until the following frame (and sometimes not even until the one after -- I've recorded delays of up to 30ms). This method works better than the timer method on Windows, but is noticeably worse on GTK.

Could there be a reliable way to change the data just before drawing the frame?

jneem:
Fixes #1060.

I'm sure there's a good reason for the current placement of AnimFrame, so I definitely expect pushback on this; I'm mostly just exploring possible solutions for #1060, and this seemed like the easiest.

I think there is part of this patch that is worth salvaging in any case: it simplifies the handling of `last_anim` in window.rs, and it makes the behavior of `AnimFrame` match the documentation, even in the case that the animation is first requested in the event context (previously, that would lead to `AnimFrame` having a non-zero initial value).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ You can find its changes [documented below](#060---2020-06-01).
 - Replaced uses of `Option<Target>` with the new `Target::Auto`. ([#1185] by [@finnerale])
 - Moved `Target` parameter from `submit_command` to `Command::new` and `Command::to`. ([#1185] by [@finnerale])
 - `Movement::RightOfLine` to `Movement::NextLineBreak`, and `Movement::LeftOfLine` to `Movement::PrecedingLineBreak`. ([#1092] by [@sysint64])
+- `AnimFrame` was moved from `lifecycle` to `event` ([#1155] by [@jneem])
 
 ### Deprecated
 
@@ -414,6 +415,7 @@ Last release without a changelog :(
 [#1145]: https://github.com/linebender/druid/pull/1145
 [#1151]: https://github.com/linebender/druid/pull/1151
 [#1152]: https://github.com/linebender/druid/pull/1152
+[#1155]: https://github.com/linebender/druid/pull/1155
 [#1157]: https://github.com/linebender/druid/pull/1157
 [#1171]: https://github.com/linebender/druid/pull/1171
 [#1172]: https://github.com/linebender/druid/pull/1172
--- a/druid/examples/anim.rs
+++ b/druid/examples/anim.rs
@@ -26,21 +26,23 @@ struct AnimWidget {
 
 impl Widget<u32> for AnimWidget {
     fn event(&mut self, ctx: &mut EventCtx, event: &Event, _data: &mut u32, _env: &Env) {
-        if let Event::MouseDown(_) = event {
-            self.t = 0.0;
-            ctx.request_anim_frame();
+        match event {
+            Event::MouseDown(_) => {
+                self.t = 0.0;
+                ctx.request_anim_frame();
+            }
+            Event::AnimFrame(interval) => {
+                ctx.request_paint();
+                self.t += (*interval as f64) * 1e-9;
+                if self.t < 1.0 {
+                    ctx.request_anim_frame();
+                }
+            }
+            _ => (),
         }
     }
 
-    fn lifecycle(&mut self, ctx: &mut LifeCycleCtx, event: &LifeCycle, _data: &u32, _env: &Env) {
-        if let LifeCycle::AnimFrame(interval) = event {
-            ctx.request_paint();
-            self.t += (*interval as f64) * 1e-9;
-            if self.t < 1.0 {
-                ctx.request_anim_frame();
-            }
-        }
-    }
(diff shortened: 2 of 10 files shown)


raphlinus:
I don't have very strong feelings whether animframe is an event or a lifecycle; I think I originally had it as the former. I believe @cmyr was the main driver behind moving it to lifecycle, and might have a good argument why it should be kept there.

cmyr:
the animation event is definitely one of the awkward corners of druid, as it currently exists.

The reason I don't like `AnimFrame` as an event is pretty simple: it behaves differently from every other event.

In general, in druid, execution follows a very simple cycle:

- the runloop sits idle, waiting for an event to arrive from the OS
- an event arrives, and we call `event` on the root widget
  - various `LifeCycle` events possibly happen as a result of changes in the state of the framework triggered by the incoming event
- we finish handling the original event, and if `Commands` have been submitted we dispatch these (as events; `Command` could also very reasonably be a separate method, but it isn't clear that this is a win)
- when commands are finished, we send `update` to every window in the application; this is an opportunity for widgets that did not see the original event to see the changes in the data, and invalidate/update state as needed. **Once update has been sent, widgets know that no more data changes are possible before the next layout & paint calls.**
- we send `layout`, and figure out where widgets are in the window
- we send `paint` and draw the widgets.

The problem with `AnimFrame` is that it does not arrive at druid-shell like other events. This might be the problem we need to solve.

Druid keeps a flag indicating whether or not it is currently animating. When you request an animation frame, we set this flag, and then ask `druid-shell` to paint us. The `paint` call returns a bool, indicating whether we're animating; if we return `true`, `druid-shell` schedules another paint call.

The thing that is weird about the `AnimFrame` event is that it is generated *while we are handling the paint call*. This means that if we mutate data during the handling of this event, that mutation isn't visible to other widgets before they get a chance to paint; it isn't visible until the next event arrives that triggers the normal event/update cycle.

This is why `AnimFrame` doesn't quite fit as an `Event`; because of how it is delivered, it cannot provide the same guarantees.


### A possible approach:

It would be *nice* if we could make `AnimFrame` an event. One possible approach to this would be to add a 'vsync' event; this might be a mode that you turn on and off, and when it is on it means that druid gets called back with an event each time the window can draw a frame, and we can handle this _before_ the paint call arrives. I'm not sure how feasible this is, but if we could do it it would really help clear all this up; animation would behave just like other events.

jneem:
Some of what you wrote I changed in #1057, which is what made this PR feasible. The current situation is that when the OS tells druid-shell that it's time to repaint, druid-shell calls into druid twice. The first time, it calls `WindowHandler::prepare_paint` (which is allowed to cause invalidations and request animation frames). When `prepare_paint` returns, `druid-shell` calls `paint`.

This PR delivers the `AnimFrame` during `prepare_paint`, meaning that widgets can still `request_paint` or `request_anim_frame`. After the `AnimFrame` is delivered, there's an `update` call also. So unless I've misunderstood something, it appears to be doing pretty much what your 'vsync' proposal says.

cmyr:
oh cool, yes, this does sound similar, and might be enough to make this change.

One other concern I would have is that often when the OS is telling us to paint, we're really not expected to be doing anything else, and we possibly  risk missing frames if we're doing a bunch of other stuff while handling that event; but I'm happy to just write that down and put the onus on the programmer to not be too silly.

cmyr:
And to clarify: really the only basis of my having moved `AnimFrame` out of `Event` was that it just didn't follow the same rules as other events. If that is changed, I have no complaints.

jneem:
Ok, cool. I will do some cleanup and testing, and then remove the draft status.

> One other concern I would have is that often when the OS is telling us to paint, we're really not expected to be doing anything else, and we possibly risk missing frames if we're doing a bunch of other stuff while handling that event; but I'm happy to just write that down and put the onus on the programmer to not be too silly.

This is a good point, and I've added a comment to `AnimFrame`. One related remark is that the X11 backend (which despite being half-finished in many respects is IIUC the only backend that explicitly knows about vsync timing) takes the most conservative approach to timing its paint: it starts painting as soon as the previous vsync is over. I think this means that it doesn't matter whether we do the expensive computation in the paint call or as a response to, for example, a different event.

jneem:
Great, thanks! @ForLoveOfCats are you close to landing #1107? If so, I can wait and rebase off that.

cmyr:
I'll take another pass at #1107 soon.

ForLoveOfCats:
This should no longer be blocked!

PR #1158, opened 2020-08-25
Implemented scaling for widgets

rhzk:
This adds scaling support to widgets

One can control the scaling a widget recieves with the Scale widget "Scale::new(value, child)".
The scale is then propagated down the tree unless overwritten again by a new Scale widget.

I have updated Button, CheckBox, Label, Padding, ProgressBar, RadioGroup, Scroll, SizedBox, Slider, Spinner, Stepper, Switch and TextBox to support scaling.

I did not update the image drawing widgets as I am unsure if they should be scaled, and in that case it needs to be done on the gpu-side?

I did think about a few different ways to implement this but I wanted to avoid any breaking change and I wanted the widgets to have control over how they scale, but if anyone thinks of a better way feel free to suggest it.

There is also a example showing dynamic and static scaling on most widgets.


jneem:
This is an interesting approach that I wouldn't have thought of! My main concern is that it requires explicit support for all downstream widgets, whereas the approach in one of @rjwittams 's branches (or PRs, I don't remember, sorry) doesn't. Can you comment on why you chose this approach instead of that one? Also, is there a reason that you decided to put it directly in the contexts instead of using `Env` and `EnvScope`?

rhzk:
I dont think @rjwittams has a scale implementation, atleast what I gathered from zulip.

I attempted to write a wrapper between widgets and piet, but it proved difficult since all the calls took a trait and I couldnt get access to the data directly. Maybe its possible, I am not that experienced with rust yet. I also thought that widgets might want to control how they scale, incase widgets have some elements that shouldnt scale, or maybe in one direction or maybe repeat instead of scaling or they have complex shapes.

I think I just followed up on something I tried earlier and didnt think about using EnvScope, I will update it to use EnvScope instead

rjwittams:
I did a Rotated widget as thats what I needed at the time, but it is likely be a degenerate form of a more general transform widget. But I haven't done that yet.
Basically transforms events, renderctxts and invalidation regions on the way in and out. I doubt its entirely robust, and likely would benefit from context support. 

I don't think it makes sense to do things in every widget, it smacks of major composability issues. I will take a look at the details when I can.

This could be another instance where 'not covering the ground' means wasted effort down strange avenues.... will see.

rjwittams:
Yeah, afraid to say I think this is probably not the right approach.

It ends up imposing a global requirement on every widget author to do some lookups and multiplies, so increases code noise. 

It is an unstable invariant - not enforced in the type system (and if it could be it would be onerous) so you'd just end up with a mishmash (some widgets support scaling and some don't). I also think it will end up inconsistent - eg fonts not scaled? - in Switch padding isn't, other things are, but how does anyone know what should be or shouldn't?  If you have 'system level' scaling, it is possible to test it at one level and know that there is a linear transformation to any other scale. If things scale in different ways, it is more like doing responsive CSS design - you need to test from a UX pov at every sane scale, which I'm not convinced we want to do at this level and really has to be way more involved in understanding the relationship between pixel sizes and human perception in different situations/on different devices.

https://github.com/rjwittams/druid/blob/binding-scroll/druid/src/widget/rotated.rs 

This is the Rotated widget I did. 
It is not a complete arbitrary transform widget, it was just enough for tab buttons, but you can see the idea.

derekdreery:
I think my preferred solution to this problem would be to have scaling being transparent to child widgets. This could be done, maybe, by having the layout algorithm do the opposite of the requested transform (so that if you need something that is 100px for example, and you are going to scale it by 0.5, then you need 200px rendering so that it ends up being 100px). Then the paint method would need to apply the corresponding affine transform to perform the scaling.

I see no reason why this approach would need to be restricted to scaling - you could equally perform rotation by multiples of 90 degrees. Probably only linear transforms (that preserve the origin) make sense, since translations would just end up looking like you've done nothing. Also, non-90 degree rotations would break the box model.

An alternative would be to not try to "do the opposite" in the layout algo, which would make all rotations make sense, but would make scaling less easy to use.

On a different note, this PR is valuable even if it doesn't end up getting merged. It helps to explore the design space and stimulate discussion as to the best approach.

rjwittams:
Might help to look at the Rotated linked above which literally does what you just mentioned. To reiterate it is for a single purpose but could be generalised. The layout there isn't transformed generally but you get the idea.

I don't know what you mean wrt scaling vs rotation. They are affine transforms, you can invert both in the same way.

General transforms that don't maintain rectilinear shapes eg skews and non n*pi/2 rotations leave extra space so you have to decide what to do with it.

derekdreery:
> General transforms that don't maintain rectilinear shapes eg skews and non n*pi/2 rotations leave extra space so you have to decide what to do with it.

Aah that's a good solution to the problem, just take the smallest bounding rectangle.

rhzk:
I would like a way for widgets to control this behavior, I will see if I can find a solution that does it automatically but gives a option for intercepting it and doing it manual.

Scaling with affine has a few problems, some of being that some widgets use a static size to handle some events so some widgets will most likely need to be updated in any case.

rjwittams:
No they won't. The events get transformed.

rhzk:
Does not look like it from my testing

rjwittams:
Testing of what?

rhzk:
For example a widget was originaly 100 pixels in width and I scaled it so its now 150 pixels, If I click on the end mouse.pos reports 150 pixels, but the widget uses the constant env.get(theme::WIDE_WIDGET_WIDTH), that reports 100 pixels in its mouse handling, updating the widget to use its own ctx.size() instead of this constant fixes the problem

rjwittams:
I am talking about what Rotated does.

raphlinus:
I'm also pretty skeptical of this patch, for reasons @rjwittams has expressed. It puts an additional burden on widgets and doesn't seem to compose smoothly. I also confess, I'm not seeing the motivation clearly. While scaling functionality *is* important (for example, scaling the content area in Runebender is essential), it's not obvious that scaling widgets is a good way to do this. What precedent do we have for this functionality in other GUI toolkits?

I could be convinced, but right now I'm leaning towards not going this direction.

rhzk:
I am working on a solution that hopefully would scale them without any work on the widget side, tho it is a bit tricky. But if you think we shouldnt scale widgets I dont mind closing this

raphlinus:
I'm not sure yet - if it can be done in a simple, clean way, it might be worth it. But if it exports complexity to the widget interface, then my threshold is a lot higher, and I'd want to consider evidence that the functionality is a close match for what applications actually need.

rjwittams:
Thats also one reason I didn't really push to generalise Rotated. The other realistic transforms all seem pretty niche.

rhzk:
In that case I will close this for now, I didnt like the idea of upscaling things behind the scenes as just upscaling everything could cause things to not look as pretty or right

PR #1160, opened 2020-08-26
Tabs widget and the Rotated widget that it uses.

rjwittams:
The tabs widget and dependencies extracted from binding-scroll. 

This does static and dynamic tabs, and also includes a Rotated widget that it uses to rotate tab labels.

Things that maybe should change: 
  - The way Rotated messes with the invalidation region is a bit rough and ready.
     I didn't want to come up with an elaborate scheme as the best way to do it needs discussion 
  - Axis probably should be promoted out of Flex
  - CrossAxisAlignment for tabs might be an over-re-use. But not sure.
  - Possibly the event methods in tabs (when events should be passed on to children) should move onto the Event enum. 
  - Possibly the event transform in Rotated should move there too.
 
 I left both where they were as its probably easier to review


rhzk:
Some things I noticed from the example:

If I click on a new tab I see the content of the previous tab I was on. I.e if I click on "Advanced" I see the content of "Basic" and If I then click on "Page 3" I see the content of "Advanced". But then the button does not work, I guess because I am on "Page 3".

I am not sure that the child widget size should be set to the entire tab space, now I cant have fixed width on a label, and if I add a background color to the label it fills the tab.

I am not a fan of downward text but if we where to have it I dont like to have each letter rotated, I think it would be better in that case for it to be like this:

P
a
g
e

3

otherwise it feels like I should tilt my head to the side to read it

rjwittams:
On rotation: this is fairly standard. E.g IntelliJ does it. I have literally never seen your recommendation outside of motel signs. If you 'don't like' the rotation, don't use it. 
I have not seen the behaviour you describe at all wrt tabs. Possibly this is a platform issue? I'll test it again but it sounds very odd.
I don't get why you wouldn't add your own spacing to the child, ie put it in a Flex or whatever. Having dead space invented by the tab widget seems strange to me.

rjwittams:
I can't recreate the behaviour you describe. I hope its fairly obvious I would have noticed that... is this on Windows? I wonder if its a platform event handling difference or something? Really doesn't make a lot of sense. Does the same happen with the dynamic tabs?

rhzk:
The issue is also with the dynamic tabs, I am on windows.

Sure you can put widgets inside a flex but I tought it was odd to overwrite the widgets behavior, ie a textbox filling the entire width of the tab instead of the width you'd expect it to have.

For the suggestion to have downwards tabs without the text being rotated, if you dont want to have that thats fine, but saying "if you dont like it dont use it" is not the wording I would use if somebody asked for a feature.

rjwittams:
You are asking to remove a feature. You literally don't have to use it - leave the tabs not rotated by setting the rotation to None.

I have no idea what you mean by overriding the widgets behaviour. It does exactly the same thing as e.g ViewSwitcher. 

If this is a Windows only issue it must be fairly fundamental.

rhzk:
What I ment was this, top one being a textbox in viewswitcher and bottom one in a tab

img

rjwittams:
I managed to get something like it to happen on GTK in a linux VM. It might be some difference in the way the animations work. 
I'll investigate tomorrow.

rjwittams:
I think its because I do .expand() on the tabs body (which does work like ViewSwitcher internally), this was I think also to make the animations look sensible (using the whole width of the tab). I'll see if there is a better way to do it...

rjwittams:
@rhzk I think both of those issues might be fixed now - no longer use expand, the tab body expands itself (which leaves children alone). 
The animation one was a bit odd, I had to add in another call to request_paint. I think it might be that the mac shell ends up painting anyway when animation frames happen and I was accidentally relying on that. So possibly a bug in the mac shell. 
I guess the lesson is to always try at least two platforms!
I haven't tested Windows as I don't have it. 
I will try out the web again at some point


I also noticed on Linux that the circle with an X on it symbol wasn't in the default font. This was pretty much a placeholder anyway. I may paint it directly or use images - but not sure we have a good answer on shipping around images in the druid library? Probably painting is better. Can do some kind of hover then too

rhzk:
Looks like its all working on Widows now!

rjwittams:
http://chil.rice.edu/research/pdf/byrne2002hfes.pdf <- On rotated text. There is always a creaky old paper to be wheeled out

druid/src/widget/rotated.rs
```diff
@@ -0,0 +1,111 @@
...
+        let mut me = me.clone();
+        me.pos = inverse * me.pos;
+        me
+    }
+}
+
+impl<T, W: Widget<T>> Widget<T> for Rotated<W> {
+    fn event(&mut self, ctx: &mut EventCtx, event: &Event, data: &mut T, env: &Env) {
+        if let Some((transform, inverse)) = self.transforms {
+            ctx.widget_state.invalid.transform_by(inverse);
```
jneem, line 51: Every context with a `request_paint` method can modify the invalid region, so this also needs to be in `update` and `lifecycle`. But I think it would actually be nicer (and maybe less fragile) to handle this in `WidgetState`. That is, replace `WidgetState::viewport_offset` by, say, `WidgetState::viewport_transform`,(which would be an `Affine`), and then handle the transformation of invalid regions during `WidgetState::merge_up`.

druid/src/widget/rotated.rs
```diff
@@ -0,0 +1,111 @@
...
+use std::f64::consts::PI;
+
+pub struct Rotated<W> {
+    child: W,
+    quarter_turns: u8,
+    transforms: Option<(Affine, Affine)>,
+}
+
+impl<W> Rotated<W> {
+    pub fn new(child: W, quarter_turns: u8) -> Self {
```
jneem, line 15: Doc comment here, please, including the fact that `quarter_turns` is clockwise.

druid/examples/widget_gallery.rs
```diff
@@ -23,7 +23,7 @@ use druid::{
     AppLauncher, Color, Data, Lens, Rect, Widget, WidgetExt, WidgetPod, WindowDesc,
 };
 
-#[cfg(not(target_arch = "wasm32"))]
+#[cfg(feature = "svg")]
```
jneem, line 26: This seems reasonable, but unrelated?

druid/src/lib.rs
```diff
@@ -149,6 +149,7 @@ mod data;
 mod env;
 mod event;
 mod ext_event;
+#[macro_use]
 pub mod lens;
```
jneem, line 153: I suggest making `lens` the first module listed, because the macros are only available for the modules listed after this. Which is fine for now, but it could lead to some head-scratching down the road when the macros don't work in another module.

druid/src/widget/flex.rs
```diff
@@ -195,8 +195,8 @@ pub struct FlexParams {
     alignment: Option<CrossAxisAlignment>,
 }
 
-#[derive(Clone, Copy)]
-pub(crate) enum Axis {
+#[derive(Data, Debug, Clone, Copy, PartialEq)]
+pub enum Axis {
```
jneem, line 199: Doc comments here, please, and also elsewhere in this file. Also, if this isn't just used for Flex, maybe it makes sense to put it elsewhere?

druid/src/widget/tabs.rs
```diff
@@ -0,0 +1,960 @@
...
+    theme, Affine, BoxConstraints, Data, Env, Event, EventCtx, Insets, LayoutCtx, Lens, LifeCycle,
+    LifeCycleCtx, PaintCtx, Point, Rect, SingleUse, Size, UpdateCtx, Widget, WidgetExt, WidgetPod,
+};
+use instant::Duration;
+
+type TabsScope<TP> = Scope<TabsScopePolicy<TP>, Box<dyn Widget<TabsState<TP>>>>;
+type TabBodyPod<TP> = WidgetPod<<TP as TabsPolicy>::Input, <TP as TabsPolicy>::BodyWidget>;
+type TabBarPod<TP> = WidgetPod<TabsState<TP>, Box<dyn Widget<TabsState<TP>>>>;
+type TabIndex = usize;
+type Nanos = u64; // TODO: Make Duration Data?
```
jneem, line 38: Sounds reasonable to me. Maybe `Instant` too, while we're at it.

druid/src/widget/tabs.rs
```diff
@@ -0,0 +1,960 @@
...
+    /// The common type for all body widgets in this set of tabs.
+    /// A flexible default is Box<dyn Widget<Self::Input>>
+    type BodyWidget: Widget<Self::Input>;
+
+    /// The common type for all label widgets in this set of tabs
+    /// Usually this would be Label<Self::Input>
+    type LabelWidget: Widget<Self::Input>;
+
+    /// This policy whilst it is being built.
+    /// This is only be useful for implementations supporting AddTab, such as StaticTabs.
```
jneem, line 68: "This is only useful"

druid/src/widget/tabs.rs
```diff
@@ -0,0 +1,960 @@
...
+    /// It can be filled in with () by other implementations until associated type defaults are stable
+    type Build;
+
+    /// Have the tabs changed. Expected to be cheap, eg pointer or numeric comparison.
```
jneem, line 72: Have the tabs changed?

druid/src/widget/tabs.rs
```diff
@@ -0,0 +1,960 @@
...
+    fn tabs_changed(&self, old_data: &Self::Input, data: &Self::Input) -> bool;
+
+    /// What are the current tabs set in order.
```
jneem, line 75: ?

druid/src/widget/tabs.rs
```diff
@@ -0,0 +1,960 @@
...
+    fn tabs(&self, data: &Self::Input) -> Vec<Self::Key>;
+
+    /// Presentation information for the tab
+    fn tab_info(&self, key: Self::Key, data: &Self::Input) -> TabInfo;
+
+    /// Body widget for the tab
```
jneem, line 81: Can you explain the meaning of a `None` return value in the doc comment?

druid/src/widget/tabs.rs
```diff
@@ -0,0 +1,960 @@
...
+
+    fn build(build: Self::Build) -> Self {
+        StaticTabs {
+            tabs: Rc::new(build),
+        }
+    }
+}
+
+pub trait AddTab: TabsPolicy {
+    fn add_tab(
```
jneem, line 168: docs?

druid/src/widget/tabs.rs
```diff
@@ -0,0 +1,960 @@
...
+        self.children.iter_mut().flat_map(|x| x.1.as_mut())
+    }
+}
+
+fn hidden_should_receive_event(evt: &Event) -> bool {
+    match evt {
+        Event::WindowConnected
+        | Event::WindowSize(_)
+        | Event::Timer(_)
+        | Event::Command(_)
```
jneem, line 515: Just a comment, that if I close the tabs example before clicking on the various tabs, I get a bunch of warnings like
```
WARN  [druid::core] Widget 'druid::widget::label::Label<tabs::Advanced>' received an event (Command(Command { symbol: "druid-builtin.close-window", payload: Any })) without having been laid out. This likely indicates a missed call to set_layout_rect.
```

My inclination is that everything is fine, and the warning is overzealous. But at some point it's probably worth thinking (and documenting) what should happen in general for hidden widgets.

druid/src/widget/tabs.rs
```diff
@@ -0,0 +1,960 @@
...
+                })
+            }
+        } else if let Some(ref mut child) = Self::child(&mut self.children, data.selected) {
+            child.paint_raw(ctx, &data.inner, env);
+        }
+    }
+}
+
+// This only needs to exist to be able to give a reasonable type to the TabScope
+pub struct TabsScopePolicy<TP> {
```
jneem, line 665: Does this need to be pub?

druid/src/widget/tabs.rs
```diff
@@ -0,0 +1,960 @@
...
+    pub fn new(name: impl Into<String>, child: impl Widget<T> + 'static) -> Self {
+        InitialTab {
+            name: name.into(),
+            child: SingleUse::new(Box::new(child)),
+        }
+    }
+}
+
+enum TabsContent<TP: TabsPolicy> {
+    Building {
```
jneem, line 758: Can you explain why you went this route instead of having a separate `TabsBuilder` struct? (I also didn't understand the need for a `Build` associated type on `TabsPolicy`, as opposed to using a separate builder.)

druid/src/widget/tabs.rs
```diff
@@ -0,0 +1,960 @@
...
+                // TODO: Type inference fails here because both sides of the lens are dependent on
+                // associated types of the policy. Needs changes to lens derivation to embed PhantomData of the (relevant?) type params)
+                // of the lensed types into the lens, to type inference has something to grab hold of
+                .lens::<TabsState<TP>, tabs_state_derived_lenses::inner>(TabsState::<TP>::inner)
+                .padding(Insets::uniform_xy(9., 5.));
+
+            if info.can_close {
+                let row = Flex::row()
+                    .with_child(label)
+                    .with_child(Label::new("ⓧ").on_click(
```
jneem, line 257: On my configuration (GTK, but unsure what font druid is using), this doesn't show up.

jneem (dismissed):
Very nice! My main complaint, as usual, is that it's missing API docs. I gave up on pointing them out one-by-one, but anyway I think that the CI is now configured to complain about them :)

druid/src/widget/flex.rs
rjwittams: I mentioned that in the PR description - and I agree... not sure where exactly it should go though.

druid/examples/widget_gallery.rs
rjwittams: I think it was angering IntelliJ
Can split out

druid/src/widget/rotated.rs
rjwittams: Yes - as I mentioned in the description,  I didn't want to do too much with this region stuff before the general idea was thought to make sense.

druid/src/widget/tabs.rs
rjwittams: Makes sense.

druid/src/widget/tabs.rs
rjwittams: Garbled comments from refactorings...

druid/src/widget/tabs.rs
rjwittams: I originally had a builder, but my impression was that the 'integrated builder' style was heavily favoured in druid.

druid/src/widget/tabs.rs
rjwittams: I noticed that too on Linux (when working out rzhks issues). I guess the options are to use images or paint it - probably worth doing the latter I suppose.

druid/src/widget/tabs.rs
rjwittams: This might be that I need more detailed info on the events, ie which ones should be delivered to widgets that haven't been seen.

druid/src/widget/tabs.rs
rjwittams: possibly not

druid/src/widget/tabs.rs
rjwittams: Left over comment from a method I got rid of

cmyr:
@rjwittams and I discussed this earlier today, and we agreed to split out the rotation work into a subsequent PR.

druid/src/widget/flex.rs
rjwittams: I still do not know where to put it. 
I think it should be dealt with in a subsequent PR by someone who does.

cmyr (commented):
Great, happy to see the ambition here. I have a bunch of questions and comments inline, but overall I think this is close to ready to get in?

Some additional unordered thoughts:

- I'd like to see some clarification around the various types whose names begin with `Tab`; this means some docs, but also some careful consideration around what needs to be public and what does not.

- I'm also seeing a bunch of warnings when I play around with the example, which should get looked at; either we're missing some event or the warnings are too aggressive.

- I haven't done a careful review of the layout and drawing code; if there are issues in there they'll shake out over time.

- I'm concerned about `String` in places; ideally we would be using `LabelText` for any text that is shown the user.

- we're still missing docs on public items; if these aren't showing up in CI it probably means the items aren't actually being exported, and shouldn't be `pub`.

- I think this is an interesting exploration of some new patterns, which I'm happy about. It's also nice that it is quite compartmentalized, and so we can always revisit in the future without much disruption.

@rjwittams if you'd like to go through the various comments and either reply or mark them as resolved that would be great, and then please ping me and switch the label back to 'ready-for-review' when you think you've addressed everything, and I'll take another pass. thanks!

CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ You can find its changes [documented below](#060---2020-06-01).
 - `Menu` commands can now choose a custom target. ([#1185] by [@finnerale])
 - `Movement::StartOfDocument`, `Movement::EndOfDocument`. ([#1092] by [@sysint64])
 - `TextLayout` type simplifies drawing text ([#1182] by [@cmyr])
+- 'Rotated' widget to allow widgets to be rotated by an integral number of quarter turns ([#1160] by [@rjwittams])
```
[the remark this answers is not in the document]
rjwittams: Yep. I seem to have left some bits I thought had been removed.

druid-shell/src/region.rs
```diff
@@ -81,6 +81,12 @@ impl Region {
         }
         self.rects.retain(|r| r.area() > 0.0)
     }
+
```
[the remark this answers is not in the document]
rjwittams: This was for rotated I think, I will check.

druid/examples/tabs.rs
```diff
@@ -0,0 +1,212 @@
+use druid::im::Vector;
+use druid::widget::{
+    Axis, Button, CrossAxisAlignment, Flex, Label, MainAxisAlignment, Padding, RadioGroup,
+    SizedBox, Split, TabInfo, Tabs, TabsPolicy, TabsTransition, TextBox, ViewSwitcher,
+};
+use druid::{theme, AppLauncher, Color, Data, Env, Lens, LensExt, Widget, WidgetExt, WindowDesc};
+use instant::Duration;
+
+#[derive(Data, Clone)]
+struct Basic {}
```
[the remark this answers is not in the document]
rjwittams: Agreed, I think this was something I removed a field from

druid/examples/tabs.rs
```diff
@@ -0,0 +1,212 @@
...
+    AppLauncher::with_window(main_window)
+        .use_simple_logger()
+        .launch(initial_state)
+        .expect("Failed to launch application");
+}
+
+fn build_root_widget() -> impl Widget<AppState> {
+    fn decor<T: Data>(label: Label<T>) -> SizedBox<T> {
+        label
+            .padding(5.)
```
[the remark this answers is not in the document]
rjwittams: I'll check that, not sure I knew about it when I wrote this

druid/src/widget/flex.rs
```diff
@@ -278,68 +355,31 @@ impl<T> ChildWidget<T> {
...
-    }
-}
-
 impl<T: Data> Flex<T> {
-    /// Create a new horizontal stack.
-    ///
-    /// The child widgets are laid out horizontally, from left to right.
-    pub fn row() -> Self {
+    /// Create a new Flex oriented along the provided axis.
+    pub fn for_axis(axis: Axis) -> Self {
```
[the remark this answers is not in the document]
rjwittams: It gets used in tabs as the direction is parameterized, and possibly table (but maybe not .. can check later). I think if it is used in that way internally it should be public, so out of crate widgets aren't disadvantaged.

druid/src/widget/tabs.rs
```diff
@@ -0,0 +1,970 @@
...
+type TabBodyPod<TP> = WidgetPod<<TP as TabsPolicy>::Input, <TP as TabsPolicy>::BodyWidget>;
+type TabBarPod<TP> = WidgetPod<TabsState<TP>, Box<dyn Widget<TabsState<TP>>>>;
+type TabIndex = usize;
+type Nanos = u64; // TODO: Make Duration Data?
+
+/// Information about a tab that may be used by the TabPolicy to
+/// drive the visual presentation and behaviour of its label
+pub struct TabInfo {
+    /// Name of the tab
+    pub name: String,
```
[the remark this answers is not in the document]
rjwittams: I'll try it, haven't thought about the dynamic one. Possibly works as expected without much trouble

druid/src/widget/tabs.rs
```diff
@@ -0,0 +1,970 @@
...
+    type Key: Hash + Eq + Clone;
+
+    /// The input data that will
+    /// a) be used to derive the tabs present
+    /// b) also be the input data for all of the child widgets.
+    type Input: Data;
+
+    /// The common type for all body widgets in this set of tabs.
+    /// A flexible default is Box<dyn Widget<Self::Input>>
+    type BodyWidget: Widget<Self::Input>;
```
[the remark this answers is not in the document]
rjwittams: I don't know. I had the feeling people in Rust land in general didn't like forced boxing and (for me) the cost doesn't seem very high. I may be more accustomed to associated types than some, but they do seem completely integrated into the rust standard library (iterators, ops, etc).

druid/src/widget/tabs.rs
```diff
@@ -0,0 +1,970 @@
...
+    pub fn new(inner: TP::Input, selected: usize, policy: TP) -> Self {
+        TabsState {
+            inner,
+            selected,
+            policy,
+        }
+    }
+}
+
+pub struct TabBar<TP: TabsPolicy> {
```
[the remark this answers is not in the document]
rjwittams: I'll check.

druid/src/widget/tabs.rs
rjwittams: I had a builder, knew that it would be complained about due to the stated preference for integrated builder methods - its been explicitly talked about as a design choice several times. 

So I used standard Rust features to avoid it and fit the stated design choices. 

And now that is a point of contention, but so far the comments don't seem to have any actual content, other than 'I didn't get this then I did'. As far as I'm aware it acheives its goals from an API pov, and doesn't corrupt the 'we don't do explicit builders for widgets' stated design choice. 

🤷‍♂️ I don't care either way other than it always feels a bit ... suboptimal ... to go back to a previous version of something. This bit of the api might go away with Crochet anyway.

druid/src/widget/tabs.rs
```diff
@@ -0,0 +1,970 @@
...
+///     .with_tab("Connection", Label::new("Connection information"))
+///     .with_tab("Proxy", Label::new("Proxy settings"))
+///     .lens(AppState::name);
+///
+///
+/// ```
+///
+pub struct Tabs<TP: TabsPolicy> {
+    axis: Axis,
+    cross: CrossAxisAlignment, // Not sure if this should have another enum. Middle means nothing here
```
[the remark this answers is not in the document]
rjwittams: Which end of the cross axis does your tab bar live on. Should be fairly easy to get from running the example.

CHANGELOG.md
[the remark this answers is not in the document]
rjwittams: Turns out I didn't push a cleanup commit a while back.

druid/examples/tabs.rs
```diff
@@ -0,0 +1,212 @@
...
+    Axis, Button, CrossAxisAlignment, Flex, Label, MainAxisAlignment, Padding, RadioGroup,
+    SizedBox, Split, TabInfo, Tabs, TabsPolicy, TabsTransition, TextBox, ViewSwitcher,
+};
+use druid::{theme, AppLauncher, Color, Data, Env, Lens, LensExt, Widget, WidgetExt, WindowDesc};
+use instant::Duration;
+
+#[derive(Data, Clone)]
+struct Basic {}
+
+#[derive(Data, Clone, Lens)]
```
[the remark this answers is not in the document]
rjwittams: I don't really know what you are looking for here

druid/src/widget/tabs.rs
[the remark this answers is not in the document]
rjwittams: It can't because an associated type is used in a member. 

tabs: Vec<(TP::Key, TabBarPod<TP>)>,

druid/src/widget/tabs.rs
```diff
@@ -0,0 +1,970 @@
...
+    axis: Axis,
+    cross_axis_alignment: CrossAxisAlignment,
+    tabs: Vec<(TP::Key, TabBarPod<TP>)>,
+    hot: Option<TabIndex>,
+    phantom_tp: PhantomData<TP>,
+}
+
+impl<TP: TabsPolicy> TabBar<TP> {
+    /// Create a new TabBar widget.
+    pub fn new(axis: Axis, cross_axis_alignment: CrossAxisAlignment) -> Self {
```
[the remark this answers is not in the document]
rjwittams: Yes, this is just the tab bar. I think it doesn't need to be public as I'm not sure in its current form its sensible to use on its own.

druid-shell/src/region.rs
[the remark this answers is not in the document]
rjwittams: > * I'm also seeing a bunch of warnings when I play around with the example, which should get looked at; either we're missing some event or the warnings are too aggressive.

I can see these now. I don't think they used to happen when I made it, so possibly affected by other changes. I'll see why that is..

druid/examples/tabs.rs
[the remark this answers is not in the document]
rjwittams: Trying it, this seems to increase the code noise of layout noise quite a bit, because factoring it out is harder (multiple calls onto a flex instead of a returned value ), meaning cut and paste is more likely.

Why is the spacer better than padding, or why does it matter? Also it isn't doing the same thing is it - padding is on all edges, and the spacer is only in the direction of the flex.

druid/src/widget/tabs.rs
rjwittams: I've changed this to layout all the children now, even if not active or animating, which I think resolves this.

druid/examples/tabs.rs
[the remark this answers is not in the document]
rjwittams: I have changed the example a bit. Not sure if it covers what you are talking about.

druid/src/widget/tabs.rs
rjwittams: I think I'd prefer to do this in a follow up as it might get a bit involved to make this button work well.

druid/src/widget/tabs.rs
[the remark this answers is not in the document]
cmyr: ah, gotcha. I might call it something like "tabs_location"? And I agree, `CrossAxisAlignment` doesn't feel like exactly the right enum for this, I might just shrug and add something like,

```rust
enum TabsEdge {
    Leading,
    Trailing,
}
```

🤷

druid/src/widget/tabs.rs
cmyr: my inclination would be to paint this, but follow-up is fine. maybe we do something gross for now, like just use 'X' as the text?

druid/src/widget/tabs.rs
cmyr: I think that's reasonable for now. I agree at some point we need to think about how to deal with hiding and unhiding widgets, but I don't mind if we're just doing a bit of extra work for the time being.

druid/src/widget/tabs.rs
cmyr: I'm not super opinionated here. I think it's nice to match the existing API patterns, but if something else made more sense then I'm happy to talk about that. I don't mind this as is. My previous comment should be interpreted as, "this is fine if you think it's best, but if you did it reluctantly because you wanted to match the existing API then I'm open to discussion an alternative design."

druid/src/widget/tabs.rs
```diff
@@ -0,0 +1,1006 @@
...
+
+    /// Lay out the tab bar along the provided axis.
+    pub fn with_axis(mut self, axis: Axis) -> Self {
+        self.axis = axis;
+        self
+    }
+
+    /// Put the tab bar at the corresponding end of the cross axis.
+    /// Defaults to Start. Note that Middle has the same effect as Start.
+    pub fn with_cross_axis_alignment(mut self, cross: CrossAxisAlignment) -> Self {
```
cmyr, line 872: echo my earlier comment I would prefer to call this something like `with_tab_bar_location` or whatever.

druid/src/widget/tabs.rs
```diff
@@ -0,0 +1,1006 @@
...
+    fn lifecycle(
+        &mut self,
+        ctx: &mut LifeCycleCtx,
+        event: &LifeCycle,
+        data: &TP::Input,
+        env: &Env,
+    ) {
+        if let LifeCycle::WidgetAdded = event {
+            let mut temp = TabsContent::Swapping;
+            std::mem::swap(&mut self.content, &mut temp);
```
cmyr, line 956: nit, but I think the idiomatic way to do this would be to use `std::mem::replace`, something like,

```rust
let content  = std::mem::replace(&mut self.content, TabsContent::Swapping);
```

cmyr (commented):
Okay, I think this is pretty much there. a few more comments are inline, but I don't think anything is blocking, except maybe for using something other than `CrossAxisAlignment`.

Feel free to address other stuff as you see fit; I'll take one more look when you're done and then I think we're good to go. Thanks!

druid/src/lib.rs
```diff
@@ -142,6 +142,9 @@ pub use druid_shell::{kurbo, piet};
 #[doc(inline)]
 pub use im;
 
+#[macro_use]
+pub mod lens;
```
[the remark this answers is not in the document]
rjwittams: This was from when I was using the lens! macro (before I had generic lens derivation?). You need this in order for macro_rules things to work in the same crate. 

I left it because it might be a headache for someone else to find out how to sort if they need to use that lens! macro

druid/src/widget/tabs.rs
rjwittams: Promise I'll do a follow up pretty quick...

druid/src/widget/tabs.rs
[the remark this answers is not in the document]
rjwittams: Yeah I think thats better

druid/src/widget/tabs.rs
rjwittams: Yep

druid/src/widget/tabs.rs
rjwittams: Yep, I think its another case where I didn't know about replace when I wrote this!

druid/src/widget/tabs.rs
rjwittams: I think its worth trying this one, if we find its not working and there isn't a better idea it can always be moved back. 
I expect people are more likely to copy examples to get started and the example here ends up looking reasonably straight forward.

rjwittams:
I think I have changed the essential bits.

cmyr (approved):
Okay this looks good to me, happy to get this in as-is and we can shake out any issues as they arise. Thank you for your patience!

PR #1161, opened 2020-08-27
Windows: pass the expanded region to the paint call

jneem:
Previously, we were expanding the invalid region to the nearest
pixel when invalidated, but passing the non-expanded region to
WinHandler::paint. This could cause graphical glitches.

### DIFF
--- a/druid-shell/examples/invalidate.rs
+++ b/druid-shell/examples/invalidate.rs
@@ -14,12 +14,12 @@
 
 use std::any::Any;
 
-use std::time::{Duration, Instant};
+use std::time::{Instant};
 
 use druid_shell::kurbo::{Point, Rect, Size};
 use druid_shell::piet::{Color, Piet, RenderContext};
 
-use druid_shell::{Application, Region, TimerToken, WinHandler, WindowBuilder, WindowHandle};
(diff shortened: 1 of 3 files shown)


jneem:
One of us has this backwards, and after double-checking the scale docs, I think it isn't me :wink: 

"dp" is "display points", which is our logical pixels. "px" is the device pixels. So it's doing the expanding in device pixels.

raphlinus:
> "dp" is "display points", which is our logical pixels. "px" is the device pixels. So it's doing the expanding in device pixels.

Oh god, I am confused, but I have a very good defense: we have a terminological nightmare on our hands. The use of the abbreviation "px" is, in CSS contexts, understood to mean [CSS pixel](https://developer.mozilla.org/en-US/docs/Glossary/CSS_pixel). This is also called "point" in macOS language, but that's a historical artifact from the time that Macintosh screens actually had 72dpi, because the word "point" has meaning in typographical circles as 1/72", and, on Windows generally has a value of 1.333 px.

The abbreviations in [Scalable](https://docs.rs/druid-shell/0.6.0/druid_shell/trait.Scalable.html) are backwards imho: "px" reads to me as CSS:px, and "dp" can read as "device pixel". I recommend we change them to something less confusing, or I think people will continue to be confused.

PR #1163, opened 2020-08-27
Add #[deny(missing_docs)] lint, add missing docs

cmyr:
I've been wanting this for a while, to ensure that PRs document their public API.

This is missing a few docs in `scope.rs`; @rjwittams is going to add those in a separate PR, since I don't know what's going on there. :)

### DIFF
--- a/druid/src/widget/common.rs
+++ b/druid/src/widget/common.rs
@@ -15,6 +15,7 @@
 use crate::{Affine, Size};
 
 // These are based on https://api.flutter.dev/flutter/painting/BoxFit-class.html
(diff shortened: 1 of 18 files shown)


druid/src/widget/common.rs
```diff
@@ -15,6 +15,7 @@
 use crate::{Affine, Size};
 
 // These are based on https://api.flutter.dev/flutter/painting/BoxFit-class.html
+/// Strategies for inscribing a rectange inside another rectange.
```
raphlinus, line 18: rectangle

PR #1168, opened 2020-08-30
Fix unresolved docstring reference in hotkey.rs Fix doc build

nvll:
The wrapper type for keyboard events was changed, but a docstring still referred to `keyboard_types` which is not able to be resolved from hotkeys.rs.

### DIFF
--- a/druid-shell/src/hotkey.rs
+++ b/druid-shell/src/hotkey.rs
@@ -109,7 +109,7 @@ impl HotKey {
 
     /// Returns `true` if this [`KeyboardEvent`] matches this `HotKey`.
     ///
-    /// [`KeyboardEvent`]: keyboard_types::KeyEvent
+    /// [`KeyboardEvent`]: KeyEvent
     pub fn matches(&self, event: impl Borrow<KeyEvent>) -> bool {
         // Should be a const but const bit_or doesn't work here.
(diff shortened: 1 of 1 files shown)


druid-shell/src/hotkey.rs
```diff
@@ -109,7 +109,7 @@ impl HotKey {
 
     /// Returns `true` if this [`KeyboardEvent`] matches this `HotKey`.
     ///
-    /// [`KeyboardEvent`]: keyboard_types::KeyEvent
+    /// [`KeyboardEvent`]: KeyEvent
```
raphlinus, line 112: This fixes the link, but the text should read "KeyEvent" as well. Good catch though!

druid-shell/src/hotkey.rs
```diff
@@ -109,7 +109,7 @@ impl HotKey {
 
     /// Returns `true` if this [`KeyboardEvent`] matches this `HotKey`.
     ///
-    /// [`KeyboardEvent`]: keyboard_types::KeyEvent
+    /// [`KeyEvent`]: KeyEvent
```
raphlinus, line 112: The text above as well, now the link doesn't match. And I'm very puzzled why this isn't being caught in CI, maybe we don't have our warnings tuned properly.

nvll: Good point, let me look at this locally. I'm not very good with github edits.

PR #1169, opened 2020-08-30
Correct typo in image.rs

nvll:
wideget -> widget


nvll:
> 
> 
> Yup! If I didn't know you personally I might suspect you're trying to game your github commit count for resume padding purposes :) :) :)

:) It's more a sign of the order I'm digging through the documentation. If you'd like I can consolidate them, and hold off for a bit? The broken compilation in the other issue felt more immediate.

PR #1170, opened 2020-08-31
Small fixes to docs / comments

nvll:
Sorry for the noise, I don't do pull request on GH a great deal and realized I could just maintain a branch at once*.

* At least unless the policy is auto-squash?


raphlinus:
Thanks!

PR #1171, opened 2020-09-01
Add Ref lens and use move semantics for lens! macro index.

### LINKED ISSUE #1138: lens::Index does not work on arrays
As per the title, when attempting to use a lens::Index on an array value I receive errors about Index and IndexMut not being implemented for the array. This is technically true, as they're only implemented on slices, not arrays. Arrays should coerce into slices, but something in all the generics is preventing that from happening. Using lens::Deref doesn't help here, as this is a real coercion and not a Deref impl. I ended up creating a lens::Field that simply indexes in its two closures, which feels unnecessary but works.

Finnerale:
Closes #1138.

See the linked issue for the reasoning behind this.

### DIFF
--- a/druid/src/lens/lens.rs
+++ b/druid/src/lens/lens.rs
@@ -125,6 +125,51 @@ pub trait LensExt<A: ?Sized, B: ?Sized>: Lens<A, B> {
         self.then(Deref)
     }
 
+    /// Invoke a type's `AsRef` and `AsMut` impl.
+    ///
+    /// It also allows indexing arrays with the [`index`] lens as shown in the example.
+    ///
+    /// # Examples
+    ///
+    /// Using `ref` this works:
+    ///
+    /// ```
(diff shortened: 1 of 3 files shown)


druid/src/lens/lens.rs
```diff
@@ -125,6 +125,51 @@ pub trait LensExt<A: ?Sized, B: ?Sized>: Lens<A, B> {
         self.then(Deref)
     }
 
+    /// Invoke a type's `AsRef` and `AsMut` impl.
+    ///
+    /// It also allows indexing arrays with the [`index`] lens as shown in the example.
```
jneem, line 130: maybe worth explaining that this is because `Index` is implemented for slices, and not arrays. Otherwise it's a little mysterious (to me, anyway) why reffing allowing indexing to work

Finnerale: Good idea! I've added a note.

cmyr:
this will need a rebase to pass CI.

PR #1172, opened 2020-09-02
Add MissingKeyError to Env

cmyr:
This is a little bit of cleanup while I was looking into making
another addition to the env. This is technically a breaking change
(we return Result instead of Option) but these API are not heavily
used and I'm not too concerned about the breakage.

### DIFF
--- a/druid/src/env.rs
+++ b/druid/src/env.rs
@@ -156,6 +156,16 @@ pub struct ValueTypeError {
     found: Value,
 }
 
+/// An error type for when a key is missing from the [`Env`].
+///
+/// [`Env`]: struct.Env.html
+#[derive(Debug, Clone)]
+#[non_exhaustive]
+pub struct MissingKeyError {
+    /// The raw key.
+    key: Arc<str>,
+}
+
 impl Env {
     /// State for whether or not to paint colorful rectangles for layout
     /// debugging.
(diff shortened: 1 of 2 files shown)


druid/src/env.rs
```diff
@@ -236,22 +252,23 @@ impl Env {
...
-            panic!("key for {} not found", key.key)
+        match self.try_get_untyped(key) {
+            Ok(val) => val,
+            Err(err) => panic!("{}", err),
         }
     }
 
     /// Gets a value from the environment, in its encapsulated [`Value`] form,
-    /// returning None if a value isn't found.
+    /// returning `None`` if a value isn't found.
```
raphlinus, line 262: Small markdown typo here.

druid/src/env.rs
```diff
@@ -205,24 +215,30 @@ impl Env {
...
+    pub fn try_get<'a, V: ValueType<'a>>(
+        &'a self,
+        key: impl Borrow<Key<V>>,
+    ) -> Result<V, MissingKeyError> {
         self.0
             .map
             .get(key.borrow().key)
             .map(|value| value.to_inner_unchecked())
+            .ok_or(MissingKeyError {
+                key: key.borrow().key,
```
raphlinus, line 240: I'm having trouble seeing how this can work if the key is not in fact `&'static str`, as it seems `MissingKeyError` will constrain the lifetime. But if CI accepts it, I'm not going to complain.

cmyr: Good point, I think I'm counting on the fact that keys are currently always static, but that we'd like non-static ones to exist. I should have a test case for that...

PR #1174, opened 2020-09-02
Update simple_logger, fix deprecation warnings

cmyr:
This was causing  our CI to fail, since  we deny all warnings.


raphlinus (approved):
I'm happy if CI is :)

PR #1178, opened 2020-09-03
Add FontDescriptor type

cmyr:
This adds a new type for describing fonts; it aggregates family, weight,  style, and size, and will be the default way of representing fonts in the environment.

This isn't used in this patch, but it will be in some follow-up work shortly.


raphlinus (approved):
This seems reasonable, and convenient. Lemme review the other one also, which I'm sure is more potentially interesting.

PR #1182, opened 2020-09-04
New text integration

cmyr:
This is a first pass at bringing the new piet stuff into druid.

Mostly this involves a new `TextLayout` type in druid, that is intended to be used as a component of other widgets that display text. This is intended to handle the bookkeeping around invalidating and recomputing layouts, resolving attributes  from the environment, and retaining other layout attributes.

This first patch is intended mostly to get us to parity with the existing code; it doesn't expose things like wrap width or alignment, yet.

**retaining the text**

~I chose not to have this object own a copy of the text, because that is already retained in the piet layout, and also already exists in whatever form it was passed to the widget in. This is annoying! It would be much nicer if you could pass in the initial text at construction time.~

~This may change if we settle on some data type(s) for strings that are clone friendly, which I think we should do but which is out of scope for now.~

This now retains the text, which is shared between this object and piet as an `Arc<str>`.

### DIFF
--- a/druid/src/text/layout.rs
+++ b/druid/src/text/layout.rs
@@ -1,0 +1,267 @@
+// Copyright 2020 The xi-editor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! A type for laying out, drawing, and interacting with text.
+
+use std::ops::Range;
+
+use crate::kurbo::{Line, Point, Rect, Size};
+use crate::piet::{
+    Color, PietText, PietTextLayout, Text as _, TextAttribute, TextLayout as _,
+    TextLayoutBuilder as _,
+};
+use crate::{ArcStr, Data, Env, FontDescriptor, KeyOrValue, PaintCtx, RenderContext};
+
+/// A component for displaying text on screen.
+///
+/// This is a type intended to be used by other widgets that display text.
+/// It allows for the text itself as well as font and other styling information
+/// to be set and modified. It wraps an inner layout object, and handles
+/// invalidating and rebuilding it as required.
+///
+/// This object is not valid until the [`rebuild_if_needed`] method has been
+/// called. Additionally, this method must be called anytime the text or
+/// other properties have changed, or if any  items in the [`Env`] that are
+/// referenced in this layout change. In general, you should just call this
+/// method as part of your widget's `update` method.
+///
+/// [`rebuild_if_needed`]: #method.rebuild_if_needed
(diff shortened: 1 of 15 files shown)


druid/src/text/layout.rs
```diff
@@ -0,0 +1,257 @@
...
+    /// This overrides the size in the [`FontDescriptor`] provided to [`set_font`].
+    ///
+    /// [`set_font`]: #method.set_font.html
+    /// [`FontDescriptor`]: struct.FontDescriptor.html
+    pub fn set_text_size(&mut self, size: impl Into<KeyOrValue<f64>>) {
+        self.text_size_override = Some(size.into());
+        self.layout = None;
+    }
+
+    /// The size of the laid-out text.
```
raphlinus, line 121: Maybe should mention that it's only meaningful after `rebuild_if_needed` has been called?

druid/src/text/layout.rs
```diff
@@ -0,0 +1,257 @@
...
+///
+/// This wraps an inner layout object, and handles invalidating and rebuilding
+/// it as required.
+///
+/// This object is not valid until the [`rebuild_if_needed`] method has been
+/// called. Additionally, this method must be called anytime the text or
+/// other properties have changed, or if any  items in the [`Env`] that are
+/// referenced in this layout change.
+///
+/// In general, you should just call this method as part of your widget's
```
raphlinus, line 36: "this method" should probably be "the `rebuild_if_needed` method"? Seems like maybe class and method got mixed up in cut'n'paste?

druid/src/text/layout.rs
```diff
@@ -0,0 +1,257 @@
...
+    /// will check to see if any used environment items have changed,
+    /// and invalidate itself as needed.
+    ///
+    /// Returns `true` if an item has changed, indicating that the text object
+    /// needs layout.
+    ///
+    /// # Note
+    ///
+    /// After calling this method, the layout may be invalid until the next call
+    /// to [`rebuild_layout_if_needed`], [`layout`], or [`paint`].
```
raphlinus, line 190: "rebuild_if_needed"?

druid/src/text/layout.rs
```diff
@@ -0,0 +1,257 @@
...
+                    .default_attribute(TextAttribute::ForegroundColor(text_color))
+                    .build()
+                    .unwrap(),
+            )
+        }
+    }
+
+    ///  Draw the layout at the provided `Point`.
+    ///
+    ///  You must call [`rebuild_if_needed`] at some point before you first
```
raphlinus, line 232: This is probably a good place to mention that the point references the top left of the text layout area.

druid/src/widget/checkbox.rs
```diff
@@ -25,6 +25,7 @@ use crate::{
 
 /// A checkbox that toggles a `bool`.
 pub struct Checkbox {
+    //FIXME: this should be a TextUi struct
```
raphlinus, line 28: Is this called "TextUi" or "TextLayout"?

druid/src/widget/radio.rs
```diff
@@ -43,6 +43,7 @@ impl RadioGroup {
 /// A single radio button
 pub struct Radio<T> {
     variant: T,
+    //FIXME: this should be using a TextUi struct
```
raphlinus, line 46: Same name issue.

raphlinus (approved):
This looks good. Most of what I found was comment issues. Looking forward to see it all land!

PR #1185, opened 2020-09-06
Move Target inside Command.

Finnerale:
This moves the `Target` into `Command`.

Currently, widgets receiving commands can't tell what the commands target actually was, and internally we have a similar problem, because we have to manually work with `(Target, Command)` tuples when ever we need access to the target.

Now `Command::target` can be used to access it at any time.

@cmyr If I remember correctly, you proposed this? I think it will work well.


cmyr (approved):
Honestly, I like this a lot; in particular I like how it more explicitly communicates how commands work, and brings the concept of a target to the foreground.

I have some sort of back-of-mind concerns about whether this reduces flexibility in some ways, like having a single command that a container sends to all of its children (what is the target?) but I don't think this is a serious problem in practice.

Finnerale:
Glad to hear. Once the checks passed, I'll merge it.

> I have some sort of back-of-mind concerns about whether this reduces flexibility in some ways, like having a single command that a container sends to all of its children (what is the target?) but I don't think this is a serious problem in practice.

I don't think that will be an issue. You can still change a command's target at any time, if you want to.

PR #1190, opened 2020-09-07
Add set_focus to EventCtx.

smmalis37:
A simple method that allows transferring focus to any arbitrary WidgetId, instead of just next and previous. I know focus is being reworked somewhat, but given how small this is I'd love if it could be added.

Open question: What would happen if a WidgetId is passed in whose widget has not registered_for_focus? Should documentation be added about this case? Should a check and warning/panic be added?

### DIFF
--- a/druid/src/contexts.rs
+++ b/druid/src/contexts.rs
@@ -449,6 +449,21 @@ impl EventCtx<'_, '_> {
         self.widget_state.request_focus = Some(FocusChange::Focus(id));
     }
 
+    /// Transfer focus to the widget with the given `WidgetId`.
+    ///
+    /// This should only be called by a widget that currently has focus.
+    ///
+    /// See [`is_focused`] for more information about focus.
+    ///
+    /// [`is_focused`]: struct.EventCtx.html#method.is_focused
+    pub fn set_focus(&mut self, target: WidgetId) {
+        if self.is_focused() {
+            self.widget_state.request_focus = Some(FocusChange::Focus(target));
+        } else {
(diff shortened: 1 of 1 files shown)


druid/src/contexts.rs
```diff
@@ -449,6 +449,21 @@ impl EventCtx<'_, '_> {
...
 
+    /// Transfer focus to the widget with the given `WidgetId`.
+    ///
+    /// This should only be called by a widget that currently has focus.
+    ///
+    /// See [`is_focused`] for more information about focus.
+    ///
+    /// [`is_focused`]: struct.EventCtx.html#method.is_focused
+    pub fn set_focus(&mut self, target: WidgetId) {
+        if self.is_focused() {
```
raphlinus, line 460: I'm not sure this is really needed. It is relevant for "next" style navigation, but I can imagine some action that's not necessarily from a focused widget causing focus to be directed to, eg, a textbox.

raphlinus (commented):
This makes sense to me. Anyone else working on focus want to chime in? (it's possible there's something I'm missing)

druid/src/contexts.rs
smmalis37: Good point.

cmyr:
`register_for_focus` (which may get removed) isn't about whether or not a widget *can* be focused; it's about whether or not it is participating in the 'focus chain', that is with tabbing around.

cmyr (approved):
This seems like a good addition, thanks!

PR #1191, opened 2020-09-07
Use a simpler approach to Direct2D presentation

### LINKED ISSUE #1144: widget_gallery example crash on Windows [scroll, resize]
On Windows (works well on Linux)

`error: process didn't exit successfully: '...\druid-master\target\debug\examples\widget_gallery.exe' (exit code: 0xc000041d)`

To reproduce the issue, just scroll down to the end then try to resize the window. I reproduced the error on two different computer.

raphlinus:
Instead of the convoluted approach of switching between Hwnd and Dxgi flip-mode presentation swapchains, just always use a Dxgi swapchain in sequential mode. This is compatible all the way back to Windows 7 with platform update (verified on reference machine), and is free of resize artifacts. In addition, it gets rid of the crashes when resources created with one D2D render target are used with another; all render targets are created with the same factory. It should also clear up some flashing artifacts on resize that I was observing on AMD graphics cards.

Note: I plan to get this in after #1037 (willing to fix expected merge conflicts myself), but it should be mergeable now.

Fixes #1144


cmyr (approved):
Thanks, always happy to delete code.

PR #1194, opened 2020-09-08
Fix breakage caused by a piet update.

jneem:


druid/src/widget/textbox.rs
```diff
@@ -104,7 +104,7 @@ impl TextBox {
             .unwrap_or(FontFamily::SYSTEM_UI);
 
         piet_text
-            .new_text_layout(&text.to_string())
+            .new_text_layout(text.to_string())
```
cmyr, line 107:
```suggestion
            .new_text_layout(text)
```

cmyr (approved):
Ah yes I understand, there's a new pre-release of piet and it's getting used here because it's semver compatible with the previous pre-release.

Thanks!

druid/src/widget/textbox.rs
cmyr: going to ignore my own suggestion so we don't need to run CI again, this will get changed in #1182 anyway.

Issue #1192, opened 2020-09-08
Text integration tracking issue


cmyr:
A tracking issue, largely for my own benefit, detailing the various bits of work involved in incorporating the new piet text stuff into druid.

- [ ] a basic `TextLayout` component that can be used to draw simple text.
- [ ] expose max width and line breaking
- [ ] expose text alignment
- [ ] support in layout for indicating baseline positions, add baseline option to flex container
- [ ] Add a rich-text type based on xi-rope
- [ ] improve documentation (maybe a book chapter?) covering various text API use cases

sysint64:
Probably known issue, but just in case - there are a lot of `TextAlignment not supported by cairo toy text` messages.

In my realtime system, the time needed to render a frame dropped from 1ms to 188ms because of that :(

sysint64:
Oh well, that's funny, I have steady 6ms-10ms if I run from the terminal, but 188ms if I run from emacs.

PR #1195, opened 2020-09-08
Expose word wrap in TextLayout and Label

cmyr:
With this patch, the TextLayout object can now be given a width
for wrapping words. This is exposed in the Label via a new
LineBreaking enum, with three options. With LineBreaking::WordWrap,
the TextLayout's wrap-width is set to the maximum width of the label's
BoxConstraints, and lines are broken appropriately. The other two
options (LineBreaking::Overflow and LineBreaking::Clip) both disable
line-breaking; in the former case if text exceeds the width of
the label it is painted outside of the label's bounds, and in the
latter case the text is clipped to the bounds of the label.

It would be nice if, in the clipping case, we could use a gradient
mask or something to fade the edges of the clipped text, but I don't
believe this is currently possible in piet. A simple alternative
of drawing a gradient from (transparent -> label background color)
doesn't work, because label's do not have an explicit background color,
which I think is also correct.

progress on #1192
Screen Shot 2020-09-10 at 12 08 45 PM
Screen Shot 2020-09-10 at 12 08 50 PM
Screen Shot 2020-09-10 at 12 10 17 PM
Screen Shot 2020-09-10 at 12 10 23 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ You can find its changes [documented below](#060---2020-06-01).
 - `Movement::StartOfDocument`, `Movement::EndOfDocument`. ([#1092] by [@sysint64])
 - `TextLayout` type simplifies drawing text ([#1182] by [@cmyr])
 - Implementation of `Data` trait for `i128` and `u128` primitive data types. ([#1214] by [@koutoftimer])
+- `LineBreaking` enum allows configuration of label line-breaking ([#1195] by [@cmyr])
 
 ### Changed
 
@@ -424,9 +425,10 @@ Last release without a changelog :(
 [#1171]: https://github.com/linebender/druid/pull/1171
 [#1172]: https://github.com/linebender/druid/pull/1172
 [#1173]: https://github.com/linebender/druid/pull/1173
-[#1182]: https://github.com/linebender/druid/pull/1185
+[#1182]: https://github.com/linebender/druid/pull/1182
 [#1185]: https://github.com/linebender/druid/pull/1185
 [#1092]: https://github.com/linebender/druid/pull/1092
+[#1195]: https://github.com/linebender/druid/pull/1195
 [#1204]: https://github.com/linebender/druid/pull/1204
 [#1205]: https://github.com/linebender/druid/pull/1205
 [#1214]: https://github.com/linebender/druid/pull/1214
--- a/druid/examples/text.rs
+++ b/druid/examples/text.rs
@@ -1,0 +1,101 @@
+// Copyright 2020 The Druid Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! An example of various text layout features.
+
+use druid::widget::{Controller, Flex, Label, LineBreaking, RadioGroup, Scroll};
+use druid::{
+    AppLauncher, Color, Data, Env, Lens, LocalizedString, UpdateCtx, Widget, WidgetExt, WindowDesc,
+};
+
+const WINDOW_TITLE: LocalizedString<AppState> = LocalizedString::new("Text Options");
+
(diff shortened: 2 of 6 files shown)


rjwittams:
This looks really good!

CHANGELOG.md
```diff
@@ -418,6 +419,7 @@ Last release without a changelog :(
 [#1171]: https://github.com/linebender/druid/pull/1171
 [#1172]: https://github.com/linebender/druid/pull/1172
 [#1173]: https://github.com/linebender/druid/pull/1173
+[#1182]: https://github.com/linebender/druid/pull/1185
```
raphlinus: link doesn't match
```suggestion
[#1182]: https://github.com/linebender/druid/pull/1182
```

druid/examples/custom_widget.rs
```diff
@@ -81,21 +84,18 @@ impl Widget<String> for CustomWidget {
...
+        layout.set_font(FontDescriptor::new(FontFamily::SERIF).with_size(24.0));
+        layout.set_text_color(fill_color);
+        layout.rebuild_if_needed(&mut ctx.text(), env);
 
         // Let's rotate our text slightly. First we save our current (default) context:
         ctx.with_save(|ctx| {
             // Now we can rotate the context (or set a clip path, for instance):
             ctx.transform(Affine::rotate(0.1));
-            ctx.draw_text(&layout, (80.0, 40.0));
+            layout.draw(ctx, (80.0, 40.0));
```
raphlinus: Not going to suggest it for this PR, but seeing this makes me wonder whether we might add some ext methods on our `PaintCtx` to make this a little smoother.

druid/examples/styled_text.rs
```diff
@@ -131,4 +133,5 @@ fn ui_builder() -> impl Widget<AppData> {
         .with_child(mono_checkbox)
         .with_spacer(8.0)
         .with_child(input.padding(5.0))
+        .debug_widget_id()
```
raphlinus: I'm assuming this is intentional to leave in? Flagging in case it's left over from development.

druid/examples/text.rs
```diff
@@ -0,0 +1,96 @@
+// Copyright 2019 The Druid Authors.
```
raphlinus, line 1: Following Google practice: current year for new files.

```suggestion
// Copyright 2020 The Druid Authors.
```

druid/examples/text.rs
```diff
@@ -0,0 +1,96 @@
...
+//! An example of various text layout features.
+
+use druid::widget::{Controller, Flex, Label, Scroll, Slider};
+use druid::{
+    AppLauncher, Color, Data, Env, Lens, LocalizedString, UpdateCtx, Widget, WidgetExt, WindowDesc,
+};
+
+const WINDOW_TITLE: LocalizedString<AppState> = LocalizedString::new("Text Examples");
+
+const TEXT: &str = r#"The idea of "structurelessness," however, has moved from a healthy counter to those tendencies to becoming a goddess in its own right. The idea is as little examined as the term is much used, but it has become an intrinsic and unquestioned part of women's liberation ideology. For the early development of the movement this did not much matter. It early defined its main goal, and its main method, as consciousness-raising, and the "structureless" rap group was an excellent means to this end. The looseness and informality of it encouraged participation in discussion, and its often supportive atmosphere elicited personal insight. If nothing more concrete than personal insight ever resulted from these groups, that did not much matter, because their purpose did not really extend beyond this."#;
```
raphlinus, line 24: Good choice of sample text, but I'm not sure it's the best possible. Maybe we should have a Zulip discussion or file an issue where we can drive consensus on possibly a better excerpt to include?

druid/src/widget/label.rs
```diff
@@ -141,6 +141,15 @@ impl<T: Data> Label<T> {
         self
     }
 
+    /// Builder-style method to set the wrap width.
```
raphlinus, line 144: So I think this is touching on a much deeper topic. With this, the word wrap is set explicitly and the widget layout follows, but if you look at something like Flutter, it's really the other way around; the text layout is wrapped to conform to whatever the widget layout resolved to. There's other moving parts to it, including the intrinsic widths etc., but I think if we are going to do "word wrap in a label" then maybe now is the time to hit those expectations.

raphlinus (commented):
See inline, I think this PR sets off a much deeper discussion than was planned...

cmyr:
I'm going to rework this to have a different API.

cmyr:
Okay, I've reworked this significantly so that instead of exposing an explicit wrap-width in label, we expose a `LineBreaking` enum that manages the wrap-width of the internal `TextLayout` object as appropriate. The PR description and screenshots have been updated appropriately.

ForLoveOfCats:
That looks glorious

raphlinus (approved):
This looks good, thanks!

PR #1196, opened 2020-09-08
Reinstate the Scroll::scroll function.

jneem:
It seems like the `Scroll::scroll` function was removed in #1107, which I'm hoping was unintentional because I was using it. This PR reinstates it.


ForLoveOfCats (approved):
That was indeed unintentional! It was moved into `ScrollComponent` instead of simply being duplicated. Sorry about that! :sweat_smile:

PR #1197, opened 2020-09-09
Set the Command's target in context.

jneem:
After recently updating druids, many of my commands stopped working. This fixed them, and according to the documentation for `submit_command` I think it has a chance of being correct.


Finnerale (approved):
Thanks, I knew I would mess it up somewhere...

PR #1198, opened 2020-09-09
Don't make Scroll bigger than its child on the non-scrolling axis.

jneem:
This undoes another change from #1107 that I think was a regression.

For example, if I have a horizontal scroll whose child has a finite height, I think that the scroll should inherit that same height. After #1107, the scroll would instead make its height as large as possible.


ForLoveOfCats (commented):
Yep, this seems right and matches how the old Scroll calculated its size. If it needs to expand that's what SizeBox is for.

In my defense that patch went through a lot of iterations. I'm glad these things which fell through the cracks are being caught before 0.7 drops :)

ForLoveOfCats (approved):

Issue #1167, opened 2020-08-29
LifeCycle::EnvChanged


cmyr:
This was mentioned by someone (@rjwittams?) during a previous discussion but I'd like to open a tracking issue; I think that we need an `EnvChanged` lifecycle event, so that widgets that use the env can invalidate themselves appropriately, and I would also like some mechanism for determining which env keys have changed, so that invalidation can be done only as necessary.

At some point we may want to investigate *diffing* the `Env`, but I think that is future work; what is important is that a widget is able to determine, when the `Env` changes, whether any of the changes involve keys that widget is interested in. The trivial version of this is just to pass the old and new `Env`s, and have the widget just directly check the keys it uses, like:

```rust
fn needs_to_rebuild_layout(old: &Env, new: &Env) -> bool {
    !old.get(KEY_I_USE).same(new.get(KEY_I_USE)) 
        || !old.get(OTHER_KEY_I_USE).same(new.get(OTHER_KEY_I_USE))
}
```

This is a relatively straight-forward project that involves some familiarity with druid internals, but should not pose serious problems; I would start with just adding a `LifeCycle::EnvChanged` event that is dispatched as part of `update` in `WidgetPod`. This should not recurse; it should be sent directly by each pod by comparing the new `Env` in update with the stored `Env`.

raphlinus:
This seems reasonable to me. Are you thinking that environment changes would supplant `update`, so that the latter would only be called in the presence of data changes?

Re diffing (which I agree should be a future investigation), the im crate provides an "efficiently diffable map," and I think it might make sense to adopt something like that for Env storage. The code fragment above doesn't really show the value in such a thing, as a query to find out if a single key has changed should be cheap no matter what the data structure.

cmyr:
The problem with update is that it only provides the new `Env`, so widgets can't tell what has changed. An alternative to this proposal would be to pass both old and new `Env` in `update`; I'm not necessarily opposed to this.

Another option (which I think has some problems, but might as well mention) is that the `Env` itself could get the functionality to be store a list of changed keys, and this would be available in `update`.

raphlinus:
[...]

PR #1201, opened 2020-09-09
Rebuild label more consistently

cmyr:
This improves the logic around when we rebuild a label's text layout;
in particular it fixes a potential bug where we would not rebuild the
layout if some property (text, font, color) was changed, but the
update method wasn't called.

### DIFF
--- a/druid/src/widget/label.rs
+++ b/druid/src/widget/label.rs
@@ -14,7 +14,7 @@
 
 //! A label widget.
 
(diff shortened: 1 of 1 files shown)


jneem (approved):
Looks good to me! I'm guessing the FIXME will be dealt with after #1167...

cmyr:
that's the plan, actually trying to get a patch together for that now. Thanks for the review!

PR #1202, opened 2020-09-09
Add debug_assert in TextLayout::draw

cmyr:
This will hopefully help diagnose issues around not rebuilding
layout objects appropriately.

It's arguable we should also panic in release; I don't have strong feelings in either direction. 🤷


raphlinus (approved):
I think a debug assert here is fine for now.

PR #1203, opened 2020-09-09
Remove old Text typealias from druid-shell

cmyr:
This was a vestige from long ago? In any case it was confusing since
it was exactly the same as the PietText type alias in piet_common.


raphlinus (approved):
Agree, this is a significant cleanup.

Issue #1159, opened 2020-08-26
Implement close request interception


cmyr:
#1118 added the ability to intercept close requests, but this is currently only working for gtk; other backends should support this as well.

- [x] mac
- [x] windows
- [x] x11

tay64:
Hello,
I've created a PR that adds the Windows part ([PR1238](https://github.com/linebender/druid/pull/1238)). This is the first time I'm trying to contribute here, hope I do not commit any faux pas in the process.

rjwittams:
https://github.com/linebender/druid/pull/1254 ends up fixing this on mac

PR #1204, opened 2020-09-09
Implement the close hook for x11

psychon:
Commit 7b6f83e8c48 added a close request hook and implemented it. This
commit implements this request for the x11 platform. Either I did
something wrong, or it is way easier to do this on x11 than on Gtk.

CC @jneem

Edit: Related to #1159


jneem (approved):
Thanks!

Yeah, in GTK you need to do some extra bookkeeping to figure out when to inhibit the default behavior. Sometimes it's nice to just not have any default behavior to deal with :)

PR #1205, opened 2020-09-09
Pass shared PietText by ref instead of cloning

cmyr:
Currently we create a copy of the PietText object every time we
need one. This changes that so that there is a single instance
in the ContextState, and the various contexts just pass out
a reference to this object as needed.

This should have few practical implications, except that in
trees with lots of text we will save a few microseconds. It's
mostly an API improvement, since the methods that tend to use
a PietText object generally want it passed by ref anyway.


raphlinus (approved):
Yes, looks good. When you say a few microseconds, did you measure that?

In theory, switching to &mut might break something, but in practice I think it's a cleaner type. It's causing flashbacks to the early days of xi-win, where it was a bit of a struggle to get the lifetimes right when rebuilding resources (tons of cloning of these factories), but I think generally the current state is right.

cmyr:
no, microseconds was just me mentally multiplying the cost of cloning an arc by a few hundred text objects.

raphlinus:
Yeah, the new changes reflect breakage I would have expected, but it doesn't seem bad at all to fix them. +1

PR #1206, opened 2020-09-09
Make sure the default Target is Global when calling process_ext_events

MTRNord:
`ExtEventSink::submit_command` was before when using with `Target::Auto` not calling the Widget Controllers. This behaviour was before implementing `Target::*` expected to happen.

(As discussed in zulip. This change was suggested by @cmyr )

### DIFF
--- a/druid/src/win_handler.rs
+++ b/druid/src/win_handler.rs
@@ -525,7 +525,7 @@ impl<T: Data> AppState<T> {
         loop {
             let ext_cmd = self.inner.borrow_mut().ext_event_host.recv();
             match ext_cmd {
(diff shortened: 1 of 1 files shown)


cmyr (approved):
I think this is right? Happy to merge for now, and @Finnerale can chime in if there's a better solution.

jneem:
I think this change makes sense, but it is in conflict with the documentation on `ExtEventSink`, so that should be updated.

cmyr:
Now I'm trying to remember what the rationale is/was here. Can this not be global for some reason, like that we're using the window's idle handler to notify of the event?

MTRNord:
Closed in favor of #1215

PR #1207, opened 2020-09-09
Add UpdateCtx::env_changed and UpdateCtx::env_key_changed

cmyr:
This is my attempt to address #1167.

I'm took one of the alternative approaches that came up in that discussion, which is to just include the previous `Env` in the `UpdateCtx`, and expose methods there for checking for changes. This goes slightly further and also includes an extra reference to the _current_ `Env` in `UpdateCtx`; this is just for ergonomics, so that you can check for changes without needing to pass the current `Env` explicitly.

This has one other 'just for ergonomics' addition; I've added a trait, `EnvResolveable`, that encapsulates the idea of a type that can check if it has changed between two versions of an `Env`. The rationale here is that it is implemented both just for `Key` as well as for `KeyOrValue`; I also hope to expand it in future work to encapsulate the idea of a type that can compute a value of some type from the `Env`, which will be useful in some cases for synthesizing env values.

This also includes updates to `TextLayout` and `Label` in order to actually use this functionality, but it does not adopt it anywhere else; I'll either do that as part of a follow-up or I'll open an issue outlining the work.

I'd also like to update the docs of the `Widget::update` method in order to cover handling changes in the `Env`.

### DIFF
--- a/druid/src/env.rs
+++ b/druid/src/env.rs
@@ -133,6 +133,35 @@ pub enum KeyOrValue<T> {
     Key(Key<T>),
 }
 
+/// A trait for anything that can resolve a value of some type from the [`Env`].
+///
+/// This is a generalization of the idea of [`KeyOrValue`], mostly motivated
+/// by wanting to improve the API used for checking if items in the [`Env`] have changed.
+///
+/// [`Env`]: struct.Env.html
+/// [`KeyOrValue`]: enum.KeyOrValue.html
+pub trait KeyLike<T> {
+    /// Returns `true` if this item has changed between the old and new [`Env`].
+    ///
+    /// [`Env`]: struct.Env.html
+    fn changed(&self, old: &Env, new: &Env) -> bool;
+}
+
+impl<T: ValueType> KeyLike<T> for Key<T> {
+    fn changed(&self, old: &Env, new: &Env) -> bool {
+        !old.get_untyped(self).same(new.get_untyped(self))
+    }
+}
+
+impl<T> KeyLike<T> for KeyOrValue<T> {
+    fn changed(&self, old: &Env, new: &Env) -> bool {
+        match self {
+            KeyOrValue::Concrete(_) => false,
+            KeyOrValue::Key(key) => !old.get_untyped(key).same(new.get_untyped(key)),
+        }
+    }
+}
+
 /// Values which can be stored in an environment.
 pub trait ValueType: Sized + Into<Value> {
     /// Attempt to convert the generic `Value` into this type.
@@ -244,7 +273,7 @@ impl Env {
     /// Panics if the key is not found.
     ///
     /// [`Value`]: enum.Value.html
-    pub fn get_untyped(&self, key: impl Borrow<Key<()>>) -> &Value {
+    pub fn get_untyped<V>(&self, key: impl Borrow<Key<V>>) -> &Value {
         match self.try_get_untyped(key) {
             Ok(val) => val,
             Err(err) => panic!("{}", err),
@@ -259,7 +288,7 @@ impl Env {
     /// e.g. for debugging, theme editing, and theme loading.
     ///
     /// [`Value`]: enum.Value.html
-    pub fn try_get_untyped(&self, key: impl Borrow<Key<()>>) -> Result<&Value, MissingKeyError> {
+    pub fn try_get_untyped<V>(&self, key: impl Borrow<Key<V>>) -> Result<&Value, MissingKeyError> {
         self.0.map.get(key.borrow().key).ok_or(MissingKeyError {
             key: key.borrow().key.into(),
         })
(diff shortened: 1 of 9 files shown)


raphlinus (approved):
As discussed extensively on Zulip, I'm approving on the basis that this should improve performance in real cases, and the API seems pretty reasonable. There is more architecture work needed on env, but that's for a future cycle.

druid/src/env.rs
```diff
@@ -133,6 +133,35 @@ pub enum KeyOrValue<T> {
...
 }
 
+/// A trait for anything that can resolve a value of some type from the [`Env`].
+///
+/// This is a generalization of the idea of [`KeyOrValue`], mostly motivated
+/// by wanting to improve the API used for checking if items in the [`Env`] have changed.
+///
+/// [`Env`]: struct.Env.html
+/// [`KeyOrValue`]: enum.KeyOrValue.html
+pub trait EnvResolveable<T> {
```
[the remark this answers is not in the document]
cmyr: I feel you, and I struggled to come up with another name. 'Resolve' does not sound right to me– perhaps `EnvResolver` would be better. I suppose you don't like `Enviable` either then, huh?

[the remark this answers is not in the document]
cmyr: This is now `KeyLike`. 🤷

druid/src/env.rs
```diff
@@ -133,6 +133,35 @@ pub enum KeyOrValue<T> {
...
+pub trait KeyLike<T> {
+    /// Returns `true` if this item has changed between the old and new [`Env`].
+    ///
+    /// [`Env`]: struct.Env.html
+    fn changed(&self, old: &Env, new: &Env) -> bool;
+}
+
+impl<T: ValueType> KeyLike<T> for Key<T> {
+    fn changed(&self, old: &Env, new: &Env) -> bool {
+        !old.get_untyped(self).same(new.get_untyped(self))
```
mmitteregger, line 152: I've got a panic on this line.
I'm not sure if I am doing anything wrong, but I think this would be better written as:
```suggestion
        match (old.try_get_untyped(self), new.try_get_untyped(self)) {
            (Err(_), Err(_)) => false,
            (Err(_), Ok(_)) => true,
            (Ok(_), Err(_)) => true,
            (Ok(old_value), Ok(new_value)) => !old_value.same(new_value),
        }
```

druid/src/env.rs
```diff
@@ -133,6 +133,35 @@ pub enum KeyOrValue<T> {
...
+    }
+}
+
+impl<T> KeyLike<T> for KeyOrValue<T> {
+    fn changed(&self, old: &Env, new: &Env) -> bool {
+        match self {
+            KeyOrValue::Concrete(_) => false,
+            KeyOrValue::Key(key) => !old.get_untyped(key).same(new.get_untyped(key)),
```
mmitteregger, line 160: Same as above.

cmyr: huh, is your panic that some key is missing from the env?

mmitteregger: Yeah, pretty sure that's it.

I am trying to change the color of a menu label based on the current selection and hot state.
I might be doing something wrong, because I'm in the middle of a upgrade to the latest master (including this PR) and haven't yet looked at all the changes and only tried to get everything in a working state again.

I've created a somewhat "minimal" example repo for this, if you want to take a look:
https://github.com/mmitteregger/druid-example/blob/master/src/widget/menu/menu_item.rs#L123-L125

rjwittams:
I think LensWrap needs to check if the environment has changed or it will skip environment only updates (meaning e.g label children of LensWrap would not get the update).

druid/src/env.rs
cmyr: In your example, I wouldn't use a `Key` at all for setting the color; I would just explicitly pass the color you want to the label.

I also wouldn't use `WidgetPod` for your label; if you are just using it for layout, you can draw the label where you want in `Paint` by using `Label::draw_at` (this is new).

cmyr: Having dug into this a bit more: you have three keys in the `Env`, `MENU_ITEM_SELECTED_COLOR` and three other variants. you then also have a key `Color` that you're setting on the label, and then you're dynamically updating this key based on the other keys.

when you set a `Key` as the value for something like text color, it means that when drawing we will figure out what that key's current color is, and then we'll use that color. So if you want to change the color, one option is to never change the key assigned to the label, and instead change the value for that key in the `Env`. An alternative is to change the key assigned to the label; for instance when your text is selected, you could do `label.set_text_color(MENU_ITEM_SELECTED_COLOR)`. In your particular case, I think this is probably the easiest approach?

The rationale for this is that you could have different themes that provide different values for various keys, and it would 'just work'. Similarly, if you were using built-in keys we could support things like dark mode, or high-contrast colors for accessibility, and it would 'just work'.

Your crash is happening because the `COLOR` key you're using doesn't always exist in the `Env`;https://linebender.org/druid/env.html#keys-values-and-themes. I think this crash is okay; by the time `update` happens, any key you've assigned to a label must exist, and we crashing if it doesn't is reasonable behaviour.

Issue #1199, opened 2020-09-09
Strange artifacts on `Spinner` at some scroll offsets


ForLoveOfCats:
At most scroll offsets past a certain point I've noticed strange artifacts on the Spinner widget. Scrolling around past this min value results in different artifacts and some offsets where the issue disappears again.

![1](https://user-images.githubusercontent.com/30666851/92559212-21f7bf00-f23e-11ea-9763-df8982758ad2.gif)
![2](https://user-images.githubusercontent.com/30666851/92559219-22905580-f23e-11ea-9954-2b5875cec8bb.gif)
![3](https://user-images.githubusercontent.com/30666851/92559220-23c18280-f23e-11ea-9e90-105b281d6266.gif)

Perhaps this is an invalidation rect issue? (cc @jneem)

Information:
GTK Shell (Cairo Piet backend)
Built from commit ddd668613807cf6b2489da935116aa38e20734fb

PR #1208, opened 2020-09-10
Fix the invalidation region for Scroll.

jneem:
The scroll widget needs to call `child.set_viewport_offset` every time the scroll offset changes.

### DIFF
--- a/druid/src/widget/scroll.rs
+++ b/druid/src/widget/scroll.rs
@@ -97,7 +97,9 @@ impl<T, W: Widget<T>> Scroll<T, W> {
     ///
     /// Returns `true` if the scroll offset has changed.
     pub fn scroll(&mut self, delta: Vec2, layout_size: Size) -> bool {
-        self.scroll_component.scroll(delta, layout_size)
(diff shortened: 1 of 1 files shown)


druid/src/widget/scroll.rs
```diff
@@ -97,7 +97,9 @@ impl<T, W: Widget<T>> Scroll<T, W> {
     ///
     /// Returns `true` if the scroll offset has changed.
     pub fn scroll(&mut self, delta: Vec2, layout_size: Size) -> bool {
-        self.scroll_component.scroll(delta, layout_size)
+        let scrolled = self.scroll_component.scroll(delta, layout_size);
```
rjwittams, line 100: Feels like an interaction that other potential users of scroll component are going to get wrong. Not sure if theres a better way though..

jneem: Yeah, I'm not wild about the fact that it needs to be handled in three separate places. Maybe I'll find some time next weekend to take a stab at a `Viewport`...

ForLoveOfCats (approved):
Fixes https://github.com/linebender/druid/issues/1199

PR #1211, opened 2020-09-10
Fix issues with placeholder TextBox text

cmyr:
This was broken slightly by #1182. This patch also cleans up a few
changes from that patch in the styled_text example that I had
not intended to commit.

As reported by @totsteps


totsteps:
Thanks for the fix, works as intended.

Finnerale (approved):

PR #1215, opened 2020-09-11
Use Global target for external events

cmyr:
This fixes the issue addressed in #1206, but also includes a bunch of doc cleanup.

@MTRNord can you double-check that this also fixes your problem? It's a slightly different change than what you'd done.


MTRNord:
This fixes my issue. Closing my original PR in favor of this

raphlinus (approved):
Glad this is fixed, and also happy to see doc improvements, as this stuff can be confusing. Thanks!

PR #1218, opened 2020-09-11
Fix SimpleLogger not getting initialized

cmyr:
This was a bug in c892631f.


raphlinus (approved):
Seems reasonable, if a bit of a footgun in SimpleLogger. Should they perhaps add a `#[must_use]` on their `new` method?

cmyr:
An entirely reasonable idea, opened a PR over there.

PR #1219, opened 2020-09-11
Fix incorrect size on window creation

raphlinus:
This is a followup to #1037 that fixes the problem where a window
created with default size ends up too big in hi-dpi contexts. It is a
minimal change, and doesn't address the flashing issue.


cmyr (approved):
Looks good!

Issue #1216, opened 2020-09-11
Standardize control widget padding


cmyr:
This is coming up as I look to implement baseline alignment for widgets; several widgets have internal padding that complicates getting them to line up as expected with other widgets. The `RadioGroup`, for instance, adds 5 units of padding around each item in the group; putting a `RadioGroup` in a column with another widget requires and getting it to look right involves going and reading the source to figure out the correct padding value.

I'd like to try and simplify this as much as possible; this will mean removing any magic internal padding values, and trying to rationalize the rest of the padding values we use into a few items in the `Env`. In particular I would like to have a single standard 'inter-widget-padding' value.

PR #1220, opened 2020-09-11
Add default horizontal & vertical padding to theme

cmyr:
This also uses these values in the Flex container, via new
add_default_spacer and with_default_spacer methods. This should
be the preferred way of spacing the standard control widgets.

This is progress on #1216; I also want to go and remove any other
internal padding constants from existing widgets.


Finnerale (approved):
Seems like a good idea.

Maybe `with_theme_spacer` would be a more obvious name? Because I have no concept of a "default spacer" but I would immediately know what's meant by a "theme spacer".

cmyr:
I was thinking about making this `with_spacer` and making `with_spacer` be `with_fixed_spacer`, but that would be annoying breakage. maybe `with_auto_spacer`?

Finnerale:
Hmm, I've the same problem with `with_auto_spacer` as with `with_default_spacer`, actually I've even more of an issue with 'auto' because it makes me think of something adaptive...

Actually, now that I think about it again, `with_default_spacer` seems fine to me. Calling it `with_theme_spacer` makes no sense if the user (or even Druid) would define multiple different "default" sizes.

Maybe we could call them `with_spacer` and `with_custom_spacer`?. I don't think changing `with_spacer` would be an issue, as it is a single find-and-replace to fix it.

This is one of the rather rare moments where I wish Rust had overloading...

cmyr:
I'm not sure there's anything I like better than `with_default_spacer`, so I think I'm just going to keep that as-is for now 🤷

PR #1222, opened 2020-09-11
Standardize checkbox and radio padding

cmyr:
This is based off of #1216; these two widgets now use padding constants from the environment internally, and so you can manually mimic the look of `RadioGroup` by creating a `Flex::column()` and then adding default spacers between each child.

This also includes some other cleanup in these two widgets; in particular they no longer user `WidgetPod` internally (which was only used so they could position their labels) and instead I've added a `draw_at` method to label to handle this case, and updated the documentation to mention it.

If anyone is confused: all of this layout and padding stuff is motivated by wanting to implement baseline alignment; that's going to touch a bunch of layout code, and so I'm trying to clean it up a bit before hand.

### DIFF
--- a/druid/src/widget/radio.rs
+++ b/druid/src/widget/radio.rs
@@ -14,14 +14,16 @@
 
 //! A radio button widget.
 
-use crate::kurbo::{Circle, Point, Rect, Size};
+use crate::kurbo::{Circle, Size};
 use crate::theme;
-use crate::widget::{CrossAxisAlignment, Flex, Label, LabelText, Padding};
+use crate::widget::{CrossAxisAlignment, Flex, Label, LabelText};
 use crate::{
     BoxConstraints, Data, Env, Event, EventCtx, LayoutCtx, LifeCycle, LifeCycleCtx, LinearGradient,
-    PaintCtx, RenderContext, UnitPoint, UpdateCtx, Widget, WidgetExt, WidgetPod,
+    PaintCtx, RenderContext, UnitPoint, UpdateCtx, Widget,
 };
 
+const DEFAULT_RADIO_RADIUS: f64 = 7.0;
+const INNER_CIRCLE_RADIUS: f64 = 2.0;
 /// A group of radio buttons
 #[derive(Debug, Clone)]
 pub struct RadioGroup;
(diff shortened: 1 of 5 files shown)


druid/src/widget/radio.rs
```diff
@@ -14,14 +14,16 @@
...
-use crate::widget::{CrossAxisAlignment, Flex, Label, LabelText, Padding};
+use crate::widget::{CrossAxisAlignment, Flex, Label, LabelText};
 use crate::{
     BoxConstraints, Data, Env, Event, EventCtx, LayoutCtx, LifeCycle, LifeCycleCtx, LinearGradient,
-    PaintCtx, RenderContext, UnitPoint, UpdateCtx, Widget, WidgetExt, WidgetPod,
+    PaintCtx, RenderContext, UnitPoint, UpdateCtx, Widget,
 };
 
+const DEFAULT_RADIO_RADIUS: f64 = 7.0;
+const INNER_CIRCLE_RADIUS: f64 = 2.0;
```
Zarenor, line 26: These are consts? Is it reasonable to make them keys in the Env? Or fields? 
Right now, if I customize BASIC_WIDGET_HEIGHT, I get a centered circle based on these consts - if I make it larger, my circle doesn't grow. If I make it smaller, there's a point at which the circle overflows the bounds. I do see that it's probably unlikely to get so small, but it's certainly possible.

Zarenor (approved):
I like these changes! The code is overall cleaner, just by catching up these implementations. And the added API on label (and the better docs) seem like a big win for little cost.

druid/src/widget/radio.rs
cmyr: I'm avoiding making any interventions in how this is actually drawn; these were previously defined in `paint` and I just moved them to the top-level. I think the desired behaviour here is up for discussion; I don't think it's necessarily wrong to say that these have a fixed size they will draw regardless of the layout size, which is similar to how text works. I can see the other side as well. Any changes to that behaviour would be a separate PR and discussion, though?

Zarenor: I think I missed them being extant in the old paint code - moving them to a more visible scope is definitely a good thing. But it seems like putting them in the Env accords with the BASIC_WIDGET_SIZE and TEXT_SIZE_NORMAL being Env keys. I'll consider writing up an issue over this and some of the other bits and bobs we discussed in the Zulip to see if there's some discussion to be had.

PR #1223, opened 2020-09-11
Cleanup geometry code in progress bar

cmyr:
There have been various API improvements in kurbo since this
was last touched, and I was looking at it anyway.


raphlinus (approved):
Nice! Really good to see how higher-level APIs leads to cleaner code.

PR #1224, opened 2020-09-11
Mac and GTK implementation of Screen module and related window methods

rjwittams:
This is extracted from sub-windows and was a fairly minimal implementation to work for those.

The main caveat that I know of here is that the mac implementation of get_position / set_position won't work properly if you have multiple monitors with a large monitors baseline lower than the first monitors baseline . I plan to fix this, but would prefer to do it after sub-windows as I have a decent example there (but it relies on sub-windows).


raphlinus (approved):
This looks good to me, though I didn't track down all the API references. I'm happy with shaking out problems as we use it. Thanks!

PR #1226, opened 2020-09-12
Add `Button::from_label`

ForLoveOfCats:
When using Button I wanted to be able to use the builder fns on Label. Rather than stub them all on Button and try to keep it up to date with Label as new fns are added it felt much simpler and cleaner to simply allow constructing a Button with a provided Label. There is no corresponding `dynamic_from_label`, one can use `Label::dynamic` to achieve that.

### DIFF
--- a/druid/src/widget/button.rs
+++ b/druid/src/widget/button.rs
@@ -33,7 +33,7 @@ pub struct Button<T> {
 impl<T: Data> Button<T> {
     /// Create a new button with a text label.
     ///
-    /// Use the `.on_click` method to provide a closure to be called when the
+    /// Use the [`.on_click`] method to provide a closure to be called when the
     /// button is clicked.
     ///
     /// # Examples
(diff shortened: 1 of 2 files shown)


ForLoveOfCats:
Something's going nuts with the nightly CI, it appears to be failing to install the toolchain

raphlinus:
Probably fallout from https://rust-lang.github.io/rustup-components-history/x86_64-apple-darwin.html ; our CI scripts are not robust against failures there.

druid/src/widget/button.rs
```diff
@@ -52,6 +52,27 @@ impl<T: Data> Button<T> {
...
+    ///
+    /// Use the `.on_click` method to provide a closure to be called when the
+    /// button is clicked.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use druid::widget::Button;
+    ///
+    /// let button = Button::from_label(Label::new("Increment")).on_click(|_ctx, data: &mut u32, _env| {
```
cmyr, line 65: I think including an example is good, but I would prefer the example to demonstrate the value of this method, e.g. I would try to use some `Label` methods that are not available on `Button`.

druid/src/widget/button.rs
```diff
@@ -52,6 +52,27 @@ impl<T: Data> Button<T> {
         }
     }
 
+    /// Create a new button with the provided label.
```
cmyr, line 55: I would link to `Label` here.

druid/src/widget/button.rs
```diff
@@ -52,6 +52,27 @@ impl<T: Data> Button<T> {
...
+    ///
+    /// Use the `.on_click` method to provide a closure to be called when the
```
cmyr, line 57: and I'd probably link to `on_click` here? I'm trying to err on the side of linking to things from docs, I think it's ultimately a better experience.

druid/src/widget/button.rs
```diff
@@ -52,6 +52,27 @@ impl<T: Data> Button<T> {
...
+    ///
+    /// let button = Button::from_label(Label::new("Increment")).on_click(|_ctx, data: &mut u32, _env| {
+    ///     *data += 1;
+    /// });
+    /// ```
+    pub fn from_label(label: Label<T>) -> Button<T> {
+        Button {
+            label,
+            label_size: Size::ZERO,
+        }
```
cmyr, line 73: You might consider making `Button::new()` now just call this under the covers?

cmyr (approved):
A few suggestions but nothing blocking, this looks like a good addition!

druid/src/widget/button.rs
ForLoveOfCats: Fair point. I was worried that setting the font would be to much fluff for this but setting the text color seems like a reasonable thing to demonstrate.

druid/src/widget/button.rs
ForLoveOfCats: Sure, I'll update the doc comment on `Button::new` as well

druid/src/widget/button.rs
ForLoveOfCats: Sure, I'll update the doc comment on `Button::new` as well

druid/src/widget/button.rs
ForLoveOfCats: Sure, why not :shrug:

PR #1227, opened 2020-09-12
Add tests to ensure that invalidation regions are transformed by scroll.

jneem:
This adds some tests to ensure that the invalidation regions are being correctly propagated up and down the tree, including in the presence of scrolling.

### DIFF
--- a/druid/src/tests/invalidation_tests.rs
+++ b/druid/src/tests/invalidation_tests.rs
@@ -1,0 +1,128 @@
+// Copyright 2020 The Druid Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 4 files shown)


druid/src/tests/invalidation_tests.rs
```diff
@@ -0,0 +1,128 @@
...
+        y1: 50.,
+    };
+
+    struct Invalidator {
+        invalid: bool,
+    }
+
+    impl<T: Data> Widget<T> for Invalidator {
+        fn event(&mut self, ctx: &mut EventCtx, _: &Event, _: &mut T, _: &Env) {
+            dbg!(self.invalid);
```
rjwittams, line 69: Is this dbg meant to still be in here?

jneem: No, sorry...

rjwittams (approved):

PR #1229, opened 2020-09-12
Fix bug with textbox placeholder selectable

totsteps:
This fixes an issue where the `TextBox` placeholder was selectable.

Before this fix:
Untitled

### DIFF
--- a/druid/src/widget/textbox.rs
+++ b/druid/src/widget/textbox.rs
@@ -434,13 +434,11 @@ impl Widget<String> for TextBox {
 
     fn paint(&mut self, ctx: &mut PaintCtx, data: &String, env: &Env) {
         // Guard against changes in data following `event`
-        let content = if data.is_empty() {
-            &self.placeholder
+        if data.is_empty() {
+            self.selection = Selection::caret(0);
         } else {
-            data
+            self.selection = self.selection.constrain_to(data);
         };
-
-        self.selection = self.selection.constrain_to(content);
 
         let height = ctx.size().height;
         let background_color = env.get(theme::BACKGROUND_LIGHT);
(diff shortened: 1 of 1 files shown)


druid/src/widget/textbox.rs
```diff
@@ -434,14 +434,13 @@ impl Widget<String> for TextBox {
 
     fn paint(&mut self, ctx: &mut PaintCtx, data: &String, env: &Env) {
         // Guard against changes in data following `event`
-        let content = if data.is_empty() {
-            &self.placeholder
+        if data.is_empty() {
+            self.selection = Selection::caret(0);
+            self.hscroll_offset = 0.;
```
cmyr, line 439: I don't think `paint` is the correct place for us to be modifying this state; we should just not allow it in the first place.

I would look into the events that modify selection, and try to make it so that we always keep our selection at 0 if we don't have real text?

cmyr (changes requested):
This is definitely a bug, but I worry that this fix might cause unexpected problems; we should just never allow a non-zero selection when the data is empty.

Thanks for opening, this is definitely something we want to fix!

druid/src/widget/textbox.rs
totsteps: My bad.

The latest changes guard against`selection` being changed when receiving `EditAction::Click` and `EditAction::Drag` events on the empty text.

cmyr (changes requested):
I'm going to be more nitty here; reading through the code I think the correct fix is at the end of `event`, where we check if `data.is_empty()` and set the placeholder; when we set the placeholder we should also set the selection to `0`.

Sorry for the runaround!

cmyr (approved):
great, thanks!

PR #1230, opened 2020-09-12
Set default label line break mode to Overflow to avoid Clip perf overhead

smmalis37:
Addresses https://github.com/linebender/piet/issues/289#issuecomment-691416962 (kinda).


cmyr (approved):
Thanks!

PR #1232, opened 2020-09-13
Unit lens.

rjwittams:
From sub windows branch.
Used to allow the sub window host to be a direct child of the app state.


ForLoveOfCats:
Having recently built a widget to do this exact job I feel that this will be very handy to have included out of the box :)

Perhaps add a changelog entry?

jneem (approved):

PR #1233, opened 2020-09-13
Adds a viewport widget.

jneem:
This is my first stab at adding a Viewport widget.

The idea is that `Viewport` will be a building-block for various scroll-like widgets, and it will correctly handle the bits that are easy to mess up or impossible to do out-of-tree (like correctly propagating invalid regions).

### DIFF
--- a/druid/src/scroll_component.rs
+++ b/druid/src/scroll_component.rs
@@ -64,37 +64,6 @@ pub enum BarHeldState {
     Horizontal(f64),
 }
 
-/// Backing struct for storing scrollbar state
-#[derive(Debug, Copy, Clone)]
-pub struct ScrollbarsState {
-    /// Current opacity for both scrollbars
-    pub opacity: f64,
-    /// ID for the timer which schedules scrollbar fade out
-    pub timer_id: TimerToken,
-    /// Which if any scrollbar is currently hovered by the mouse
-    pub hovered: BarHoveredState,
-    /// Which if any scrollbar is currently being dragged by the mouse
-    pub held: BarHeldState,
-}
-
-impl Default for ScrollbarsState {
-    fn default() -> Self {
-        Self {
-            opacity: 0.0,
-            timer_id: TimerToken::INVALID,
-            hovered: BarHoveredState::None,
-            held: BarHeldState::None,
-        }
-    }
-}
-
-impl ScrollbarsState {
-    /// true if either scrollbar is currently held down/being dragged
-    pub fn are_held(&self) -> bool {
-        !matches!(self.held, BarHeldState::None)
-    }
-}
-
 /// Embeddable component exposing reusable scroll handling logic.
 ///
 /// In most situations composing [`Scroll`] or [`List`] is a better idea
@@ -108,39 +77,53 @@ impl ScrollbarsState {
 /// logic to a custom widget when the need arises.
 ///
 /// It should be used like this:
-/// - Store an instance of `ScrollComponent` in your widget's struct.
-/// - During layout, set the [`content_size`] field to the child's size.
-/// - Call [`event`] and [`lifecycle`] with all event and lifecycle events before propagating them to children.
+/// - Store an instance of `ScrollComponent` in your widget's struct, and wrap the child widget to
+///   be scrolled in a [`Viewport`].
+/// - Call [`event`] and [`lifecycle`] with all event and lifecycle events before propagating them
+///   to children.
 /// - Call [`handle_scroll`] with all events after handling / propagating them.
-/// - And finally perform painting using the provided [`paint_content`] function.
+/// - Call [`draw_bars`] to draw the scrollbars.
 ///
 /// Also, taking a look at the [`Scroll`] source code can be helpful.
 ///
 /// [`Scroll`]: ../widget/struct.Scroll.html
 /// [`List`]: ../widget/struct.List.html
-/// [`content_size`]: struct.ScrollComponent.html#structfield.content_size
+/// [`Viewport`]: ../widget/struct.Viewport.html
 /// [`event`]: struct.ScrollComponent.html#method.event
 /// [`handle_scroll`]: struct.ScrollComponent.html#method.handle_scroll
 /// [`lifecycle`]: struct.ScrollComponent.html#method.lifecycle
-/// [`paint_content`]: struct.ScrollComponent.html#method.paint_content
 #[derive(Debug, Copy, Clone)]
 pub struct ScrollComponent {
-    /// Current state of both scrollbars
-    pub scrollbars: ScrollbarsState,
+    /// Current opacity for both scrollbars
+    pub opacity: f64,
+    /// ID for the timer which schedules scrollbar fade out
+    pub timer_id: TimerToken,
+    /// Which if any scrollbar is currently hovered by the mouse
+    pub hovered: BarHoveredState,
+    /// Which if any scrollbar is currently being dragged by the mouse
+    pub held: BarHeldState,
 }
 
 impl Default for ScrollComponent {
     fn default() -> Self {
-        ScrollComponent::new()
+        Self {
+            opacity: 0.0,
+            timer_id: TimerToken::INVALID,
+            hovered: BarHoveredState::None,
+            held: BarHeldState::None,
+        }
     }
 }
 
 impl ScrollComponent {
     /// Constructs a new [`ScrollComponent`](struct.ScrollComponent.html) for use.
     pub fn new() -> ScrollComponent {
-        ScrollComponent {
-            scrollbars: ScrollbarsState::default(),
-        }
+        Default::default()
+    }
+
+    /// true if either scrollbar is currently held down/being dragged
+    pub fn are_bars_held(&self) -> bool {
+        !matches!(self.held, BarHeldState::None)
     }
 
     /// Makes the scrollbars visible, and resets the fade timer.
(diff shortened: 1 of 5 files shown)


druid/src/scroll_component.rs
```diff
@@ -124,11 +125,6 @@ impl ScrollbarsState {
 /// [`paint_content`]: struct.ScrollComponent.html#method.paint_content
 #[derive(Debug, Copy, Clone)]
 pub struct ScrollComponent {
```
rjwittams, line 127: This name now seems even less appropriate than it did before. Maybe in a subsequent PR it could be renamed to ScrollbarOverlay or something. I'm also not entirely sure why its on the top level - even though it is not a widget itself, it is entirely concerned with widget matters as far as I can tell. But can totally see leaving it for this PR to make the diff easier to read

druid/src/scroll_component.rs
```diff
@@ -124,11 +125,6 @@ impl ScrollbarsState {
...
-    /// The size of the scrollable content, make sure to keep up this
-    /// accurate to the content being scrolled
-    pub content_size: Size,
-    /// Current offset of the scrolling content
-    pub scroll_offset: Vec2,
     /// Current state of both scrollbars
     pub scrollbars: ScrollbarsState,
```
rjwittams, line 129: Should ScrollbarsState be collapsed into the parent struct?

jneem: Yes, that seems to work nicely.

druid/src/scroll_component.rs
jneem: I basically agree. But as you know, my dream is to eventually move this logic into a nicely-composable widget. In the meantime, I'm not super inclined to do lots of moving and renaming...

rjwittams (approved):

ForLoveOfCats:
I'm sorry I wasn't able to review this in time but I *really* don't like this change. It undermines the entire usefulness of ScrollComponent (which was identified in Robert's review in the mentioning that the name doesn't match up with the function very well after this change). The explicit goal of the ScrollComponent as stated numerous times was to patch up the hole in the layout system where a scrolling child needs more information than can be provided by its Scroll parent in order to perform the desired layout behavior. This change prevents that use case entirely by tying portions of the ScrollComponent logic to this Viewport widget which does not fit into the situation where a widget has no child and needs to scroll itself which is the whole goal of this ScrollComponent in the first place. I recognize the usefulness of generic viewport functionality where a child widget can be painted at a specific offset and all the details handled internally. However that should not come at a determent of the usefulness of another system. I would have preferred individual switches on Scroll to disable scrollbars, disable scroll input, and stub methods out to manipulate the offset. Baring that if the desire really is to use a viewport widget as the child of Scroll then don't tie the usage of ScrollComponent to this viewport widget.

rjwittams:
Is it tied to it?

ForLoveOfCats:
Yes, see https://github.com/linebender/druid/blob/master/druid/src/scroll_component.rs#L222

ForLoveOfCats:
I have nothing against having a viewport widget and I feel it has great utility for downstream widget development. As a result I feel like all that really needs to be done here is to un-tie ScrollComponent from Viewport in a second PR. Having scrollbar drawing in a separate call is fine to me. My main issue is just that these changes require the usage of Viewport to interact with ScrollComponent which removes the entire usage case it was originally designed for.

rjwittams:
I think it can be parameterized over exactly what it is controlling. 
In terms of turning scroll bars on and off I have that in a branch that I will manually merge yet again when the dust settles.

jneem:
Sorry for jumping the gun on this; I am preparing a revert commit.

I do want to say a bit more about the motivation for this, though, because I don't see a nice way to reconcile it with what I think you're proposing.

I have two main issues with the previous state of things:
- it was not obvious how to get all the transformations right
- it was impossible to get the transformations right in an out-of-tree widget.

The point of viewport was to centralize the translation logic in one place. Then you can get correct scrolling by just embedding a viewport widget. Note that prior to this patch, I had to insert calls to `WidgetPod::set_viewport_offset` in three different parts of `Scroll` (as part of #1208). Each time, I had to know that the method I was calling on `ScrollComponent` had a chance of changing the offset. The advantage of integrating the `Viewport` is that we only have one location for storing and updating the viewport size and offset.

Also, you mention widgets that "scroll themselves", but this is currently not compatible with the way invalidation regions are propagated. You need to have a `WidgetPod` child that you can call `set_viewport_offset` on in order for the invalidation region to get propagated up correctly. My idea for how "self-scrolling" widgets would work was that they should be composite widgets that contain a viewport with a private inner `WidgetPod`.

ForLoveOfCats:
The way I see it a self-scrolling widget is no different from any other widget in that it generally will invalidate its entire painting rect. The architecture where one has a two custom widgets with a scrolling/viewport widget sandwiched in the middle was already possible prior to ScrollComponent (and was something I argued against in the discussions about ScrollComponent, if this were to be the desired approach to the problem of needing to know more layout information when having scroll behavior then we would rip out ScrollComponent entirely).

I see utility in the viewport widget outside the issue of scrolling. The invalidation rect issue with an offset child is something I agree is ungainly but I don't think it's as big of an issue in practice with how we currently only have one container which offsets a child widget. If and when that number increases I think that it can be revisited as then there will be an actual usage case to consider when designing a solution.

rjwittams:
I think the separation of offset management from scroll bar drawing & interaction does make sense in and of itself, for example when you have scroll bars over a whole composite, but various children scroll differently (e.g frozen regions in a spreadsheet). 
So I don't think it should be ripped out exactly, if it ended up similar to what Joe had before. 

I don't think the fear of widget composition should hold a huge amount of weight given the architecture we have. IE if something is possible, but clunky, that is ok, if we are aware of the clunkiness and know it has to be addressed at some point (perhaps by easier state management or extended layout etc). Something being *impossible* to get right isn't really tenable however...

jneem:
> The way I see it a self-scrolling widget is no different from any other widget in that it generally will invalidate its entire painting rect. 

When it scrolls, sure, but what if you have a list, say, and only one of the children invalidates itself?

I think my main high-level problem is that I'm having trouble imagining a situation where you want scroll bars but not a viewport. Or maybe a better way of saying it is: if you have scroll bars but not a viewport, then you need to re-implement viewport's logic anyway.

On a more down-to-earth level, the reason I made those changes to `ScrollComponent` is that after writing `Viewport` I found that it duplicated a lot of the state and logic from `ScrollComponent`. Besides the fact that the duplication made me unhappy, this also made it tricky to use, because using both `Viewport` and `ScrollComponent` required you to manually synchronize the state between the two of them. Because of this, I'd rather not just undo the `ScrollComponent` changes from this patch. I'd prefer to revert completely and then spend some time finding a better way of reducing the duplication.

ForLoveOfCats:
The ScrollComponent was not designed for containers of children, that's what Scroll is for. I'll use the example of a multi-line text edit widget as that is what I am most familiar with and drove my initial interest in the topic.

This widget wants to paint in all available space. Scroll has to provide infinite constraints on at least one axis for it to perform its job. This widget also might want to access the current scroll offset during layout to determine where in the document it is located and then calculate what the longest visible line is and then determine the "scrollable" width is based on that. It wants to get the scroll offset at paint time to determine which portion of the document to paint. This widget has no children, when it paints it paints its entire rect in the window. It doesn't want to paint one or more children at a specific offset, it *is* the child. These things are *much* more clunky to accomplish by way of working around the layout behavior of Scroll.

After playing around with List some more I no longer believe that it should handle its own scrolling but instead continue to be placed in a Scroll instance just like Flex needs to when used for a list of items. I think that for the foreseeable future Scroll is our only widget which needs to paint a child at a modifiable offset.

I think the main source of miscommunication here is on what problems ScrollComponent was originally designed to solve. It's not anti-composition nor is it designed to make Scroll simpler though that is a nice plus which would allow for this viewport type usage be added to it without complicating it to much. It patches a hole in the layout system for specific use cases of which there were no other decent solutions proposed (iirc I was the one who brought up the three widget stack as a clunky and annoying solution which I wanted to avoid in the original PR).

It would be helpful for me as well to understand the specific use cases you were thinking of when implementing this PR so I may better understand what situations you are trying to improve. My understanding currently is that it is to make it easier to implement scrolling containers other than Scroll which as I mentioned before I do not believe to have a significant use.

jneem:
Thanks, that description helps. I think I'm going to try a version where the "viewport state" (i.e. child size and viewport rect) has its own struct, which can (but doesn't have to) be owned by the `Viewport` widget. Probably not until the weekend at the earliest, though, and I don't even know how well it will work. My offer to revert still stands...

I am a bit surprised by this, though:
> This widget has no children, when it paints it paints its entire rect in the window
When the cursor blinks, don't you want to repaint only the cursor?

> It would be helpful for me as well to understand the specific use cases you were thinking of when implementing this PR
My main motivation was #1208 rather than my particular use-cases. I wanted to make it easy to write a correct scrolling widget. I do have one situation where I would use a viewport, though: a timeline for a video-editing program. I'm currently using a `Scroll` with its scrollbars disabled and I'm overriding `Scroll`'s built-in event handling by wrapping it in a `Controller`.

ForLoveOfCats:
If you think viewport is best suited to be a specialized widget instead of a mode of Scroll then that sounds like a good middle ground as I do definitely see usefulness in having a widget like that for various usage cases such as your own mention of a timeline. I'm game for the revert in that case until you feel comfortable with a PR with the adapted changes.

The mention of a blinking cursor is a good point but I don't feel that a widget knowing to offset a specific invalidation rect that it has produced is to much to ask especially when that widget is being built to be scroll aware in the first place.

PR #1234, opened 2020-09-13
Restore broken rustfmt

raphlinus:
Switching to the cfg_if! macro broke rustfmt in the druid-shell platform
modules. This patch restores the individual #[cfg] attributes on the
items, and also applies rustfmt, demonstrating how much entropy can
accumulate if you let it.


jneem (approved):

smmalis37:
github needs a 😱 reaction for things like this

cmyr (approved):
😱 indeed.

PR #1235, opened 2020-09-14
Split up WindowDesc to allow sub-windows and window reconfiguration.

rjwittams:
This is extracted from sub-windows. 

It allows reconfiguration of windows and paves the way for construction of non-top level ones.

### DIFF
--- a/druid/src/app.rs
+++ b/druid/src/app.rs
@@ -16,11 +16,11 @@
 
 use crate::ext_event::{ExtEventHost, ExtEventSink};
 use crate::kurbo::{Point, Size};
-use crate::shell::{Application, Error as PlatformError, WindowBuilder, WindowHandle};
+use crate::shell::{Application, Error as PlatformError, WindowBuilder, WindowHandle, WindowLevel};
 use crate::widget::LabelText;
 use crate::win_handler::{AppHandler, AppState};
 use crate::window::WindowId;
-use crate::{AppDelegate, Data, DruidHandler, Env, LocalizedString, MenuDesc, Widget, WidgetExt};
+use crate::{AppDelegate, Data, Env, LocalizedString, MenuDesc, Widget};
 
 use druid_shell::WindowState;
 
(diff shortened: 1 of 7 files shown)


druid/src/app.rs
```diff
@@ -37,27 +35,75 @@ pub struct AppLauncher<T> {
...
     /// The `WindowId` that will be assigned to this window.
     ///
     /// This can be used to track a window from when it is launched and when
     /// it actually connects.
     pub id: WindowId,
 }
 
+/// The parts of a window, pending construction, that are dependent on top level app state.
+/// This includes the boxed root widget, as well as other window properties such as the title.
+pub struct PendingWindow<T> {
```
cmyr, line 63: Does this need to be public?

druid/src/app.rs
```diff
@@ -37,27 +35,75 @@ pub struct AppLauncher<T> {
...
+    pub(crate) root: Box<dyn Widget<T>>,
+    pub(crate) title: LabelText<T>,
+    pub(crate) menu: Option<MenuDesc<T>>,
+}
+
+impl<T: Data> PendingWindow<T> {
+    /// Create a pending window from a boxed widget.
+    pub fn new_from_boxed(root: Box<dyn Widget<T>>) -> PendingWindow<T> {
```
cmyr, line 71: this jumps out at me; I'm sure there's a rationale it just isn't immediately clear. Is there a reason for this to be public API? Does it make sense to only have `fn new`, and make that do the boxing? Are there places where the incoming widget is already boxed, and we want to avoid boxing in that case?

druid/src/app.rs
```diff
@@ -142,6 +188,156 @@ impl<T: Data> AppLauncher<T> {
...
+    /// window.window_size((1000.0, 500.0));
+    /// ```
+    ///
+    /// The actual window size in pixels will depend on the platform DPI settings.
+    ///
+    /// This should be considered a request to the platform to set the size of the window.
+    /// The platform might increase the size a tiny bit due to DPI.
+    ///
+    /// [`Size`]: struct.Size.html
+    /// [display points]: struct.Scale.html
```
cmyr, line 221: appreciate the docs 👍

cmyr (approved):
This looks good to me, a few questions inline but nothing blocking. Thanks!

druid/src/app.rs
```diff
@@ -37,27 +35,75 @@ pub struct AppLauncher<T> {
...
+            title: LocalizedString::new("app-name").into(),
+            menu: MenuDesc::platform_default(),
+        }
+    }
+
+    /// Create a pending window from any widget.
+    pub fn new<W, F>(root: F) -> PendingWindow<T>
+    where
+        W: Widget<T> + 'static,
+        F: FnOnce() -> W + 'static,
```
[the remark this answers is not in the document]
cmyr: this is my doing, and I actually forget the exact rationale. 😕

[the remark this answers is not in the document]
cmyr: (except that it was working around some issue with lifetimes or reuse or something)

rjwittams:
I'll fix some bits later/tomorrow - a lot of the docs are just moved around from WindowDesc, I don't think fixing the content of those docs really has much to do with this PR. Some things are public because they need to be for subwindows I think (which has to construct windows slightly differently from the Druid side - ie not using WindowDesc).

cmyr:
my strategy around `pub` would be to make things non-pub now, and then change them to `pub` in the follow-up PRs, as they're needed; this is a nice forcing-function, so that we are only marking `pub` the things that need to be.

druid/src/app.rs
rjwittams: The answer was yes, it did need to be public because it is what the win_handler in druid holds onto

druid/src/app.rs
rjwittams: Got rid of it. Not sure if that is good or bad. It was like this just to preserve existing functionality as far as I remember.

PR #1236, opened 2020-09-14
Revert "Adds a viewport widget. (#1233)"

jneem:
This reverts commit 307025b699d0ecc48c281e1367023834ff57cc93.


ForLoveOfCats:
I think we can wait on this until the discussion in https://github.com/linebender/druid/pull/1233 reaches a conclusion as my current stance is that a viewport widget has merit outside this specific situation as long as the embeddedable scrolling behavior is not directly tied to it.

ForLoveOfCats (approved):
See https://github.com/linebender/druid/pull/1233#issuecomment-692306940 and https://github.com/linebender/druid/pull/1233#issuecomment-692310709

PR #1237, opened 2020-09-15
Create default env with Env::default()

### LINKED ISSUE #1181: env setup is confusing
We have `Env::default` but the actual way to create the default env is `theme::init`. I think it probably _should_ be `Env::default`, but maybe there's a method in theme that adds the theme items, and `Env::default` should just call this?

totsteps:
Fixes #1181

### DIFF
--- a/druid/src/app.rs
+++ b/druid/src/app.rs
@@ -20,9 +20,7 @@ use crate::shell::{Application, Error as PlatformError, WindowBuilder, WindowHan
 use crate::widget::LabelText;
 use crate::win_handler::{AppHandler, AppState};
 use crate::window::WindowId;
-use crate::{
-    theme, AppDelegate, Data, DruidHandler, Env, LocalizedString, MenuDesc, Widget, WidgetExt,
-};
+use crate::{AppDelegate, Data, DruidHandler, Env, LocalizedString, MenuDesc, Widget, WidgetExt};
 
 use druid_shell::WindowState;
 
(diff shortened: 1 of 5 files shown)


druid/src/app.rs
```diff
@@ -118,7 +116,7 @@ impl<T: Data> AppLauncher<T> {
     pub fn launch(mut self, data: T) -> Result<(), PlatformError> {
         let app = Application::new()?;
 
-        let mut env = theme::init();
+        let mut env = crate::Env::default();
```
cmyr, line 119:
```suggestion
        let mut env = Env::default();
```

druid/src/core.rs
```diff
@@ -995,7 +995,7 @@ mod tests {
             state: &mut state,
         };
 
-        let env = crate::theme::init();
+        let env = crate::Env::default();
```
cmyr, line 998:
```suggestion
        let env = Env::default();
```
I think `Env` is already in scope, here?

druid/src/theme.rs
```diff
@@ -74,9 +74,8 @@ pub const SCROLLBAR_EDGE_WIDTH: Key<f64> =
     Key::new("org.linebender.druid.theme.scrollbar_edge_width");
 
 /// An initial theme.
-pub fn init() -> Env {
-    Env::default()
-        .adding(WINDOW_BACKGROUND_COLOR, Color::rgb8(0x29, 0x29, 0x29))
+pub fn init(env: Env) -> Env {
```
cmyr, line 77: I would do something slightly fancy here, since ultimately I think we want to make this private API:

- keep the function signature the same (`pub fn init() -> Env`)
- the impl should just call `Env::default()`
- add `#[deprecated(since = "0.7.0", note = "use Env::default() instead")]`
- add a new `pub (crate) fn` called something like `add_to_env` that is exactly what you have, here


Does that make sense?

cmyr (changes requested):
Thanks! have a few little suggestions but this looks like a good bit of cleanup. :)

cmyr (approved):
Looks great, thanks!

PR #1241, opened 2020-09-18
Make GTK dependency optional.

Finnerale:
This adds a new feature `GTK` that gates all its related dependencies, which allows building the X11 backend without GTK.

`GTK` is a default feature, which means this won't change anything for existing usages, but with
```toml
druid = { version = "0.7", default-features = false, features = ["x11"] }
```
one can use Druid with X11 and no GTK bits involved :tada: 

This seems like a simple working solution, if it compiles on Windows and Mac.

Also, I had to give the feature an upper case name, otherwise it would have conflicted with the `gtk` feature...
Maybe a different name would be less ambiguous, but I could not come up with one.

### DIFF
--- a/druid-shell/Cargo.toml
+++ b/druid-shell/Cargo.toml
@@ -14,8 +14,8 @@ rustdoc-args = ["--cfg", "docsrs"]
 default-target = "x86_64-pc-windows-msvc"
 
 [features]
-x11 = ["x11rb", "nix", "cairo-sys-rs", "cairo-rs"]
-gtk = ["cairo-rs", "gio", "gdk", "gdk-sys", "glib", "glib-sys", "gtk-sys", "gtk-rs"]
+x11 = ["x11rb", "nix", "cairo-sys-rs"]
+gtk = ["gio", "gdk", "gdk-sys", "glib", "glib-sys", "gtk-sys", "gtk-rs"]
 default = ["gtk"]
 
 [dependencies]
@@ -32,9 +32,6 @@ instant = { version = "0.1.6", features = ["wasm-bindgen"] }
 anyhow = "1.0.32"
 keyboard-types = { version = "0.5.0", default_features = false }
 
-# TODO(x11/dependencies): only use feature "xcb" if using X11
-cairo-rs = { version = "0.9.1", default_features = false, features = ["xcb"], optional = true }
-
 [target.'cfg(target_os="windows")'.dependencies]
 wio = "0.2.2"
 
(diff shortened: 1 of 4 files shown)


druid-shell/Cargo.toml
```diff
@@ -14,7 +14,9 @@ rustdoc-args = ["--cfg", "docsrs"]
 default-target = "x86_64-pc-windows-msvc"
 
 [features]
-x11 = ["x11rb", "nix", "cairo-sys-rs"]
+x11 = ["x11rb", "nix", "cairo-sys-rs", "cairo-rs"]
+GTK = ["cairo-rs", "gio", "gdk", "gdk-sys", "glib", "glib-sys", "gtk-sys", "gtk"]
```
cmyr, line 18: can we just call this `gtk`? It's okay if this collides with the `gtk` crate I believe.

druid-shell/Cargo.toml
```diff
@@ -59,16 +48,19 @@ core-graphics = "0.22.0"
...
-cairo-rs = { version = "0.9.1", default_features = false, features = ["xcb"] }
-gio = "0.9.1"
-gdk = "0.13.2"
-gdk-sys = "0.10.0"
-glib = "0.10.1"
-glib-sys = "0.10.0"
-gtk-sys = "0.10.0"
-gtk = { version = "0.9.2", features = ["v3_22"] }
+# TODO(x11/dependencies): only use feature "xcb" if using X11
+cairo-rs = { version = "0.9.1", default_features = false, features = ["xcb"], optional = true }
```
cmyr, line 53: we currently allow cairo to be used, opt-in, on macOS. I'm not sure we should, but we do; I think this is why these deps are in two places.

cmyr (commented):
This should be a nice improvement for x11 users, thanks for digging in. I have a few little concerns inline, happy to merge this when we've figured those things out.

druid-shell/Cargo.toml
Finnerale: I also prefer just calling it `gtk`, but Cargo does not agree with us :(

(It will error about two features with the same name)

druid-shell/Cargo.toml
Finnerale: Huh I did not know that. I'll move Cairo back to the "normal" dependencies then.

druid-shell/Cargo.toml
ratmice: what about `platform-gtk`?

cmyr: Okay, what I propose:

in `druid-shell/Cargo.toml`, rename the *actual* gtk crate to `gtk-rs`, and then call the *feature* `gtk`:

```toml
[features]
x11 = ["x11rb", "nix", "cairo-sys-rs", "cairo-rs"]
gtk = ["cairo-rs", "gio", "gdk", "gdk-sys", "glib", "glib-sys", "gtk-sys", "gtk-rs"]
default = ["gtk"]

[target.'cfg(target_os="linux")'.dependencies]
# other deps here
gtk-rs = { version = "0.9.2", features = ["v3_22"], package = "gtk", optional = true }
```

druid-shell/Cargo.toml
Finnerale: This proved to be a bit tricky, but it should work now.

druid-shell/Cargo.toml
Finnerale: I think that did the job :+1:

cmyr (approved):
This looks good. I think we should actually remove the cairo-on-mac stuff, it isn't even wired up to druid anymore (it only works for druid-shell) and it just isn't worth the ongoing maintenance cost; I'll open an issue for that though, it's definitely a separate patch.

Thanks!

PR #1244, opened 2020-09-18
Set wrap width before rebuilding label

### LINKED ISSUE #1243: Blank Label with word wrap post #1239
After upgrading to and past 23cc3a48983078f6212125130b56f72ee67de41b enabling word wrap results in the label not displaying any text. I've tested that it works as expected with ebcbe8a8e8b82d5d2463011501463bc915477070 but not 23cc3a48983078f6212125130b56f72ee67de41b. I am running the GTK shell and by extension the Cairo Piet backend

image

cc @cmyr

cmyr:
- fixes #1243


ForLoveOfCats (commented):
This fixes the issue initially but the text becomes blank upon window resize

cmyr:
oops, that should be working as well now. :)

ForLoveOfCats (approved):
Seems to fully resolve the issue!

PR #1245, opened 2020-09-18
Use ArcStr in LocalizedString and LabelText

cmyr:
This lets us share our text between druid and the piet layout
object, and starts the process of moving us away from using String
where possible.

This was motivated by explorations around making Label optionally work
as a `Widget<ArcStr>`, and using lens-like adapters to generate
that text from some other data types as needed; this would be a small
improvement of the label API.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@ You can find its changes [documented below](#060---2020-06-01).
 - `TextLayout` type simplifies drawing text ([#1182] by [@cmyr])
 - Implementation of `Data` trait for `i128` and `u128` primitive data types. ([#1214] by [@koutoftimer])
 - `LineBreaking` enum allows configuration of label line-breaking ([#1195] by [@cmyr])
-- `TextAlignment` support in `TextLayout` and `Label` ([#1210] by [@cmyr])`
+- `TextAlignment` support in `TextLayout` and `Label` ([#1210] by [@cmyr])
 - `UpdateCtx` gets `env_changed` and `env_key_changed` methods ([#1207] by [@cmyr])
 - `Button::from_label` to construct a `Button` with a provided `Label`. ([#1226] by [@ForLoveOfCats])
 - Lens: Added Unit lens for type erased / display only widgets that do not need data. ([#1232] by [@rjwittams])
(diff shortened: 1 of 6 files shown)


druid/src/widget/label.rs
```diff
@@ -17,25 +17,36 @@
...
+    Static {
+        /// The text.
+        string: ArcStr,
+        /// Whether or not the `resolved` method has been called yet.
+        ///
+        /// We want to return `true` from that method when it is first called,
+        /// so that callers will know to retrieve the text. This matches
+        /// the behaviour of the other variants.
+        resolved: bool,
+    },
```
Finnerale, line 49: Maybe move this into its own `Static` struct the way we did with `Dynamic`? Then we could hide the `resolved` field.

druid/src/widget/label.rs
```diff
@@ -45,13 +56,13 @@ pub enum LabelText<T> {
 #[doc(hidden)]
 pub struct Dynamic<T> {
```
Finnerale, line 57: Not really related to this PR, but I think it is a bit irritating that `Dynamic` is `doc(hidden)`. It makes the `LabelText` documentation slightly weird, because `Dynamic` looks like a broken link.

druid/src/widget/label.rs
```diff
@@ -310,8 +316,8 @@ impl<T: Data> Label<T> {
 impl<T> Dynamic<T> {
     fn resolve(&mut self, data: &T, env: &Env) -> bool {
         let new = (self.f)(data, env);
-        let changed = new != self.resolved;
-        self.resolved = new;
+        let changed = new.as_str() != &*self.resolved;
```
Finnerale, line 319: I usually prefer `as_ref` over `&*`
```suggestion
        let changed = new.as_str() != self.resolved.as_ref();
```

CHANGELOG.md
```diff
@@ -57,7 +57,7 @@ You can find its changes [documented below](#060---2020-06-01).
 - Moved `Target` parameter from `submit_command` to `Command::new` and `Command::to`. ([#1185] by [@finnerale])
 - `Movement::RightOfLine` to `Movement::NextLineBreak`, and `Movement::LeftOfLine` to `Movement::PrecedingLineBreak`. ([#1092] by [@sysint64])
 - `AnimFrame` was moved from `lifecycle` to `event` ([#1155] by [@jneem])
-- Contexts' `text()` methods return `&mut PietText` instead of cloning ([#1205] by [@cmyr])
+- `LocalizedString` and `LabelText` use `ArcStr` instead of String ([#1245] by [@cmyr])
```
Finnerale, line 60: This diff looks weird… Did you delete the `#1205` entry intentionally?

Finnerale (approved):
A few nits, but seems like a good change.

druid/src/widget/label.rs
cmyr: ah totally if `AsRef` is available, I thought this was only available via deref. :)

CHANGELOG.md
cmyr: good catch!

PR #1248, opened 2020-09-20
Another try at a viewport widget.

jneem:
Here's another stab at a viewport widget. This one differs from the previous one in that it doesn't make `ScrollComponent` depend on the viewport widget. It does make other changes to `ScrollComponent`, though. Specifically, it separates the "viewport state" (which I've called `RectPort`) from the scrollbar state. The scroll component no longer stores the viewport state; in methods that need to modify the viewport state, it gets a mutable reference. The reason for this was that I wanted to make these modifications explicit for the caller, because the caller will typically need to do something in response to the viewport changing.


ForLoveOfCats:
I see a few terminology things I'm not fond of and I can't do an in-depth review at this moment. However at first glance this seems to be a reasonable compromise!

jneem:
If you have better naming suggestions, I'd be happy to hear them! Like, I know that `RectPort` is a bad name, but `Viewport` was already taken...

I'm also in no rush to merge.

derekdreery:
Just saying thanks for all the work on a big PR.

druid/src/widget/viewport.rs
```diff
@@ -0,0 +1,259 @@
...
+        Point::new(x, y)
+    }
+
+    /// Changes the viewport offset by `delta`.
+    ///
+    /// Returns true if the offset actually changed. Even if `delta` is non-zero, the offset might
+    /// not change. For example, if you try to move the viewport down but it is already at the
+    /// bottom of the child widget, then the offset will not change and this function will return
+    /// false.
+    pub fn scroll_by(&mut self, delta: Vec2) -> bool {
```
ForLoveOfCats, line 50: This is confusing terminology as Viewport is not only for Scroll, perhaps `offset_by`?

druid/src/widget/viewport.rs
```diff
@@ -0,0 +1,259 @@
...
+        self.scroll_to(self.rect.origin() + delta)
+    }
+
+    /// Sets the viewport origin to `pos`.
+    ///
+    /// Returns true if the position changed. Note that the valid values for the viewport origin
+    /// are constrained by the size of the child, and so the origin might not get set to exactly
+    /// `pos`.
+    pub fn scroll_to(&mut self, origin: Point) -> bool {
```
ForLoveOfCats, line 59: Similar to above, perhaps `offset_to` or `pan_to`?

druid/src/widget/viewport.rs
```diff
@@ -0,0 +1,259 @@
...
+    }
+}
+
+/// A widget exposing a rectangular view into its child, which can be used as a building block for
+/// widgets that scroll their child.
+pub struct Viewport<T, W> {
+    child: WidgetPod<T, W>,
+    port: RectPort,
+    constrain_horizontal: bool,
+    constrain_vertical: bool,
```
ForLoveOfCats, line 76: Minor nit, these could be an enum

druid/src/widget/viewport.rs
```diff
@@ -0,0 +1,259 @@
...
+        self.constrain_vertical = constrain;
+    }
+
+    /// Changes the viewport offset by `delta`.
+    ///
+    /// Returns true if the offset actually changed. Even if `delta` is non-zero, the offset might
+    /// not change. For example, if you try to move the viewport down but it is already at the
+    /// bottom of the child widget, then the offset will not change and this function will return
+    /// false.
+    pub fn scroll_by(&mut self, delta: Vec2) -> bool {
```
ForLoveOfCats, line 167: See comment on `RectPort`

druid/src/widget/viewport.rs
```diff
@@ -0,0 +1,259 @@
...
+        self.scroll_to(self.viewport_origin() + delta)
+    }
+
+    /// Sets the viewport origin to `pos`.
+    ///
+    /// Returns true if the position changed. Note that the valid values for the viewport origin
+    /// are constrained by the size of the child, and so the origin might not get set to exactly
+    /// `pos`.
+    pub fn scroll_to(&mut self, origin: Point) -> bool {
```
ForLoveOfCats, line 176: See comment on `RectPort`

druid/src/widget/scroll.rs
```diff
@@ -42,7 +41,7 @@ enum ScrollDirection {
 /// [`vertical`]: struct.Scroll.html#method.vertical
 /// [`horizontal`]: struct.Scroll.html#method.horizontal
 pub struct Scroll<T, W> {
-    child: WidgetPod<T, W>,
+    viewport: Viewport<T, W>,
     scroll_component: ScrollComponent,
     direction: ScrollDirection,
```
ForLoveOfCats, line 46: `Scroll::direction` is no longer used

druid/src/scroll_component.rs
```diff
@@ -475,24 +444,4 @@ impl ScrollComponent {
             self.reset_scrollbar_fade(|d| ctx.request_timer(d), &env);
         }
     }
-
-    /// Helper function to paint a closure at the correct offset with clipping and scrollbars
-    pub fn paint_content(
```
ForLoveOfCats, line 480: This function has value for scroll-able widgets which are not containers for children so I'm inclined to keep it even though Scroll no longer uses it

ForLoveOfCats (commented):
I'm sorry I can't give a very thorough review at this moment but I wanted to give at least a bit of feedback. I don't have anything against the name of `RectPort`, what I was referring to are the references to scrolling in `Viewport` which doesn't line up with the potential usage cases.

druid/src/scroll_component.rs
rjwittams: That seems to be a different standard for inclusion than everything else ( might be useful for some people but not used here)? I guess if there was an example that used it... (not that I want to impose work on anyone)

druid/src/widget/viewport.rs
rjwittams: I think the word scroll does indicate to most people familiar with gui terminology any movement around a larger area, not just that widget. I guess another word would be panning but not sure if that is just changing the name for the sake of it.

druid/src/widget/viewport.rs
rjwittams: Snap!

druid/src/widget/viewport.rs
```diff
@@ -0,0 +1,259 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::kurbo::{Affine, Point, Rect, Size, Vec2};
+use crate::widget::prelude::*;
+use crate::{Data, WidgetPod};
+
+/// Represents the size and position of a rectangular "viewport" into a larger area.
+#[derive(Clone, Copy, Default, Debug, PartialEq)]
+pub struct RectPort {
```
rjwittams, line 21: I feel like there must be a better name than this but I can't think of one right now...

druid/src/scroll_component.rs
ForLoveOfCats: Yeah honestly I'm not *very* strongly opposed to removing this as it's honestly not much code

druid/src/widget/viewport.rs
cmyr: We could call *this* `Viewport` and call the widget `ClipContainer` or something? just spitballing...

ForLoveOfCats: Viewport is rapidly becoming a very overloaded term. In some cases it can be the size of the window, in other situations it refers to the paint rect of a widget, and now we are adding two new things which each reference related topics with similar terminology. I don't have much against `RectPort` but I do like renaming the widget to `ClipContainer` or something else.

druid/src/widget/viewport.rs
jneem: I'm going to try `offset`. There's no breakage since it's a new widget...

jneem: I decided I liked `pan` better :shrug:

druid/src/widget/viewport.rs
jneem: The logic turns out nicer with the bools, actually, since I need to treat them independently. Ok, the enum could save a byte but I'm not sure that's important...

druid/src/scroll_component.rs
jneem: :shrug: no strong opinion from me, but since I already removed it I'm going to be lazy and leave it removed for now. Can always put it back in later.

druid/src/widget/viewport.rs
jneem: Hmm, `ClipContainer` is missing half the story for me, since I see this widget as being about both clipping and panning. I'm going to try sticking with `Viewport` for now and renaming `RectPort` to `ViewportPos`...

druid/src/scroll_component.rs
ForLoveOfCats: Fair enough!

druid/src/widget/viewport.rs
ForLoveOfCats: This is kinda a case of the five monkeys experiment, when I implemented a similar thing ages ago I was asked to use an enum. I'm fine with it as it is

jneem:
@ForLoveOfCats Did I imagine it or did you leave some comments on some of the API docs? I could have sworn you did but I can't find it now, either here or in the old PR.

ForLoveOfCats:
Nope, I've barely skimmed the updated doc comments

druid/src/widget/viewport.rs
```diff
@@ -0,0 +1,259 @@
...
+use crate::widget::prelude::*;
+use crate::{Data, WidgetPod};
+
+/// Represents the size and position of a rectangular "viewport" into a larger area.
+#[derive(Clone, Copy, Default, Debug, PartialEq)]
+pub struct ViewportPos {
+    /// The size of the area that we have a viewport into.
+    pub content_size: Size,
+    /// The view rectangle. This should be contained in `content_size.to_rect()`.
+    pub rect: Rect,
```
cmyr, line 25: What about the case of a scroll + zoom view that lets you decrease the zoom level so that the content is smaller than the allowed size? This is actually a case I encounter in runebender, and is a common pattern in things like image viewers?

It may be that I can handle that at another level of the stack, but wanted to raise the question; it's a case I'd like to support.

jneem: Good point. I think it definitely makes sense for `ViewportPos` to support this. For `Viewport`, it probably raises some layout issues that I'd like to punt on for now.

druid/src/widget/viewport.rs
cmyr: Can you clarify how the distinction between clipping and panning? I think of panning as just a particular subjective experience of something being clipped. 🤔

jneem: I don't really have a *very* strong opinion on this, but when I hear "clip" I just think of the graphical operation. Like, even without a transformation, a `ClipContainer` might be able to clip a child so that it has rounded corners or something? That does admittedly sound pretty contrived...

druid/src/widget/viewport.rs
jneem: This seems to be mostly a documentation issue, since I'm neither enforcing nor relying on the view being smaller than the content. The panning methods are intended for that use case, so I've tried to make that clear in their docstrings.

druid/src/widget/viewport.rs
cmyr: Cool, just curious. I think it's fairly common for UI toolkits to call this sort of thing a 'clip view'; at least, that's what it's [called in AppKit](https://developer.apple.com/documentation/appkit/nsclipview).

ForLoveOfCats:
I've poked at this some more and while I feel good about this PR on the whole I do have a few concerns which I've voiced a bit earlier inline but I want to cover a bit more. However I don't want to block on these concerns alone.

I'm worried about the term "viewport". At this point we have quite a few meanings for the word and this PR adds another new meaning to it. Ironically in several locations in the impl for the `Viewport` widget it constructs a local rect variable called `viewport`. Furthermore the `ViewportPos` feels non-descriptive as it does not directly contain a position value (the closest is `ViewportPos::rect.origin()`) which leads to `Viewport` having the `.pos(&self)` method which does not return an actual position. Again I don't want to explicitly block on any of this. Naming things is hard and I don't really have any ready suggestions at this exact moment. My worry is the increased mental load of keeping track of these things and the non-specificity which the term viewport is heading towards.

cmyr:
I agree that the names don't feel perfect. The only real suggestion I have is `ClipBox` for the widget, which gives us `Viewport` or the thing describing the visible region. 🤷

ruffson:
I am really interested in using this. Are there any blockers?

jneem:
No, I've just been busy on other things, sorry. I'll try changing the names to `ClipBox`/`Viewport` and see if that flies.

ForLoveOfCats (approved):
I feel more comfortable with the data structure being called `Viewport` over the widget. I do still have concerns as we generally refer to the paint rect as the "viewport" leading to code which touches two separate values with the name viewport but containing completely different data and different purposes mere lines apart. That said I do feel like clarifying our terminology in this area can be a separate conversation in the same vein as the discussions on resolution terminology.

PR #1251, opened 2020-09-23
Move LensWrap to widget module

cmyr:
This is some maintenance I was inspired to do in the course of exploring a particularly deep rabbit hole 😒


raphlinus (approved):
Sure, that bothered me slightly too, but my activation energy for such moves is higher :)

PR #1252, opened 2020-09-23
Separate Label and RawLabel

cmyr:
(based off of #1245)

This makes label work more like other widgets; the RawLabel
widget is a thing that has a text data type (currently ArcStr)
and displays this text.

The existing Label widget is rewritten as a wrapper around
RawLabel; it is basically a sort of specialized, env-aware
LensWrap.


Finnerale (approved):
It's a bit annoying that we have to duplicate the builder methods, but I suppose there is now good way around it.

PR #1256, opened 2020-09-25
Alerts and their macOS implementation, take 2

cmyr:
This supersedes #1039, and brings it up to date with current druid.


cmyr:
My personal take on this, as-written: I think it's good, but I think that we will ultimately want to have a wrapper api in druid that does things like allow for localization and perhaps allows the caller to associate `Command`s with different alert actions. My general inclination would be to merge as-is and then worry about that stuff down the road, though, given that localization isn't really working yet anyway.

jneem:
That sounds reasonable to me. I will play with this over the weekend, but it might take a little while because I'll have to do a gtk version first...

druid-shell/src/platform/mac/window.rs
```diff
@@ -133,14 +137,18 @@ enum IdleKind {
...
     nsview: WeakPtr,
-    handler: Box<dyn WinHandler>,
+    pub(crate) handler: Box<dyn WinHandler>,
     idle_queue: Arc<Mutex<Vec<IdleKind>>>,
     /// Tracks window focusing left clicks
     focus_click: bool,
     // Tracks whether we have already received the mouseExited event
     mouse_left: bool,
+    /// Pending alert requests
+    alert_queue: RefCell<VecDeque<AlertRequest>>,
```
jneem, line 149: Is it correct that this is a queue? I'd have expected a stack (i.e. last alert fired is the first one to be dismissed)

jneem:
Ok, I have a more-or-less working GTK version [here](https://github.com/jneem/druid/tree/alerts-gtk). The main issue I had is that the way the buttons are specified (primary, cancel, alternatives) doesn't map particularly cleanly to the GNOME HIG: there, it is optional whether to specify a default/primary action (because if the only non-cancel button does something destructive and non-reversible then it shouldn't be the default).

I still need to look at the druid side of the patch...

jneem:
I'm not wild about how the `AlertResponse` is plumbed through druid -- I think it would be cleaner to use commands instead (and more flexible, because the widget that requests the alert can choose where to target the command). But if you want to get this off your plate, I'm also just happy to have it merged as is and I can make some PRs later.

jneem (approved):

druid-shell/src/platform/mac/window.rs
xStrom: Alert dismissing order can be arbitrary.

xStrom:
I discussed it with @cmyr and am moving this work back to #1039 - sorry for the ping-pong everyone.

PR #1259, opened 2020-09-27
Guarantee that WidgetAdded is the first thing a widget sees.

### LINKED ISSUE #1127: Ensure that widgets receive LifeCycle::WidgetAdded as first "event".
We currently have the rule, that widgets can rely on `LifeCycle::WidgetAdded` being the first "event" (/update/layout/…) a widget receives. Yet we take no measures to ensure this is the case,  and I would like to change that.

My first approach to this was preventing calls to `update`/`event/etc. by doing a "debug panic", i.e. only panic in debug mode, but log a message when running in release mode.

After implementing that I realized that it might be possible to just initialize any widget during those events if it has not been initialized yet, and it turned out that this would be possible as well.

But now I'm unsure which one would be the better solution. The first one allows easy pinpointing of where you did something wrong, while the second basically ignores it and just fixes your mess. The letter however would not fix the whole problem, because if you forgot to call `children_changed` (which is the main cause for a missing `WidgetAdded`), you also end up with a missing `request_layout`, which can not be solved lazily.

I think option one (panicking) is better, because it requires you to take proper care of it. The latter is mostly nice for prototyping I guess, but unless it makes `children_changed` completely redundant it would make things more difficult to debug instead.

### LINKED ISSUE #1258: Lifecycle issues on dynamic widget change
This is a crash when using master Druid with the Crochet prototype. The lifecycle problems were happening before, but only giving warnings. Now text layout is highly dependent on getting lifecycle right. Though it's observed in Crochet, I'm pretty sure it's a problem in Druid, and will probably bite us any time we're trying to reconfigure children dynamically.

The following println-debugging trace should give a clue what's going wrong. It's from a simplified version of the "counter" example, stripped down to just be a label. I can post a gist of the example, and a diff adding the println, if it'd be useful.

```
anywidget lifecycle WidgetAdded
anywidget event WindowConnected
children changed
anywidget layout
anywidget layout
label layout
anywidget lifecycle Size(4.0W├ù16.0H)
label lifecycle Size(4.0W├ù16.0H)
anywidget lifecycle Size(1427.0W├ù726.0H)
anywidget lifecycle Internal(RouteWidgetAdded)
anywidget lifecycle WidgetAdded
label lifecycle WidgetAdded
anywidget paint
anywidget paint
label paint
thread 'main' panicked at 'TextLayout::draw called without rebuilding layout object. Text was 'Foo'', C:\Users\raph\github\druid\druid\src\text\layout.rs:291:9
```

The "children changed" log is when the child widget (a label) gets added to the (modified) flex container; it's [this line](https://github.com/raphlinus/crochet/blob/5043f3283429e5ff0c511b2718523a25f9156ad2/src/widget/flex.rs#L588) in particular.

The underlying problem seems pretty simple: it's doing the WidgetAdded lifecycle *after* the layout, which seems pretty clearly wrong to me. I think this problem may be hidden in other examples because there might be an intervening `update` or something, but I also wouldn't be surprised if it were possible to trigger in extant code even outside Crochet.

I haven't investigated the lifecycle ordering logic carefully yet. It feels like a bit of a haunted graveyard. But I am willing to dig in.

(PS: yes, the × symbol is mojibake. Because it's 2020 and all software is broken)

Finnerale:
Closes #1127 and hopefully closes #1258 as well.

With this, `WidgetPod` will panic whenever a widget receives anything but `LifeCycle::WidgetAdded` as its first event.

The advantage of a panic is that the backtrace makes it in my experience very easy to track down where things went wrong.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@ You can find its changes [documented below](#060---2020-06-01).
 - `Checkbox::set_text` to update the label. ([#1346] by [@finnerale])
 - `Event::should_propagate_to_hidden` and `Lifecycle::should_propagate_to_hidden` to determine whether an event should be sent to hidden widgets (e.g. in `Tabs` or `Either`). ([#1351] by [@andrewhickman])
 - `set_cursor` can be called in the `update` method. ([#1361] by [@jneem])
+- `WidgetPod::is_initialized` to check if a widget has received `WidgetAdded`. ([#1259] by [@finnerale])
 
 ### Changed
 
@@ -118,6 +119,7 @@ You can find its changes [documented below](#060---2020-06-01).
 - Improve Windows 7 DXGI compatibility ([#1311] by [@raphlinus])
 - Fixed `Either` not passing events to its hidden child correctly. ([#1351] by [@andrewhickman])
 - Don't drop events while showing file dialogs ([#1302], [#1328] by [@jneem])
+- Ensure that `LifeCycle::WidgetAdded` is the first thing a widget sees. ([#1259] by [@finnerale])
 
 ### Visual
 
@@ -138,6 +140,7 @@ You can find its changes [documented below](#060---2020-06-01).
 
 - Standardized web targeting terminology. ([#1013] by [@xStrom])
 - X11: Ported the X11 backend to [`x11rb`](https://github.com/psychon/x11rb). ([#1025] by [@jneem])
+- Add `debug_panic` macro for when a backtrace is useful but a panic unnecessary. ([#1259] by [@finnerale])
 
 ### Outside News
 
@@ -524,6 +527,7 @@ Last release without a changelog :(
 [#1328]: https://github.com/linebender/druid/pull/1328
 [#1346]: https://github.com/linebender/druid/pull/1346
 [#1351]: https://github.com/linebender/druid/pull/1351
+[#1259]: https://github.com/linebender/druid/pull/1259
 [#1361]: https://github.com/linebender/druid/pull/1361
 
 [Unreleased]: https://github.com/linebender/druid/compare/v0.6.0...master
--- a/druid/src/core.rs
+++ b/druid/src/core.rs
@@ -161,6 +161,13 @@ impl<T, W: Widget<T>> WidgetPod<T, W> {
         &self.state
     }
 
+    /// Returns `true` if the widget has received [`LifeCycle::WidgetAdded`].
+    ///
+    /// [`LifeCycle::WidgetAdded`]: ./enum.LifeCycle.html#variant.WidgetAdded
+    pub fn is_initialized(&self) -> bool {
+        self.old_data.is_some()
+    }
+
     /// Query the "active" state of the widget.
     pub fn is_active(&self) -> bool {
         self.state.is_active
@@ -423,6 +430,14 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
             return;
         }
 
+        if !self.is_initialized() {
+            debug_panic!(
+                "{:?}: paint method called before receiving WidgetAdded.",
+                ctx.widget_id()
+            );
+            return;
+        }
+
         ctx.with_save(|ctx| {
             let layout_origin = self.layout_rect().origin().to_vec2();
             ctx.transform(Affine::translate(layout_origin));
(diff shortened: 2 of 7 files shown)


druid/src/core.rs
```diff
@@ -527,11 +550,14 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
                 Event::Timer(_) => (),
                 Event::WindowConnected => (),
                 Event::WindowSize(_) => (),
+                // Should commands require layout?
+                // If yes, it will cause problems with our test suite.
+                Event::Command(_) => (),
                 _ => {
-                    log::warn!(
-                        "Widget '{}' received an event ({:?}) without having been laid out. \
```
Finnerale, line 532: I had to exclude `Event::Command` here, because otherwise our tests would break.
There we often send test commands without doing any layout.

This can of course be changed back, but I wanted to push this now without meddling too much with our test stuff.

Finnerale:
@raphlinus does this help you with #1258?

druid/src/core.rs
jneem: This came up in #1160 recently, because unviewed tabs don't get laid out but they might still receive commands. I think there's a bigger decision that needs to be made (and documented) about which events and lifecycle events hidden widgets should expect to receive, but for now I think it's completely reasonable for them to get commands.

druid/src/core.rs
```diff
@@ -411,6 +418,14 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
             return;
         }
 
+        if !self.is_initialized() {
+            debug_panic!(
+                "{:?} is supposed to paint without having first received WidgetAdded.",
```
cmyr, line 423: I find this wording slightly awkward? might just say something like, paint method called before WidgetAdded."

druid/src/core.rs
```diff
@@ -411,6 +418,14 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
             return;
         }
 
+        if !self.is_initialized() {
+            debug_panic!(
```
cmyr, line 422: Although I think `debug_panic!` is an excellent addition, I wonder if this shouldn't actually be a `debug_assert!`?

druid/src/core.rs
cmyr: I agree, it's even totally reasonable for a `Command` to be sent while handling `Event::WindowConnected`, and in that case layout should *not* have been called.

druid/src/util.rs
```diff
@@ -18,6 +18,31 @@ use std::collections::HashMap;
 use std::hash::Hash;
 use std::mem;
 
+/// Panic in debug and log::error in release mode.
+///
+/// This macro is in some way a combination of `panic` and `debug_assert`,
+/// but it will log the provided message instead of ignoring it in release builds.
+///
+/// It's useful when a backtrace would aid debugging but a crash can be avoided in release.
+macro_rules! debug_panic {
```
cmyr, line 27: this is an excellent addition.

cmyr (approved):
A few notes but this looks like it should be quite helpful with debugging some of this subtler lifecycle stuff.

druid/src/core.rs
Finnerale: `debug_assert!` does nothing in release, while `debug_panic!` at least logs an error. Thus, I would prefer `debug_panic`.

cmyr: makes sense!

druid/src/core.rs
rjwittams: There are methods in the tabs code that try to say whether an event or lifecycle should be delivered to hidden widgets. Possibly should be moved to the enums themselves.

cmyr:
Just checking in, is this ready to go? If not let us know what you need.

Finnerale:
I'm waiting for @raphlinus to see if this helps with #1258

cmyr:
What's the status here? does it still hopelessly break crochet?

Finnerale:
No, Crochet would work just fine with it, since I fixed the context handling. This is just waiting for someone to say yay or nay.

CHANGELOG.md
```diff
@@ -109,6 +110,7 @@ You can find its changes [documented below](#060---2020-06-01).
 - Multi-click on Windows, partial fix for #859 ([#1157] by [@raphlinus])
 - Windows: fix crash on resize from incompatible resources ([#1191 by [@raphlinus]])
 - GTK: Related dependencies are now optional, facilitating a pure X11 build. ([#1241] by [@finnerale])
+<<<<<<< HEAD
```
cmyr, line 113: oops

cmyr (approved):
Okay, my inclination is to merge this, I think it may uncover some problems but ultimately that will be good for us? Worse case scenario we can always revert.

Finnerale:
Ok. Once the checks are done, I'll merge it.

PR #1260, opened 2020-09-27
Fix failure to update dynamic label

cmyr:
We were failing to set the text in the update case; this
was a consequence of #1252


druid/src/widget/label.rs
```diff
@@ -477,6 +477,7 @@ impl<T: Data> Widget<T> for Label<T> {
         self.text_should_be_updated = false;
         if data_changed {
             let new_text = self.text.display_text();
+            dbg!(&new_text);
```
Finnerale, line 480: I don't think this `dbg!` is intentional?

cmyr: 😅

Finnerale (approved):

PR #1261, opened 2020-09-27
Fix for flashing on window creation

rhzk:
Small PR to fix the flashing issue left over from #1037

This should also fix any flashing that might happen when creating a window minimized or maximized.

Also updated setting position and size to use Into

### DIFF
--- a/druid-shell/src/window.rs
+++ b/druid-shell/src/window.rs
@@ -178,8 +178,8 @@ impl WindowHandle {
     /// [`position`] The position in pixels.
     ///
     /// [`position`]: struct.Point.html
-    pub fn set_position(&self, position: Point) {
-        self.0.set_position(position)
+    pub fn set_position(&self, position: impl Into<Point>) {
+        self.0.set_position(position.into())
     }
 
     /// Returns the position in virtual screen coordinates.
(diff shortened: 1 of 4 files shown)


druid-shell/src/window.rs
```diff
@@ -141,7 +141,7 @@ impl WindowHandle {
     /// Sets the state of the window.
     ///
     /// [`state`]: enum.WindowState.html
-    pub fn set_window_state(&mut self, state: WindowState) {
+    pub fn set_window_state(&self, state: WindowState) {
```
rjwittams, line 144: It seems odd to do this as a backend may not use interior mutability (especially if we had a completely virtual backend for tests etc).
Does it work on every platform?

rhzk: Hmm, having mutable here made the borrow checker complain

rhzk: I will revert this, maybe something to look into in the future when all backends are supported

rjwittams: Anything set_* not taking &mut self or self seems suspect to me... (even if all current back ends do use interior mutability, it logically makes little sense)

rhzk:
There is some edge case that this solution cant handle, I will need to find another solution..

PR #1263, opened 2020-09-27
[ci] Run GTK tests as separate job

cmyr:
The linux builds are by far the slowest part of CI, and we
should see *some* modest gain by splitting them in two.


cmyr:
Looks like this gives us a reasonable speedup; testing wasm is now the bottleneck.

Finnerale (approved):
I've only skimmed the `ci.yml` but it looks good.

PR #1266, opened 2020-09-28
env::KeyOrValue stores raw T, not Value

cmyr:
This is slightly cleaner, and allows interacting with the `Concrete`
item directly without needing to downcast from `Value`.

Split this out as a separate commit because I want to run bloatcmp.


cmyr:
Cool, I was curious about this; `Value` is taking up a bunch of space in various places, likely because of the size of `FontDescriptor`.

rjwittams (approved):
Looks sensible to me

PR #1267, opened 2020-09-28
Store placeholder text as a separate TextLayout

cmyr:
This lets us remove a bunch of logic for setting the placeholder
text in the layout, changing the color, etc.

I also updated the `hello` example to use non-default fonts and sizes, so I could check that this was working as expected.

### DIFF
--- a/druid/src/widget/textbox.rs
+++ b/druid/src/widget/textbox.rs
@@ -40,7 +40,10 @@ const CURSOR_BLINK_DURATION: Duration = Duration::from_millis(500);
 /// A widget that allows user text input.
 #[derive(Debug, Clone)]
 pub struct TextBox {
-    placeholder: String,
+    // this has a different inner type just to keep us honest; in a future
+    // patch this will have to be the case, which means we can't use the two layouts
+    // interchangeably.
+    placeholder: TextLayout<String>,
     text: TextLayout<ArcStr>,
     width: f64,
     hscroll_offset: f64,
@@ -57,6 +60,8 @@ impl TextBox {
     /// Create a new TextBox widget
     pub fn new() -> TextBox {
         let text = TextLayout::new();
+        let mut placeholder = TextLayout::from_text("");
+        placeholder.set_text_color(theme::PLACEHOLDER_COLOR);
         Self {
             width: 0.0,
             hscroll_offset: 0.,
@@ -64,13 +69,13 @@ impl TextBox {
             selection: Selection::caret(0),
             cursor_timer: TimerToken::INVALID,
             cursor_on: false,
-            placeholder: String::new(),
+            placeholder,
         }
     }
 
     /// Builder-style method to set the `TextBox`'s placeholder text.
     pub fn with_placeholder(mut self, placeholder: impl Into<String>) -> Self {
-        self.placeholder = placeholder.into();
+        self.placeholder.set_text(placeholder.into());
         self
     }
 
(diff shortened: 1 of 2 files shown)


druid/src/widget/textbox.rs
```diff
@@ -435,13 +435,7 @@ impl Widget<String> for TextBox {
...
     fn paint(&mut self, ctx: &mut PaintCtx, data: &String, env: &Env) {
         // Guard against changes in data following `event`
-        let content = if data.is_empty() {
-            &self.placeholder
-        } else {
-            data
-        };
-
-        self.selection = self.selection.constrain_to(content);
+        self.selection = self.selection.constrain_to(data);
```
rjwittams, line 438: I think this means that if you have a selection, then your text gets disappeared by an update, it will wind up not drawing any selection as the selection will also disappear. It might be clearer to explicitly not do any of the selection stuff if we are using the placeholder - not sure if the same applies for the cursor?

cmyr: this is a questionable line, and maybe it should be removed? As I understand it it's an attempt to catch some problem with an invalid selection that was not correctly handled during update.

The problem is that the selection is stored in the widget, but it references the app data, so when the app data is changed in another widget, the selection can end up invalid. This *should* be handled in `update`, but maybe there was some problem?

rjwittams: Yes it does seem a bit odd

jneem: Maybe this could be an assert instead of trying to silently fix it? It seems to me like a serious bug in druid if we get a paint call with different data from our last update...

Zarenor: I'm not quite sure how we could land in paint without having hit update to constrain the selection either. That definitely sounds like an odd case. Do we want to drop this line and see if we have crashes again, or would we rather leave this as a backstop? I don't see how this could be terribly expensive.

raphlinus: Feels to me this line should be removed, just on the basis that paint really should be as close to a pure function as possible. If there is a user-visible problem here (and I'm not sure there is, but haven't done detailed code analysis or experimentation), it should probably be handled in update.

Zarenor (approved):
This looks good to me. Fairly straight forward changes, and I think it's a significant improvement in readability in TextBox's code to boot.

druid/src/widget/textbox.rs
```diff
@@ -360,71 +369,59 @@ impl Widget<String> for TextBox {
...
+            self.selection = self.selection.constrain_to(data);
+            self.text.set_text(data.as_str().into());
+            ctx.request_layout();
         }
 
-        self.text.rebuild_if_needed(ctx.text(), env);
-        ctx.request_paint();
+        if ctx.env_changed()
+            && self.placeholder.needs_rebuild_after_update(ctx)
+                | self.text.needs_rebuild_after_update(ctx)
```
raphlinus, line 401: Just curious why single `|` here? It feels to me like `||` is more idiomatic to indicate "or".

druid/src/widget/textbox.rs
cmyr: I'm in the process of rewriting this stuff, and it will be obsoleted soon. My inclination is to just leave it as is for now?

raphlinus (approved):
Generally looks good, I'd be happier with the mutation in `paint` removed. Thanks!

PR #1271, opened 2020-09-30
Updated broken links to a new ones

### LINKED ISSUE #1268: Broken link in README
The ["Linux platform notes section of the README"](https://github.com/linebender/druid#linux) links to [this dead page](http://gtk-rs.org/docs/requirements.html). Link text is "gtk-rs dependencies".

I looked around the gtk-rs website a bit but couldn't the correct link. Upstream README also has bad link: https://github.com/gtk-rs/gtk/issues/1063

Dmitry-Borodin:
Fixes #1268 

Looks like this link from upstream is the best one, despite they are suggesting to use latest versions of dependencies, not ones from your distro. I think our users will see that and we don't need to clarify.

I'm not sure whether it make sense to add this pure documentation PR to a changelog, I think this is not what users want to see when they check changes for a new version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,8 +90,7 @@ druid = { git = "https://github.com/linebender/druid.git" }
 
 #### Linux
 
-On Linux, Druid requires gtk+3; see [gtk-rs dependencies] for installation
-instructions.
+On Linux, Druid requires gtk+3; see [GTK installation page].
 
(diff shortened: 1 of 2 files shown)


Finnerale (approved):
There is indeed no need for a changlog entry :+1: 
Thanks for fixing this!

cmyr:
Much appreciated!

PR #1272, opened 2020-10-02
Reduce flashing on window creation

rhzk:
Sets the correct window size directly after the window is built instead of adding it to the deferred queue.
Since this is done before ShowWindow is called it should not be any visible "flashing" occuring.
Updated Show() to handle maximized and minimized window, as this was also deferred and might have caused flashing.

### DIFF
--- a/druid-shell/src/platform/windows/window.rs
+++ b/druid-shell/src/platform/windows/window.rs
@@ -84,7 +84,7 @@ pub(crate) struct WindowBuilder {
     show_titlebar: bool,
     size: Option<Size>,
     min_size: Option<Size>,
-    position: Point,
+    position: Option<Point>,
     state: window::WindowState,
 }
 
@@ -582,7 +582,6 @@ impl WndProc for MyWndProc {
                 } else {
                     1.0
                 };
-
                 let scale = Scale::new(scale_factor, scale_factor);
                 self.set_scale(scale);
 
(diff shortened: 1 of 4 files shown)


druid-shell/src/platform/windows/window.rs
```diff
@@ -1234,7 +1214,12 @@ impl WindowBuilder {
                 present_strategy: self.present_strategy,
             };
 
+            let position = match self.position {
+                Some(pos) => pos,
+                None => Point::new(CW_USEDEFAULT as f64, CW_USEDEFAULT as f64),
```
raphlinus, line 1219: I am slightly uncomfortable with CW_USEDEFAULT being converted into f64. I know it's going to work, but feels like it's not quite the right type for the job.

Would it work for `position` to be `(i32, i32)` here?

raphlinus (approved):
Looks good, one small concern inline. I agree that the synchronous call to SetWindowPos in the window creation call is the best way to solve this.

druid-shell/src/platform/windows/window.rs
rhzk: Good catch, makes more sense to cast self.position to i32

raphlinus:
Any reason not to merge this now? There are no conflicts, and I've verified again that it works on my machine.

jneem:
I'm guessing it's because @rhzk doesn't have write access?

PR #1274, opened 2020-10-02
Start updating TextBox for piet text changes

cmyr:
This is an attempt to integrate a bit of the new text stuff in to the `TextBox` widget.

Most of this code is a refactor that splits the actual manipulation of the buffer and selections into a new `Editor` struct that can be used to implement other text-editing widgets.

Apart from that, there are two other commits; one of them changes `TextBox` from being `Widget<String>` to be `Widget<TextStorage + EditableText>`, and the second adds a `TextBox::multiline` constructor, as a demo of multiline editing.

The multiline stuff is half-baked; the textbox does not support vertical scrolling. Ultimately it should be rewritten (maybe using @jneem's #1248 work?) so that it uses more standard scroll machinery; but this will be a bit fiddly, and I don't want to sink into it right now. I'll write up an issue detailing how I think this should work, though, and maybe someone can take it up?

This also includes a tweak to `examples/styled_text.rs` to show a multiline texbox; I'm going to replace this with a simple markdown preview app, but that work is mostly done but it's on another computer, and will have to wait til I get to the studio (perhaps later today?)

### DIFF
--- a/druid/src/text/editor.rs
+++ b/druid/src/text/editor.rs
@@ -1,0 +1,311 @@
+// Copyright 2020 The Druid Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! A component for building text editing widgets
+
+use super::{
+    movement, offset_for_delete_backwards, EditAction, EditableText, MouseAction, Movement,
+    Selection, TextLayout, TextStorage,
+};
+use crate::kurbo::Line;
+use crate::piet::PietText;
+use crate::{Application, Env, MouseEvent, PaintCtx, Point, Rect, UpdateCtx};
+
+/// A component for widgets that offer text editing.
+///
+/// `Editor` manages an [`EditableText`] type, applying edits and maintaining
+/// selection state.
+///
+/// [`EditableText`]: trait.EditableText.html
+#[derive(Debug, Clone)]
+pub struct Editor<T> {
+    layout: TextLayout<T>,
+    selection: Selection,
+    multiline: bool,
+    fixed_width: f64,
+}
+
+impl<T> Editor<T> {
+    /// Create a new `Editor`.
+    pub fn new() -> Self {
+        Editor {
+            layout: TextLayout::new(),
+            selection: Selection::caret(0),
+            multiline: false,
+            fixed_width: f64::INFINITY,
+        }
+    }
+
+    /// Set whether the editor supports multi-line text. Default to false.
+    ///
+    /// If this is false, inserted text will only insert up to the first
+    /// included newline.
+    pub fn set_multiline(&mut self, multiline: bool) {
+        self.multiline = multiline;
+    }
+
+    /// Set an explicit wrap width for this editor.
+    ///
+    /// By default the editor will not wrap lines; this is suitable for
+    /// cases such as a a single-line text field, where the containing
+    /// widget will scroll the editor as required.
+    pub fn set_wrap_width(&mut self, width: f64) {
+        self.layout.set_wrap_width(width);
+    }
+
+    //TODO discussion: should this deref/deref_mut?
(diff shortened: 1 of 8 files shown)


druid/src/text/editor.rs
```diff
@@ -0,0 +1,311 @@
...
+                } else {
+                    self.selection
+                };
+                data.edit(to_delete.range(), "");
+                self.selection = Selection::caret(to_delete.min());
+            }
+            EditAction::Move(mvmt) => self.selection = movement(mvmt, self.selection, data, false),
+            EditAction::ModifySelection(mvmt) => {
+                self.selection = movement(mvmt, self.selection, data, true)
+            }
```
totsteps, line 185: I think you forgot `EditAction::SelectAll`?

druid/src/widget/textbox.rs
```diff
@@ -310,129 +203,92 @@ impl Widget<String> for TextBox {
...
-        }
     }
 
-    fn lifecycle(&mut self, ctx: &mut LifeCycleCtx, event: &LifeCycle, data: &String, _env: &Env) {
+    fn lifecycle(&mut self, ctx: &mut LifeCycleCtx, event: &LifeCycle, data: &T, env: &Env) {
         match event {
             LifeCycle::WidgetAdded => {
                 ctx.register_for_focus();
-                self.text.set_text(data.as_str().into());
+                self.editor.set_text(data.to_owned());
```
totsteps, line 250: Now since we have `multiline` support, what if the initial state lensed by a Textbox contains a `\n` char?

druid/src/text/editor.rs
```diff
@@ -0,0 +1,311 @@
...
+
+    /// Update the editor if the data or env has changed.
+    ///
+    /// The widget owning this `Editor` must call this method during its own
+    /// [`update`] call.
+    ///
+    /// [`update`]: ../trait.Widget.html#tymethod.update
+    pub fn update(&mut self, ctx: &mut UpdateCtx, new_data: &T, env: &Env) {
+        if self.data_is_stale(new_data) {
+            self.layout.set_text(new_data.clone());
```
totsteps, line 149: Same here. If the textbox is not multi-lined, we should guard against `new_data` containing `\n` characters!

This issue can be seen when a non-multi lined textbox and another multi-lined textbox lens the same data!

cmyr: this is... a problem.

I'm not sure how best to think about this. Piet has no concept of a single-line layout, and trying to validate external data changes is difficult. It isn't clear what we should do when they happen.

I think my (bad) answer to this, for now, is that it is up to the programmer to ensure that newlines are not inserted externally. We could also definitely debug_panic if a newline exists in the initial text. 😒

druid/src/text/editor.rs
cmyr: good catch, thanks!

druid/src/text/editor.rs
```diff
@@ -0,0 +1,311 @@
...
+            data.edit(self.selection.range(), "");
+            self.selection.min()
+        };
+
+        self.selection = Selection::caret(cursor_pos);
+    }
+
+    fn delete_forward(&mut self, data: &mut T) {
+        let to_delete = if self.selection.is_caret() {
+            movement(Movement::Right, self.selection, data, false)
```
justinmoon, line 243: Delete button doesn't do anything when text isn't selected.

The `self.selection` starts and ends at the same index, e.g. `Selection { start: 16, end: 16 }`. I think you might need something like `offset_for_delete_forward`.

justinmoon:
Besides the delete button comment, everything here worked (on Windows).

cmyr:
Thanks @justinmoon I'll take a look!

druid/src/text/editor.rs
cmyr: oops, the `false` should have been `true`.

justinmoon (approved):
61a2c62 looks good. "select all" and "delete" both work now.

druid/src/text/editor.rs
```diff
@@ -0,0 +1,311 @@
...
+            text
+        } else {
+            text.split('\n').next().unwrap_or("")
+        };
+        let sel = self.selection.range();
+        data.edit(sel, text);
+        self.selection = Selection::caret(self.selection.min() + text.len());
+    }
+
+    /// Delete to previous grapheme if in caret mode.
```
raphlinus, line 225: This comment is imprecise: the backspace logic sometimes uses backspace granularity, but other times uses finer granularity to match expected behavior (individual vowels and diacritics in Indic scripts, harakat in Arabic, etc).

raphlinus (approved):
Generally this looks good. Obviously text editing will continue to be an area of ongoing work.

Issue #1068, opened 2020-06-29
Re-entrancy and RefCells


jneem:
`druid-shell` has an issue with re-entrancy, particularly with file dialogs. The background is probably familiar to many of you, but let me include it anyway.

The main entry point from `druid-shell` to `druid` is the `WinHandler` trait, which `druid-shell` defines and `druid` implements. In the other direction, `druid-shell` provides `druid` with a `WindowHandle`, which `druid` can use to interact with the shell in various ways. An important difference between `WinHandler` and `WindowHandle` is that `WinHandler`'s methods take `&mut self`, while `WindowHandle`'s take `&self`. In particular, `WindowHandle` is cloneable and shareable and generally uses interior mutability to get stuff done.

A typical call stack looks like:
1) the system calls into `druid-shell`, maybe because it got a mouse event or something
2) `druid-shell` calls into druid through some method on `WinHandler`, like maybe `WinHandler::mouse_up`. Note that this usually involves borrowing the `WinHandler` using `RefCell::borrow_mut`.
3) `druid` does a bunch of complicated stuff, including maybe calling some methods in `druid-shell` like `WindowHandle::invalidate` or `WindowHandle::request_timer`

All of this is fine, because (for example) `WindowHandle::request_timer` does its interior mutation quickly and returns right away. But `WindowHandle` has two problematic functions, `open_file_sync` and `save_as_sync`. As the names suggest, these are blocking methods. In particular, they return control to the system immediately. So now the call stack can continue like:
3) `druid` does a bunch of complicated stuff, including calling `WindowHandle::open_file_sync`
4) the system calls into `druid-shell`, maybe because it got a timer event or something
5) `druid-shell` *wants* to call some method on `WinHandler`, but it can't because its reference to `WinHandler` has already been `RefCell::borrow_mut`ed.

The current behavior is to ignore any events that arrive while the `WinHandler` is borrowed, but this isn't great. This problem comes up on several of the backends. I think only x11 and web are immune, and only because they don't implement the problematic methods on `WindowHandle` yet.

[...]

PR #1275, opened 2020-10-03
Centralize the WinHandler borrowing in gtk/x11 backends

jneem:
This is in preparation for dealing with #1068.

The idea is to make the RefCell borrowing of the WinHandler centralized (and also more principled, and better documented). Then we'll be able to slot in the various not-quite-synchronous operations at the place where the WinHandler borrow is dropped.


cmyr (approved):
This is not an area I have been thinking about or working in, but from a distance this looks like a reasonable bit of consolidation and cleanup. Thanks!

Issue #1125, opened 2020-08-11
TextBox doesn't reset position after unfocused


MGlolenstine:
Within the same example from #1124 I've found yet another problem.

If we type more text than there's space for in the first `TextBox`, the text starts shifting. If we switch to another `TextBox` and start deleting from the beginning, the weird shift starts showing.
![image](https://user-images.githubusercontent.com/9406770/89904308-0bcbf400-dbe9-11ea-84a9-26c45e162eb1.png)
In my opinion the text's position should reset when unfocused, at least that's how most of the other applications do it.

# EDIT:
This is what happens, when there's no text and only hints remain.
![image](https://user-images.githubusercontent.com/9406770/89904710-8e54b380-dbe9-11ea-9b12-c710668bfe3e.png)

cmyr:
Thanks, will keep an eye open for this when I start redoing text stuff in druid (shortly)

MGlolenstine:
> Thanks, will keep an eye open for this when I start redoing text stuff in druid (shortly)

You're welcome and thanks for working on it 😉

cmyr:
I think this is really a dupe of your previous issue? As in this describes the problem, and the other issue is capturing aa symptom of that. I would close one of them, maybe the other one?

MGlolenstine:
I wouldn't say it's a dupe, as they describe two different problems, which are both connected to TextBoxes.
If you want to close any of them, feel free to do so.
If I had a say, I'd leave the first one, as it has a MinimalRunnableExample attached.

cmyr:
I think they describe different manifestations of the same problem? If a textbox deselects its contents when it loses focus, (this issue) then that fixes the problem of multiple textboxes having selections simultaneously?

MGlolenstine:
You are indeed correct.

PR #1276, opened 2020-10-03
Fix #1124, #1125

### LINKED ISSUE #1124: Able to select text in multiple TextBoxes at once
I don't think that this should be happening, but I can select text in multiple `TextBox`es at once.
image
If I start typing I only replace the text in the focused one, but I don't think it's aesthetically pleasing to be able to select text in multiple Boxes.

sysint64:
PR fixes #1124 and #1125.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,6 +98,8 @@ You can find its changes [documented below](#060---2020-06-01).
 - Windows: fix crash on resize from incompatible resources ([#1191 by [@raphlinus]])
 - GTK: Related dependencies are now optional, facilitating a pure X11 build. ([#1241] by [@finnerale])
 - `widget::Image` now computes the layout correctly when unbound in one direction. ([#1189] by [@JAicewizard])
+- TextBox doesn't reset position after unfocused. ([#todo] by [@sysint64])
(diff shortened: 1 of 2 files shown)


cmyr (approved):
Thanks! this stuff is definitely going to be in flux for the next little while but good to get this fixed.

totsteps:
```diff
+ self.move_selection(Movement::StartOfDocument, data, false);
```

I don't think this is what we want. One needs to click twice in order to enter new text. This also results in some weird behavior(happens randomly), where the whole text is selected when the textbox receives focus by the first click(gif below).

The textbox should preserve the selection movement and only draw the selection if it is being focused. As seen in CLion,  vscode, JavaFX, etc.

I am not sure what the best solution is, but again not resetting the selection does not resolve  #1125. Maybe resetting `hscroll_offset` when `FocusChanged` solves this!

Screen Recording 2020-10-05 at 18 34 19

cmyr:
so I think the desired behaviour is platform dependant. 

- On all platforms, if a widget does not have focus it should not draw its selection.
- on mac, when a textbox gains focus, it should select all text.
- in windows, a textbox should preserve its previous selection state.

additionally:

- on mac, a textbox should not draw the cursor if the selection is not a caret.

I think it's best to do this work on top of #1274.

PR #1278, opened 2020-10-04
Fix #1129 and add request_timer to LayoutCtx

### LINKED ISSUE #1129: Show scroll when changing size
If we scroll and show the scrollbar, we can see it's size changing, but if we keep on changing the height of the content, it eventually fades away.
Would there be a way to keep it if the content is changing. That's how most other frameworks do it, just do that you know that something's changing.

Majora320:
This resolves #1129 and also adds request_timer to LayoutCtx. Please let me know if this would have any negative implications of this, although it seems like something that would be useful to have in the public api for similar use cases to this one.


cmyr (approved):
Thanks for this, always feels good to fix these little rough edges!

PR #1279, opened 2020-10-05
Fix update on delegate commands.

jneem:
This is a small fix for a bug I encountered while working on #1068. The first problem is that update wasn't called when a command was handled by the delegate. The second is that the `open_save` example had `true` and `false` mixed up.

On a related note, I'd like to change some `bool`s to something like `enum Handled { Handled, Unhandled }`. Is there any objection to that?


cmyr (approved):
`Handled/Unhandled` sounds very reasonable!

PR #1280, opened 2020-10-05
Implement vertical movement (for text editing)

cmyr:
Based off of #1274 

This implements vertical motion for text editing.

I think the traits we use here may change over time, but this gets something working for now.

(this is slightly buggy because it relies on a pending piet release)

### DIFF
--- a/druid/src/text/movement.rs
+++ b/druid/src/text/movement.rs
@@ -14,7 +14,9 @@
 
 //! Text editing movements.
 
-use crate::text::{EditableText, Selection};
+use crate::kurbo::Point;
+use crate::piet::TextLayout as _;
+use crate::text::{EditableText, Selection, TextLayout, TextStorage};
 
 /// The specification of a movement.
 #[derive(Debug, PartialEq, Clone, Copy)]
@@ -23,6 +25,10 @@ pub enum Movement {
     Left,
     /// Move to the right by one grapheme cluster.
     Right,
+    /// Move up one visible line.
+    Up,
+    /// Move down one visible line.
+    Down,
     /// Move to the left by one word.
     LeftWord,
     /// Move to the right by one word.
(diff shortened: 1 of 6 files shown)


druid/src/text/selection.rs
```diff
@@ -28,13 +28,20 @@ pub struct Selection {
 
     /// The active edge of a selection, as a byte offset.
     pub end: usize,
+
+    /// The saved horizontal position, during vertical movement.
+    pub h_pos: Option<f64>,
```
justinmoon, line 33: It's a little weird how this variable is called `h_pos` here and `x_pos` on 105

justinmoon (approved):
Besides the one variable naming nit, this works great for me on Windows

druid/src/text/selection.rs
cmyr: I agree :)

druid/src/text/movement.rs
```diff
@@ -37,44 +43,98 @@ pub enum Movement {
...
 }
 
-/// Compute the result of movement on a selection .
-pub fn movement(m: Movement, s: Selection, text: &impl EditableText, modify: bool) -> Selection {
-    let offset = match m {
+/// Compute the result of movement on a selection.
+///
+/// returns a new selection representing the state after the movement.
+///
+/// If `modify` is true, only the leading edge (the 'end') of the selection
```
raphlinus, line 50: Is "leading" the right word here? It seems to me "trailing".

druid/src/text/movement.rs
```diff
@@ -37,44 +43,98 @@ pub enum Movement {
...
+                (s.max(), None)
             }
         }
 
-        Movement::PrecedingLineBreak => text.preceding_line_break(s.end),
-        Movement::NextLineBreak => text.next_line_break(s.end),
+        Movement::Up => {
+            let cur_pos = layout.hit_test_text_position(s.end);
+            if cur_pos.line == 0 {
+                (0, Some(cur_pos.point.x))
```
raphlinus, line 90: I think this is the behavior of our existing code, but I think `Some(s.h_pos.unwrap_or(cur_pos.point.x))` is slightly more precise. And that might lead to logic simplification because I think it's the same in both branches.

Note that the code as you have it matches xi2, so I certainly can't complain :)

druid/src/text/movement.rs
```diff
@@ -37,44 +43,98 @@ pub enum Movement {
...
+                let h_pos = s.h_pos.unwrap_or(cur_pos.point.x);
+                let point_above = Point::new(h_pos, cur_pos.point.y - lm.height);
+                let up_pos = layout.hit_test_point(point_above);
+                (up_pos.idx, Some(point_above.x))
+            }
+        }
+        Movement::Down => {
+            let cur_pos = layout.hit_test_text_position(s.end);
+            if cur_pos.line == layout.line_count() - 1 {
+                (text.len(), Some(cur_pos.point.x))
```
raphlinus, line 102: Same concern as above.

raphlinus (approved):
Generally this looks good, very small nits inline.

It's fun to see the capabilities slowly grow :)

druid/src/text/movement.rs
```diff
@@ -37,44 +43,98 @@ pub enum Movement {
...
+            } else {
+                let lm = layout.line_metric(cur_pos.line).unwrap();
+                let h_pos = s.h_pos.unwrap_or(cur_pos.point.x);
+                // may not work correctly for point sizes below 1.0
```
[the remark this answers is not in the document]
cmyr: This comment was intended to be somewhat tongue in cheek 😅.

Wouldn't adding half the line-height would fail if our current line-height is more than 2x the height of the preceding line, which doesn't seem impossible?

This seemed like the most robust approach, but there may be issues I'm not considering.

druid/src/text/movement.rs
cmyr: fair, I guess "active" is the better word.

druid/src/text/movement.rs
cmyr: Good call, I think I prefer preserving the existing h_pos if possible.

PR #1282, opened 2020-10-05
Only draw selection when focused

### LINKED ISSUE #1124: Able to select text in multiple TextBoxes at once

cmyr:
- fixes #1124 (again)


totsteps (approved):
Tested on mac. LGTM

PR #1283, opened 2020-10-05
Improve selection behaviour on focus change

cmyr:
On windows, we persist the existing selection when focus
changes. On mac, the behaviour depends on whether or not there
was a click. If there was a click, we use the click to determine
the selection; otherwise we select the whole buffer.

In addition, on mac, we no longer draw the cursor if the selection
is not a caret.

- fixes #1225 (again)

### DIFF
--- a/druid/src/widget/textbox.rs
+++ b/druid/src/widget/textbox.rs
@@ -44,6 +44,12 @@ pub struct TextBox<T> {
     cursor_timer: TimerToken,
     cursor_on: bool,
     multiline: bool,
+    /// true if a click event caused us to gain focus.
+    ///
+    /// On macOS, if focus happens via click then we set the selection based
+    /// on the click position; if focus happens automatically (e.g. on tab)
+    /// then we select our entire contents.
+    was_focused_from_click: bool,
 }
 
 impl TextBox<()> {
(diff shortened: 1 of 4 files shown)


totsteps (approved):
Tested on macOS and Linux. LGTM.

cmyr:
oops! note to self: I need a changelog for this.

druid/src/widget/textbox.rs
```diff
@@ -250,7 +269,12 @@ impl<T: TextStorage + EditableText> Widget<T> for TextBox<T> {
                 self.editor.set_text(data.to_owned());
                 self.editor.rebuild_if_needed(ctx.text(), env);
             }
-            LifeCycle::FocusChanged(_) => {
+            LifeCycle::FocusChanged(_is_focused) => {
+                #[cfg(target_os = "macos")]
+                if *_is_focused && !self.was_focused_from_click {
+                    self.editor.select_all(data);
+                }
```
Finnerale, line 276: Why not use `cfg!` and avoid the `_`?
```suggestion
            LifeCycle::FocusChanged(is_focused) => {
                if cfg!(target_os = "macos") && *is_focused && !self.was_focused_from_click {
                    self.editor.select_all(data);
                }
```

PR #1285, opened 2020-10-06
Respect double-click rectangle

raphlinus:
Previously the Windows double click detection logic would register any
two clicks within the time threshold to be a double click. The correct
logic for Windows also takes into account a rectangle constraining the
maximum mouse movement.

### DIFF
--- a/druid-shell/src/platform/windows/window.rs
+++ b/druid-shell/src/platform/windows/window.rs
@@ -329,6 +329,7 @@ struct WndState {
     has_mouse_focus: bool,
     //TODO: track surrogate orphan
     last_click_time: Instant,
+    last_click_pos: (i32, i32),
     click_count: u8,
 }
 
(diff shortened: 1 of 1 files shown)


raphlinus:
The false double clicks are extremely annoying in the pen tool in Runebender.

jneem (approved):

druid-shell/src/platform/windows/window.rs
```diff
@@ -1027,15 +1028,20 @@ impl WndProc for MyWndProc {
                         let mods = s.keyboard_state.get_modifiers();
                         let buttons = get_buttons(wparam);
                         let dct = unsafe { GetDoubleClickTime() };
-                        let threshold = Duration::from_millis(dct as u64);
                         let count = if down {
                             // TODO: it may be more precise to use the timestamp from the event.
                             let this_click = Instant::now();
-                            if this_click - s.last_click_time >= threshold {
+                            let thresh_x = unsafe { GetSystemMetrics(SM_CXDOUBLECLK) };
+                            let thresh_y = unsafe { GetSystemMetrics(SM_CYDOUBLECLK) };
```
rhzk, line 1035: Might not have any impact but maybe use the dpi version for windows 10?
```
let (thresh_x, thresh_y) = 
if let Some(func) = OPTIONAL_FUNCTIONS.GetSystemMetricsForDpi {
    let dpi = (self.scale().x() * SCALE_TARGET_DPI) as u32;
    (func(SM_CXDOUBLECLK, dpi), func(SM_CYDOUBLECLK, dpi))
 }else {
     (GetSystemMetrics(SM_CXDOUBLECLK), GetSystemMetrics(SM_CYDOUBLECLK))
 };
```

raphlinus: That's interesting, thanks, as I was wondering about dpi. In my testing, the value was the same (always 4, whether from GSM or using 96 or 192 from GSMFD). I'm actually not sure how much difference it makes, when you are actually double-clicking it's usually 0.

Also relevant: https://devblogs.microsoft.com/oldnewthing/20041018-00/?p=37543 which is an authoritative source. That suggests I should be resetting the click counter on WM_ACTIVATE and also when the button changes.

raphlinus:
Merging, as I think this is good enough as-is, and there wasn't feedback on the dpi question.

PR #1287, opened 2020-10-07
Use widget::prelude::* where appropriate

cmyr:
This will encourage other people to use the prelude, which in
turn makes it slightly less annoying when (and if) we change the
signatures of widget methods.

This includes a few small cleanups that came about while
updating imports; most of these involve replacing things like

`let rect = Rect::from_origin_size(Point::ZERO, my_size);`

with,

`let rect = my_size.to_rect();`.


justinmoon (approved):
LGTM

Issue #1242, opened 2020-09-18
Clarification of behaviour and expectations of the layout API


cmyr:
*what does it mean for a widget to be given a size different from the one it returned from its `layout` method?*

## the problem

In doing the baseline work, I have been exploring making Widget::layout return a more general type than `Size`. Overtime, we have added additional metrics that are computed in layout. The first of these was the 'paint insets', which describe the area outside of the view's layout size where the view may draw; and now I've been adding baseline positions, which are offsets relative to the edge of the layout's size that are used to align horizontal text.

Currently, these additional metrics are added through calls on the `LayoutCtx`; for instance, `LayoutCtx::set_paint_insets`. The idea was to hopefully simplify this, by having all of these metrics be returned directly from the layout call, when applicable.

Of course, there's a problem.

This problem isn't specific to the redesign; it relates to the fact that the *actual* size of a widget is not the necessarily the size returned from the widget's `layout` method; it is a size chosen by whichever ancestor is managing the widget's `WidgetPod`. *In the general case* this should be the size returned by the widget, but this is not enforced; the parent can in theory chose an arbitrary size.

What does it mean for the parent to choose a size different from that returned by the child? This is not really written down anywhere; nor is there a clear guideline for how the child is expected to adapt when drawing to an unexpected size.

 This potential conflict poses problems for things like baselines and paint insets. These metrics are relative to the size calculated in layout, and will be invalid if the widget ends up being provided with a different size.
 
## tentative proposal

 I think we can improve on this situation, by clarifying our invariants and how we will behave when they are not met. I propose the following:
 
 - If a widget returns a valid size from `layout` (defined as a size that satisfies the provided constraints) the widget will *always* be given the same size in `paint`.
 
 - If the parent widget *needs* to paint the child in a size other than what the child requested, the child's `paint` method will be called with the child's expected size, and the parent will chose to either clip or overflow the content.
 
 - All metrics (such as baselines and paint insets) will be relative to the 'layout size', that is the size returned from `Widget::layout`.
 
 - `WidgetPod::set_layout_rect` may be replaced with `WidgetPod::set_layout_pos`?

The main advantage of this is that it significantly simplifies layout; The `WidgetPod` itself can stash the result of its child's `layout` method, and know that the provided values are valid.

## questions

[...]

PR #1289, opened 2020-10-07
Add WidgetPod::set_origin, soft-deprecate set_layout_rect

cmyr:
As outlined in https://github.com/linebender/druid/issues/1242.

Allowing a parent to give a widget a size different from the size
it had computed in its `layout` method had the potential to cause
subtle breakage. With this change, we now always use the size
returned by the widget's layout method, and expect the container
to set the origin only, using WidgetPod::set_origin.

This does not remove or (yet) officially deprecate
WidgetPod::set_layout_rect; I am curious to see if there is any
existing code where a widget was being given a size different
from what it had returned, and deprecating this method would
cause us to fail in CI. I expect to deprecate it before 0.7.0, though.

This is motivated by the work on baselines; baseline offsets are one
of the things that would be subtly broken if the child's chosen
size and the final size differed.

### DIFF
--- a/druid/src/core.rs
+++ b/druid/src/core.rs
@@ -71,10 +71,14 @@ pub struct WidgetPod<T, W> {
 #[derive(Clone)]
 pub(crate) struct WidgetState {
     pub(crate) id: WidgetId,
-    /// The frame of this widget in its parents coordinate space.
-    /// This should always be set; it is only an `Option` so that we
-    /// can more easily track (and help debug) if it hasn't been set.
-    layout_rect: Option<Rect>,
+    /// The size of the child; this is the value returned by the child's layout
+    /// method.
+    size: Size,
+    /// The origin of the child in the parent's coordinate space; together with
+    /// `size` these constitute the child's layout rect.
+    origin: Point,
+    /// A flag used to track and debug missing calls to set_origin.
+    is_expecting_set_origin_call: bool,
     /// The insets applied to the layout rect to generate the paint rect.
     /// In general, these will be zero; the exception is for things like
     /// drop shadows or overflowing text.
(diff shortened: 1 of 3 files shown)


druid/src/core.rs
```diff
@@ -189,34 +193,33 @@ impl<T, W: Widget<T>> WidgetPod<T, W> {
...
+
+    /// Set the origin of this widget, in the parent's coordinate space.
+    ///
     /// A container widget should call the [`Widget::layout`] method on its children in
-    /// its own [`Widget::layout`] implementation, then possibly modify the returned [`Size`], and
-    /// finally call this `set_layout_rect` method on the child to set the final layout [`Rect`].
+    /// its own [`Widget::layout`] implementation, and then call `set_origin` to
+    /// position those children.
     ///
     /// The child will receive the [`LifeCycle::Size`] event informing them of the final [`Size`].
```
jneem, line 210: Do we still need `LifeCycle::Size` if your size is always determined by your own layout function?

cmyr: I was wondering about this too. I guess it depends on how this event is actually used? I could see it making sense in some cases, because it means the widget doesn't need to internally hold on to past sizes. I'm not actually sure if this matters, though? I don't believe I use this event anywhere.

jneem: I guess I can't imagine anything that I would want to do in `lifecycle` that I couldn't just do in `layout` :shrug: Given our lack of API stability, my inclination would be to get rid of it and then put it back later if someone finds a use. But I'll give you the green button anyway...

jneem (approved):

druid/src/core.rs
cmyr: Maybe @xStrom would like to chime in? Removing this would also let us remove some special-case code we use to send this notification..

PR #1291, opened 2020-10-08
Fixes for rustc 1.47.0

cmyr:
Locally I have a panic in clippy when running against the wasm target; I've opened https://github.com/rust-lang/rust-clippy/issues/6140 for that, and if it shows up here I will set that CI item as an allowed failure 🤷


raphlinus (approved):
Every 6 weeks, like clockwork :)

I was worried the renamed lint would bump our MSRV (which I think would be pretty aggressive), but it looks like it's just an "unknown lint" warning. That's fine.

Also, I think the clippy issue is https://github.com/rust-lang/rust-clippy/issues/6140.

PR #1292, opened 2020-10-08
Fix panic in TextBox when data changes externally

cmyr:
We were not correctly updating the selection as required.

This will fail CI until #1291 is merged.

cc @totsteps, this is your bug from yesterday, thanks again for the testing!


jneem (approved):
Oops :)

PR #1293, opened 2020-10-08
Implement TextStorage and EditableText for Arc<String>

cmyr:
This isn't the perfect long-term solution, but in the near-term it
provides a more data-friendly option for editable text.

### DIFF
--- a/druid/src/text/editable_text.rs
+++ b/druid/src/text/editable_text.rs
@@ -15,7 +15,8 @@
 //! Traits for text editing and a basic String implementation.
 
 use std::borrow::Cow;
-use std::ops::Range;
+use std::ops::{Deref, Range};
+use std::sync::Arc;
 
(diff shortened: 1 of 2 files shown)


druid/src/text/editable_text.rs
```diff
@@ -190,6 +191,52 @@ impl EditableText for String {
...
+    fn next_codepoint_offset(&self, offset: usize) -> Option<usize> {
+        self.deref().next_codepoint_offset(offset)
+    }
+    fn preceding_line_break(&self, offset: usize) -> usize {
+        self.deref().preceding_line_break(offset)
+    }
+    fn next_line_break(&self, offset: usize) -> usize {
+        self.deref().next_line_break(offset)
+    }
+
```
raphlinus, line 231: Blank lines inconsistent here :/

raphlinus (approved):
Seems pretty harmless, if boilerplatey.

PR #1294, opened 2020-10-08
Fix with parse widget failing to update child

### LINKED ISSUE #1290: Numeric textboxes break when given non-numeric data in recent git revisions
Run `cargo run --example parse`. Click on text box. Press Backspace, followed by any letter key.

Expected behavior: The 0 in the text box is deleted, the letter appears in the text box.
Actual behavior: The 0 in the text box is deleted correctly. The letter never appears. The text box now ignores all further key and mouse input. Warning messages such as the following appear in the log on all input, whether keyboard or mouse.

`WARN  [druid::text::editor] editor data changed externally, skipping event Insert("h")`

Known working commit: b38e873f7fceff5068c11e950aa2202b472729a2
Known broken commit: 694af59d63318c7197f81eedba56de3071182321
Git bisect says first broken commit is c64372879de3e28cc2b7f1deeb97352362f8f2be, though the exact behavior of these broken text boxes varies by commit.

Notably, this does _not_ occur on text boxes which are bound to String fields.

I can also rule out a regression in the `Parse` type itself, since I've written my own independent, extended alternative to Parse and it displays the same bug. This seems to be related to recent work on TextBox.

cmyr:
Because the TextLayout object backing the TextBox keeps
internal track of the text, we need to send it `update`
explicitly.

- fixes #1290

### DIFF
--- a/druid/src/widget/parse.rs
+++ b/druid/src/widget/parse.rs
@@ -37,8 +37,15 @@ impl<T> Parse<T> {
 
 impl<T: FromStr + Display + Data, W: Widget<String>> Widget<Option<T>> for Parse<W> {
     fn event(&mut self, ctx: &mut EventCtx, event: &Event, data: &mut Option<T>, env: &Env) {
+        let pre_data = self.state.clone();
         self.widget.event(ctx, event, &mut self.state, env);
         *data = self.state.parse().ok();
+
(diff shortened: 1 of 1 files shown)


cmyr:
I have a better fix for this.

PR #1295, opened 2020-10-08
Expose baseline on TextLayout, add Baseline alignment to Flex

cmyr:
This introduces the idea of baseline alignment as a component
of layout.

During their `layout` calls, widgets can specify their baseline_offset,
which is the distance from the bottom of their reported size to their
baseline.

Generally, the baseline will be derived from some text object,
although widgets that do not contain text but which appear next
to widgets that do can specify an arbitrary baseline.

This also adds CrossAxisAlignment::Baseline to Flex; this is
only meaningful in a Flex::row, in which case it aligns all of
its children based on their own reported baselines.

The best place to play around with this code is in examples/flex.rs.

Screen Shot 2020-10-08 at 6 20 10 PM
Screen Shot 2020-10-08 at 6 20 21 PM
Screen Shot 2020-10-08 at 6 20 28 PM
Screen Shot 2020-10-08 at 6 20 36 PM

This is also the last of the major outstanding text stuff??

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ You can find its changes [documented below](#060---2020-06-01).
 - `RichText` and `Attribute` types for creating rich text ([#1255] by [@cmyr])
 - `request_timer` can now be called from `LayoutCtx` ([#1278] by [@Majora320])
 - TextBox supports vertical movement ([#1280] by [@cmyr])
+- Widgets can specify a baseline, flex rows can align baselines ([#1295] by [@cmyr])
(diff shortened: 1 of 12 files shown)


raphlinus (approved):
I'm going to approve this. I have some concerns about how well it will work in a multi-line configuration, but also I experimented a bit with the flex demo and find it works nicely for single line, and that's probably a good enough goal for this cycle; it's hugely useful for the kind of UI stuff we're doing, where labels are realistically single line.

Thanks!

PR #1296, opened 2020-10-09
[ci] Disable clippy + wasm in ci

cmyr:
This will still the tests against wasm, but clippy is disabled
until further notice. This can be turned back on the next time
we get an update to cargo.

See https://github.com/rust-lang/rust-clippy/issues/6140


raphlinus (approved):
Seems like a fairly minimal patch, and glad to see it's fixed in beta upstream.

PR #1298, opened 2020-10-10
Use Handled instead of bools.

jneem:
This makes some "is this event handled"-style return types more typey, and it also includes a better fix for #1279. I originally went with `Handled::Handled` and `Handled::Unhandled`, but it seemed a bit redundant.

### DIFF
--- a/druid/src/app_delegate.rs
+++ b/druid/src/app_delegate.rs
@@ -122,7 +122,7 @@ pub trait AppDelegate<T: Data> {
     /// This function is called with each ([`Target`], [`Command`]) pair before
     /// they are sent down the tree.
     ///
-    /// If your implementation returns `Handled::Unhandled`, the command will be sent down
+    /// If your implementation returns `Handled::No`, the command will be sent down
     /// the widget tree. Otherwise it will not.
     ///
     /// To do anything fancier than this, you can submit arbitary commands
--- a/druid/src/app_delegate.rs
+++ b/druid/src/app_delegate.rs
@@ -17,7 +17,7 @@
 use std::any::{Any, TypeId};
 
 use crate::{
-    commands, core::CommandQueue, Command, Data, Env, Event, MenuDesc, SingleUse, Target,
+    commands, core::CommandQueue, Command, Data, Env, Event, Handled, MenuDesc, SingleUse, Target,
     WindowDesc, WindowId,
 };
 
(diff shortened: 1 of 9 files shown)


druid/src/util.rs
```diff
@@ -50,3 +50,28 @@ where
...
+    /// An event was already handled, and shouldn't be propagated to other event handlers.
+    Yes,
+    /// An event has not yet been handled.
+    No,
+}
+
+impl Handled {
+    /// Has the event been handled yet?
+    pub fn is_handled(self) -> bool {
+        matches!(self, Handled::Yes)
```
Finnerale, line 66: Considering that `Handled` implements `Eq`:
```suggestion
        self == Handled::Yes
```

druid/src/util.rs
```diff
@@ -50,3 +50,28 @@ where
...
+    }
+
+    /// Returns `Handled` if `handled` is true, and `Unhandled` otherwise.
+    pub fn from_handled(handled: bool) -> Handled {
+        if handled {
+            Handled::Yes
+        } else {
+            Handled::No
+        }
+    }
```
Finnerale, line 76: Maybe implement `Into<bool>` and `From<bool>` instead?
Then it can just be used as `Handled::from(my_bool)`.

druid/src/app_delegate.rs
```diff
@@ -122,7 +122,7 @@ pub trait AppDelegate<T: Data> {
     /// This function is called with each ([`Target`], [`Command`]) pair before
     /// they are sent down the tree.
     ///
-    /// If your implementation returns `true`, the command will be sent down
+    /// If your implementation returns `Handled::Unhandled`, the command will be sent down
```
Finnerale, line 125:
```suggestion
    /// If your implementation returns `Handled::No`, the command will be sent down
```

druid/src/win_handler.rs
```diff
@@ -319,23 +319,25 @@ impl<T: Data> Inner<T> {
...
 
-    /// Returns `true` if the command was handled.
-    fn dispatch_cmd(&mut self, cmd: Command) -> bool {
-        if !self.delegate_cmd(&cmd) {
-            self.do_update();
-            return true;
+    fn dispatch_cmd(&mut self, cmd: Command) -> Handled {
+        let handled = self.delegate_cmd(&cmd);
+        // We do the update whether or not the command was handled, because the delegate has
+        // arbitrary user code in it: it could change the data but return Handled::No.
```
Finnerale, line 325: This comment seems a bit irritating to me, outside the context of this diff.
At first, I wanted to suggest a different phrasing, but I think removing this one would be best.
The code is quite clear about running `do_update` after a call to `delegate_cmd` and afterwards to an early return if the delegate handled the event.

Finnerale (approved):
Thanks for getting this done, I tried before but could not decide what to call the enum variants :sweat_smile: 

I do have a bunch of nits, but those aside I'm happy to get this merged.

jneem:
Thanks for the review! I'm not sure that these names are the best either, but even imperfect names are better than `bool`...

PR #1299, opened 2020-10-12
Fix `Either` using wrong paint insets

andrewhickman:
Spotted this while reading the code, looks like a copy-paste error!


jneem:
Thanks for the patch! Would you mind adding an entry to the "Fixed" section of the CHANGELOG? (Note that the issue and contributor links there aren't automatic: they're defined at the bottom of the file)

cmyr (approved):
Nice catch, thanks!

Issue #1297, opened 2020-10-10
Crash in text box layout when text box is too narrow


andrewhickman:
I ran into this on Windows 10, using version e01c99502c8c500e2f7f663e01789ffd4b68ca13

Repro steps:
- In the "text" example, increase the padding a bit on line 115: `.padding((SPACER_SIZE * 8.0, SPACER_SIZE))`
- Run the example
- Set line break mode to "wrap"
- Decrease the width of the window

On my system this crashes with the following backtrace:

<details><summary>Backtrace</summary>

```

thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: BackendError(hresult 80070057)', C:\Users\andre\Repositories\druid\druid\src\text\layout.rs:301:72
stack backtrace:
   0: std::panicking::begin_panic_handler
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39\/library\std\src\panicking.rs:475
   1: core::panicking::panic_fmt
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39\/library\core\src\panicking.rs:85
   2: core::option::expect_none_failed
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39\/library\core\src\option.rs:1221
   3: core::result::Result<piet_direct2d::text::D2DTextLayout, piet::error::Error>::unwrap<piet_direct2d::text::D2DTextLayout,piet::error::Error>
             at C:\Users\andre\.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\core\src\result.rs:973
   4: druid::text::layout::TextLayout<text::AppState>::rebuild_if_needed<text::AppState>
             at .\src\text\layout.rs:301
   5: druid::widget::label::{{impl}}::layout<text::AppState>
             at .\src\widget\label.rs:534
   6: druid::widget::controller::{{impl}}::layout<text::AppState,druid::widget::label::RawLabel<text::AppState>,text::LabelController>
             at .\src\widget\controller.rs:125
   7: druid::widget::widget::{{impl}}::layout<text::AppState>
             at .\src\widget\widget.rs:251
   8: druid::core::WidgetPod<text::AppState, alloc::boxed::Box<Widget<text::AppState>>>::layout<text::AppState,alloc::boxed::Box<Widget<text::AppState>>>
             at .\src\core.rs:491
   9: druid::widget::container::{{impl}}::layout<text::AppState>
[...]

PR #1304, opened 2020-10-12
Ensure text wrap width is at least 0.0

andrewhickman:
Implementing the fix suggested [here](https://github.com/linebender/druid/issues/1297#issuecomment-707235539)


cmyr (approved):
Thanks!

Issue #1306, opened 2020-10-13
Cross-platform menu issues (for Runebender)


raphlinus:
I'm doing most my Runebender development on Windows, and have come across some issues related to menus. This is a tracking issue for those.

* The Runebender source uses `platform_menus::mac` menus directly, without any kind of configuration check. Those use `RawMods::Meta`, which bind to Windows, and that doesn't work.

* The shortcuts ("Ctrl-S") are not displayed in Windows menus. I think that was intended for a while, but I didn't quite get around to implementing them. There was churn around `HotKey` but I think that's stabilized reasonably well now thanks to the keyboard work.

* The provided menus are missing alt-key shortcuts. These appear to work, by adding an ampersand in the localized string (`&File`, `E&xit`, etc), and are stripped out (at least on mac, haven't tested Linux yet).

The most expedient fix to the first problem is to switch from `RawMods` to `SysMods::Cmd`. That would result in reasonable behavior if menus for the wrong platform were used, if not the most platform-appropriate keybindings. A deeper change would be to expand the `platform_menus::common` choices, and more aggressively enforce the use of actual platform specific menus.

I have local patches with expedient fixes for most of this, and plan to PR them. I'm open to discussion about the right way to proceed, and of course if people want to take on more extensive rework.

kud1ing:
Not sure whether it fits here, but on Mac the "Quit" enter should be [the last entry](https://developer.apple.com/design/human-interface-guidelines/macos/menus/menu-bar-menus/). While on Mac Rundender it is the 3rd last.

![Uploading Bildschirmfoto 2020-10-13 um 20.57.41.png…]()

raphlinus:
Technically, that's a Runebender issue, but it's relevant to discuss here, as the point is how we can best support a native-feeling menu structure appropriate to each platform.

kud1ing:
> they are not included in the Edit menu, which I believe is where they should be.

Indeed, [you're right](https://developer.apple.com/design/human-interface-guidelines/macos/menus/menu-bar-menus/#edit-menu) about this.

kud1ing:
Another thing is that it should be "runebender" instead of "Druid" obviously.

raphlinus:
[...]

PR #1310, opened 2020-10-14
Display hotkeys for menu items in Windows

raphlinus:
This patch adds a string to describe hotkeys for menu items in the
Windows back-end.

Part of work tracked in #1306


cmyr (approved):
Windows definitely hasn't had the attention that mac has for menu stuff, good to see some of this get cleaned up.

PR #1312, opened 2020-10-14
Parse widget should keep track of old and new state

### LINKED ISSUE #1290: Numeric textboxes break when given non-numeric data in recent git revisions

cmyr:
- fixes #1290

This is a much more correct approach than was #1294.

### DIFF
--- a/druid/src/widget/parse.rs
+++ b/druid/src/widget/parse.rs
@@ -22,6 +22,10 @@ use crate::Data;
 /// Converts a `Widget<String>` to a `Widget<Option<T>>`, mapping parse errors to None
 pub struct Parse<T> {
     widget: T,
+    // because we are synthesizing the data for our inner widget, we keep track
+    // of 'old' and 'new' data as widgetpod would, so that our child can handle
+    // update() correctly.
+    prev_state: String,
     state: String,
 }
 
@@ -30,6 +34,7 @@ impl<T> Parse<T> {
     pub fn new(widget: T) -> Self {
         Self {
             widget,
+            prev_state: String::new(),
             state: String::new(),
         }
     }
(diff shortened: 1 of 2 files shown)


druid/src/widget/parse.rs
```diff
@@ -38,7 +43,9 @@ impl<T> Parse<T> {
 impl<T: FromStr + Display + Data, W: Widget<String>> Widget<Option<T>> for Parse<W> {
     fn event(&mut self, ctx: &mut EventCtx, event: &Event, data: &mut Option<T>, env: &Env) {
         self.widget.event(ctx, event, &mut self.state, env);
-        *data = self.state.parse().ok();
+        if self.state != self.prev_state {
```
RaoulHC, line 46: Would it make more sense to have `self.prev_state = self.state;` just before passing state to the inner widget here, and then scrapping lines 68-70? This would still ensure that the widget is updated properly with `prev_state` and `state`, while not setting the state via `data.to_string()`

RaoulHC (commented):
the `mem::replace` line still causes panics parsing floating numbers.

druid/src/widget/parse.rs
cmyr: Thanks for checking this. There are some frustrating limitations to using `parse()` for this, instead of some dedicated validation/transformation mechanism, but I hope I have the main cases covered now!

cmyr:
I'm going to close this and focus on replacing `Parse` entirely.

PR #1314, opened 2020-10-14
Remove unused cairo-rs dependency from macOS

jrmuizel:


cmyr (approved):
Agree, it's probably time for this to go.

Previously you could optionally use GTK on mac, which was helpful for testing, but I think is not worth maintaining.

PR #1315, opened 2020-10-14
Fix incorrect cursor position when hscroll > 0.0

### LINKED ISSUE #1313: textbox cursor position incorrect when text is wider than textbox area
This is a regression that happened sometime in the last week or two, presumably while I was fixing other things:

cmyr:
Had been testing the multiline case and ignoring the single when
I fixed the cursor to always clamp to the right bound of the layout
when necessary.

- fixes #1313


totsteps (approved):
Thanks for the fix. Works as intended.

PR #1316, opened 2020-10-14
Allow access to Editor from TextBox

cmyr:
This is directly motivated by the EditableLabel in runebender;
I need access to the layout in my new implementation.

This also tweaks the api for getting access to the TextLayout from
within the Editor to better conform with rust API conventions.


totsteps (approved):
LGTM

PR #1317, opened 2020-10-15
Count clicks in Gtk mouse handler

raphlinus:
Do our own click counting rather than relying on the count given by Gtk.
Among other things, this avoids the problems of multiple click events.

Progress towards #859


cmyr (approved):
Great, good to get these fundamentals in place.

PR #1321, opened 2020-10-15
Fix small bug in TextBox hscroll calculation

cmyr:
This was not correctly accounting for padding, which would
cause us to not add hscroll as needed if the width of
the text was less than our width, but more than our width
exclusive of padding.

before:

after:


totsteps (approved):
Had this on my TODO list since last month.

Thanks for the fix.

PR #1324, opened 2020-10-16
Improve frameless window on windows

rhzk:
Some changes to the frameless window on windows, reducing the resize artifacts and the behavior is more simular to electron.
Added a wrapper function for GetSystemMetric since it ended up being called a lot

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ You can find its changes [documented below](#060---2020-06-01).
 - Added documentation to resizable() and show_titlebar() in WindowDesc. ([#1037] by [@rhzk])
 - Windows: Added internal functions to handle Re-entrancy. ([#1037] by [@rhzk])
 - Windows: WindowDesc: Create window with disabled titlebar, maximized or minimized state, and with position. ([#1037] by [@rhzk])
-- Windows: WindowHandle: Change window state. Toggle titlebar. Change size and position of window. ([#1037] by [@rhzk])
+- Windows: WindowHandle: Change window state. Toggle titlebar. Change size and position of window. ([#1037], [#1324] by [@rhzk])
 - Windows: WindowHandle: Added handle_titlebar(), Allowing a custom titlebar to behave like the OS one. ([#1037] by [@rhzk])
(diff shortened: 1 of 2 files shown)


cmyr:
Played around with this a bit in a VM and wasn't able to really see any visual difference with `master`, but I'm also not very sure what I should be looking for. Is there anything in particular I should do to test this?

rhzk:
The difference here is pretty smal.
The master version have 3 colored edges on left right and bottom side (something that was not visible to me originally until i played around with theme/effects settings in windows) this gets rid of those weird edges.

The other thing is resizing, and there is a few different things here: (This could potentially be different on other systems?)

Using ```PresentStrategy::Sequential``` mode with ```present_after = 1```, ie waiting for vblank removed all resize artifacts with titlebar visible / normal mode, and made resizing seem a lot less "jiggly" in borderless mode.

In borderless mode the best option might actually be to use ```PresentStrategy::Flip``` and call ```handler.prepare_paint()``` inside ```WM_SIZE``` after the ```handler.size()``` call, atleast that gave very nice results on my system. I could update the PR to use this mode in borderless, unless there is some reason we want to avoid using ```Flip``` and/or not call ```prepare_paint()``` in ```WM_SIZE ```?

I tried to make some recordings to show the resizing differences but the recording software was not good at capturing it, I might try some different software later

cmyr (approved):
I'm really just not up on the windows side enough to offer much useful feedback. Given that, and given that I don't see any actual *regressions* when playing around, I'm inclined to defer to you.

cmyr:
@rhzk I've invited you as a collaborator on the druid repo, so feel free to merge your own approved PRs and also to provide approving reviews of others' PRs in circumstances where you feel comfortable doing so. :)

rhzk:
Thanks for the invite @cmyr. I will let this PR simmer a bit more, just in case I might realize a better or simpler way to do something or something I missed.

cmyr:
sounds good!

derekdreery:
@cmyr @rhzk I would suggest merging this, because we won't get any data on its effects until then, and we can always reverse it out later if there are problems.

PR #1325, opened 2020-10-16
Only account for baseline in Flex if needed

cmyr:
Previously we would always leave enough space for all widgets
to potentially be baseline aligned, which could cause unintended
layout problems when baseline alignment wasn't used.

### DIFF
--- a/druid/src/widget/flex.rs
+++ b/druid/src/widget/flex.rs
@@ -619,13 +619,12 @@ impl<T: Data> Widget<T> for Flex<T> {
         // these two are calculated but only used if we're baseline aligned
         let mut max_above_baseline = 0f64;
         let mut max_below_baseline = 0f64;
-        let mut any_use_baseline = matches!(self.cross_alignment, CrossAxisAlignment::Baseline);
+        let mut any_use_baseline = self.cross_alignment == CrossAxisAlignment::Baseline;
 
         // Measure non-flex children.
         let mut major_non_flex = 0.0;
         for child in &mut self.children {
-            any_use_baseline &=
-                matches!(child.params.alignment, Some(CrossAxisAlignment::Baseline));
+            any_use_baseline &= child.params.alignment == Some(CrossAxisAlignment::Baseline);
 
             if child.params.flex == 0.0 {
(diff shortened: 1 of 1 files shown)


druid/src/widget/flex.rs
```diff
@@ -619,10 +619,14 @@ impl<T: Data> Widget<T> for Flex<T> {
         // these two are calculated but only used if we're baseline aligned
         let mut max_above_baseline = 0f64;
         let mut max_below_baseline = 0f64;
+        let mut any_use_baseline = matches!(self.cross_alignment, CrossAxisAlignment::Baseline);
```
Finnerale, line 622:
```suggestion
        let mut any_use_baseline = self.cross_alignment == CrossAxisAlignment::Baseline;
```

druid/src/widget/flex.rs
```diff
@@ -619,10 +619,14 @@ impl<T: Data> Widget<T> for Flex<T> {
...
 
         // Measure non-flex children.
         let mut major_non_flex = 0.0;
         for child in &mut self.children {
+            any_use_baseline &=
+                matches!(child.params.alignment, Some(CrossAxisAlignment::Baseline));
```
Finnerale, line 628:
```suggestion
            any_use_baseline &= child.params.alignment == Some(CrossAxisAlignment::Baseline);
```

Finnerale (approved):
This should be using `==` instead of `matches!`, then it looks good to me.

druid/src/widget/flex.rs
derekdreery: What is the policy on `matches!` vs `==` given they will probably result in the same assembly.

Finnerale: `matches!` is more verbose, but powerful. When I see it, I expect a more complicated enum comparison, probably ignoring some parts of the variant. If `==` does the job perfectly, using it makes more sense to me.

cmyr: I agree with this, I think if `PartialEq` is implemented and does the job that is clearer code.

PR #1327, opened 2020-10-16
Make FontDescriptor construction const

cmyr:


raphlinus (approved):
This makes sense for now, but I wonder if FontDescriptor might become a more complex object later, for example containing a unique identifier to a custom-loaded font, as well as drop logic. But I think we can cross that bridge when we get to it, don't want yagni to get in the way.

cmyr:
I think *some* font descriptors will always be available in a `const` context? Custom fonts already won't work in `const`, but that's okay, you can make them at runtime. 🤷

PR #1328, opened 2020-10-17
Make dialogs on windows non-sychronous.

jneem:
This does #1068 for windows.


druid-shell/src/platform/windows/window.rs
```diff
@@ -1083,28 +1044,17 @@ impl WndProc for MyWndProc {
...
-            }
+            WM_CLOSE => self
+                .with_wnd_state(|s| s.handler.request_close())
+                .map(|_| 0),
             DS_REQUEST_DESTROY => {
                 unsafe {
                     DestroyWindow(hwnd);
                 }
                 Some(0)
             }
```
rhzk, line 1055: If we make WindowHandle::close() deferred I think we could get rid of this custom message now,
unless there is some other reason we need it?

jneem:
Now that mac support has landed, I've updated this PR to remove the backwards-compatibility glue.

jneem:
Now that mac support has landed, I've updated this PR to remove the backwards-compatibility glue.

cmyr (approved):
Great. Looks like there might be some clippy stuff to clean up now, but happy to have this all sorted out, thanks again for taking this on!

PR #1331, opened 2020-10-19
[mac] Adopt new save/open API

cmyr:
This implements the new save/open (#1068) API on macOS.

The implementation here is slightly different; Cocoa uses the idea
of 'blocks' (a c extension for callbacks) in most of their async API,
so we don't need to to manually defer the work.

### DIFF
--- a/druid-shell/src/platform/mac/window.rs
+++ b/druid-shell/src/platform/mac/window.rs
@@ -22,6 +22,7 @@ use std::mem;
 use std::sync::{Arc, Mutex, Weak};
 use std::time::Instant;
 
+use block::ConcreteBlock;
 use cocoa::appkit::{
     CGFloat, NSApp, NSApplication, NSAutoresizingMaskOptions, NSBackingStoreBuffered, NSEvent,
     NSView, NSViewHeightSizable, NSViewWidthSizable, NSWindow, NSWindowStyleMask,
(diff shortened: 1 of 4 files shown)


druid-shell/src/platform/mac/window.rs
```diff
@@ -933,26 +934,56 @@ impl WindowHandle {
...
+    pub fn save_as_sync(&mut self, _options: FileDialogOptions) -> Option<FileInfo> {
+        log::warn!("save_as_sync should no longer be called on mac!");
         None
     }
 
+    pub fn save_as(&mut self, options: FileDialogOptions) -> Option<FileDialogToken> {
+        let token = FileDialogToken::next();
+        let self_clone = self.clone();
+        unsafe {
+            let panel = dialog::build_panel(FileDialogType::Save, options);
```
jneem, line 971: Looks like except for `FileDialogType::Save` this is identical to `open_file`?

jneem (approved):
Great! I have no way of testing this, but at least it *looks* good.

druid-shell/src/platform/mac/window.rs
cmyr: also different is the method we call on the handler inside the block. I could probably come up with some way to share this code more but I'm not sure it's worth the typing.

cmyr: (okay fine I rewrote it to share an implementation 😛)

Issue #1277, opened 2020-10-04
Animation interferes with timer


bheisler:
Platform: Windows 10

Code: https://gist.github.com/bheisler/f1185c6de8d8ee064fe2b3b215050aae

Druid Version: Latest git
```
[dependencies]
druid = { git = "https://github.com/linebender/druid.git", version = "0.6" }
```

I noticed that timer events don't seem to be delivered while a widget is animating. I found this surprising, as did several others on the Druid zulip, so I wrote a small test case to demonstrate this behavior.

To reproduce, run the example code and click the "Start Timer and Animation" button. About five seconds later, it will print something like this to the command line:
`Timer completed after 5.0059584s. Expected time: 1s. Animation time: 5s.`

The timer is registered for one second, not five, so this is surprising. However, if you comment out `ctx.request_anim_frame();` on either line 32 or 41, then the timer completes after about one second, as expected:
`Timer completed after 0.9909591s. Expected time: 1s. Animation time: 5s.`

You can also adjust ANIMATION_TIME to control how long the custom widget animates for; the timer event will not be delivered until after the animation stops.

rhzk:
So when request_anim_frame() is called a WM_PAINT message is sent to the systems message queue. When we handle the WM_PAINT message Event::AnimFrame() is called which causes request_anim_frame() to get called again creating a WM_PAINT message and so on.

I think the problem here is that WM_TIMER messages have a low priority and therefore dont get put in the message queue as long as there is a WM_PAINT message there, which there constantly is because of the loop. Causing the WM_TIMER message to only be added after the loop ends and thats why your timer ends up lasting as long as the animation

jneem:
Can you check how often `paint` is called on your system? Last time I checked on an actual windows machine, it seemed like we were only getting WM_PAINT about once per monitor refresh, but I remember possibly seeing some other behavior on a VM.

Also (as someone who knows nothing about programming on windows) is it possible to change the message priorities, or alternatively to just retrieve all the available messages and decide on our how how to prioritize them?

bheisler:
Regarding the number of paint calls - it varies a bit but it's around 300 for a 5 second animation, so it looks like it's animating at or around 60fps. The CPU usage hovers around 1-2% as well (on a machine with 4 physical/8 logical cores) so it definitely isn't pinning a core repainting over and over.

Unfortunately I don't know anything about the windows API so I can't help you there.

raphlinus:
[...]

PR #1332, opened 2020-10-19
Try using the PeekMessage trick to make timers less laggy.

jneem:
This is an attempt at fixing #1277.

### DIFF
--- a/druid-shell/src/platform/windows/application.rs
+++ b/druid-shell/src/platform/windows/application.rs
@@ -27,9 +27,9 @@ use winapi::shared::winerror::HRESULT_FROM_WIN32;
 use winapi::um::errhandlingapi::GetLastError;
 use winapi::um::shellscalingapi::PROCESS_PER_MONITOR_DPI_AWARE;
 use winapi::um::winuser::{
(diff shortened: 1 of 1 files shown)


raphlinus (approved):
I am fine with this, but do want to caution that *any* change to the `run` method is best-effort, and in fact will not be effective when the application is not running its own message loop. That will include when live-resizing (including just moving the window), when a modal dialog is open, and possibly other cases. I'd be entirely open to a bit warning going into the code to this effect, as it is very tempting to fix things there.

I think in the much longer term we'll move away from using WM_PAINT and schedule our own repainting, but that's a low priority (heh heh) as there's a bunch of things that can go wrong, and one of the main motivations I had for going there (low-latency wait objects) apparently requires flip-mode presentation, which in turn doesn't work with smooth resizing.

jneem:
Ok, I've added a little warning and I'll push if the CI clears it. I confess to knowing basically nothing about how the message loop works on windows, and so for example I don't really understand your comment about `WM_PAINT` (because I thought we had established that we weren't clogging up the message queue with `WM_PAINT` messages, although I'm still mystified as to why the timer events weren't getting through).

PR #1333, opened 2020-10-20
Rework anim

JAicewizard:
This is a really small one. I also retroactively added `use druid::widget::prelude::*` to the identity because I am pretty sure that is the recommended way of doing it, and `anim.rs` is a very small example.

This PR just moves the animation state to `data` and adds some documentation. I thought the state should usually be in `data`, but I might be wrong.


druid/examples/anim.rs
```diff
@@ -12,52 +12,62 @@
...
             _ => (),
         }
     }
 
-    fn lifecycle(&mut self, _ctx: &mut LifeCycleCtx, _event: &LifeCycle, _data: &u32, _env: &Env) {}
+    fn lifecycle(&mut self, _ctx: &mut LifeCycleCtx, _event: &LifeCycle, _data: &f64, _env: &Env) {}
 
-    fn update(&mut self, _ctx: &mut UpdateCtx, _old_data: &u32, _data: &u32, _env: &Env) {}
+    fn update(&mut self, ctx: &mut UpdateCtx, old_data: &f64, data: &f64, _env: &Env) {
+        //this is just some margin of error, not very meaningfull in any real way.
```
jneem, line 53: meaningful

druid/examples/anim.rs
```diff
@@ -12,52 +12,62 @@
...
 use std::f64::consts::PI;
 
 use druid::kurbo::{Circle, Line};
 use druid::widget::prelude::*;
 use druid::{AppLauncher, Color, LocalizedString, Point, Vec2, WindowDesc};
 
-struct AnimWidget {
-    t: f64,
-}
+struct AnimWidget {}
```
jneem, line 29: I think having the time in the data makes sense for this tiny example, but I think it's a better practice in larger apps to have the animation state in the widget, not the data. And then because people look at the examples to get an idea of best practices, I think it's better to have it in the widget here.

The reason animation state usually lives in the widget is because it's a piece of transient data that isn't usually interesting for the app as a whole. You can see the built-in `Switch` widget for an example -- it has a `bool` related to the animation state, and it would be annoying if every time you used `Switch`, you had to reserve a `bool` for it in your app's data.

jneem (commented):
Thanks for continuing with this! I'm not convinced right now about the `Data` part, and I've tried to explain why. If my reasoning doesn't make sense, though, feel free to push back on it.

druid/examples/anim.rs
JAicewizard: That makes sense, thanks for letting me know. Maybe something for in the book to talk about as well?
I can open a issue about this if that's the way to go about this.

cmyr: that would be good, this is something we've discussed in the past, and it we should write it down.

druid/examples/anim.rs
```diff
@@ -74,7 +83,6 @@ pub fn main() {
             .with_placeholder("You spin me right round..."),
     );
     AppLauncher::with_window(window)
-        .use_simple_logger()
```
cmyr, line 77: I think we should use logging in all examples; it's important that people have logging enabled during development, so that we can do things like print warnings about incorrect layout.

JAicewizard: I don't really mind, I thought that nobody would really use the examples for real development, but I guess people copy paste examples all the time. I will also put it back for the identity in this PR as well then.

druid/examples/anim.rs
JAicewizard: #1337

druid/examples/anim.rs
```diff
@@ -12,7 +12,14 @@
...
 
-//! An example of an animating widget.
+//! An example of an animating widget. It is just a widget that
+//! requests an animation frame when it needs to, and draws the frame in the
+//! `paint` method.
+//! Once the animation is over it simply stops requesting animation frames.
+//! Usually we would put the state in the `Data`, but for things like animation
+//! we don't. This is because the animation state is not useful to know for the
+//! rest of the app. If this is something the rest of your widgets should know
+//! about, you could put it in the data put it in the `data`.
```
jneem, line 22: put it in the in the put it in :)

jneem:
Looks good, I just found one more typo. The `()` thing is definitely better than `u32`!

JAicewizard:
I dont think thats my code? I cant rerun them I think.

jneem (approved):

jneem:
Thanks! The CI thing seems like it was related to version changes in piet.

PR #1334, opened 2020-10-20
Use '=' on pre-release dependencies

cmyr:
I'm going to do another piet pre-release shortly, and would like
to avoid breakage.


jneem (approved):

raphlinus:
A warning, I'm pretty sure this won't work unless piet *internally* uses '=' dependencies. I think pre5 might be a lost cause, but we can make things more robust for pre6.

PR #1336, opened 2020-10-21
Rework blocking functions

JAicewizard:
Mostly reordering/commenting.
I think most of these examples are like that, but for someone unfamiliar to the example/druid it can really help guide them through the example (at least from my experience.)

I also installed a spell checker, never needed it before since my projects weren't open source :tada:


druid/examples/blocking_function.rs
```diff
@@ -12,45 +12,73 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! An example of a blocking function running in another thread.
+//! An example of a blocking function running in another thread. We have
+//! we give the other thread some data and then we also pass some data back
+//! to the main thread using commands.
```
cmyr, line 17:
```suggestion
//! An example of a blocking function running in another thread. We give
//! the other thread some data and then we also pass some data back
//! to the main thread using commands.
```

druid/examples/blocking_function.rs
```diff
@@ -12,45 +12,73 @@
...
 }
 
+// Pretend this is downloading a file, or doing heavy calculations...
+fn slow_function(number: u32) -> u32 {
+    let a_while = time::Duration::from_millis(2000);
+    thread::sleep(a_while);
+    number + 1
+}
+
+struct Delegate {}
```
cmyr, line 80: for structs with no fields, we can declare them without braces; this can just be,

```rust
struct Delegate;
```

cmyr (commented):
Couple little things, looks good though!

jneem (approved):
Thanks!

PR #1338, opened 2020-10-21
Update piet

### LINKED ISSUE #1265: label/button rendering issues on web
On web the label/button text rendering is offset to the top. If the button is big it is not noticeable but if you resize the window to be small the text overflows the textbox.
There is also an issue in Piet that the colour (or any other attribute) is not settable, but that is not what this issue is about. Due to this it is hard to read the text.
misplaced_button_text
For context the top is the search-bar and bottom is dev tools slided up to make the window smaller artificially.
This is 1 window with 1 child widget, the Button.

### LINKED ISSUE #1309: On Windows 10, TextBox in hello.rs example does not resize to fit placeholder text
On Windows 10, I modified the bleeding edge Hello World example (`hello.rs`) so that it looks like 

```Rust
// a textbox that modifies `name`.
    let textbox = TextBox::new()
        .with_placeholder("Who are we greeting?")
        .with_text_size(22.0) // Used to be `with_text_size(18.0)`
        .fix_width(TEXT_BOX_WIDTH)
        .lens(HelloState::name);
```
The text box does resize correctly to the height of the place holder text: 

hello-world

cmyr:
This also adds back a method that was removed from Region, with
a deprecation notice.

- fixes #1265 
- fixes #1309

### DIFF
--- a/druid-shell/Cargo.toml
+++ b/druid-shell/Cargo.toml
@@ -21,8 +21,8 @@ x11 = ["x11rb", "nix", "cairo-sys-rs"]
 [dependencies]
 # NOTE: When changing the piet or kurbo versions, ensure that
 #       the kurbo version included in piet is compatible with the kurbo version specified here.
-piet-common = "=0.2.0-pre5"
-kurbo = "0.6.3"
+piet-common = "=0.2.0-pre6"
+kurbo = "0.7.0"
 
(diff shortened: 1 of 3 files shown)


raphlinus (approved):
Looking forward to seeing all these goodies land!

PR #1339, opened 2020-10-21
Make platform-specific menus more portable

raphlinus:
If an application were to misuse a platform-specific for the wrong
platform, it would get unusable keybindings. With this patch, it might
not be the most appropriate for the platform, but it would work.

The linked bug below suggests a more systematic approach, but this is a
minimal patch to get things to a less broken state.

Part of work tracked in #1306.

### DIFF
--- a/druid/src/menu.rs
+++ b/druid/src/menu.rs
@@ -47,7 +47,7 @@
 //! Creating the default Application menu for macOS:
 //!
 //! ```
-//! use druid::{Data, LocalizedString, MenuDesc, MenuItem, RawMods};
+//! use druid::{Data, LocalizedString, MenuDesc, MenuItem, SysMods};
 //! use druid::commands;
 //!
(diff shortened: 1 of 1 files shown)


jneem (approved):
This certainly seems like an improvement, but I'm a bit confused about the intended cross-platform usage of `SysMods` and `RawMods`. At least on linux, it's common for "meta" to refer to the "windows" key, but it seems that `SysMods` doesn't have it...

raphlinus:
The lack of "meta" or "windows" in SysMods is intentional. On macOS, the "apple" or "command" key maps to "meta," which is `SysMods::Cmd` on that platform. It's also the case that the windows key on Windows can't be used as a hotkey modifier, as it seems to be intercepted by the system. A good way to think about SysMods is that it's the lowest common denominator of modifier keys.

This area can definitely use more refinement and documentation.

PR #1340, opened 2020-10-22
Optional image features

### LINKED ISSUE #1305: minimise features imported from `image` crate.
Right now using images in druid causes all default features of image to be imported.
This means that I cannot disable features on my end, because druid will "need" them. I am sure druid doesn't need to encode/decode in any image format for me.
The biggest one is "jpeg/rayon" which imports 20 crates or so, which is really not needed for druid.
If this is something you agree with I can minimize the amount of features enabled for image and send in a PR.

JAicewizard:
This is a follow up to https://github.com/linebender/piet/pull/342.
For my simple use-case (https://martens-icons-website.netlify.app/) the WASM binary shrunk from 1.4MiB -> 0.8MiB uncompressed, making this a 60% WASM binary savings for this simple app not needing any image format.
fixed #1305


JAicewizard:
I just wanted to say that locally I hit a clippy bug in one of the tests where clippy crashes, I might report that as a bug if it happens here too

druid-shell/Cargo.toml
```diff
@@ -18,6 +18,23 @@ default = ["gtk"]
 gtk = ["gio", "gdk", "gdk-sys", "glib", "glib-sys", "gtk-sys", "gtk-rs", "gdk-pixbuf"]
 x11 = ["x11rb", "nix", "cairo-sys-rs"]
 
+# passing on all the image features. AVIF is not supported because it does not
+# support decoding, and thats al we use `Image` for.
```
jneem, line 22: that's all

druid/Cargo.toml
```diff
@@ -26,6 +26,23 @@ image = ["druid-shell/image"]
 svg = ["usvg", "harfbuzz-sys"]
 x11 = ["druid-shell/x11"]
 
+# passing on all the image features. AVIF is not supported because it does not
+# support decoding, and thats al we use `Image` for.
```
jneem, line 30: that's all

jneem (approved):
Nice!

cmyr:
@JAicewizard was your crash [this one](https://github.com/rust-lang/rust-clippy/issues/6140)? If it's not you should definitely consider reporting it back to clippy. :)

PR #1341, opened 2020-10-23
use Self where possible

adamnemecek:


cmyr:
Is there any consensus in rust that `Self` is preferred wherever possible? I'm personally indifferent, and tend to only use `Self` in places where there are generic params or otherwise using the concrete type name is a hassle.

adamnemecek:
As of right now there are places in the druid codebase where Self was used and where it wasn't. I think that the only reason why 

```rust
impl A {
    fn new() -> Self {
        A { }
    }
}
```

is still used over 

```rust
impl A {
    fn new() -> Self {
        Self { }
    }
}
```
is that the latter wasn't possible until relatively recently. I find the latter much cleaner. It also makes renaming types easier.

Finnerale:
[Zulip discussion](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/use.20Self.20where.20possible.20%28druid.231341%29/near/214758291)

cmyr:
I think inertia wins here and we'll just leave things as they are, it doesn't seem to be a real problem. In any case thanks for the patch @adamnemecek, I appreciate the work.

PR #1343, opened 2020-10-24
Rework calc

JAicewizard:
I also reworked the implementation of the calculator, because I think using just floats is easier to read/reason with, even if it might not create a perfect calculator.


cmyr (commented):
I overall like the idea of this, but I have a few concerns about the patch. In particular the behaviour has changed in important ways; if I press `92 ÷ 10 × 5` I now end up with `0` instead of the correct result.

The second issue here is that in moving things around we've changed the alignment of the main display label, which should (I think) always be right aligned; previously we were achieving this by setting the `MainAxisAlignment` of the main flex column, although today we would probably do it by setting the text alignment of the label, and making it expand to fill the row.

This brings up maybe a slightly more meta point, about structuring patches: when writing a patch that introduces behaviour changes, where possible it is good to avoid moving too much code around, because it makes the diff more complicated to read. In general if there's a bunch of code that needs to be moved around, doing that in a separate patch is nice, because that patch can be skimmed and merged without concern, and then the patch that actually needs more careful review ends up being much shorter and easier to review.

I'd encourage you to think about this while doing some of this cleanup; ultimately it will let us get these changes merged more quickly and easily. Let me know if any of this doesn't make sense!

JAicewizard:
I am currently on mobile, but I agree with basically all you mentioned. I tried a lot of combinations including all operations and they worked, I will look into that tho it might not have to be perfect but it should at least work in basic example.
I didn't notice the alignment change, probably becausetthee display is empty by default, I will try to make sure to test the layout more in the future.

And do you mean diferent commits? Or diferent prs.
Diferent prs seems very a but much. I guess I never though of commits as a helping hand for code review,iI always look at the final dif.
If putting things in diferent commits helps I will try and do that, I should probably think more about how to structure the example before hand in that case, now I just see how to make it simpler/easier to read/more logically structure and have incremental steps of that and rewriting is just one of those steps.

If the behaviour change (floating point rounding mostly) is too much I can undo that but it mostly felt like magic when I first read it, so that really made me feel it wasn't simple enough for an example.

JAicewizard:
I restructured the file to be more like I would normally structure a druid application since this example doesn't really show anything off. Maybe I tried to hard to find something to show off before but it indeed just doesn't work well.
There ware some issues with the C/CE buttons and the = sign, which led to the second formula being filled in always resulting in 0. Both issues are fixed now.

cmyr:
to clarify: I wasn't talking so much about separate commits, but more about how to structure PRs; I keep on opening this to review and it's a big diff and I know that most of it is just moving things around, but some of it is clearly changing behaviour, and it's very hard to review the behaviour changing stuff because it's all mixed in with the moving things around stuff.

What I would suggest: if there are aspects of the *behaviour* of this example that you would like to change, open up a PR with just those changes, which should be easy to review; and then afterwards we can do a PR that moves things around how you like. :)

JAicewizard:
Ah yes, I will keep that in mind. I am currently almost done with the event viewer.
No offense, but it was a bit messy for en example, so I did change some bits and pushed the logic a bit up again.
In that case there are no behavioral changes, only splitting things off/inlining things, do you want those to also be split?

I am sorry the calc one is such a big one, I can definitely split off the moving things around from the behavior changes from now on.

cmyr:
yea, `event_viewer` is intended more as a debugging tool than as a real 'example', so it may be a bit fast & loose.

JAicewizard:
After some cleanup I kinda liked the way you did it. I am also definitely seeing how useful `Controllers` actually are.

JAicewizard:
Maybe it is an idea to just remove this from the examples?
There could be a druid showcase repo/sub-directory with some "real-world" example applications.
Then the examples directory can be purely to show off how to do certain things in druid, and the showcase can be for things like "look we made X with druid!! how cool" kind of things. Those would look like a normal application and could be multi-file/bigger then the examples.

That way the examples are purely for "learn how to render using a paintCtx" etc, more like elementary building blocks.

cmyr:
I think it is worth including some examples that are simple real-world applications, but I agree that it would be nice to arrange things so that they're easier to explore, similar to how [iced does this](https://github.com/hecrj/iced/tree/master/examples).

cmyr:
@JAicewizard what are your feelings about this? My current inclination is that we should close this, and if we want to revisit in the future we can do it as a pair of new PRs?

JAicewizard:
thats cool by me, anything wrong with it ending up the same in the end? just split up in 2

cmyr:
Split up is fine! I'm going to close this for now, if you revisit it again feel free to reopen. :)

PR #1346, opened 2020-10-25
Add Checkbox::set_text function.

Finnerale:
I need this to port `Checkbox` to Crochet and it seems useful enough to go into Druid directly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,7 +50,7 @@ You can find its changes [documented below](#060---2020-06-01).
 - TextBox supports vertical movement ([#1280] by [@cmyr])
 - Widgets can specify a baseline, flex rows can align baselines ([#1295] by [@cmyr])
 - `TextBox::with_text_color` and `TextBox::set_text_color` ([#1320] by [@cmyr])
-- `Checkbox::set_label` to update the label. ([#1346] by [@finnerale])
+- `Checkbox::set_text` to update the label. ([#1346] by [@finnerale])
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


ForLoveOfCats:
Would the usage case in Crotchet still be covered by this if it was a builder fn `with_label` or a constructor `from_label` like is on `Button`?

ForLoveOfCats:
Ah nevermind, I see that this is about updating the text value

Finnerale:
@ForLoveOfCats ~No, it wouldn't. I want to change the label without having to recreate the widget.~ You noticed that already

On adding setters in general to Druid:

I believe Druid would benefit from such functionality. Recreating widgets to change their properties mostly works for widgets without internal state, but even those then require being properly reinitialized with `WidgetAdded`and so forth. And for widgets with internal state, recreating them will even cause user-visible problems. And if we add setters to those widgets, it would seem inconsistent not to have them on cheap-to-recreate widgets.

ForLoveOfCats:
I agree that mutating widgets in-tree is a reasonable thing to be doing.

As this is not actually being passed a `Label` widget but instead the `LabelText` perhaps it would be more accurately called `set_label_text` or `set_text`? Otherwise this looks good to me now that I've actually looked at the usage case.

Finnerale:
@ForLoveOfCats Sounds good! I renamed it to `set_text` and also updated the doc comments.

ForLoveOfCats:
Missed the changelog but then it looks good to me :+1:

Finnerale:
Oops, now it should be right!

ForLoveOfCats (approved):

PR #1348, opened 2020-10-26
Rework cursor

JAicewizard:
This mostly moved added comments and moved the cursor creation to `Event::WindowConnected`.


druid/examples/cursor.rs
```diff
@@ -61,11 +45,17 @@ impl<W: Widget<AppState>> Controller<AppState, W> for CursorArea {
...
     ) {
-        if data.custom.is_none() {
-            data.custom = ctx.window().make_cursor(&data.custom_desc).map(Arc::new);
-        }
-        if matches!(event, Event::MouseMove(_)) {
-            ctx.set_cursor(&data.cursor);
+        match event {
+            Event::WindowConnected => {
+                let cursor_image =
+                    ImageBuf::from_data(include_bytes!("./assets/PicWithAlpha.png")).unwrap();
```
cmyr, line 51: while I agree that it makes sense to move the creation of the actual cursor into `WindowConnected`, I would leave the logic around creating the `ImageBuf` where it was.

How I would think about this is in terms of a principle that may have a name, but is basically: "all else being equal, state should be declared at the earliest opportunity." If something is known at compile time, and can be `const`, it should be `const`; if something can be computed in `main` it should be in `main`; etc. There are of course exceptions, such as if something has runtime or binary-size costs, but I think this is a good rule of thumb because it makes code easier to reason about; you can assume that if something is being done in a given place it is because it can't be done earlier, which makes it easier to reason about how a program works.

druid/examples/cursor.rs
```diff
@@ -13,44 +13,28 @@
...
+    WindowDesc,
 };
 
 use druid::widget::prelude::*;
 use druid::widget::{Button, Controller};
 
-use std::sync::Arc;
+use std::rc::Rc;
 
-#[derive(Clone, Data, Lens)]
```
cmyr, line 27: These patches are making me think more generally about how I would structure these files, in terms of where to put `main()`, where to put `AppState`, etc.

As a *general* rule, when I write rust, I put all of the public types at the top of the file, with their implementations below. I think this makes it easier to get a general sense of what a file will contain by looking at the start of the file. I don't always follow this exactly; especially for things like iterator types or other less important things, I will sometimes put them elsewhere in a file.

If i had to pick a rule for `main()` I would say it should always just go at the end of the example. 🤷

druid/examples/cursor.rs
```diff
@@ -13,44 +13,28 @@
...
-            } else {
-                Cursor::Arrow
-            }
-        }
-        Cursor::Custom(_) => Cursor::Arrow,
-    }
-}
-
-struct CursorArea {}
+/// This Controller sets the cursor to whichever one is selected
```
cmyr, line 32:
```suggestion
/// This Controller switches the current cursor based on the selection.
```

druid/examples/cursor.rs
```diff
@@ -74,7 +64,9 @@ impl<W: Widget<AppState>> Controller<AppState, W> for CursorArea {
 fn ui_builder() -> impl Widget<AppState> {
     Button::new("Change cursor")
         .on_click(|ctx, data: &mut AppState, _env| {
-            data.cursor = Arc::new(next_cursor(&data.cursor, &data.custom));
+            data.cursor = next_cursor(&data.cursor, data.custom.clone());
+            // Just like in the Controller we have to set the cursor then clicked as well
+            // otherwise the cursor will only change once we move.
```
cmyr, line 69:
```suggestion
            // After changing the cursor, we need to update the active cursor
            // via the context in order for the change to take effect immediately.
```

druid/examples/cursor.rs
```diff
@@ -61,11 +45,17 @@ impl<W: Widget<AppState>> Controller<AppState, W> for CursorArea {
...
-            ctx.set_cursor(&data.cursor);
+        match event {
+            Event::WindowConnected => {
+                let cursor_image =
+                    ImageBuf::from_data(include_bytes!("./assets/PicWithAlpha.png")).unwrap();
+                let custom_desc = CursorDesc::new(cursor_image, (0.0, 0.0));
+                data.custom = ctx.window().make_cursor(&custom_desc).map(Rc::new);
+            }
+            Event::MouseMove(_) => {
+                ctx.set_cursor(&data.cursor);
```
cmyr, line 56: I would move a bit of the docs from up top down here, and would say something like,

```
/// The cursor is reset by Druid at the start of each `MouseMove` event;
/// If we wish to use a non-default cursor we have to override this explicitly.

**Review by cmyr — COMMENTED** [review]
Added a few notes!

File: druid/examples/cursor.rs (hunk 1 above)
Comment (JAicewizard) [reply druid/examples/cursor.rs:51]: I get the point, I thought that since the variable would only be used once, it seems a bit useless to use the `data` for that, ad usually the state goes into the `data`. But I also see you point of doing as much as early as possible.

File: druid/examples/cursor.rs (hunk 2 above)
Comment (JAicewizard) [reply druid/examples/cursor.rs:27]: I agree with you for a normal druid application, but examples usually show something off, I have tried to put whatever is being shown off at the top and leave the less important things at the bottom.

With examples people always copy paste of course, but I think file layout is something people will always change to fit their liking, so I didn't prioritise the notnal layout I would use for structuring the program (or what was already there).

Maybe I am thinking too much about this, maybe I should just structure them like a normal program, but my experience is that examples that do it that way don't help me. These are short examples so they won't live or die by it, but I still want to make them to the point.

File: druid/examples/cursor.rs (hunk 5 above)
Comment (JAicewizard) [reply druid/examples/cursor.rs:56]: Makes sense, I should have probably asked this in zulip but I was wondering if it was actually needed _every_ mouse move event. I will put something like that in.

File: druid/examples/cursor.rs (hunk 2 above)
Comment (cmyr) [reply druid/examples/cursor.rs:27]: I think your reasoning makes sense!

File: druid/examples/cursor.rs (hunk 5 above)
Comment (cmyr) [reply druid/examples/cursor.rs:54]: I think this is really valuable thing about having you look at the examples: if something is confusing, it will likely be confusing to other readers, and it means we should explain it. Please feel encouraged to ask any questions that occur to you!

File: druid/examples/cursor.rs (hunk 1 above)
Comment (cmyr) [reply druid/examples/cursor.rs]: I agree the data might not be perfect, and I would probably put a `#[data(ignore)]` annotation on that field? I think in a perfect world it would actually be *widget* state...

Comment (JAicewizard) [reply druid/examples/cursor.rs]: It cannot be in a widgets state because both the button and the `CursorArea` need acces to it, I can create a custom widget/controller but that would be very elaborate. IMO the perfect would be one where `CursorDesc::new` and `ImageBuf::from_data` are const, but they are not.

also should I open an issue about documenting `#[data(ignore)]`? because it seems very useful for things like this, fields that are static/never change but need to be known to multiple widgets.

Comment (JAicewizard) [reply druid/examples/cursor.rs]: Just in general, should I open documentation issues? Maybe its worth having one issue with a checklist instead of a bunch of small issues.

Comment (Finnerale) [reply druid/examples/cursor.rs]: Opening an issue for that sounds good. In my experience, multiple small issues work better than large checklists, and feel more satisfying to close :)

Comment (JAicewizard) [reply druid/examples/cursor.rs]: I can relate to that

File: druid/examples/cursor.rs [hunk 6]
```diff
@@ -83,16 +73,52 @@ fn ui_builder() -> impl Widget<AppState> {
         .padding(50.0)
 }
 
+#[derive(Clone, Data, Lens)]
+struct AppState {
+    cursor: Rc<Cursor>,
+    custom: Option<Rc<Cursor>>,
+    // TODO refer to docs once they exist
```
cmyr, line 80: We can link to the docs for `Data`, for now?

cmyr (commented):
We can fix up that one `TODO`, otherwise this looks good. Thanks!

druid/examples/cursor.rs
JAicewizard: do you know of any way to link to the latest docs instead of 0.6? I feel like linking to 0.6 would just become a forgotten link that will get outdated.

cmyr: I don't think there's a way to do that for docs in examples. :(

cmyr (approved):
Thanks!

PR #1349, opened 2020-10-26
[linux] Textbox should select all text when tabbed to

### LINKED ISSUE #1342: Textbox should be selected when tabed into it.
I noticed the patch that selected a textbox on tab, and it only applies to macos,  I tried some GTK3 and QT applications, and both select all text on tab, I think that druid should do the same on linux as it does on macos

cmyr:
This is the same is the current behaviour for macOS.

- closes #1342

### DIFF
--- a/druid/src/widget/textbox.rs
+++ b/druid/src/widget/textbox.rs
@@ -26,6 +26,8 @@ use crate::{
     Selector, SysMods, TimerToken,
 };
 
+const MAC_OR_LINUX: bool = cfg!(any(target_os = "mac", target_os = "linux"));
+
 const BORDER_WIDTH: f64 = 1.;
 const TEXT_INSETS: Insets = Insets::new(4.0, 2.0, 0.0, 2.0);
 
@@ -304,7 +306,7 @@ impl<T: TextStorage + EditableText> Widget<T> for TextBox<T> {
                 self.editor.rebuild_if_needed(ctx.text(), env);
             }
             LifeCycle::FocusChanged(is_focused) => {
-                if cfg!(target_os = "macos") && *is_focused && !self.was_focused_from_click {
+                if MAC_OR_LINUX && *is_focused && !self.was_focused_from_click {
                     self.editor.select_all(data);
                 }
                 self.was_focused_from_click = false;


druid/src/widget/textbox.rs
```diff
@@ -31,6 +31,8 @@ const TEXT_INSETS: Insets = Insets::new(4.0, 2.0, 0.0, 2.0);
 
 const CURSOR_BLINK_DURATION: Duration = Duration::from_millis(500);
 
+const MAC_OR_LINUX: bool = cfg!(any(mac, windows));
```
ForLoveOfCats, line 34: Should this be `cfg!(any(mac, linux));`?

ForLoveOfCats:
Off topic comment but this is the sort of thing that makes me do a double take and question the fact that this isn't the behavior on Windows, kinda like how shift+mouse wheel for horizontal scrolling isn't actually a convention on Windows

druid/src/widget/textbox.rs
cmyr: um.

yes, definitely.

ForLoveOfCats (approved):

ratmice:
Hmm, I tried with the `styled_text` example, but didn't seem to trigger it.
was curious about the select-entire-textbox vs select just text-region, and the behavior when tabbing into an empty textbox.

cmyr:
> was curious about the select-entire-textbox vs select just text-region, and the behavior when tabbing into an empty textbox.

confused by this a bit, what do you mean by 'select-entire-textbox' versus 'text-region'?

'Select all' when the textbox is empty should be the same as setting the cursor to position `0`.

ratmice:
> > was curious about the select-entire-textbox vs select just text-region, and the behavior when tabbing into an empty textbox.
> 
> confused by this a bit, what do you mean by 'select-entire-textbox' versus 'text-region'?
> 
> 'Select all' when the textbox is empty should be the same as setting the cursor to position `0`.

Whether selection extends to the edge of the textbox, or merely extends to the end of the text within the textbox.
I debugged it and can ctrl-a to select the entire text in `styled_text`, and it behaves by selecting the text *within* the textbox with the selection not extending to the border.

Usually i've seen with systems which extend the selected area to the border of the `TextBox`, the empty selection needs special casing to do as you say to have the same behavior as setting the cursor position to `0`

I couldn't get it to trigger without:
`const MAC_OR_LINUX: bool = cfg!(any(target_os = "mac", target_os = "linux"));`

Anyhow, and empty selection works as one would expect, like you say above.

ForLoveOfCats (changes requested):
I just checked in the Rust playground and 
```rust
const MAC_OR_LINUX: bool = cfg!(any(target_os = "mac", target_os = "linux"));
```
is true while
```rust
const MAC_OR_LINUX: bool = cfg!(any(mac, linux));
```
is indeed false. Good catch @ratmice, I completely missed that. I wonder why it wasn't a compiler error

cmyr:
Glad someone around here is paying attention 😅

ForLoveOfCats (approved):
Just ran it locally and it appears to work as described

ratmice (approved):

PR #1351, opened 2020-10-27
Propagate events to hidden widgets in `Either`

### LINKED ISSUE #1350: Excessive invalidation when using `Either`
Repro steps:

- Enable `debug_invalidation` in the `either` example
- Hover over the "click to reveal slider" label

The entire screen is invalidated repeatedly, even when the mouse is not moving.

andrewhickman:
Fixes #1350 

The `should_propagate_to_hidden` methods are taken from the tabs widget. I made them public since they seem useful for external widget implementors.


jneem (approved):
Makes sense to me.

andrewhickman:
This currently has an issue when using a `Spinner` (or other animated widget). `Either` doesn't lay out its hidden child, so there is a warning when it gets `AnimFrame` events. I'm not sure whether it should lay out both children (as `Tabs` does) or not send `AnimFrame` events to the child

andrewhickman:
I've changed the warning to only be emitted for events that have `should_propagate_to_hidden() == false`. @cmyr does that sound reasonable? It means `Event::Command` and `Event::AnimFrame` will no longer generate the warning.

Finnerale (changes requested):
This should have a changlog entry (probably multiple)

Finnerale (approved):
Thanks!

andrewhickman:
CI had a spurious error but that seems to have resolved itself now

Finnerale:
@andrewhickman this looks good to me, should we merge it?

andrewhickman:
Yep, its ready to merge 👍

PR #1352, opened 2020-10-28
Add Crochet feature and a method to WidgetPod.

Finnerale:
This PR will allow me to fix Crochet and update it to the newest Druid.

It does
- Add a feature gate "crochet"
- Add `WidgetPod::with_event_context` method to run a function with the widgets `EventCtx`. This is the most minimal change to allow Crochet to function properly.
- (Fix `LabelText::Static` `resolve` method)


raphlinus (approved):
Looks good. And thanks for taking on the crochet update task, it'll be really good to remove that friction point.

PR #1355, opened 2020-10-28
Fix `TextLayout::set_text` not always setting the text when it should

### LINKED ISSUE #1354: "editor data changed externally" when using `Arc<String>` with `TextBox`
Repro steps:
- Create a textbox using `Arc<String>` for its data
- Run the app and hit <delete> and then another key in the textbox

The second key is ignored and a warning is generated like
```
WARN  [druid::text::editor] editor data changed externally, skipping event Insert("s")    
```

andrewhickman:
Fixes #1354 

The check on `as_str` here wasn't quite equivalent to `Data::same` when using `Arc<String>` as the datatype. Empty edits like pressing delete at the end of a textbox wouldn't update the text

Most of the calling code seems to check `Data::same` before calling this method so I just removed the extra checks. Alternatively we could skip empty edits in the impl of `EditableText` for `Arc<String>`?

### DIFF
--- a/druid/src/text/editable_text.rs
+++ b/druid/src/text/editable_text.rs
@@ -196,7 +196,10 @@ impl EditableText for Arc<String> {
         <String as EditableText>::cursor(self, position)
     }
     fn edit(&mut self, range: Range<usize>, new: impl Into<String>) {
-        Arc::make_mut(self).edit(range, new)
+        let new = new.into();
+        if !range.is_empty() || !new.is_empty() {
+            Arc::make_mut(self).edit(range, new)
+        }
     }
     fn slice(&self, range: Range<usize>) -> Option<Cow<str>> {
         Some(Cow::Borrowed(&self[range]))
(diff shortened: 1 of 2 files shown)


cmyr:
The intent of that being a string compare instead of a data compare was to prevent unnecessary invalidation in cases like this, but maybe I was being overly clever?

Finnerale (commented):
Maybe it would make sense to change `Editor::data_is_stale` from
```rust
    fn data_is_stale(&self, data: &T) -> bool {
        self.layout.text().map(|t| !t.same(data)).unwrap_or(true)
    }
```
to
```rust
    fn data_is_stale(&self, data: &T) -> bool {
        self.layout.text().map(|t| t.as_str() != data.as_str()).unwrap_or(true)
    }
```
Instead? This would be more accurate I suppose.

andrewhickman:
Thinking about it some more, I don't think comparing the strings is sufficient because the style attributes might have changed.

Druid doesn't provide any implementations of `EditableText` with styling, but a user could create one. (I've created one that does syntax highlighting, for example)

cmyr:
@andrewhickman the `set_text` method should only care about the text; there should be some other way to invalidate styles.

andrewhickman:
I've changed it to use @Finnerale's suggestion of checking string equality in `data_is_stale`.

Finnerale (approved):
Looks good, thanks!

Issue #1323, opened 2020-10-16
Segmentation fault (core dumped) when using MenuDesc widget and svg feature


ivanceras:
I got `Segmentation fault (core dumped)` crash when I enabled `svg` feature and attaching menu to `WindowDesc`.

- Using `svg` feature while NOT attaching a menu to window run just fine.
- Attaching menu to window while NOT using svg will run just fine.

Replicated in 
- 63e34fa2 ( latest master, when this issue is filed)
- 93620d1 

To quickly replicate:
Add `svg` to the default feature in druid's `Cargo.toml`
```toml
[features]
default = ["gtk","svg"]
```
then run the example which uses menu:
```
cargo run --example multiwin
```
```log
DEBUG [druid::localization] available locales [], current en-US
DEBUG [druid::localization] resolved: [en-US]
INFO  [multiwin] Window added, id: WindowId(1)
Segmentation fault (core dumped)
```
OS used:
Ubuntu 18.04.5 LTS x86_64 
Kernel: 4.15.0-118-generic 

I wasn't able to replicate this in the published 0.6.0 crate.

jneem:
Thanks for the report! I'm having trouble reproducing, though, even with the exact same commit. Are you able to produce a backtrace?

ivanceras:
@jneem there was no backtrace.
I am also able to reproduce this problem on a freshly installed ubuntu 20.04

futurepaul:
I'm also getting this segfault with no backtrace after updating my app to master (was on 0.6). I'm on pop_os latest which is based on ubuntu 20.10.

justinmoon:
I also get this on Ubuntu 20.04, but not Windows 10

jneem:
Are any of you able to get a backtrace using gdb? I found some instructions [here](https://wiki.freepascal.org/Creating_a_Backtrace_with_GDB).

futurepaul:
I followed the gdb instructions, got this:
```
Starting program: /home/futurepaul/dev/rust/druid/target/debug/examples/multiwin-3d3680e1e9e4ce7a 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7ffff5e8a640 (LWP 192921)]
[New Thread 0x7ffff5689640 (LWP 192922)]
[New Thread 0x7ffff4e28640 (LWP 192923)]
[New Thread 0x7fffef100640 (LWP 192924)]
[...]

PR #1356, opened 2020-10-29
Try reverting usvg.

jneem:
@justinmoon can you try this one?


justinmoon (approved):
`cargo run --example multiwin --features svg` now works on Ubuntu, whereas `master` segfaults as described in https://github.com/linebender/druid/issues/1323. Great detective work @jneem!

justinmoon (approved):

PR #1357, opened 2020-10-29
Fixed broken doc links due to typo in container.rs

StarfightLP:
Noticed some broken links in the Container widget docs and fixed them.

Tested locally with `cargo doc` and Firefox.


jneem (approved):
Thanks! Looks like intra-docs links stabilized on [Sept 24](https://github.com/rust-lang/rust/pull/74430#issuecomment-698130298), so it should be in the next stable?

PR #1358, opened 2020-10-29
The gallery example now requires a new feature.

derekdreery:
The image create now requires a feature to be passed through to enable loading `png`s, which the gallery example uses.


jneem (approved):
Thanks!

PR #1359, opened 2020-10-29
Defer context menus on gtk and windows.

jneem:
Windows is currently untested -- I'll check it if CI passes.

### DIFF
--- a/druid-shell/src/platform/windows/window.rs
+++ b/druid-shell/src/platform/windows/window.rs
@@ -147,6 +147,7 @@ enum DeferredOp {
     SetSize(Size),
     SetResizable(bool),
     SetWindowState(window::WindowState),
+    ReleaseMouseCapture,
 }
 
 #[derive(Clone)]
@@ -561,6 +562,16 @@ impl MyWndProc {
                         }
                     }
                 }
+                DeferredOp::ReleaseMouseCapture => unsafe {
+                    if ReleaseCapture() == FALSE {
+                        let result = HRESULT_FROM_WIN32(GetLastError());
+                        // When result is zero, it appears to just mean that the capture was already released
+                        // (which can easily happen since this is deferred).
+                        if result != 0 {
+                            warn!("failed to release mouse capture: {}", Error::Hr(result));
+                        }
+                    }
+                },
             }
         } else {
             warn!("Could not get HWND");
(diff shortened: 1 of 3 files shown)


cmyr:
Tested on windows; I don't get the crash I would get otherwise, but i do get lots of warnings like,

```
WARN  [druid_shell::platform::windows::window] failed to release mouse capture: HRESULT 0x0: The operation completed successfully.
```

Not sure if this is a concern?

jneem:
I'm not sure either, but I noticed that also. I haven't had time to dig in yet, though.

jneem:
Ok, so I've disabled the warning. I'm not completely confident that it's the right thing, but my understanding of the situation is:
- the mouse goes down, and so we capture it
- the mouse goes up, and so we schedule a menu to appear when the handler is dropped. We also see that all mouse buttons were released, so we schedule a call to release the capture.
- we drop the handle, and run the context menu. While the menu is being run, the capture gets released somehow.
- we finish running the menu, and try to drop the capture but we get an error because there isn't a mouse capture active.

I'm not sure if there is much to be done here besides disabling the warning when we try to release a non-existent capture. According to the previous comments in the source, we need to defer the capture. And once we've done that, there isn't really any way to stop the capture being released before we actually run the deferred thing.

cmyr:
I guess I don't even really understand what the capture is doing, in this case? Maybe whoever is managing the menu handles capture?

jneem:
Yeah, it's pretty mysterious to me. The usual code path is that we take capture when the first button goes down and release it when the last button goes up. But when the menu pops up in between those two things happening, I think that we lose control over the event loop? And then I wasn't able to track what's going on. Maybe someone more familiar with windows programming can weigh in...

druid-shell/src/platform/windows/window.rs
```diff
@@ -1003,25 +1060,13 @@ impl WndProc for MyWndProc {
                             s.handler.mouse_down(&event);
                         } else {
                             s.handler.mouse_up(&event);
-                            should_release_capture = s.exit_mouse_capture(button);
+                            if s.exit_mouse_capture(button) {
+                                self.handle.borrow().defer(DeferredOp::ReleaseMouseCapture);
+                            }
```
cmyr, line 1065: I'm curious what happens if we continue to release here like normal? Would we have already handled the deferred `ContextMenu` action?

jneem: Prior to this PR, we would do the context menu synchronously in `handler.mouse_up` above. As it is in this PR right now, we run the context menu just below this, at the closing brace on line 1067.

The latest changes to `ReleaseCapture` don't actually change the ordering between the context menu and releasing the capture. Before the changes, we ran the context menu when dropping the handle borrow, and then a few lines later we have the release capture. After the changes, we run the context menu *and* the release capture (in that order) when dropping the handle borrow. The main reason for the changes to `ReleaseCapture` in the latest patch is that we have a general mechanism for deferring now, so I figured we should use it. The only actual behavior change is the silencing of the warning.

cmyr: i'll experiment a bit.

cmyr (approved):
Okay after closer reading I think we were just incorrect to treat the `0` return value as an error. I'm happy with this going in, and we will continue to iterate in any case. Thanks!

PR #1361, opened 2020-10-30
Allow setting the cursor in the update method.

jneem:
Motivation is that I have a widget that needs to change its cursor in response to the app data, but the data can get changed by keyboard shortcuts that don't cause an event to the relevant widget (e.g. keyboard shortcuts for menu entries).


cmyr (approved):
Sounds reasonable. Ultimately I don't love how we handle custom cursors, but it works for now.

Edit to clarify: what I don't like is how we set it on every event, which feels inelegant. It would be nice if a widget could just indicate what cursor should be used when it is hot, and then it could override this as needed? or something?

JAicewizard:
yeah having to set it explicitly every event seems off and confusing for newcomers as well.

jneem:
Yeah, that makes total sense. I'm going to go ahead and push this, though, and I'll open an issue for the other thing.

PR #1363, opened 2020-10-31
Rework custom widget

JAicewizard:
I mostly added some helpful comments and a z-index example. I couldn't get the .draw_text example to work (yet) but I will try some more. I also made sure that the beizier curves don't go outside the minimum size druid of druid.


JAicewizard:
Added the second way of drawing text, I think this covers basically all methods on `PaintCtx` now and pretty well actually.

druid/examples/custom_widget.rs
```diff
@@ -51,54 +55,92 @@ impl Widget<String> for CustomWidget {
...
+        }
     }
 
     // The paint method gets called last, after an event flow.
     // It goes event -> update -> layout -> paint, and each method can influence the next.
     // Basically, anything that changes the appearance of a widget causes a paint.
     fn paint(&mut self, ctx: &mut PaintCtx, data: &String, env: &Env) {
-        // Let's draw a picture with Piet!
-
+        println!("{:?}", ctx.size());
```
jneem, line 75: Did you mean to put the println here?

jneem (commented):
Looks good, I think you might just have left in a println by mistake.

druid/examples/custom_widget.rs
JAicewizard: no, that was me playing around with the image size.

jneem (approved):

PR #1364, opened 2020-11-01
Re-design dynamic length lists.

derekdreery:
I ended up redesigning this widget, moving it away from how `Flex` works. It could replace `List`, live alongside it, or live out of tree in the project I'll use it in (`jack-mixer`).

## Original notes

I need this for some work that I'm doing where I want a horizontal dynamic list with spacing. It might need some tidying before merge (i.e. the TODOs).

A future addition would be `{Main|Cross}AxisAlignment`. However I'm not sure if you actually need `MainAxisAlignment` since you can do stuff with wrapping widgets.

### DIFF
--- a/druid/src/widget/flex.rs
+++ b/druid/src/widget/flex.rs
@@ -262,7 +262,12 @@ impl Axis {
     }
 
     /// Generate constraints with new values on the major axis.
-    fn constraints(self, bc: &BoxConstraints, min_major: f64, major: f64) -> BoxConstraints {
+    pub(crate) fn constraints(
+        self,
+        bc: &BoxConstraints,
+        min_major: f64,
+        major: f64,
+    ) -> BoxConstraints {
         match self {
             Axis::Horizontal => BoxConstraints::new(
                 Size::new(min_major, bc.min().height),
@@ -789,7 +794,7 @@ impl CrossAxisAlignment {
     /// Given the difference between the size of the container and the size
     /// of the child (on their minor axis) return the necessary offset for
     /// this alignment.
-    fn align(self, val: f64) -> f64 {
+    pub(crate) fn align(self, val: f64) -> f64 {
         match self {
             CrossAxisAlignment::Start => 0.0,
             // in vertical layout, baseline is equivalent to center
@@ -799,7 +804,7 @@ impl CrossAxisAlignment {
     }
 }
 
-struct Spacing {
+pub(crate) struct Spacing {
     alignment: MainAxisAlignment,
     extra: f64,
     n_children: usize,
@@ -813,7 +818,7 @@ impl Spacing {
     /// this returns an iterator of `f64` spacing,
     /// where the first element is the spacing before any children
     /// and all subsequent elements are the spacing after children.
-    fn new(alignment: MainAxisAlignment, extra: f64, n_children: usize) -> Spacing {
+    pub(crate) fn new(alignment: MainAxisAlignment, extra: f64, n_children: usize) -> Spacing {
         let extra = if extra.is_finite() { extra } else { 0. };
         let equal_space = if n_children > 0 {
             match alignment {
--- a/druid/src/widget/list.rs
+++ b/druid/src/widget/list.rs
@@ -12,60 +12,104 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! Simple list view widget.
+//! List view widget.
+
+mod iter;
 
 use std::cmp::Ordering;
 use std::f64;
-use std::sync::Arc;
 
+pub use self::iter::ListIter;
 #[cfg(feature = "im")]
 use crate::im::Vector;
-use crate::kurbo::{Point, Rect, Size};
-use crate::widget::Axis;
+use crate::kurbo::{common::FloatExt, Point, Rect, Size};
+use crate::widget::{flex::Spacing, Axis, CrossAxisAlignment, MainAxisAlignment};
 use crate::{
     BoxConstraints, Data, Env, Event, EventCtx, LayoutCtx, LifeCycle, LifeCycleCtx, PaintCtx,
     UpdateCtx, Widget, WidgetPod,
 };
+
+#[derive(Copy, Clone)]
+pub enum ListMainAlignment {
+    /// Inherit options from `MainAxisAlignment`.
+    NoFlex(MainAxisAlignment),
+    /// Force children to use up all space evenly.
+    ///
+    /// Spacing will be added between items, but not at ends. If you don't want spacing set it to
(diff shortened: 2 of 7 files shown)


futurepaul:
I can't speak to the flex innards but I'm very happy with the API. I already has a use where I needed a horizontal list with 1 pixel spacers and this works great for that.

derekdreery:
I'm really happy with how this turned out. I  welcome any comments, good or bad.

derekdreery:
I would probably squish before any commit - to get rid of the earlier prototypes that would just complicate the history.

derekdreery:
@futurepaul I encourage you to have a look at the new version.

druid/src/widget/flex.rs
```diff
@@ -799,7 +804,7 @@ impl CrossAxisAlignment {
     }
 }
 
-struct Spacing {
+pub(crate) struct Spacing {
```
derekdreery, line 807: I actually ended up not using this, so it could go back to how it was.

druid/src/widget/list.rs
```diff
@@ -12,194 +12,198 @@
...
     UpdateCtx, Widget, WidgetPod,
 };
 
-/// A list widget for a variable-size collection of items.
-pub struct List<T> {
-    closure: Box<dyn Fn() -> Box<dyn Widget<T>>>,
-    children: Vec<WidgetPod<T, Box<dyn Widget<T>>>>,
+/// How to space out elements of the list
+#[derive(Debug, Copy, Clone)]
+pub enum Spacing {
```
cmyr, line 33: This spacing stuff is my primary current hesitation with this patch, but my concern is mostly a *feeling*; i'll try and express it best I can.

I *do* like the idea of having some mechanism for better expressing default item spacing, but something in this design feels slightly garbled, slightly un-systematic. This mostly comes down to the concept of `flex`, and what this would mean in this context. My basic concern is that the `flex_items` stuff feels slightly bolted on.

Maybe a good way to explore this is for me to state some things I believe and then the conclusions that leads me to, and if there's anything that doesn't sound right to you, you can point that out and we can talk about it?

- I think that a list should almost always be expected to exist in a `Scroll` container.
- By definition, a `Scroll` container provides an infinite constraint on the scrolling axis.
- A `Scroll` (well, now a `ClipBox`) container also provides `0` as its minimum size. (Maybe this is wrong? If, for instance, the `Scroll` itself has a non-zero incoming min constraint, that should be passed on to its child?)
- The idea of a `flex` layout is about determining how to distribute extra space.
- A list should never have 'extra' space, since the size of the list is purely a function of the size of the children.

Given this, flex in lists just doesn't make sense to me, but I do think I'm missing *something* here, so maybe if you can help me understand the rationale a bit better we can come to an understanding?

I *do* think there's a good argument that `Scroll` should not necessarily have no minimum constraint, which would go a long way to making this make more sense to me.

In any case, any clarification would be very helpful!

cmyr (commented):
Okay I'm going to focus just now on the high-level design, and in particular the question of what 'flex' should mean in the context of a list, which isn't clear to me; I elaborate more on a comment inline.

Thanks for your patience with this, happy to talk it out!

druid/src/widget/list.rs
derekdreery: I think I might be thinking ahead to a theoretical future where we also have wrapping, so it's actually the cross axis that scrolls, not the main. In that case flexing the main axis makes sense.

I do understand what you're saying re. expecting a list to be scrolled. It would be nice to have choice about how to arrange the widgets in the case where they all fit in the clip box, but my patch doesn't work like that now - it will fail when the user asks for flex and unbounded main constraint. Maybe the solution is to finish the job and implement wrapping as well.

derekdreery: Or alternatively remove all the flex stuff, but it feels like it will be less effort to implement wrapping.

cmyr: yea honestly I think that having v/h lists is of marginal benefit, but having a collection view that can act like a h-list makes total sense, and has a bunch of other benefits as well, I would love to see this go in that direction.

derekdreery: Ok I'm going to spew out my thinking for a kind of `WrappingList`.

What would be neat is if we could have some kind of wrapping collection, and then the way to avoid the wrapping was to wrap the widget in something with unbounded constraints on the main axis (i.e. `Scroll`). 

 1. The parent asks the children to choose their sizes, with the (0, parent_max) on the main axis and (0, infty) on the cross axis (i.e. cross axis is unbounded). Provide a knob on the widget to bound the cross-axis if desired (e.g. to make everything the same size).
    - We don't want any children bigger than the parent, because then we wouldn't be able to lay them out even with wrapping.
 2. As part of sizing the children, also decide which line each child goes on. The rule is that every time there isn't enough space on the previous line, a new line is started.
 3. Draw the children, aligning them based on a `MainAxisAlignment` parameter of the widget.

derekdreery: I've experimented with wrapping list-like widgets (see my `list_wrap` branch). I've kinda convinced myself that the original idea might have merit. My argument is:

There are times where you want a flex layout, but you don't know at compile time exactly how many items there will be. Say they are defined is a config file (as is the case with `mixjack`). Without the flex options, you have no way of laying out in this kind of situation. Without the patch I made, I'm not sure how I would do layout.

cmyr: There's not *no* way; you can always just dynamically rebuild a flex layout, based on some data, There's an example of this in the `flex` example. This would be the approach I would suggest for this case.

In terms of a layout API for a grid/collection, I would consider taking a closer look at some other similar sorts of collections, like the flutter `GridView` or UIKit's `UICollectionView`.

there are a few ways to handle this, from an API perspective. one way is to specify a number of columns, and then you have an arbitrary number of rows, based on how many items you have. Another approach would be to specify a max width, and then you just lay items out until you hit that width, and then start a new row.

These are huge simplifications, of course. in UIKit, there's an entire class that is responsible for describing how your items should be laid out in a collection view, [`UICollectionViewLayout`](https://developer.apple.com/documentation/uikit/uicollectionviewlayout).

One interesting thing about doing a thorough collection view API is that `List` just becomes a specialization of that.

derekdreery: Maybe in the short term I could just add axis and spacing to list.

cmyr: I think that would also be reasonable, with spacing just being a `KeyOrValue<f64>`.

derekdreery: Ok I will make a new simple patch, and then turn this patch into an exploration of a collection widget, with discussion on zulip.

PR #1365, opened 2020-11-01
RFC: Virtual scrolling list

jpochyla:
PoC of a virtualized scrolling container. Tries to be as simple as possible, so only vertical scrolling is supported for now, and rows are constant-height. Efficient iteration of large collections is done through a new `RangeIter` trait.

It needs two small changes to druid:

- `request_update_child` that takes a child WidgetPod. It was proposed in the original `request_update` PR but decided against (AFAIK because a use-case wasn't available).
- Exposing child `WidgetPod` in `ClipBox`, so the scrolling wrapper can request updates of the list container.

**Problems:**
- All the data invalidation paths need to be reviewed and tested more throughly, I'm not confident that i.e. `update` or `layout` can't be missed.
- Code is duplicated with `Scroll` and `List` widgets.

**Todo:**
- ~~Example~~
- Better docs

What do you think?

### DIFF
--- a/druid/src/widget/virt_list.rs
+++ b/druid/src/widget/virt_list.rs
@@ -1,0 +1,401 @@
+// Copyright 2020 The Druid Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Widget that displays a long collection of items without keeping all item
+//! widgets in memory.
+
+use std::cmp::Ordering;
+use std::ops::Range;
+use std::sync::Arc;
+
+#[cfg(feature = "im")]
+use crate::im::Vector;
+
+use crate::kurbo::{Point, Rect, Size, Vec2};
+
+use crate::scroll_component::ScrollComponent;
+use crate::widget::ClipBox;
+use crate::{
+    BoxConstraints, Data, Env, Event, EventCtx, LayoutCtx, LifeCycle, LifeCycleCtx, PaintCtx,
+    Selector, UpdateCtx, Widget, WidgetPod,
+};
+
+/// A virtualized scrolling widget for a (possibly large) collection of items.
+pub struct VirtList<C, T> {
+    clip: ClipBox<C, VirtListInner<T>>,
+    scroll_component: ScrollComponent,
+}
+
+impl<C: RangeIter<T>, T: Data> VirtList<C, T> {
+    /// Create a new vertical list widget. Closure will be called every time when a new child
+    /// needs to be constructed. Child height needs to be constant.
+    pub fn vertical<W: Widget<T> + 'static>(
+        child_height: f64,
+        closure: impl Fn() -> W + 'static,
+    ) -> Self {
+        let inner = VirtListInner::new(
+            Size::new(0.0, child_height),
+            Box::new(move || Box::new(closure())),
+        );
+        Self {
+            clip: ClipBox::new(inner)
+                .constrain_vertical(false)
+                .constrain_horizontal(true),
+            scroll_component: ScrollComponent::new(),
+        }
+    }
+}
+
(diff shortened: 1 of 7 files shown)


jpochyla:
One thing I'd like to get rid of is the tight coupling between the scrolling container and the virtual list itself, to support situations like multiple virtual lists side-by-side in one viewport. I'm not really sure how to approach this, as there can be arbitrary number of widgets between the scroll and the list. @rjwittams could `Scope` be a possible solution here?

cmyr (commented):
My head is a bit fuzzy right now but want to at least acknowledge this: the code looks great.

My understanding is that this implements the same functionality as is currently available in `List`, but with the potential to perform better for very large collections? I don't think we will want to provide two different variants of the same basic functionality, but I could imagine this replacing the current `List` if that was something that made sense.

jpochyla:
Thanks for looking into this! Current `List` supports items of variable height, that is one major difference. I'm not sure how a variable-height virtualized list widget would work, although it exists elsewhere and should be possible, but I think that very large collections will usually have a constant row height (table or a grid). Another difference would be the inherent trade-off between creating all widgets beforehand and having best-performing scroll (`List`), and updating only the visible window, saving memory and widget traversal, but paying the cost when scrolling (`VirtList`).

cmyr:
oh right, I wanted to address this:

I think that variable height should be at least an option, with the caveat that it will have performance implications. If the user wanted variable height, we could ask them to provide some sort of 'height hint' closure that would take the data and return an `f64`, and then we would avoid needing to instantiate widgets and doing a fully layout call?

cmyr:
@jpochyla What are your current thoughts on this?

jpochyla:
I think that adding an option for variable-height rows is totally doable and I can implement it. But I'm unhappy about the tight coupling between the scrolling wrapper (`VirtList`) and the actual list container (`VirtListInner`). It's fine for a single scrolling list, but in my project I'd like to have two lists side-by-side with a single scrollbar, with some other widgets in the middle (`Flex`, padding, and so on). I had one idea about how to solve it and simplify virtual scrolling further, but it needs another changes in Druid:

- removing the special scrolling wrapper (`VirtList`)
- propagating the visible rect down the tree from regular `Scroll`, maybe replacing the `set_viewport_offset` with it.

I was paying attention to your notification work, as a system for reporting generic information up the tree, as this situation is similar, although the direction is reversed. Maybe it could as well be implemented with `Scope`, passing the viewports in app data and having `VirtList` accept `(Rect, impl RangeIter)`.

What do you think? I will work on this further today and try to implement one of the proposals.

druid/src/widget/virt_list.rs
```diff
@@ -0,0 +1,401 @@
...
+    fn paint(&mut self, ctx: &mut PaintCtx, data: &C, env: &Env) {
+        data.for_in(self.visible_data_range(data), |child_data, i| {
+            self.child(i).paint(ctx, child_data, env);
+        });
+    }
+}
+
+/// This iterator-like trait enables using `VirtList` with any kind of `Data`,
+/// with efficient iteration of sub-ranges.
+pub trait RangeIter<T>: Data {
```
rjwittams, line 290: I'm not quite sure why you want ranges here, rather than direct indexed access from outside. 
Its similar to my IndexedItems in druid_table.  
If you think about sorting you can see why random access is the expected case. 
https://github.com/rjwittams/druid_table/blob/123b2d9ca8a446a551b6c2edf87c87075dea5bce/src/data.rs#L12

Also associated type for the iterated item (T) should be preferred - as you are not going to implement RangeIter for different T on the same type it will make complex inference situations simpler. (Worth looking round the Rust library types on this kind of thing).

jpochyla: > I'm not quite sure why you want ranges here, rather than direct indexed access from outside.

Well, the access is sequential, `narrow()`ing `im::Vector` is faster than direct `.get()`, and the index irrelevant, so this seemed natural.

> as you are not going to implement RangeIter for different T on the same type

Actually I do that, but I agree it probably wasn't the intended case.

rjwittams:
Maybe this woul

rjwittams:
Sorry, slipped!

rschulman:
I'm running into a situation where I have a Scroll containing a list that can have anywhere up to 10k or more items, and the MouseMove events alone are bogging down the app, so a widget along these lines is something I very much need. Is there any interest in revisiting something along these lines?

rschulman:
Would druid be up for taking a PR that only makes the changes to `core.rs` and `contexts.rs` that this PR proposes? I believe everything else could be done in one's own custom widgets, and those two changes seem fairly minor.

PR #1367, opened 2020-11-03
Ensure that WindowHandle and IdleHandle have right Send/Sync markers

jneem:
This (hopefully) clears up some unsoundness in the GTK backend, and also enforces cross-platform Send/Sync markers for WindowHandle and IdleHandle.

### DIFF
--- a/druid-shell/src/platform/gtk/window.rs
+++ b/druid-shell/src/platform/gtk/window.rs
@@ -90,6 +90,8 @@ macro_rules! clone {
 #[derive(Clone, Default)]
 pub struct WindowHandle {
     pub(crate) state: Weak<WindowState>,
+    // Ensure that we don't implement Send, because it isn't actually safe to send the WindowState.
+    marker: std::marker::PhantomData<*const ()>,
 }
 
 /// Builder abstraction for creating new windows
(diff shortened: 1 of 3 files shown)


druid-shell/src/platform/gtk/window.rs
```diff
@@ -90,6 +90,8 @@ macro_rules! clone {
 #[derive(Clone, Default)]
 pub struct WindowHandle {
     pub(crate) state: Weak<WindowState>,
+    // Ensure that we don't implement Send, because it isn't actually safe to send the WindowState.
+    marker: std::marker::PhantomData<*const ()>,
```
cmyr, line 94: I think we should enforce this for all platforms, to ensure the behaviour is consistent, so maybe put it in the common `WindowHandle`, not the platform ones?

cmyr (commented):
I think this is the sort of thing we need to enforce for all platforms, let me know if you think I've missed something?

druid-shell/src/platform/gtk/window.rs
jneem: Yes, that's the purpose of the static assert on the common WindowHandle: to enforce the cross-platform behavior. It just so happens that only the GTK one needed an extra marker.

cmyr (approved):
Great, thanks!

PR #1370, opened 2020-11-03
Allow user to specify different min-sizes for each side of a Split.

bheisler:
Right now the Split widget only accepts a single "min_size: f64" value which is applied to both halves of the split. In situations where the two child widgets have different min-sizes, the user has no way to specify this, leading to one component being overlarge or the other sometimes forced off the edge of the Split widget. This patch modifies the relevant functions to accept separate min-size values for the two child widgets.


Zarenor (approved):
Looks straightforward, and like the easy way to do this.
Thanks!

PR #1372, opened 2020-11-04
(slightly) rework either

JAicewizard:
This is a very simple example with a very simple change.
I am thinking about just removing al the localization from every example and make a special example with that, it is very nice someone translated all of them but a dedicated example probably works best. This could showcase things like localization arguments etc that are not really shown off in any of the current examples.


druid/examples/either.rs
```diff
@@ -12,42 +12,43 @@
...
+        Label::new("Click to reveal slider").padding(5.0),
     );
-    col.add_child(either);
-    col
+    Flex::column().with_child(button).with_child(either)
+}
+
+pub fn main() {
+    let main_window =
+        WindowDesc::new(ui_builder).title(druid::widget::LabelText::from("Switcheroo"));
```
cmyr, line 49: you should be able to just write `.title("Switcheroo")`.

druid/examples/either.rs
```diff
@@ -12,42 +12,43 @@
...
+        .lens(AppState::which)
+        .padding(5.0);
 
-    let mut col = Flex::column();
-    col.add_child(
-        Checkbox::new("Toggle slider")
-            .lens(AppState::which)
-            .padding(5.0),
-    );
+    // The either widget has 2 children, one of which is visible.
```
cmyr, line 35:
```suggestion
    // The Either widget has two children, one of which is visible at a time.
```

english convention varies, but a good guideline for technical writing is to write numbers under 10 (so up to nine, inclusive 😉) as words.

druid/examples/either.rs
```diff
@@ -12,42 +12,43 @@
...
+    // you have to pass in a closure which takes `Data` and and `Env`.
+    // This closure determines which widget gets displayed based on the
+    // return value (a bool). False is the second widget and true is the first.
```
cmyr, line 38:
```suggestion
    // The `Either` widget has two children, only one of which is visible at a time.
    // To determine which child is visible, you pass it a closure that takes the
    // `Data` and the `Env` and returns a bool; if it returns `true`, the first
    // widget will be visible, and if `false`, the second.
```

english convention varies, but a good guideline for technical writing is to write numbers under 10 (so up to nine, inclusive 😉) as words.

cmyr (commented):
Few little things, but looks good!

druid/examples/either.rs
JAicewizard: Yeah it is dutch convention too for under 20, I usually forget about it.
If anything I prefer just numbers as numbers to keep consistency.

druid/examples/either.rs
JAicewizard: If my writing is very bad let me know, I can put more effort into it.
I don't have a preference here but maybe you view that completely different.

druid/examples/either.rs
```diff
@@ -12,42 +12,43 @@
...
+        .padding(5.0);
 
-    let mut col = Flex::column();
-    col.add_child(
-        Checkbox::new("Toggle slider")
-            .lens(AppState::which)
-            .padding(5.0),
-    );
+    // The Either widget has two children, one of which is visible at a time.
+    // you have to pass in a closure which takes `Data` and and `Env`.
```
cmyr, line 36:
```suggestion
    // you have to pass in a closure which takes `Data` and `Env`.
```

druid/examples/either.rs
```diff
@@ -12,42 +12,43 @@
...
+        Label::new("Click to reveal slider").padding(5.0),
     );
-    col.add_child(either);
-    col
+    Flex::column().with_child(button).with_child(either)
+}
+
+pub fn main() {
+    let main_window =
+        WindowDesc::new(ui_builder).title(druid::widget::LabelText::from("Switcheroo"));
```
cmyr, line 49:
```suggestion
        WindowDesc::new(ui_builder).title("Switcheroo");
```

druid/examples/either.rs
JAicewizard: :+1:
I am no longer behind my computer so I couldn't commit myself, thanks!

druid/examples/either.rs
JAicewizard: :+1:
I am no longer behind my computer so I couldn't commit myself, thanks!

cmyr (approved):
Thanks!

Issue #1140, opened 2020-08-18
Add a textbox for numeric values.


JAicewizard:
Right now there is no way to have custom verify functions, or prevent people from typing in non-numeric characters.
It would be great if there was some kind of way to have textboxes that only allow numeric content.
It would be amazing if this could be implemented in such a way where the textbox implements `Widget<uxx/fxx>`

Finnerale:
Maybe [`Parse`](https://docs.rs/druid/0.6.0/druid/widget/struct.Parse.html) is what you are looking for? There also is an [example](https://github.com/linebender/druid/blob/master/druid/examples/parse.rs) for it.

JAicewizard:
It is sort of, but it doesn't prevent anyone from typing in non-numeric characters.
I can still type in anything, it will just result in `None(0)` as data.

Finnerale:
Hmm, in that case maybe a `Lens` filtering out non-numeric characters could work?
This is what I would imagine (just a sketch, haven't tested it):
```rust
struct NumericLens;
impl Lens<String, String> for NumericLens {
    fn with<V, F: FnOnce(&String) -> V>(&self, data: &String, f: F) -> V {
        f(data)
    }
    fn with_mut<V, F: FnOnce(&mut String) -> V>(&self, data: &mut String, f: F) -> V {
        let v = f(data);
        data.retain(|c| c.is_digit(10));
        v
    }
}
```

JAicewizard:
I got something to work using:
```
struct StringLens;
impl Lens<AppState, String> for StringLens {
    fn with<V, F: FnOnce(&String) -> V>(&self, data: &AppState, f: F) -> V {
        let mut string = data.icons.to_string();
        if data.icons == 0 {
            string = "".into();
        }
        f(&string)
    }
    fn with_mut<V, F: FnOnce(&mut String) -> V>(&self, data: &mut AppState, f: F) -> V {
        let mut string = data.icons.to_string();
        if data.icons == 0 {
            string = "".into();
        }
        let v = f(&mut string);
        string.retain(|c| c.is_digit(10));
        data.icons = string.parse().unwrap_or(0);
        v
    }
}
```

This does exactly what I want, it also allows me to have a slider point to the same value.
It is however not very elegant, and numeric input is something a lot of applications will end up needing in some way or another.
I would leave this issue open for future development, but if you feel this is good enough support feel free to close this.
If you do close this, it would be nice to have an example specifically for this, possibly linked in the textbox documentation.

[...]

PR #1377, opened 2020-11-06
Add Formatter trait for validating & formatting text

cmyr:
This is a mechanism for handling conversions back and forth
between values and their textual representations. Its major
use case is managing user input in text fields when the input
is intended to represent some other value type.

This patch includes the trait, a simple implementation based
on `FromStr` (intended as a replacement for the `Parse` widget)
and a new ValueTextBox widget that uses some Formatter to manage
a TextBox.

The behaviour of the ValueTextBox differs from the normal TextBox
in one very important way; it does not attempt to update the data
until editing has 'completed'. Editing is completed when the user
presses <kbd>return</kbd> or attempts to navigate away from the textbox;
the user can also manually cancel editing by pressing <kbd>esc</kbd>, which
discards the edit.

These interactions (especially the hard-wired key handling) feels
a bit brittle, and I'd like to try and find a more systematic
approach.

This is *very* strongly inspired by Cocoa's [`Formatter`](https://developer.apple.com/documentation/foundation/formatter#) class.

update:

This patch led me down a rabbit hole, because it encounters a bunch of situations that are hard to deal with given druid's current data model. In this particular case, this is related to error handling; the solution here is not satisfactory, (it involves setting a callback on the textbox and then manually sending commands as errors occur) but it makes error handling at least *possible*, until we're able to come up with a more robust, longer-term solution.

I intend to write up the problems i encountered and how I'm thinking about them in a subsequent document.

This now replaces the `parse` example with an example *package*, `validation`, which can  be run with,

```sh
cargo run --package=validation
```

progress on #1140

### DIFF
--- a/druid/src/text/format.rs
+++ b/druid/src/text/format.rs
@@ -107,6 +107,8 @@ pub struct ValidationError {
 ///
 /// [`Formatter`]: Formatter
 /// [`FromStr`]: std::str::FromStr
+#[non_exhaustive]
+//TODO: take a fmt string (like `{:.2}` to allow user to customize output
 pub struct ParseFormatter;
 
 impl Validation {
@@ -159,6 +161,13 @@ impl ValidationError {
     }
 }
 
+impl ParseFormatter {
+    /// Create a new `ParseFormatter`.
+    pub fn new() -> Self {
+        ParseFormatter
+    }
+}
+
 impl<T> Formatter<T> for ParseFormatter
 where
     T: FromStr + std::fmt::Display,
@@ -189,5 +198,11 @@ impl std::fmt::Display for ValidationError {
 impl std::error::Error for ValidationError {
     fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
         Some(&*self.inner)
+    }
+}
+
+impl Default for ParseFormatter {
+    fn default() -> Self {
+        ParseFormatter
     }
 }
--- a/druid/src/text/format.rs
+++ b/druid/src/text/format.rs
@@ -1,0 +1,193 @@
+// Copyright 2020 The Druid Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Creating, interpreting, and validating textual representations of values.
+
+use std::str::FromStr;
+use std::sync::Arc;
+
+use super::Selection;
+use crate::Data;
+
+/// A trait for types that create, interpret, and validate textual representations
+/// of values.
+///
+/// A formatter has two responsiblities: converting a value into an appropriate
+/// string representation, and attempting to convert a string back into the
+/// appropriate value.
+///
+/// In addition, a formatter performs validation on *partial* strings; that is,
+/// it determines whether or not a string represents a potentially valid value,
+/// even if it is not currently valid.
+pub trait Formatter<T> {
+    /// Return the string representation of this value.
+    fn format(&self, value: &T) -> String;
+    /// Return the string representation of this value, to be used during editing.
+    ///
+    /// This can be used if you want the text to differ based on whether or not
+    /// it is being edited; for instance you might display a dollar sign when
+    /// not editing, but not display it during editing.
+    fn format_for_editing(&self, value: &T) -> String {
+        self.format(value)
+    }
+
+    /// Determine whether the newly edited text is valid for this value type.
+    ///
+    /// This always returns a [`Validation`] object which indicates if
+    /// validation was successful or not, and which can also optionally,
+    /// regardless of success or failure, include new text and selection values
+    /// that should replace the current ones.
+    ///
+    ///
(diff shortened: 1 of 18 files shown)


JAicewizard:
I feel that `format_for_editing` might be a bit useless of a method. If someone really wanted to have a different formatting for editing they could just as well use some kind of internal state for that. It might also be that the user wants to format the same value differently for 2 labels even if both are not editable.

I also dont think that `Formatter` is the right name, it does way more then just formatting. `Parser` sounds more accurate to me but that might clash with `.parse`.

druid/src/text/format.rs
```diff
@@ -0,0 +1,208 @@
...
+    pub fn success() -> Self {
+        Validation {
+            result: Ok(()),
+            selection_change: None,
+            text_change: None,
+        }
+    }
+
+    /// Create a `Validation` with an error indicating the failure reason.
+    pub fn failure(err: impl std::error::Error + 'static) -> Self {
```
JAicewizard, line 125: Would also be nice if it had just a little documentation on what this does. Does this result in the edit being undone? Does something happen to the selection? Does the user see an error indicating they typed something wrong? (probably not the later 2 but it should be clear)

JAicewizard (commented):
My previous comment still stands.

In addition to that I think it might be useful to have some easy "undo", but that might be a validation error?

JAicewizard:
I forgot to put this in, but I would also change the name of the example, then we keep the 'validation' name open for a more simplified example.

druid/src/text/format.rs
```diff
@@ -0,0 +1,208 @@
...
+    ///
+    /// [`format`]: #tymethod.format
+    fn value(&self, input: &str) -> Result<T, ValidationError>;
+}
+
+/// The result of a [`Formatter`] attempting to validate some partial input.
+///
+/// [`Formatter`]: Formatter
+pub struct Validation {
+    result: Result<(), ValidationError>,
```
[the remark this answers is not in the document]
cmyr: This whole `Validation` struct is sort of an augmented result; it includes not only the `Result` itself, but also additional information that exists both in the success and failure cases. It would definitely be possible to store this as an `Option`, but I think keeping it as a `Result` more clearly reflects the semantics of the program.

druid/src/text/format.rs
cmyr: This is a function that creates a `Validation` object with a given inner error. In the context of text editing, how the textbox reacts when validation fails is dependent on how that text box is configured.

cmyr (commented):
>I feel that `format_for_editing` might be a bit useless of a method. If someone really wanted to have a different formatting for editing they could just as well use some kind of internal state for that. It might also be that the user wants to format the same value differently for 2 labels even if both are not editable.

`format_for_editing`: I can see how this method might seem a bit superfluous, but I think it will be used in common cases; for instance it is common to edit a currency value, but to not want to let the user edit the currency symbol, and I can't think of a better way of doing this. If I'm totally honest, this is pretty much directly a copy of the API of the Cocoa Formatter class, and I sort of trust that they thought this through and decided this was the best option.

Different formats for same value: this is the beauty of our approach here; you can format the same value in different ways by providing different `Formatter` instances.

> I also dont think that Formatter is the right name, it does way more then just formatting. Parser sounds more accurate to me but that might clash with .parse.

This is a challenge; the `Formatter` does both formatting and validation. You can think of it as a type that handles the conversion between a value and a string; I think that `Parser` suffers a similar problem in that it only captures half of the job. You could imagine other names like `Stringify` or something, but that's also pretty incomplete. 😒

JAicewizard:
> `format_for_editing`: I can see how this method might seem a bit superfluous, but I think it will be used in common cases; for instance it is common to edit a currency value, but to not want to let the user edit the currency symbol, and I can't think of a better way of doing this. If I'm totally honest, this is pretty much directly a copy of the API of the Cocoa Formatter class, and I sort of trust that they thought this through and decided this was the best option.
> 
> Different formats for same value: this is the beauty of our approach here; you can format the same value in different ways by providing different `Formatter` instances.

It almost seems like these are 2 completely separate concepts, one if of displaying text, like the `Display` trait but druid specific, and one is actually just a fancy lens specifically for text editing. It would be easier to implement if it was 2 separate traits that are used for separate purposes. One for textboxes and one for labels. 

This also sparked the question in me: are custom lenses still useful, besides the derived ones? but thats something for zulip

> This is a challenge; the `Formatter` does both formatting and validation. You can think of it as a type that handles the conversion between a value and a string; I think that `Parser` suffers a similar problem in that it only captures half of the job. You could imagine other names like `Stringify` or something, but that's also pretty incomplete. unamused

If the trait would be split like suggested above, maybe `Editor`?? If the trait would be split the editable one would allow you do edit the value(editing via text ofc)

cmyr:
The thing is, they *do* need to work together; in the case of a text editing box, you format the text that populates the box, and then you validate the input as it occurs, so I don't think it makes a lot of sense to separate them :(


Custom lenses are very useful; they're used heavily in runebender.

JAicewizard:
I mean split them in editable formatting + validation and normal formatting, not splitting them editable formatting + normal formatting and validating.

cmyr:
I agree that you shouldn't need to worry about the editing stuff if you don't need it; that's why there's a default impl of `format_for_editing`, and if you don't implement it it just reuses the normal `format` method.

It's also worth thinking of this as a very low-level trait, that ultimately you shouldn't need to implement manually much, because we should provide default formatters for common types of values that handle things like localization.

One way to think about this `Formatter` trait is that it is the absolutely most general interface that encompasses a very broad range of possible behaviour, and most of the details and nicer API should be filled in by specific formatters further up the stack.

JAicewizard:
ah ok then I misunderstood the goal I think

cmyr:
okay, going to merge this, can shake out problems as they arise.

Issue #1308, opened 2020-10-13
IME tracking issue


raphlinus:
This is the tracking issue for [Input Method Editor] (IME) support in Druid. As of the time of writing, it's not an immediate priority, but increasingly other things depend on it, and we've collected a fair amount of information on the topic. Also note, #267 has some info on this topic, but we'll go into more depth here. For more color on IME, the blog [Text Editing Hates You Too] makes good reading.

## Cross-platform abstraction

The main work is creating a cross-platform abstraction for IME, which would be implemented in druid-shell. The main function of this abstraction includes:

* Allowing the IME to perform editing operations such as inserting text, changing the composition region, etc.

* Giving the IME enough context of the text being edited to be useful.

* Negotiating geometry so the IME can be displayed at the correct place on the screen. In general, the IME is displayed in its own window, not directly managed by the app. (See #1301, which is a manifestation of this requirement)

* Various lifecycle events and also dealing with language changes.

## TextBox integration

The cross-platform abstraction lives in druid-shell. Layered on top of that is support for IME-driven editing in the `TextBox` widget. The work can in theory be split into two separate subprojects, but it might be difficult to validate the shell work without some kind of working text editing widget.

There are also decisions to be made whether to plumb the raw druid-shell IME interface up so (potentially user-written) widgets can use them, or whether to do some wrapping at the druid layer to take care of boilerplate and provide a somewhat higher level interface. For keyboard events, we do more of the former, but there is plenty of precedent for the latter as well. A reasonable rule of thumb is that anything platform specific (and thus unsafe) belongs in shell, but if there is logic that is common across most platforms, that belongs more in druid.

## Existing work

There are two Rust GUI toolkits that have at least some IME support. The most complete appears to be tcw3, which is part of the [Stella2] repo. The cross-platform abstraction is in [iface.rs](https://github.com/yvt/Stella2/blob/master/tcw3/pal/src/iface.rs), but IME is not split out. The major relevant traits are `Wm::new_text_input_ctx`, `Wm::text_input_*`, `TextInputCtxListener`, and `TextInputCtxEdit`. Look under `tcw3/pal/src/*/textinput` for the implementations of those traits. The tcw3 subdirectory is MIT-licensed. If we feel we want to directly use any of that code, it should be possible to iron out any licensing related kinks.

[...]

PR #1378, opened 2020-11-07
macOS text input

lord:
This change proposes a new set of APIs for text input, as well as fully implements proper text input using those APIs on macOS. Other platforms have text input commands simulated from key down events — this won't handle IME input on those platforms, but will hopefully work as a temporary solution for those platforms until we can add IME support to them.

To test it out: `cd druid-shell; cargo run --example edit_text`

Sorry for the really long commit!! Let me know if I should split it up. Most of the changes should hopefully be boilerplate and comments, but there's still a fair amount of logic.

I have a few things I'm still unsure about, although I'm sure there's plenty more I've missed:

- this i think will allow for RTL/LTR, but do we ever want specially handled vertical text? i guess the macOS APIs don't really seem to handle it specifically, so maybe OK to ignore for now
- have not spent too much time considering whether this design works with async functions on android. i've also only barely looked into non-mac platforms for this, but the similarity between this design and tcw3's design makes me feel confident that it'll work on at least gtk and windows, which I think tcw3 has a full text input implementation for.
- this is a pretty complex interface! if folks have ideas for how it could be simplified, could be a good idea to try?
- are we happy with the taxonomy of `TextInputAction`s I've created? it seems maybe better to me than a completely flat list of commands in a single `enum`, but perhaps I've over-combined commands, like how vertical movement is shared between selection movements and scrolling. these actions are also a little macOS specific, although I've tried my best to directly implement some of the more esoteric commands like "transpose". tcw3 handles this by having a bunch of post-ime keybindings on other platforms that trigger the appropriate `TextInputAction`.
- not too happy with all the `TextInput` prefixes everywhere on basic types, but was necessary given how druid-shell puts everything in a single top-level namespace. if we think it's ok to have a `text_input_types` prefix or something like that, could make types like `TextSelection` and `TextInputUpdate` look a little prettier?
- this does not yet implement killing and yanking, triggered by control-k and control-y on macOS. hopefully this is ok to defer for now?
- there are no tests right now. is this ok? i do think in the long term, would be nice to have a standard test suite that's run on all `TextInputHandler` implementations to ensure they correctly manipulate the text state, deal with graphemes correctly, etc

See also the IME tracking issue: #1308

### DIFF
--- a/druid-shell/examples/edit_text.rs
+++ b/druid-shell/examples/edit_text.rs
@@ -118,7 +118,7 @@ impl WinHandler for AppState {
         piet.draw_text(layout, (0.0, 0.0));
 
         // draw caret
-        let caret_x = layout.hit_test_text_position(doc.selection.extent).point.x;
+        let caret_x = layout.hit_test_text_position(doc.selection.active).point.x;
         piet.fill(
             Rect::new(caret_x - 1.0, 0.0, caret_x + 1.0, FONT_SIZE),
             &CARET_COLOR,
@@ -205,16 +205,16 @@ impl InputHandler for AppInputHandler {
     fn replace_range(&mut self, range: Range<usize>, text: &str) {
         let mut doc = self.state.borrow_mut();
         doc.text.replace_range(range.clone(), text);
-        if doc.selection.anchor < range.start && doc.selection.extent < range.start {
+        if doc.selection.anchor < range.start && doc.selection.active < range.start {
             // no need to update selection
-        } else if doc.selection.anchor > range.end && doc.selection.extent > range.end {
+        } else if doc.selection.anchor > range.end && doc.selection.active > range.end {
             doc.selection.anchor -= range.len();
-            doc.selection.extent -= range.len();
+            doc.selection.active -= range.len();
             doc.selection.anchor += text.len();
-            doc.selection.extent += text.len();
+            doc.selection.active += text.len();
         } else {
             doc.selection.anchor = range.start + text.len();
-            doc.selection.extent = range.start + text.len();
+            doc.selection.active = range.start + text.len();
         }
         doc.refresh_layout();
         doc.composition = None;
@@ -284,9 +284,9 @@ fn apply_default_behavior(handler: &mut AppInputHandler, action: Action) -> bool
             };
             handler.set_selection(Selection::new_caret(updated_index));
         }
-        Action::MoveExtent(movement) => {
+        Action::MoveSelecting(movement) => {
             let mut selection = handler.selection();
-            selection.extent = match apply_movement(handler, movement, selection.extent) {
+            selection.active = match apply_movement(handler, movement, selection.active) {
                 Some(v) => v,
                 None => return false,
             };
@@ -296,7 +296,7 @@ fn apply_default_behavior(handler: &mut AppInputHandler, action: Action) -> bool
             let len = handler.len();
             let selection = Selection {
                 anchor: 0,
-                extent: len,
+                active: len,
             };
             handler.set_selection(selection);
         }
@@ -307,7 +307,7 @@ fn apply_default_behavior(handler: &mut AppInputHandler, action: Action) -> bool
         }
         Action::Delete(movement) => {
             let mut selection = handler.selection();
-            selection.extent = match apply_movement(handler, movement, selection.extent) {
+            selection.active = match apply_movement(handler, movement, selection.active) {
                 Some(v) => v,
                 None => return false,
             };
--- a/druid-shell/examples/edit_text.rs
+++ b/druid-shell/examples/edit_text.rs
@@ -31,9 +31,9 @@ use druid_shell::piet::{
 };
 
 use druid_shell::{
-    keyboard_types::Key, Affinity, Application, KeyEvent, Region, TextDirection, TextInputAction,
-    TextInputHandler, TextInputToken, TextInputUpdate, TextMovement, TextSelection,
-    VerticalMovement, WinHandler, WindowBuilder, WindowHandle,
+    keyboard_types::Key, text, text::Action, text::Event, text::InputHandler, text::Selection,
+    text::VerticalMovement, Application, KeyEvent, Region, TextInputToken, WinHandler,
+    WindowBuilder, WindowHandle,
 };
 
 use druid_shell::kurbo::{Point, Rect};
@@ -56,7 +56,7 @@ struct AppState {
 #[derive(Default)]
 struct DocumentState {
     text: String,
-    selection: TextSelection,
+    selection: Selection,
     composition: Option<Range<usize>>,
     text_engine: Option<PietText>,
     layout: Option<PietTextLayout>,
@@ -79,7 +79,7 @@ impl WinHandler for AppState {
     fn connect(&mut self, handle: &WindowHandle) {
         self.handle = handle.clone();
         let token = self.handle.add_text_input();
-        self.handle.set_active_text_input(Some(token));
+        self.handle.set_focused_text_input(Some(token));
         self.text_input_token = Some(token);
         let mut doc = self.document.borrow_mut();
         doc.text_engine = Some(handle.text());
@@ -95,6 +95,7 @@ impl WinHandler for AppState {
         piet.fill(rect, &BG_COLOR);
         let doc = self.document.borrow();
         let layout = doc.layout.as_ref().unwrap();
+        // TODO(lord): rects for range on layout
         if let Some(composition_range) = doc.composition.as_ref() {
             let left_x = layout
                 .hit_test_text_position(composition_range.start)
@@ -107,14 +108,8 @@ impl WinHandler for AppState {
             );
         }
         if !doc.selection.is_caret() {
-            let left_x = layout
-                .hit_test_text_position(doc.selection.upstream_index())
-                .point
-                .x;
-            let right_x = layout
-                .hit_test_text_position(doc.selection.downstream_index())
-                .point
-                .x;
+            let left_x = layout.hit_test_text_position(doc.selection.min()).point.x;
+            let right_x = layout.hit_test_text_position(doc.selection.max()).point.x;
             piet.fill(
                 Rect::new(left_x, 0.0, right_x, FONT_SIZE),
                 &SELECTION_BG_COLOR,
@@ -146,13 +141,11 @@ impl WinHandler for AppState {
             println!("user pressed c! wow! setting selection to 0");
 
             // update internal selection state
-            self.document.borrow_mut().selection = TextSelection::new_caret(0);
+            self.document.borrow_mut().selection = Selection::new_caret(0);
 
             // notify the OS that we've updated the selection
-            self.handle.update_text_input(
-                self.text_input_token.unwrap(),
-                TextInputUpdate::SelectionChanged,
-            );
+            self.handle
+                .update_text_input(self.text_input_token.unwrap(), Event::SelectionChanged);
 
             // repaint window
             self.handle.request_anim_frame();
@@ -163,8 +156,8 @@ impl WinHandler for AppState {
         false
     }
 
-    fn text_input(&mut self, _token: TextInputToken, _mutable: bool) -> Box<dyn TextInputHandler> {
-        Box::new(AppTextInputHandler {
+    fn text_input(&mut self, _token: TextInputToken, _mutable: bool) -> Box<dyn InputHandler> {
+        Box::new(AppInputHandler {
             state: self.document.clone(),
             window_size: self.size,
             window_handle: self.handle.clone(),
@@ -188,20 +181,20 @@ impl WinHandler for AppState {
     }
 }
 
-struct AppTextInputHandler {
+struct AppInputHandler {
     state: Rc<RefCell<DocumentState>>,
     window_size: Size,
     window_handle: WindowHandle,
 }
 
-impl TextInputHandler for AppTextInputHandler {
-    fn selection(&mut self) -> TextSelection {
+impl InputHandler for AppInputHandler {
+    fn selection(&mut self) -> Selection {
         self.state.borrow().selection.clone()
     }
     fn composition_range(&mut self) -> Option<Range<usize>> {
         self.state.borrow().composition.clone()
     }
-    fn set_selection(&mut self, range: TextSelection) {
+    fn set_selection(&mut self, range: Selection) {
         self.state.borrow_mut().selection = range;
         self.window_handle.request_anim_frame();
     }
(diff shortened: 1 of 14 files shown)


cmyr:
Looks like you're plugging away at this but if you think it needs a look or you have any other concerns I'm happy to do a review or have a chat.

lord:
thanks! maybe ~a day more of work and it'll be ready for a first look?

lord:
@cmyr added some more docs, and squashed a few bugs! want to take a peek?

although I've already written a bunch of code, still happy to make bigger structural changes to how the API works if we think it's a good idea!

druid-shell/src/platform/mac/text_input.rs
```diff
@@ -0,0 +1,679 @@
...
+        // "selectToMark:" => {}          // TODO(lord): extends the selection to include the mark
+        // "setMark:" => {}               // TODO(lord): remembers index in document (but what about grapheme clusters??)
+        // "swapWithMark:" => {}          // TODO(lord): swaps current and stored mark indices
+        // "yank:" => {}                  // TODO(lord): triggered with control-y, inserts in the text previously killed deleteTo*OfParagraph
+        "transposeWords:" => {} // textedit doesn't support, so neither will we
+        "changeCaseOfLetter:" => {} // textedit doesn't support, so neither will we
+        "indent:" => {}         // textedit doesn't support, so neither will we
+        "insertContainerBreak:" => {} // textedit and pages don't seem to support, so neither will we
+        "quickLookPreviewItems:" => {} // don't seem to apply to text editing?? hopefully
+        "complete:" => {}             // intercepted by the IME; if it gets to us, noop
```
cmyr, line 625: this one makes me do a weird face.

how and where does this get sent? do we actually receive this on return in some circumstances, or does the IME translate it to an 'insert newline' or something if there's no active IME session? If we're at all unsure it might at least be worth logging since it will be potentially a confusing thing to have to debug.

druid-shell/src/text_input.rs
```diff
@@ -0,0 +1,516 @@
...
+///
+/// As an example, if the input caret is at the start of the document `hello world`, we would expect both `anchor`
+/// and `extent` to be `0`. If the user holds shift and presses the right arrow key five times, we would expect the
+/// word `hello` to be selected, the `anchor` to still be `0`, and the `extent` to now be `5`.
+#[derive(Clone, Default, PartialEq, Eq, Hash)]
+pub struct TextSelection {
+    /// This is the end of the selection that stays unchanged while holding shift and pressing the arrow keys.
+    pub anchor: usize,
+    /// This is the end of the selection that moves while holding shift and pressing the arrow keys.
+    pub extent: usize,
```
cmyr, line 78: I haven't seen these names before, are they your creation? I definitely can see wanting to avoid terms like 'start' and 'end', but I do find `extent` a bit unclear, since it feels like a measure of length or size, where it is really just another position. I'm not sure what I'd prefer though, but maybe `anchor` and `active`?

druid-shell/src/text_input.rs
```diff
@@ -0,0 +1,516 @@
...
+    pub fn new_caret(index: usize) -> TextSelection {
+        TextSelection {
+            anchor: index,
+            extent: index,
+        }
+    }
+
+    /// Corresponds to the end of the selection — either the `anchor` or `extent` — that has a lesser byte index.
+    /// Because of bidirectional text, this is not necessarily "left".
+    pub fn upstream_index(&self) -> usize {
```
cmyr, line 94: I actually think that `min()` and `max()` are okay for these, but curious to hear your thoughts. I do get the idea of making the names distinct in order to make people think harder about the trickiness, but I'm not sure in this case?

druid-shell/src/text_input.rs
```diff
@@ -0,0 +1,516 @@
...
+    fn slice_bounding_box(&mut self, range: Range<usize>) -> Option<Rect>;
+
+    /// Applies some special action to the text field.
+    ///
+    /// Requries a mutable lock.
+    fn handle_action(&mut self, action: TextInputAction);
+}
+
+#[allow(dead_code)]
+/// Simulates `TextInputHandler` calls on `handler` for a given keypress `event`. This circumvents the platform, and so can't
```
cmyr, line 248: Rustdoc treats doc comments kind of like git treats commit messages; the first line is 'special', and acts as a sort of summary; for instance if you're in the docs for a module, where it lists the contents of that module, the first line is displayed beside each item.

In practice what this means is it makes sense to try and have a simple first line (like the first sentence, here) and then a newline followed by the more detailed docs.

cmyr (commented):
This is really impressive rob, and it definitely stretches my own understanding of the cocoa text system.

To address your specific questions first:


- vertical text: happy to put this on the "things to worry about in 2023" list.
- async functions on android: not sure how to think about this, but it feels ambitious, and it also doesn't need to be a major concern right now. The fact that there is prior-art for an API like this working on gtk & windows is very compelling, and enough to satisfy my general cross-platform concerns.
- complexity: I really don't see many places where this could be simplified, and I actually feel encouraged by its relative simplicity; this is going to be an API that is principally used internally, and given that I think it's pretty good? I'll have to actually see what's involved in making this work with druid proper, but I'm pretty encouraged.
- taxonomy: I think this is excellent, and very similar to how we did this in xi. I think sharing things like movement between selection and cursor movement makes total sense.
- `TextInput` prefixes: I would not be opposed to having a `text` module in druid-shell; that's how we ended up handling it in `druid` (and a lot of the druid types will end up being replaced by these new types, I think).
- kill/yank: This is... I mean, these two (and the various `setMark:` commands) are just such a pain. They really don't make sense cross-platform. we could definitely consider maintaining a per-app killring in `druid-shell`, and then implementing these ourselves, and I don't think this would really be *that* hard, but I think it's okay if this doesn't work for now.
- no tests for now is okay. We will need to think about this "eventually", but... 😈

And then other things:

The most major thing that is missing here, for me, is a sort of 'guide level' explanation of how IME is supposed to work, and how a user of druid-shell is supposed to interact with it. Like:


- each widget that can accept text input must register with the window, by calling XXXX.
- each time a widget gains focus, it must set itself as active, by calling XXXX.
- when the (selection, text, layout properties) of a widget change, you must notify the IME by calling XXXX.
- when there is an active IME session, your widget will be given a chance to handle raw keyboard events. If the events are not handled, they will be passed to the IME, which will convert them to `TextInputAction`s.

I think that this sort of information will be invaluable, and will be a really important artifact of this work.

Other than that, I don't really have a lot else to say, and maybe a good next step is for me to try and integrate this into druid proper, and see how that feels? I think that would give me a better general sense of how this API works for us, and whether or not there are any kinks to work out.

druid-shell/src/text_input.rs
lord: [Chromium](https://source.chromium.org/chromium/chromium/src/+/master:content/browser/android/selection/selection_popup_controller.h;l=54-55;drc=659834c429b56ed3e9f6f2cc5a0636c393a72d5d) calls this base/extent, and I've seen anchor/focus in other places. Generally not partial to `focus`, since it's a different concept from focus in other UI contexts, but I don't mind `active`, can switch if nobody else has other opinions! I personally prefer `anchor` over `base` since an anchor suggests it won't move in some contexts, but also I've been told my naming suggestions have been bad in the past (originally on Fuchsia I proposed `anchor`/`head` which people hated) so probably best for me to defer to others' judgement :)

lord: one nice aspect of `extent` is it's a noun instead of an adjective, but that's a pretty minor nit

druid-shell/src/text_input.rs
```diff
@@ -0,0 +1,516 @@
...
+
+/// Corresponds to a selection of some text. This also can represent a text input caret — the blinking vertical
+/// bar where text is to be inserted. A caret is a selection with zero length, where `anchor == extent`.
+/// Indices are always expressed in UTF-8 bytes, and must be between 0 and the document length, inclusive.
+///
+/// As an example, if the input caret is at the start of the document `hello world`, we would expect both `anchor`
+/// and `extent` to be `0`. If the user holds shift and presses the right arrow key five times, we would expect the
+/// word `hello` to be selected, the `anchor` to still be `0`, and the `extent` to now be `5`.
+#[derive(Clone, Default, PartialEq, Eq, Hash)]
+pub struct TextSelection {
```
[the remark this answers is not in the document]
lord: That would definitely be ideal! I think it will require we modify this to include last-horizontal-move-pixel-x positions, but perhaps an IME will some day want that information anyway. It does mean we'll have to more explicitly specify how pixel-x positions (is there a better name for that which you've seen) change in response to edits?

druid-shell/src/platform/mac/text_input.rs
```diff
@@ -0,0 +1,679 @@
...
+use cocoa::foundation::{NSArray, NSPoint, NSRect, NSSize, NSString, NSUInteger};
+use cocoa::{appkit::NSWindow, foundation::NSNotFound};
+use objc::runtime::{Object, Sel};
+use objc::{class, msg_send, sel, sel_impl};
+
+// thanks to winit for the custom NSRange code:
+// https://github.com/rust-windowing/winit/pull/518/files#diff-61be96e960785f102cb20ad8464eafeb6edd4245ea40224b3c3206c72cd5bf56R12-R34
+#[repr(C)]
+#[derive(Debug)]
+pub struct NSRange {
```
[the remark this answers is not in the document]
lord: I think it unfortunately doesn't support `objc::Encode`; will add a comment indicating this. Ideally this change could get upstreamed into `objc` and we could delete this custom implementation.

druid-shell/examples/edit_text.rs
```diff
@@ -0,0 +1,378 @@
...
+                None => return false,
+            };
+            handler.replace_range(selection.to_range(), "");
+        }
+        _ => return false,
+    }
+    true
+}
+
+fn movement_goes_downstream(movement: TextMovement) -> bool {
```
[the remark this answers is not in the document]
lord: Definitely would be ideal (as would the next function) but right now the implementation is wrong; it doesn't take into account the context around the caret, which is necessary for knowing the direction of `Left` and `Right`. How would you feel about punting for a later commit, when we've figured out a little bit more of the rules for how bidi text must work?

druid-shell/src/text_input.rs
lord: I think you're right. Switched to `min` and `max`!

druid-shell/src/text_input.rs
lord: I think this should be fixed, but if you spot a long first line I've missed somewhere, let me know!

druid-shell/src/platform/mac/text_input.rs
lord: I've, uh, never seen them get sent. I only know this behavior by changing my macOS text editing keybindings to force this to get sent to the text input system. A log is a great idea, I've added that.

druid-shell/src/platform/mac/window.rs
```diff
@@ -419,10 +422,69 @@ lazy_static! {
...
+            super::text_input::character_index_for_point as extern fn(&mut Object, Sel, NSPoint) -> NSUInteger,
+        );
+        decl.add_method(
+            sel!(firstRectForCharacterRange:actualRange:),
+            super::text_input::first_rect_for_character_range
+                as extern fn(&mut Object, Sel, NSRange, *mut c_void) -> NSRect,
+        );
+        decl.add_method(
+            sel!(doCommandBySelector:),
+            super::text_input::do_command_by_selector as extern fn(&mut Object, Sel, Sel),
```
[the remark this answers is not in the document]
lord: They should be forwarding now, but maybe worth another look-over from you; my understanding of Objective C is pretty shaky 😅

druid-shell/examples/edit_text.rs
```diff
@@ -0,0 +1,378 @@
...
+        TextMovement::Line(dir) => direction_goes_downstream(dir),
+        TextMovement::ParagraphEnd => true,
+        TextMovement::Vertical(VerticalMovement::LineDown) => true,
+        TextMovement::Vertical(VerticalMovement::PageDown) => true,
+        TextMovement::Vertical(VerticalMovement::DocumentEnd) => true,
+        _ => false,
+    }
+}
+
+fn direction_goes_downstream(direction: TextDirection) -> bool {
```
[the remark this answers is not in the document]
lord: see above comment on `movement_goes_downstream`

druid-shell/src/window.rs
```diff
@@ -322,6 +323,31 @@ impl WindowHandle {
         self.0.text()
     }
 
+    /// Indicates there's a new editable text input in the window.
```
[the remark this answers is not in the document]
lord: I've hopefully fixed the doc comment, but do you feel like `{add, remove, update}_text_field` would be a more clear method name?

druid-shell/examples/edit_text.rs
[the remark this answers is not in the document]
cmyr: gotcha.

I think it probably makes sense even ignoring BiDi to have a `TextDirection` enum in piet and to associate this with each piet `TextLayout`, but I'm not sure where that boundary should really be.

It does feel, though, that anyone using this code is going to need *some* variant of this functionality? Happy to worry about that a bit more during the integration phase, though.

druid-shell/src/text_input.rs
cmyr: I like anchor, I think that's clear. What's the natural inverse of `anchor`? `unmoored`? `adrift`? 🤡

extent.... just doesn't feel correct to me, because i equate extent with a measurement of some kind. Like: this would make sense to me if extent were an `isize`, relative to `anchor`. `anchor` and `edge`? `anchor` and `cursor`? 😱

lord: hahaa, the opposite of an anchor is maybe a `buoy`? a `sail`? a `raft`? i do kind of like `drift`, but it maybe falls into the same problem as extent where it sounds like a delta instead of an absolute value? also like buoy but maybe it's a little extravagant. also hard to spell.

i think `edge` is sometimes used to indicate either edge of the selection, so of all these options, maybe best to just stick with your original suggestion of active?

druid-shell/examples/edit_text.rs
```diff
@@ -0,0 +1,378 @@
...
+    fn prepare_paint(&mut self) {
+        self.handle.invalidate();
+    }
+
+    fn paint(&mut self, piet: &mut piet_common::Piet, _: &Region) {
+        let rect = self.size.to_rect();
+        piet.fill(rect, &BG_COLOR);
+        let doc = self.document.borrow();
+        let layout = doc.layout.as_ref().unwrap();
+        if let Some(composition_range) = doc.composition.as_ref() {
```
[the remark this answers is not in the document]
lord: fixed!

lord:
@cmyr if you're working on this soon, would it be helpful for me to rebase it?

cmyr:
Only if you're feeling like you need a project; I was planning on starting a new working branch based off of this, and getting it up-to-date myself 🤷

lord:
That works! I'll conserve my energy for TSF

cmyr:
this has been merged as part of #1619

PR #1381, opened 2020-11-07
Move TextBox border & insets into env

jpochyla:
- Adds support for `Insets` values in `Env`.
- Moves `TextBox` border and text insets constants into `Env` under `theme::TEXTBOX_INSETS` and  `theme::TEXTBOX_BORDER_WIDTH`.
- Adjusts the stateful scroll offset computation to account for the insets and to match native input widgets as close as possible.

### DIFF
--- a/druid/src/env.rs
+++ b/druid/src/env.rs
@@ -24,7 +24,7 @@ use std::sync::Arc;
 
 use crate::localization::L10nManager;
 use crate::text::FontDescriptor;
-use crate::{ArcStr, Color, Data, Point, Rect, Size};
+use crate::{ArcStr, Color, Data, Insets, Point, Rect, Size};
 
 /// An environment passed down through all widget traversals.
 ///
@@ -104,6 +104,7 @@ pub enum Value {
     Point(Point),
     Size(Size),
     Rect(Rect),
+    Insets(Insets),
     Color(Color),
     Float(f64),
     Bool(bool),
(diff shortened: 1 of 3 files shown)


jpochyla:
Related thing: what do you think about adding another, `TEXTBOX_INSETS`-like, offset to `TextBox`, that narrows the editor area and allows building things like textboxes with icons?

druid/src/env.rs
```diff
@@ -435,10 +438,13 @@ impl Data for Value {
         use Value::*;
         match (self, other) {
             (Point(p1), Point(p2)) => p1.x.same(&p2.x) && p1.y.same(&p2.y),
+            (Size(s1), Size(s2)) => s1.width.same(&s2.width) && s1.height.same(&s2.height),
             (Rect(r1), Rect(r2)) => {
                 r1.x0.same(&r2.x0) && r1.y0.same(&r2.y0) && r1.x1.same(&r2.x1) && r1.y1.same(&r2.y1)
             }
-            (Size(s1), Size(s2)) => s1.width.same(&s2.width) && s1.height.same(&s2.height),
+            (Insets(i1), Insets(i2)) => {
```
cmyr, line 445: `Insets` has a `Data` impl lower in the file already (these impls were added after this code was written) so we can just write, `(Insets(i1), Insets(i2)) => i1.same(&i2)`, I think.

cmyr (approved):
Thanks, this looks very sensible. One little comment about `Data` impls but I think it makes sense to merge this as-is, and then clean up all of those comparisons as a separate PR.

druid/src/env.rs
jpochyla: Oh I missed this, sorry.

cmyr: no worries, I have a followup inbound

PR #1384, opened 2020-11-09
Respect platform differences when handling text input

cmyr:
This now takes different code paths for 'mac' and 'not mac' when
mapping key events to text input events; it is not especially
thorough, and mostly just fixes the things that were most viscerally
annoying me.

### DIFF
--- a/druid/src/text/text_input.rs
+++ b/druid/src/text/text_input.rs
@@ -15,7 +15,7 @@
 //! Map input to `EditAction`s
 
 use super::Movement;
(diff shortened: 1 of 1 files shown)


derekdreery (approved):
In my experience linux closely follows windows, so simply having mac and non-mac might be enough.

cmyr:
I'm actually going to hold on to this because there's serious overlap with the IME stuff in #1378.

cmyr:
This should be handled via actual IME, so I'm going  to close.

PR #1385, opened 2020-11-09
Fix TextBox problems

jpochyla:
Fixes two issues introduced in #1381.

- Cursor disappearing behind the right edge:
 Adds `x1` offset to the default insets. Previously, the inset value was calculated as `x0 * 2.0`, now as `x0 + x1`.
- Miscalculating the scroll offset:
 Clamp the offset value only in case it's bigger than needed.


cmyr (approved):
Thanks!

PR #1387, opened 2020-11-09
Add a selectable list example.

derekdreery:


druid/examples/selectable_list.rs
```diff
@@ -0,0 +1,82 @@
...
+        |data, _| data.0 == Some(data.1.position),
+        Label::raw().with_text_color(Color::BLACK).lens(l),
+        Label::raw().lens(l2),
+    )
+    .on_click(|_, data, _| {
+        if data.0 == Some(data.1.position) {
+            data.0 = None;
+        } else {
+            data.0 = Some(data.1.position);
+        }
```
cmyr, line 80: not a blocker but I'd write this as,

```rust
data.0 = if data.0 == Some(data.1.position) {
    None 
} else {
    Some(data.1.position)
};
```

I find this control flow easier to follow; you know the each block has to evaluate to a new value, instead of each block mutating state.

druid/examples/selectable_list.rs
```diff
@@ -0,0 +1,82 @@
...
+
+fn list_item() -> impl Widget<(Option<usize>, ListItem)> {
+    let l = lens!((Option<usize>, ListItem), 1).then(lens!(ListItem, label));
+    let l2 = lens!((Option<usize>, ListItem), 1).then(lens!(ListItem, label));
+    Either::new(
+        |data, _| data.0 == Some(data.1.position),
+        Label::raw().with_text_color(Color::BLACK).lens(l),
+        Label::raw().lens(l2),
+    )
+    .on_click(|_, data, _| {
```
cmyr, line 75: As this is an example, I think it might be worth adding a short comment here to the effect that this logic is handling selection.

cmyr (commented):
Cool, I appreciate how concise this is.

The main thing I wonder, looking at this, is whether we could combine this with another example? Or replace the existing list example? I do think this is worth demonstrating, but I also think a proliferation of examples has real discoverability and maintenance costs.

derekdreery:
I'll have a think how this could be merged into another example, possibly the `list` example.

derekdreery:
OK ready for review again.

druid/examples/list.rs
```diff
@@ -85,39 +119,49 @@ fn ui_builder() -> impl Widget<AppData> {
...
                             .align_vertical(UnitPoint::CENTER),
                     )
                     .padding(10.0)
                     .background(Color::rgb(0.5, 0.0, 0.5))
                     .fix_height(50.0)
-            })
-            .with_spacing(10.),
-        )
+            }
+            Either::new(
```
derekdreery, line 152: It might be possible to move the Either down the widget tree.

derekdreery:
I think the best way to review this is to run it and see if you like the way it works, and only then look at the code. That way you don't waste time on details if you don't like the overall effect.

druid/examples/list.rs
jneem: It seems fine where it is to me. I think if I were writing it, I'd just have a single `Label` and switch the color by changing `LABEL_COLOR` in an `EnvScope`, but this seems fine too :shrug:

jneem (approved):
Looks good to me! And I know you didn't put it there, but I find it much easier to see what's going on without the `debug_paint_layout`

druid/examples/list.rs
derekdreery: I'll have a think about this before I merge.

xStrom:
What's the status on this? Are you still interested in moving this forward @derekdreery?

It seems that it was reviewed and approved, but never merged, which is unfortunate.

The next step right now would be to rebase this on `master` and get a fresh successful CI run. Then I can do final review and merge.

derekdreery:
Cool I'll rebase when i get chance.

PR #1389, opened 2020-11-11
Fix scaling on GTK & X11

### LINKED ISSUE #939: X11 shell does not handle DPI scaling.
This should be done after #904 is merged.

dhardy:
WIP: fixes #939 

The first commit fixes scaling on the GTK backend, per my testing.

The second commit attempts to get the scaling factor when using the X11 backend. Unfortunately it doesn't work:
```
     Running `/home/dhardy/.cache/cargo/debug/examples/calc`
INFO  [druid_shell::platform::x11::application] X server supports Present version 1.0
INFO  [druid_shell::platform::x11::application] X server supports XFIXES version 5.0
WARN  [druid_shell::platform::x11::application] Application::get_locale is currently unimplemented for X11 platforms. (defaulting to en-US)
DEBUG [druid::localization] available locales [en-US, fr-CA, de-DE], current en-US
DEBUG [druid::localization] resolved: [en-US]
WARN  [druid_shell::platform::x11::window] WindowBuilder::resizable is currently unimplemented for X11 platforms.
[druid-shell/src/platform/x11/window.rs:233] scale = Scale {
    x: 1.0004923682914821,
    y: 1.0,
}
```
(Scale factor is actually 1.1667.) Any idea what's wrong here? @crsaracco

### DIFF
--- a/druid-shell/src/platform/gtk/window.rs
+++ b/druid-shell/src/platform/gtk/window.rs
@@ -375,6 +375,7 @@ impl WindowBuilder {
                         // Clip to the invalid region, in order that our surface doesn't get
                         // messed up if there's any painting outside them.
                         for rect in invalid.rects() {
+                            let rect = rect.to_px(scale);
                             surface_context.rectangle(rect.x0, rect.y0, rect.width(), rect.height());
                         }
                         surface_context.clip();
--- a/druid-shell/src/scale.rs
+++ b/druid-shell/src/scale.rs
@@ -100,6 +100,8 @@ impl Default for Scale {
 
 impl Scale {
     /// Create a new `Scale` based on the specified axis factors.
+    ///
+    /// Units: none (scale relative to "standard" scale)
     pub fn new(x: f64, y: f64) -> Scale {
         Scale { x, y }
     }


druid-shell/src/platform/x11/window.rs
```diff
@@ -224,6 +224,14 @@ impl WindowBuilder {
                 | EventMask::PointerMotion,
         );
 
+        // Conversion factor for scale (units = mm/pixels)
+        const FACTOR: f64 = 25.4 / 96.0;
+        let scale = Scale::new(
+            FACTOR * screen.width_in_pixels as f64 / screen.width_in_millimeters as f64,
+            FACTOR * screen.height_in_pixels as f64 / screen.height_in_millimeters as f64,
+        );
```
jneem, line 232: I wouldn't rely on this information, for two reasons. First, I have no idea where it comes from and whether its reliable. On my system, xrandr and xdpyinfo report two different physical sizes for the same screen. Second, I think it's a bad idea to default to non-integer scaling. I'd argue for the xft-based solution that @crsaracco [linked](https://github.com/linebender/druid/issues/939#issuecomment-649853164) in #939.

dhardy: Thanks, I'll try implementing the xft method.

> Second, I think it's a bad idea to default to non-integer scaling.

You're joking, right? 1080p screens are common across 13" - 17" laptops now (165 - 130 DPI). 27" 1440p screens are also common (109 DPI). Besides which, many older users like slightly larger text. This is a *huge* demand for scale factors between 1 and 2.

jneem: I wasn't actually joking, but I'm definitely happy to discuss it further. To be clear, I'm not arguing against *supporting* fractional scaling, just about defaulting to it. My main reason is that it causes anti-aliasing artifacts (like #1088).

> Besides which, many older users like slightly larger text.

Is there a reason to prefer scaling everything over just increasing the font size? (This is a genuine question -- I like *tiny* text so I'm very much not in the target audience)

dhardy: > that it causes anti-aliasing artifacts

Which is why I took a rather different approach to scaling with KAS (e.g. [here](https://github.com/kas-gui/kas/blob/master/kas-theme/src/dim.rs#L18)). But that's off topic...

> Is there a reason to prefer scaling everything over just increasing the font size?

Good question. Actually, I don't know much about how Druid likes to handle scaling, but at least the GTK backend currently is happy to use fractional scaling. This whole topic should probably get a new issue since this PR is just about fixing the currently-missing scaling support on X11 and broken support on GTK.

psychon: > First, I have no idea where it comes from and whether its reliable. On my system, xrandr and xdpyinfo report two different physical sizes for the same screen. 

The data reported by `xdpyinfo` / the connection setup will always report 96 DPI. IIRC Keith Packard wrote that in a blog post somewhen, but I am not sure. But this can definitely not report data per RandR-output and there is also no way for the X11 server to update the information (well, RandR has something about that, but then you can just use the "real" info from RandR).

druid-shell/src/platform/x11/window.rs
```diff
@@ -224,6 +224,14 @@ impl WindowBuilder {
...
+        dbg!(scale);
```
cmyr, line 233:
```suggestion
```

druid-shell/src/platform/x11/window.rs
cmyr: I'm happy to see this merged as an improvement, and then revisit the larger question of how we want to 'properly' do scaling? I'm not spending much time on linux (although I'm happy to play around in a vm as needed) so I trust @jneem and @dhardy to get this merged when there's consensus, and I'm happy to break any ties if needed. 👼

druid-shell/src/platform/x11/window.rs
dhardy: Not so simple... the scale value is wrong.

druid-shell/src/platform/x11/window.rs
dhardy: Perhaps we should also refer to the work of our peers: https://docs.rs/winit/0.23.0/winit/dpi/index.html#how-is-the-scale-factor-calculated

I'm in two minds whether it's worth bothering much here.

dhardy: Winit uses [Xlib::open](https://docs.rs/x11-dl/2.18.5/x11_dl/xlib/struct.Xlib.html#method.open) to access `Xft.dpi`. The `RustAudio` code uses the `x11` crate. I think we can't access X resources with `x11rb`, so an extra dependency would be necessary?

psychon: At least from a C point of view: You are already linking against `libcairo.so` even when using x11rb. `libcairo.so` links against `libX11`. So your extra dependency would only be a Rust wrapper for `libX11`.

Also: Due to this issue and https://github.com/psychon/x11rb/issues/550, I noticed that Xresource handling is missing in x11rb and opened https://github.com/psychon/x11rb/issues/551 about that. It might be that the next release of x11rb will provide a way to query `Xft.dpi`. Technically, it is really easy. It's just that I'd have to hand-write a parser for the necessary file format (easy but annoying) and be reasonable sure that it behaves as expected (this seems to be the hard case - Xresources seem to be quite powerful and I do not really understand all the rules (yet)).
X11-the-protocol-wise, Xresources are just a property on the root window containing a string, plus some fallbacks in case there is no such properties. Thus, only the parsing and the actual query matching are necessary.

dhardy:
Right, I removed the X11 stuff since it may make sense to wait for the next `x11rb` release as [mentioned by](https://github.com/linebender/druid/pull/1389#discussion_r525189763) @psychon.

The GTK fix there shouldn't be any problem merging I think.

jneem:
Sounds good, let's get the gtk fix in first.

jneem (approved):

druid-shell/src/platform/x11/window.rs
psychon: I certainly do not claim to know what I am doing, but I just added an example to https://github.com/psychon/x11rb/issues/550#issuecomment-729025335 that draws a 2 cm line by querying information from RandR. It works for me, so it must obviously work everywhere. ;-)
This does *not* do anything with Xft configuration and has zero user configurability. It just detects values and uses them.
Also, I screwed up the calculation of DPI three times and then gave up. Thus, the code works in dots per cm instead. This also makes it a lot easier to calculate the length of a 2 cm line. This should just be a matter of adding the right scaling factor.

Also: Urgh for X11 (and thus x11rb) mixing u16, i16 and u32 in the involved APIs.

PR #1391, opened 2020-11-11
Add possibility to set new image data

tomaszmduda:
This is useful to use a widget to show different images, e.g. as a slide show application.


cmyr (approved):
This definitely seems like a reasonable thing to have. Thanks!

PR #1392, opened 2020-11-12
Add Scroll::scroll_to and ClipBox::pan_to_visible

cmyr:
Often when manipulating a scroll view you care less
about relative movement and more about just ensuring
some portion of the content is visible; this API
supports that case.

@jneem does this make sense to you?

### DIFF
--- a/druid/src/widget/clip_box.rs
+++ b/druid/src/widget/clip_box.rs
@@ -70,6 +70,47 @@ impl Viewport {
             false
         }
     }
+
+    /// Pan the smallest distance that makes the target [`Rect`] visible.
+    ///
+    /// If the target rect is larger than viewport size, we will prioritize
+    /// the region of the target closest to its origin.
+    pub fn pan_to_visible(&mut self, rect: Rect) -> bool {
+        /// Given a position and the min and max edges of an axis,
+        /// return a delta by which to adjust that axis such that the value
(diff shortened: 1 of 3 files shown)


druid/src/widget/clip_box.rs
```diff
@@ -188,6 +232,16 @@ impl<T, W: Widget<T>> ClipBox<T, W> {
...
     }
 
+    /// Adjust the viewport to display as much of the target region as is possible.
+    ///
+    /// This will move the viewport the smallest distance that fully shows
+    /// the target region. If the target region is larger than the viewport,
+    /// we will display the portion that fits, prioritizing the portion closest
+    /// to the origin.
+    pub fn pan_to_visible(&mut self, region: Rect) -> bool {
+        self.port.pan_to_visible(region)
```
jneem, line 242: I think you'll need to do `self.child.set_viewport_offset` here, like in `pan_to`

jneem (approved):
Definitely looks like a useful thing to have! I think you do need the call to `set_viewport_offset` to get invalidation propagating correctly.

druid/src/widget/clip_box.rs
cmyr: oh yea definitely missed that, thanks :)

PR #1393, opened 2020-11-12
Add axis and spacing to list.

derekdreery:
Add spacing and axis options to list widget. Spacing can be a value or resolved from env.

Closes #1364.

### DIFF
--- a/druid/src/widget/list.rs
+++ b/druid/src/widget/list.rs
@@ -39,25 +39,19 @@ pub struct List<T> {
 impl<T: Data> List<T> {
     /// Create a new list widget. Closure will be called every time when a new child
     /// needs to be constructed.
-    pub fn new<W: Widget<T> + 'static>(closure: impl Fn() -> W + 'static, axis: Axis) -> Self {
+    pub fn new<W: Widget<T> + 'static>(closure: impl Fn() -> W + 'static) -> Self {
         List {
             closure: Box::new(move || Box::new(closure())),
             children: Vec::new(),
-            axis,
+            axis: Axis::Vertical,
             spacing: KeyOrValue::Concrete(0.),
         }
     }
 
-    /// Create a new vertical list widget. Closure will be called every time when a new child
-    /// needs to be constructed.
-    pub fn vertical<W: Widget<T> + 'static>(closure: impl Fn() -> W + 'static) -> Self {
-        List::new(closure, Axis::Vertical)
-    }
-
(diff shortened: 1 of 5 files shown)


druid/src/widget/list.rs
```diff
@@ -23,26 +24,54 @@ use crate::im::Vector;
...
     closure: Box<dyn Fn() -> Box<dyn Widget<T>>>,
     children: Vec<WidgetPod<T, Box<dyn Widget<T>>>>,
+    axis: Axis,
+    spacing: KeyOrValue<f64>,
 }
 
 impl<T: Data> List<T> {
     /// Create a new list widget. Closure will be called every time when a new child
     /// needs to be constructed.
-    pub fn new<W: Widget<T> + 'static>(closure: impl Fn() -> W + 'static) -> Self {
```
cmyr, line 39: Maybe this is heresy, but I think I would leave `new` as is (it creates a vertical list) and then just have a `horizontal` constructor that makes a horizontal list; I think the vertical case is going to be 95%+ of uses cases.

cmyr (commented):
Looks good! One thought on the API and one note below, but otherwise I'm happy for this to get in; let me know what you think.

For max correctness one other thing we might want to do now is to check in `layout` if the spacing key has changed in the env, and then call `request_update` if it has.

druid/src/widget/list.rs
derekdreery: Yep can do that.

derekdreery:
> For max correctness one other thing we might want to do now is to check in layout if the spacing key has changed in the env, and then call request_update if it has.

I suspect that this should be added to other widgets as well. Shall I have a look through?

cmyr:
yes, the ability to inspect changed keys is relatively new (7db6dd01) and this definitely isn't being done optimally by a bunch of widgets.

cmyr (approved):
I was thinking that `horizontal` would be a constructor? We could document this in `new`, and say something to the effect, "To construct a horizontal list, use [`List::horizontal`] instead?

I'm open to discussion here, but this is at least what I'd had in mind.

In any case, I'm going to approve; feel free to merge either as-is, or with that change.

derekdreery:
I guess it just felt more natural to me, if there weren't two constructors with equal weighting, then the axis should be a parameter rather than in the constructor.

derekdreery:
I will try to have a look at the `KeyOrValue` thing more comprehensively in a separate pass.

Issue #1388, opened 2020-11-10
A mechanism for child->parent communication ('notifications')


cmyr:
This issue sketches out an idea for an architectural addition to druid that would support communication from a child widget back to some parent.

### motivation

There are a number of situations where a child widget may wish to communicate with one of its ancestors. A classic example is when text is embedded in a scroll view; if the cursor moves beyond the edge of the currently visible region, you would like to tell the containing scroll view to modify the scroll offset in order to display the current cursor.

This is probably the most obvious case, but there are a number of other important ones:
- A widget that has focus may wish to relinquish focus, but wants to prioritize its direct ancestors by giving them a chance to react to this change before the root widget (another way of implementing focus scope)
- A widget that is contained in a list may wish to indicate that it has been selected (maybe this isn't the best mechanism for this?)
- a textbox that supports validation may wish to communicate a validation error up to a higher level in the tree, so that it can be displayed as an error
- two sibling widgets wish to communicate any other state or data that is not part of the explicit `AppData`; 

In all of these instances you could currently use `Command`, but with one important caveat; you have to know the `WidgetId` of the widget you are intending to communicate, which at the very least can be a major hassle. `Command` has another complication as well, which is that there are no guarantees over when it is dispatched.

### notifications

I would like to propose a new mechanism to be used in these cases, which I am calling 'notifications'. Like a `Command`, A notification is submitted by a widget in response to some event; however unlike `Command`, notifications are dispatched to parents immediately after the child's `event`/`lifecycle` etc method has returned.

The basic idea is that after the `WidgetPod`'s call to its child's `event`/`layout` method has returned, the pod checks to see if there are any outstanding notifications. If there are, it delivers them immediately to the child. If the child indicates that it has handled the notification, no further work is necessary; otherwise it will stay in the queue, and delivery will be tried again at the next ancestor widgetpod, until we reach the root of the tree; unhandled notifications will be discarded.

### implementation

There are a number of options when we consider how this should be implemented.

In any case, I think that we will add a `Notification` type; this will be very similar to `Command`, except that it *does not* have an explicit target, and it *does* have an explicit source; but it can use the same `Selector` infrastructure.

```rust
pub struct Notification {
    symbol: SelectorSymbol,
    payload: Arc<dyn Any>,
    source: WidgetId,
}
```

`Notification`s will be dispatched very much like `Command`s as well:

```rust
// in EventCtx, say:
pub fn submit_notification(&mut self, notification: impl Into<Command>) {
[...]

PR #1394, opened 2020-11-12
Notification sketch

cmyr:
This implements the notification proposal described in #1388.
The implementation is minimal and relatively non-invasive;
my inclination is to merge this and then experiment on top
of that, with the understanding that it will be fairly easy
to remove this if it isn't pulling its weight.

~This is a quick and dirty implementation of the notifications mechanism described in #1388.~

~This adds notifications (which act like an event, and can only be sent via the `EventCtx`, although I suspect they should ultimately be a `LifeCycle`) and also adds support for a 'scroll to' notification in the `Scroll` widget, that lets children modify the scroll position; it uses this in the `TextBox` so that we always display the cursor position when the cursor moves. This can be tested in the `styled_text` example.~

~Overall I'm quite happy with how little code this is, and my inclination is to push this forward, merging some version of the notification mechanism; this would be tentative but make it possible to explore various ways of using this without having to maintain a fork.~

~(this is based on #1392)~

### DIFF
--- a/druid/src/command.rs
+++ b/druid/src/command.rs
@@ -81,6 +81,30 @@ pub struct Command {
     target: Target,
 }
 
+/// A message passed up the tree from a [`Widget`] to its ancestors.
+///
+/// In the course of handling an event, a [`Widget`] may change some internal
+/// state that is of interest to one of its ancestors. In this case, the widget
+/// may submit a [`Notification`].
+///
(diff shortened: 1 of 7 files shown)


raphlinus (approved):
This looks good, thanks! Fairly clean and simple overall. Let's hope it works out well in practice.

PR #1395, opened 2020-11-13
Do layout *just* before painting.

jneem:
This ensures that there are no data changes between layout and paint. Prior to this, it was possible to have an animation frame or an `update` call in between `layout` and `paint`, which could cause Labels to panic.


cmyr (approved):
Yes, this looks right. Should've caught this in an earlier review; it should be an invariant that nothing happens between layout and paint.

I also think (and I'm less sure about this) that the only thing that should be possible between update and layout is `RouteWidgetAdded`, if a widget was added during update; that widget will see `WidgetAdded` and then should immediately see `layout` and then `paint`.

PR #1396, opened 2020-11-13
Implement Data for the NonZero integer types

bheisler:
I wanted to have a `NonZeroU32` in my data for my recipe manager (to represent the number of servings a recipe produces) but I couldn't because the Data trait is not implemented for the `NonZero` integer types. This pull request adds trivial implementations for those types.


cmyr (approved):
Looks good :)

PR #1397, opened 2020-11-13
Remove type param from Unit lens

cmyr:
Just some cleanup I noticed while doing some other work.


derekdreery:
What do you think about having a `lens::id` function? It might be more discoverable than a struct, given they are mostly to support functions in `LensExt`.

derekdreery (approved):

cmyr:
@derekdreery I don't have a strong feeling, but I *do* think that `Id` should actually be called `Identity`, I'll prepare a patch for that momentarily.

PR #1398, opened 2020-11-13
Ensure TextBox editor is up to date after event

cmyr:
Previously the editor's inner layout would be stale between
when an edit occured, and when update happened; this meant that
things like text measurement and cursor positions would be incorrect
unless the caller manually rebuilt the layout.

Since rebuilding is already lazy, and avoids doing work unless state
has changed, this should not be much more expensive and removes
a significant potential source of confusion.

This was motivated by my experiments with notifications.

### DIFF
--- a/druid/src/text/editor.rs
+++ b/druid/src/text/editor.rs
@@ -173,7 +173,12 @@ impl<T: TextStorage + EditableText> Editor<T> {
         self.layout.rebuild_if_needed(factory, env);
     }
 
-    /// Perform an [`EditAction`](enum.EditAction.html).
+    /// Perform an [`EditAction`].
+    ///
+    /// After editing, the inner layout is not valid until [`rebuild_if_needed`]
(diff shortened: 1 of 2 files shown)


jneem (approved):
I'm definitely not an expert on `TextBox`, but this looks reasonable to me.

cmyr:
I'm holding off on this because I think there are possible problems.

Basically: if we rebuild too early, we will fail a 'needs rebuild' check in `update`, which means we won't call `layout`, so this needs more thought.

derekdreery:
> There are only two hard things in Computer Science: cache invalidation and naming things.
    - Phil Karlton

Although I prefer the version
> There are only two hard things in Computer Science: cache invalidation, naming things, and off by 1 errors.

cmyr:
I'm going to close this because I'm not sure it's necessary once we aren't using parse.

PR #1399, opened 2020-11-14
Add a lens that negates a boolean.

derekdreery:
I can't remember where, but someone suggested that this would be a useful addition.


druid/src/lens/lens.rs
```diff
@@ -509,3 +537,23 @@ impl<T> Lens<T, ()> for Unit<T> {
         f(&mut ())
     }
 }
+
+#[derive(Debug, Copy, Clone)]
+pub struct Not;
```
cmyr, line 542: This should still get docs, I'm surprised we aren't failing CI? we have `#[warn(missing_docs)]` turned on, and one of our CI steps should deny warnings... 🤔

cmyr (approved):
Looks good, modulo one missing docstring 👍

PR #1401, opened 2020-11-14
Rework (and rename) ext event

JAicewizard:
This mostly just adds comments.
I also got a more satisfying colour generation comming in another PR for this and the `identity` example.
I also think that this and the `identity` one are too similair to not be based on each other. Both use comands and have colours shifting going on. Both should handle the colour shifting the same way (not using timers like the `identity` example) so I will be opening a PR about completely redoing that one. It will also make it cleaner I think.

EDIT: I also will be opening a PR to remove the `blocking_functions` example since it really just has a lot of clutter and revolves around the same concept as this example. 

sorry this took a while, my factorio addiction is acting up again.


JAicewizard:
I decided to rename form `ext_event` to `async_event` since "ext" (probably) stands for external, but the example doesnt really have anything external to it. Calling it async makes more sense as this example applies to all async events.
If you have any preference of keeping it let me know.

druid/examples/async_event.rs
```diff
@@ -0,0 +1,133 @@
...
+impl Widget<Color> for ColorWell {
+    fn event(&mut self, _ctx: &mut EventCtx, event: &Event, data: &mut Color, _env: &Env) {
+        match event {
+            // This is where we handle our command.
+            Event::Command(cmd) if cmd.is(SET_COLOR) => {
+                // We don't do much data processing in the `event` method.
+                // All we really do is just set the data. This causes a call
+                // to `update` which requests a paint. You can also request a paint
+                // during the event, but this should be reserved for changes to self.
+                // For changes to `Data` always make `update` to the paint requesting.
```
futurepaul, line 102: to -> do

druid/examples/async_event.rs
```diff
@@ -0,0 +1,133 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! An example of sending commands from another thread.
+//! This is useful when you want to have some kind of
+//! generated content(like here), or some task that just
```
futurepaul, line 17: missing space

druid/examples/async_event.rs
```diff
@@ -0,0 +1,133 @@
...
+//! (waiting on an http request, some cpu intensive work etc.)
+
+use instant::Instant;
+use std::thread;
+use std::time::Duration;
+
+use druid::widget::prelude::*;
+use druid::{AppLauncher, Color, Selector, Target, WidgetExt, WindowDesc};
+
+// If you want to send commands to over event sinks you have to give it some kind
```
futurepaul, line 28: `commands to over event sinks` not sure what you're saying here

futurepaul (approved):
This is a great improvement! I'm fine with the name change. I could see people being confused that this isn't `async` in the general Rust ecosystem use of that word, but at the same time that's probably the word people are looking for for this kind of functionality?

JAicewizard:
yeah I thought about that too, I can rename it to `threaded_event` but I'm not sure people look for that. I will add a readme with some comments on the examples helping people find them. `ext_event` might be the druid name but that is very far away from how a beginning user sees it.

druid/examples/async_event.rs
JAicewizard: `commands over event sinks` is what I mean

druid/examples/async_event.rs
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    fn paint(&mut self, ctx: &mut PaintCtx, data: &Color, _env: &Env) {
+        let rect = ctx.size().to_rounded_rect(5.0);
+        ctx.fill(rect, data);
+    }
+}
+
+fn make_ui() -> impl Widget<Color> {
+    ColorWell {}
```
cmyr, line 128: I don't think these braces are necessary?

```suggestion
    ColorWell
```

druid/examples/async_event.rs
```diff
@@ -0,0 +1,133 @@
...
+    // long time and shouldn't block main UI updates.
+    thread::spawn(move || generate_colours(event_sink));
+
+    launcher
+        .use_simple_logger()
+        .launch(Color::BLACK)
+        .expect("launch failed");
+}
+
+fn generate_colours(event_sink: druid::ExtEventSink) {
```
cmyr, line 56: Although I personally use international spellings frequently, we have decided to use `Color` instead of `Colour` in druid so I think it probably makes sense to be consistent with that here as well?

cmyr (approved):
Looks good, a few little notes but I think this is an improvement!

druid/examples/async_event.rs
JAicewizard: oh sry, I try to keep is US English as well, but my nature is very much still with colour.

cmyr: me too, but sometimes we must defer to the imperialists. 😢

cmyr (approved):
Thanks!

PR #1402, opened 2020-11-14
Remove blocking functions

JAicewizard:
As mentioned in the rework `ext_event` example this example is pretty useless and doesnt really deserve its place in this list anymore. Unless the goal is to have a clean cut example for every individual scenario this is pretty useless.


futurepaul:
Do we have another example that demonstrates a one-off thread like this rather than a long-lived one? If not this feels valuable to me as reference because it was a very common how-to question before this example was added.

JAicewizard:
no, so in that case I think we should just keep this one (so close this PR)? to me it really feels like a duplicate, the only difference being instead of having one thread with a long running time you have multiple with short running time.

futurepaul:
Without a strong replacement for this use case, such as a chapter in the book, I prefer to keep this around. I'm speaking from my own confusion on how best to do something like this before this example was added :)

Issue #1134, opened 2020-08-16
Label height is too small for upper case accented letters


alarcheres:
Label height is too small for upper case accented letters, accents overflow on the above widget (or are cropped when there is none).

![druid-wrong-label-height](https://user-images.githubusercontent.com/60073372/90342208-a4f27480-e006-11ea-8231-cde1bf5a19ce.png)

Minimal reprodution: 
```rust
use druid::widget::*;
use druid::*;

pub fn main() {
    let main_window = WindowDesc::new(|| Label::new("ÉÀÈÊ"))
        .title("")
        .window_size((400.0, 400.0));

    AppLauncher::with_window(main_window)
        .launch(())
        .expect("Failed to launch application");
}
```
System Specification: 
Druid 0.6
Debian 10 with GTK Backend

cmyr:
this is likely a limitation of cairo's toy text API, and so there isn't much we can do about it until we implement a new text system for linux.

PoignardAzur:
@cmyr Any progress?

cmyr:
should be fixed, can you verify?

PoignardAzur:
That looks fixed to me:

![textbox_screenshot](https://user-images.githubusercontent.com/8573618/128243934-1055783d-ca13-46ce-a6df-e3886ebe2059.png)

But I'm not sure how the textbox example would have looked before.

Maan2003:
yeah probably got fixed by the pango text
![image](https://user-images.githubusercontent.com/49202620/128244518-891e5e27-1c84-4d6e-aaa5-0e961ffd75d8.png)

![image](https://user-images.githubusercontent.com/49202620/128244318-7130f957-4919-4edc-afd0-cf7a91762718.png)

PR #1403, opened 2020-11-14
Request multiline textbox layout on update

gmorenz:
Currently when you add/remove a newline in a multiline textbox the textbox does not resize until you do something like resize the window. This results in your cursor going off the screen when adding a newline at the end of the textbox.

This causes the textbox to request a layout on every update, which fixes the problem.

I'm not sure this is the right fix. For one it might be possible to be more conservative here and only update when newlines are added. Alternatively I notice an issue (#1134) which suggests accents should change text height, and it's possible for reason like that we should be rerunning layout on every change regardless of whether or not the textbox is multiline. Or something inbetween the two that I'm not aware of.

### DIFF
--- a/druid/src/text/editor.rs
+++ b/druid/src/text/editor.rs
@@ -161,9 +161,9 @@ impl<T: TextStorage + EditableText> Editor<T> {
         if self.data_is_stale(new_data) {
             self.layout.set_text(new_data.clone());
             self.selection = self.selection.constrained(new_data);
-            ctx.request_paint();
+            ctx.request_layout();
         } else if self.layout.needs_rebuild_after_update(ctx) {
-            ctx.request_paint();
+            ctx.request_layout();
         }
         self.rebuild_if_needed(ctx.text(), env);
     }


JAicewizard:
I think it should be possible to keep track of the amount of newlines right? I think that would be ideal. But that's something that can be added/changed later when performance becomes more of a thing.

druid/src/widget/textbox.rs
```diff
@@ -325,7 +325,8 @@ impl<T: TextStorage + EditableText> Widget<T> for TextBox<T> {
         if !self.suppress_adjust_hscroll && !self.multiline {
             self.update_hscroll(ctx.size().width, env);
         }
-        if ctx.env_changed() && self.placeholder.needs_rebuild_after_update(ctx) {
+        if self.multiline || (ctx.env_changed() && self.placeholder.needs_rebuild_after_update(ctx))
```
cmyr, line 328: I think the issue here is that editor's `update` method is calling `request_paint` instead of `request_layout`; maybe try making that change instead (`src/text/editor.rs:L166`) and see if it fixes your problem?

cmyr (changes requested):
Cool, this sounds like a bug but I think the best solution is slightly different, can you try the change I suggest and see if that fixes your problem?

cmyr:
Okay thinking about this (and playing around) a bit more, I'm not actually able to reproduce the underlying issue you reported? That is, if I take the `hello` example and I make it a multiline text box, I get:



I do think that there are bugs here, but maybe you're encountering a platform-specific problem? Are you on linux?

gmorenz:
Hmm, windows atm but I thought I saw this behavior earlier on linux too.

Will create a minimal repro, a gif, and try your proposed solution after work hours.

I'm surprised to see that your test isn't surfacing the issue. Maybe the label resizing is forcing it to relayout?

gmorenz:
Minimal reproduction

```
use druid::{*, widget::*};

fn main() {
    AppLauncher::with_window(WindowDesc::new(
        || {
            let tb = TextBox::multiline();
            Flex::column().with_child(tb)
        }))
        .launch("".to_string())
        .expect("Failed to launch");
}
```

gmorenz:
Adding a label above it (i.e. the following) does indeed fix the issue - which explains your gif.

```
use druid::{*, widget::*};

fn main() {
    AppLauncher::with_window(WindowDesc::new(
        || {
            let label = Label::new(|s: &String, _env: &Env| s.clone());
            let tb = TextBox::multiline();
            Flex::column().with_child(label).with_child(tb)
        }))
        .launch("".to_string())
        .expect("Failed to launch");
}
```

gmorenz:
Changing line 164 to request_layout also fixes the issue, line 166 doesn't affect my code but I'm guessing I should change it as well (but don't fully understand this code). I've pushed a commit matching that guess.

gmorenz:
Gif of behavior without patch: (not that it really shows much)

image

cmyr (approved):
You're right, both of these lines should be changed; this looks correct to me.

Thanks!

PR #1404, opened 2020-11-15
Added lost focus, and got focus for gtk.

derekdreery:
I've not tested the windows or mac impls.

### DIFF
--- a/druid-shell/src/platform/gtk/window.rs
+++ b/druid-shell/src/platform/gtk/window.rs
@@ -307,7 +307,8 @@ impl WindowBuilder {
                 | EventMask::ENTER_NOTIFY_MASK
                 | EventMask::KEY_RELEASE_MASK
                 | EventMask::SCROLL_MASK
-                | EventMask::SMOOTH_SCROLL_MASK,
+                | EventMask::SMOOTH_SCROLL_MASK
+                | EventMask::FOCUS_CHANGE_MASK,
         );
 
         win_state.drawing_area.set_can_focus(true);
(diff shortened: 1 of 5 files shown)


derekdreery:
I've now added a `WindowLostFocus` event. It isn't symmetric with the way "window got focus" events are handled, but they are different beasts: you can lose focus while doing something, specifically dragging something and moving the mouse out of the window with a window manager that changes focus on mouse move (window managers like i3 do this by default, and it's sometimes enabled for accessibility).

derekdreery:
Let me know if you would like a squash.

druid-shell/src/platform/gtk/window.rs
```diff
@@ -608,6 +609,26 @@ impl WindowBuilder {
                 Inhibit(true)
             }));
 
+        win_state
+            .drawing_area
+            .connect_focus_in_event(clone!(handle => move |_widget, event| {
+                assert!(event.get_in());
```
jneem, line 615: This scares me, partly because the [gtk docs](https://gtk-rs.org/docs/gdk/struct.EventFocus.html#method.get_in) aren't telling me what it does, but mostly because I don't think we should panic just because we got something unexpected from the system. Could it be a `log::error!` instead?

druid-shell/src/platform/gtk/window.rs
```diff
@@ -608,6 +609,26 @@ impl WindowBuilder {
...
+                if let Some(state) = handle.state.upgrade() {
+                    state.with_handler(|h| h.got_focus());
+                }
+                Inhibit(true)
+            }));
+
+        win_state
+            .drawing_area
+            .connect_focus_out_event(clone!(handle => move |_widget, event| {
+                assert!(!event.get_in());
```
jneem, line 625: Same here

jneem (changes requested):

druid-shell/src/platform/gtk/window.rs
derekdreery: I'm happy to remove these.

derekdreery:
I squashed because nothing was gained from having the individual commits.

derekdreery:
@cmyr this should be a quick review - it's a small patch.

jneem (approved):

jneem:
No more complaints from me, but it would be nice if someone else could take a quick glance at the mac part

cmyr (approved):
Mac looks good, although now that we're printing in `got_focus` it looks like that's getting called twice in a row or something? Will open an issue, but not a blocker for this. Thanks!

PR #1405, opened 2020-11-16
Rename lens::Id -> lens::Identity

cmyr:
Maybe not necessary to keep around the old version to do the deprecation, but 🤷


futurepaul (approved):
This makes sense to me since "id" already has another meaning in Druid. My own preference would be rip the band-aid off instead of deprecation (it's a very easy search / replace to update my existing code, and 0.7 has other breaking changes anyway), but I don't feel strongly one way or the other.

cmyr:
I think you're right, will just do a straight-up rename. :)

PR #1407, opened 2020-11-17
Improve docs for Event and LifeCycle enums

cmyr:
This is a small tweak to try and clarify how these are different,
and to highlight the two 'special case' events, Command and
Notification.


druid/src/event.rs
```diff
@@ -119,16 +123,28 @@ pub enum Event {
...
+    /// then the [`MenuItem`]s in the menu will each correspond to a `Command`.
+    /// When the menu item is selected, that [`Command`] will be delivered to
+    /// the root widget of the appropriate window.
+    /// - If you are doing work in another thread (using an [`ExtEventSink`])
+    /// then [`Command`]s are the mechanism by which you communicate back to
+    /// the main thread.
+    /// - Widgets and other Druid components can send custom [`Command`]s at
+    /// runtime, via methods such as [`EventCtx::submit_command`].
     /// Called with an arbitrary [`Command`], submitted from elsewhere in
     /// the application.
```
jneem, line 141: Did you mean to leave this sentence in? It looks redundant, given the new part.

jneem (approved):

druid/src/event.rs
cmyr: Good catch!

PR #1410, opened 2020-11-18
Add missing call to `CloseClipboard` on windows

andrewhickman:
This could cause a druid application to steal control of the clipboard, preventing it from working in other apps


raphlinus (approved):
Seems good to me. Thanks for catching this!

ForLoveOfCats (approved):
This explains the oddities I kept seeing with attempting to take screenshots of my app on Windows! I can confirm that this fixes the issue and looks reasonable. This should get an entry under "fixed" in the changelog and then it looks good to me :)

ForLoveOfCats:
Dang it Raph!

andrewhickman:
Just spotted another case where this isn't always called, so don't merge just yet 😄

andrewhickman:
`get_format` wasn't calling `CloseClipboard` in the `None` path, so I've wrapped everything in a little helper function instead

cmyr (approved):
Thanks!

PR #1412, opened 2020-11-18
Center the individual widgets in the hello example.

### LINKED ISSUE #1406: jitter while editing TextBox
Not sure what's going on here, but it looks like something about the width of the text in the textbox is causing it to be positioned slightly differently? I see this on both linux and mac, so it's likely a druid thing.

jneem:
Fixes #1406.


cmyr (approved):
Good detective work, this makes sense. Will need to watch out for this in other situations as well.

cmyr:
going to merge this to fix that `to_rgba8` warning.

PR #1413, opened 2020-11-19
Use set_origin instead of set_layout_rect in built-in widgets

jpochyla:
This replaces usages of `WidgetPod::set_layout_rect` by `WidgetPod::set_origin` in all built-in widgets. This means that the layout rect is always left to be determined solely by the child's `layout` method. Only widget not obeying this was `tabs::TabBar`, expanding the minor axis of tab labels to the maximum (so i.e. in the vertical alignment they have the same width). This was replaced by keeping the layout as returned from the child, and expanding the minor axis later in `paint`.

### DIFF
--- a/druid/examples/timer.rs
+++ b/druid/examples/timer.rs
@@ -18,7 +18,7 @@ use std::time::Duration;
 
 use druid::widget::prelude::*;
 use druid::widget::BackgroundBrush;
-use druid::{AppLauncher, Color, LocalizedString, Point, Rect, TimerToken, WidgetPod, WindowDesc};
+use druid::{AppLauncher, Color, LocalizedString, Point, TimerToken, WidgetPod, WindowDesc};
 
 static TIMER_INTERVAL: Duration = Duration::from_millis(10);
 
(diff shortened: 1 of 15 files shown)


cmyr (approved):
Cool, thanks for this.

The reason I had avoided making this change initially because I wanted to see if any widgets were breaking this rule, because it would indicate something we needed to fix; but looks like you beat me to it.

About the tabs situation: this is a hiccup in other widgets, like `Flex`, as well; the ability to force widgets to match the size on some axis of their largest sibling. Doing this would require doing two `layout` calls to each child, but that might be something to consider at some point.

In any case, thanks!


This can be rebased to fix the `to_rgba8` warnings.

jpochyla:
> About the tabs situation: this is a hiccup in other widgets, like Flex, as well; the ability to force widgets to match the size on some axis of their largest sibling.

Oh but `Flex` isn't doing this, currently, right? Just checking if I missed something.

> This can be rebased to fix the to_rgba8 warnings.
 
Done, thanks <3

cmyr:
Correct, `Flex` does not currently provide this, but I think maybe it should?

PR #1414, opened 2020-11-19
Add source data type as param of LensWrap

cmyr:
This resolves an issue with WidgetExt being unable to appropriately
infer this parameter in certain circumstances. This is illustrated
with the included test case, which fails to compile without this
change.

### DIFF
--- a/druid/src/lib.rs
+++ b/druid/src/lib.rs
@@ -211,10 +211,6 @@ pub use window::{Window, WindowId};
 #[cfg(test)]
 pub(crate) use event::{StateCell, StateCheckFn};
 
-#[deprecated(since = "0.7.0", note = "use druid::widget::LensWrap instead")]
-#[allow(missing_docs)]
-pub type LensWrap<A, B, C> = widget::LensWrap<A, B, C>;
-
 /// The meaning (mapped value) of a keypress.
 ///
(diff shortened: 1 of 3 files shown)


jneem:
This seems like an improvement, although I wonder if it would be better for `Lens` to have associated types instead of type parameters? Then you'd presumably only need `LensWrap<L, W>` (with `impl<L, W> Widget<L::In> for LensWrap<L, W> where W: Widget<L::Out>`)

cmyr:
I was wondering about that as well. The possible limitations would be... `Lens` could only be implemented for a single concrete pair of params, per concrete lens struct? I don't know if this would be an issue. Can you think of any other concerns?

jneem:
That occurred to me too, but I couldn't think of an example. You could still have, e.g., an identity lens by allowing generic params on the struct that you're implementing `Lens` for:
```rust
struct Identity<T> { /* phantom, etc */ }

impl<T> Lens for Identity<T> {
    type In = T;
    type Out = T;
    // etc...
}
```

cmyr:
I'm open to trying that change, but I don't feel like it's particularly urgent; given that `LensWrap` is used through `WidgetExt` in the majority of cases I don't think the extra param is much of a burden. That said if you're motivated, go for it :)

jneem (approved):
I might give it a try later, to see how it works...

PR #1415, opened 2020-11-19
Fix clippy lints for 1.48

cmyr:


raphlinus (approved):
I'll approve because it's always good to resolve lints, but I'm wondering whether to deny the `rc_buffer` lint project-wide. https://github.com/rust-lang/rust-clippy/issues/6170 points out some of the issues, and Colin points out on Zulip that we often *do* mutate the interior of these buffers, thanks to `make_mut` and friends.

PR #1416, opened 2020-11-19
Revise docs in druid_shell::window.

jneem:
In celebration of the fact that intra-doc-links just stabilized, I took a stab at updating some docs.

### DIFF
--- a/druid-shell/src/window.rs
+++ b/druid-shell/src/window.rs
@@ -164,32 +164,21 @@ impl WindowHandle {
     }
 
     /// Sets the state of the window.
-    ///
-    /// [`state`]: enum.WindowState.html
     pub fn set_window_state(&mut self, state: WindowState) {
         self.0.set_window_state(state);
     }
 
     /// Gets the state of the window.
-    ///
-    /// [`state`]: enum.WindowState.html
     pub fn get_window_state(&self) -> WindowState {
         self.0.get_window_state()
     }
 
-    /// Allows the operating system to handle a custom titlebar
-    /// like the default one.
+    /// Informs the system that the current location of the mouse should be treated as part of the
+    /// window's titlebar. This can be used to implement a custom titlebar widget. Note that
+    /// because this refers to the current location of the mouse, you should probably call this
+    /// function in response to every relevant [`WinHandler::mouse_move`].
     ///
-    /// It should be used on Event::MouseMove in a widget:
-    /// `Event::MouseMove(_) => {
-    ///    if ctx.is_hot() {
-    ///        ctx.window().handle_titlebar(true);
-    ///    } else {
-    ///        ctx.window().handle_titlebar(false);
-    ///    }
-    ///}`
-    ///
-    /// This might not work or behave the same across all platforms.
+    /// This is currently only implemented on Windows.
     pub fn handle_titlebar(&self, val: bool) {
         self.0.handle_titlebar(val);
     }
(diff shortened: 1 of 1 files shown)


druid-shell/src/window.rs
```diff
@@ -253,14 +229,11 @@ impl WindowHandle {
...
-    ///
-    /// [`prepare_paint`]: trait.WinHandler.html#tymethod.prepare_paint
-    /// [`paint`]: trait.WinHandler.html#tymethod.paint
-    /// [`invalidate`]: struct.WindowHandle.html#tymethod.invalidate
-    /// [`invalidate_rect`]: struct.WindowHandle.html#tymethod.invalidate_rect
+    /// Request that [`prepare_paint`](WinHandler::prepare_paint) and [`paint`](WinHandler::paint)
+    /// be called next time there's the opportunity to render another frame. This differs from
+    /// [`invalidate`](WindowHandle::invalidate) and
+    /// [`invalidate_rect`](WindowHandle::invalidate_rect) in that it doesn't invalidate any part
+    /// of the window.
```
cmyr, line 236: my inclination would be to continue keeping the links at the bottom, to make it easier to read in the code itself; this still works with intradoc links.

druid-shell/src/window.rs
```diff
@@ -349,13 +316,11 @@ impl WindowHandle {
...
-    ///
-    /// The returned [`Scale`] is a copy and thus its information will be stale after
-    /// the platform DPI changes. A correctly behaving application should consider
-    /// the lifetime of this [`Scale`] brief, limited to approximately a single event cycle.
+    /// Get the DPI scale of the window.
     ///
-    /// [`Scale`]: struct.Scale.html
+    /// The returned [`Scale`](crate::Scale) is a copy and thus its information will be stale after
+    /// the platform DPI changes. A correctly behaving application should consider the lifetime of
+    /// this `Scale` brief, limited to approximately a single event cycle.
```
cmyr, line 323:
```suggestion
    /// the platform DPI changes. This means you should not stash it and rely on it later; it is
    /// only guaranteed to be valid for the current pass of the runloop.
```

druid-shell/src/window.rs
```diff
@@ -367,43 +332,34 @@ pub struct WindowBuilder(platform::WindowBuilder);
...
+    /// Takes the [`Application`](crate::Application) that this window is for.
     pub fn new(app: Application) -> WindowBuilder {
         WindowBuilder(platform::WindowBuilder::new(app.platform_app))
     }
 
-    /// Set the [`WinHandler`]. This is the object that will receive
-    /// callbacks from this window.
-    ///
-    /// [`WinHandler`]: trait.WinHandler.html
+    /// Set the [`WinHandler`]. This is the object that will receive callbacks from this window.
```
cmyr, line 340: Although we aren't always the best at this, a good rule of thumb for docstrings is to treat them like git commits; A very short, maximally descriptive line, with everything else after a newline. The rationale here is that the first line shows up in the rendered docs as the description of an item in a module, for instance.

druid-shell/src/window.rs
```diff
@@ -199,51 +188,38 @@ impl WindowHandle {
         self.0.show_titlebar(show_titlebar)
     }
 
-    /// Sets the position of the window in virtual screen coordinates.
-    /// [`position`] The position in pixels.
-    ///
-    /// [`position`]: struct.Point.html
+    /// Sets the position of the window in pixels, relative to the origin of the virtual screen.
```
cmyr, line 191: might as well link to `Scale` for pixels?

cmyr (approved):
Awesome, thanks for this! A few observations but feel free to merge when you're happy.

PR #1417, opened 2020-11-20
Add theme::UI_FONT_BOLD and theme::UI_FONT_ITALIC

cmyr:
had resisted these but it's nice to be able to create a bold/italic font descriptor with the same size as the normal `UI_FONT`.


raphlinus (approved):
Sure, I can see the value in these.

PR #1419, opened 2020-11-20
impl `Data` for `&'static str`

derekdreery:
The impl uses pointer equality, the same as for most other types.

### DIFF
--- a/druid/src/data.rs
+++ b/druid/src/data.rs
@@ -14,6 +14,7 @@
 
 //! Traits for handling value types.
 
+use std::ptr;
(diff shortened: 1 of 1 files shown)


derekdreery:
Interestingly, with `&'static str`s that really are part of the binary data section, `Data::same(&x, &y)` iff `*x == *y`. However this doesn't hold if the data was created on the heap and then forgotten. (A `&'static str` on the stack is UB)

cmyr (approved):
Seems reasonable, thanks!

PR #1420, opened 2020-11-20
Fix crochet helper method.

Finnerale:
In #1394 notifications were added, but the `EventCtx` in the crochet helper method `with_event_context` was not updated accordingly.


raphlinus (approved):
Thanks for being on top of this!

PR #1422, opened 2020-11-21
Add a couple of missing with_/set_ methods.

bheisler:
Most Druid widgets have builder-style with_* methods alongside setter-style set_* methods. I noticed a couple of cases where one or the other was missing and added them.

### DIFF
--- a/druid/src/widget/textbox.rs
+++ b/druid/src/widget/textbox.rs
@@ -149,6 +149,11 @@ impl<T> TextBox<T> {
         self
     }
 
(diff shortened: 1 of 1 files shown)


bheisler:
Ah, turns out that Flex does have a `cross_axis_alignment` method, I just didn't notice it because it doesn't follow the naming convention.

cmyr (approved):
Thanks! Yea, we aren't always 100% consistent with the naming conventions. :/

PR #1424, opened 2020-11-22
Base the identity example of async_event

JAicewizard:
The identity is in many ways similar to the async event one, and I feel that the current implementation of identity has a lot of clutter around the core concept. Making identity more similar to async event helps people in not having to understand 2 implementations of the same concept, having a widget change color.


cmyr (commented):
Thinking about this, I would actually be happy to remove this example, the new `validation` example will make use of widget identity, and we can point folks there if they're curious about that concept.

JAicewizard:
I think its nice for new users to have smaller examples that show how to do X. If we point users to the validation example, which is quite big, someone who doesn't know the feature yet, might get lost in the surrounding code that has to do with validation.

cmyr:
That may be true. In that case i think I would prefer to leave this example as it was before, because it just includes fewer concepts; there's no controller, no `ExtEventSink`, etcetera. I think the fewer concepts that are used in an example the clearer it is what is actually being demonstrated?

JAicewizard:
That is definitely something I thought about too, this change means that you need to understand the async_event example before you can look at this one. Everything taken into account I thought this simplified the code enough to be worth it. It also means that if you already understand the async_event widget this one will be easier. (since this one is quite messy with all the colour generations and commands/timers going around).

I might rethink this widget all together to remove the colours all together and just have it be 2 buttons that change 2 numbers.

JAicewizard:
I think I'm going to close this in favour of #1435

PR #1428, opened 2020-11-24
Create a constant lens.

derekdreery:
Following on from https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/druid-chatter/near/217758048, here is a PR to add a lens and constructor for it that always return the same value and discard changes. It's useful for when you have a widget that is usually dynamic, but sometimes can be static.

Comments welcome!

### DIFF
--- a/druid/src/lens/lens.rs
+++ b/druid/src/lens/lens.rs
@@ -553,13 +553,9 @@ where
     }
 }
 
-/// Create a lens that always gives the same value and discards changes.
-pub fn constant<T>(t: T) -> Constant<T> {
-    Constant(t)
-}
-
+/// A lens that always gives the same value and discards changes.
 #[derive(Debug, Copy, Clone)]
-pub struct Constant<T>(T);
+pub struct Constant<T>(pub T);
 
 impl<A, B: Clone> Lens<A, B> for Constant<B> {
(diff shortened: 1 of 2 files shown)


derekdreery:
An alternative would be to get rid of the free function and make the value inside the lens `pub`.

derekdreery:
Edit: I made it use a struct like the other lenses.

cmyr:
funny, I've been thinking about something similar lately.

My version of this is a "mixin lens", that converts `A -> AB` by adding some state. My vision of this, though, is that it isn't exactly a lens, but a special widget; the widget stores some state, `B`, and some pair of methods to do `A -> AB` and `AB -> B`, and then the dream (this is actually tricky, because commands are not generic) is that you can send this widget commands that will modify its internal state.

So... maybe my version doesn't make too much sense, right now.

In any case, it feels like this is sort of highlighting a larger problem, which is related to my work with errors, where the assumption that widgets are all displaying data from the global app model just doesn't really make sense.

I think the 'correct' solution to this will involve some real architectural thinking. I'm not opposed to having a lens like this in the meantime; it reminds me a lot of the work on 'scopes', and this is just a minimal, static scope?

derekdreery:
I agree with all you say. I think there is a place for this lens - I think it feels intuitive.

cmyr (approved):
sounds good!

Issue #1426, opened 2020-11-23
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value'


NatsuShio:
```
C:/Users/me/.cargo/bin/cargo.exe run --color=always --package figure_sauce --bin figure_sauce
    Finished dev [unoptimized + debuginfo] target(s) in 0.24s
     Running `target\debug\figure_sauce.exe`
DEBUG [druid::localization] available locales [], current en-US
DEBUG [druid::localization] resolved: [en-US]
DEBUG [druid_shell::platform::windows::window] dxgi factory pointer = 0x212e39d6980
DEBUG [druid_shell::platform::windows::window] 0x212e39f3650: desc = Some("NVIDIA GeForce GTX 965M"), vram = 4252762112
DEBUG [druid_shell::platform::windows::window] 0x212e39f3b00: desc = Some("Intel(R) HD Graphics 530"), vram = 134217728
DEBUG [druid_shell::platform::windows::window] 0x212e39a1ff0: desc = Some("Microsoft Basic Render Driver"), vram = 0
DEBUG [druid_shell::platform::windows::window] adapter = 0x212e39f3650
DEBUG [druid_shell::platform::windows::window] swap chain res = 0x0, pointer = 0x212e5902ed0
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', C:\Users\me\.cargo\git\checkouts\druid-f71533f3d81c0bc8\e677854\druid-shell\src\platform\windows\window.rs:366:46
stack backtrace:
   0: std::panicking::begin_panic_handler
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4\/library\std\src\panicking.rs:483
   1: core::panicking::panic_fmt
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4\/library\core\src\panicking.rs:85
   2: core::panicking::panic
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4\/library\core\src\panicking.rs:50
   3: core::option::Option<mut piet_direct2d::d2d::DeviceContext*>::unwrap<mut piet_direct2d::d2d::DeviceContext*>
             at C:\Users\me\.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\core\src\option.rs:383
   4: druid_shell::platform::windows::window::WndState::render
[...]

PR #1429, opened 2020-11-24
Attempt to provide more info to @NatsuShio

derekdreery:
Provide more error information for #1426.

I haven't checked this as I'm on linux.


cmyr:
Has this outlived its useful life?

derekdreery:
@cmyr I think this patch is an improvement that stands on its own irrespective of the original need. It provides more information on an error than previously.

cmyr (approved):
Fair enough, better error reporting makes sense.

PR #1434, opened 2020-11-28
Remove ImageBuf from druid-shell

jneem:
We've got a piet release containing `ImageBuf`, so remove it from druid-shell.


jneem:
Never mind, this is too soon: piet-common doesn't forward image yet.

cmyr (approved):
Great, this seems like a good bit of unification.

PR #1437, opened 2020-11-28
slight cleanup of the flex showcase

JAicewizard:
I decided to not really comment the code. Most of it is UI code and it is more of a widget showcase (like the GTK3 widget factory) then a real example.


druid/examples/flex.rs
```diff
@@ -13,17 +13,39 @@
 // limitations under the License.
 
 //! Demonstrates alignment of children in the flex container.
+//! This is more of a widget showcase where you can turn all the knobs
+//! and see how it behaves and less of an example.
```
derekdreery, line 17:
```suggestion
//! This example showcases the full set of functionality of flex, giving you 
//! knobs to change all the parameters. 99% of the time you will want to
//! hard-code these parameters, which will simplify your code considerably.
```

What do you think? Is this what you're getting at with your comment? Feel free to push back if you prefer the original.

druid/examples/flex.rs
```diff
@@ -13,17 +13,39 @@
...
+    ("Around", MainAxisAlignment::SpaceAround),
+];
+const CROSS_AXIS_ALIGNMENT_OPTIONS: [(&str, CrossAxisAlignment); 4] = [
+    ("Start", CrossAxisAlignment::Start),
+    ("Center", CrossAxisAlignment::Center),
+    ("End", CrossAxisAlignment::End),
+    ("Baseline", CrossAxisAlignment::Baseline),
+];
+const FLEX_TYPE_OPTIONS: [(&str, FlexType); 2] =
+    [("Row", FlexType::Row), ("Column", FlexType::Column)];
```
derekdreery, line 48: This is much neater IMO. :smile:

derekdreery (changes requested):
Thanks for the PR. Only minor comments. Feel free to push back.

derekdreery (approved):
Thanks!

druid/examples/flex.rs
JAicewizard: yeah having them inline wasn't very nice, just a shame it cant be a `Vec` but so be it.

PR #1438, opened 2020-11-28
Add default minimum size to `WindowConfig`

### LINKED ISSUE #1303: Have a sane default minimum window size
Currently by default windows can be made arbitrarily small. I think there should be a default minimum size in the 400x400 range, and if you want your window to be resizeable below that you need to explicitly set the minimum size to something smaller, using [`WindowDesc::with_min_size`](https://docs.rs/druid/0.6.0/druid/struct.WindowDesc.html#method.with_min_size).

Just as a note: you can set a minimum size for a window, and I think this should be set for most actual applications? I actually think this should have reasonable defaults, and if you want to be able to resize windows to be very small you should have to specify that manually, but that's also a separate issue. 😁

_Originally posted by @cmyr in https://github.com/linebender/druid/issues/1297#issuecomment-707235539_

colinfruit:
fixes #1303.

### DIFF
--- a/druid/src/app.rs
+++ b/druid/src/app.rs
@@ -41,7 +41,7 @@ pub struct AppLauncher<T> {
 /// It does not include anything related to app data.
 pub struct WindowConfig {
     pub(crate) size: Option<Size>,
-    pub(crate) min_size: Option<Size>,
+    pub(crate) min_size: Size,
     pub(crate) position: Option<Point>,
     pub(crate) resizable: Option<bool>,
     pub(crate) show_titlebar: Option<bool>,
@@ -189,7 +189,7 @@ impl Default for WindowConfig {
     fn default() -> Self {
         WindowConfig {
             size: None,
-            min_size: Some(WINDOW_MIN_SIZE),
+            min_size: WINDOW_MIN_SIZE,
             position: None,
             resizable: None,
             show_titlebar: None,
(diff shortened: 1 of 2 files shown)


druid/src/app.rs
```diff
@@ -187,7 +189,7 @@ impl Default for WindowConfig {
     fn default() -> Self {
         WindowConfig {
             size: None,
-            min_size: None,
+            min_size: Some(WINDOW_MIN_SIZE),
```
derekdreery, line 192:
```suggestion
            min_size: WINDOW_MIN_SIZE,
```

derekdreery, line 192: You also need to change the type definition on line 44 and the function on line 236.

derekdreery, line 192:
```suggestion
            min_size: WINDOW_MIN_SIZE,
```

derekdreery (changes requested):
Hi @colinfruit, thanks for the PR.

All looks good, I just have 1 question that others might need to address:

With this change, I don't think it's possible to construct the `None` variant of the minimum window size any more. We should either

 1. Make it a `Size` instead of `Option<Size>`.
 2. Allow unsetting the minimum size.

I'm almost sure (1) is the correct approach, because you can still do `set_min_size(Size::new(0., 0.))`. So if you make this change I'll merge the PR. It can always be changed to (2) later.

derekdreery (approved):
Thanks!

PR #1440, opened 2020-12-02
Add TextBox::text_position

cmyr:
This returns the offset in the text fields frame where
the text is actually drawn; I need this for the hacky
EditableLabel widget in Runebender.


derekdreery (approved):
Looks great. Just out of interest, what do you envisage the "non-hacky" solution being?

cmyr:
Good question. I think maybe just allowing you to specify a 'style' for the textbox for different states?

PR #1441, opened 2020-12-02
WidgetPod: change not laid out panic to warning

Maan2003:
This changes `debug_panic` about widgets recieving events without being laid, back to warning. This was introduced in 87157577b532b9c3422cc9c3897aec02cb733898.


cmyr (approved):
Thanks, I don't think we need to keep crashing all the time until we've figured out what our actual invariants are. 👍

PR #1442, opened 2020-12-02
(game of life) Fix fps not updating imidiatly

JAicewizard:
This is pretty self-explanatory, the non behaviour-changing things coming in another PR


cmyr (approved):
Makes sense, thanks!

PR #1443, opened 2020-12-02
cleanup game of life

JAicewizard:
This cleans some things up like using ? where applicable and removing unnecessary code.


JAicewizard:
I need to get the vscode rust extension to work again. It just doesnt do anything useful.

druid/examples/game_of_life.rs
```diff
@@ -57,7 +57,6 @@ impl Grid {
                 // death by loneliness or overcrowding
                 if life && (n_lives_around < 2 || n_lives_around > 3) {
                     indices_to_mutate.push(pos);
-                    continue;
```
derekdreery, line 150: Whilst it is true that the code will perform identically without this `continue`, keeping it in might communicate intent better. What do you think?

JAicewizard: oh I actually thought it did the opposite, having a break/continue within a loop makes it harder to reason about.
After looking at the idiomatic implementation (https://rustwasm.github.io/book/game-of-life/implementing.html) I think using a match might actually be easiest to reason about?

druid/examples/game_of_life.rs
```diff
@@ -12,125 +12,35 @@
...
 use druid::widget::{Button, Flex, Label, Slider};
 use druid::{
-    AppLauncher, Color, Data, Lens, LocalizedString, MouseButton, Point, Rect, TimerToken,
-    WidgetExt, WindowDesc,
+    AppLauncher, Color, Data, Lens, MouseButton, Point, Rect, TimerToken, WidgetExt, WindowDesc,
 };
 use std::sync::Arc;
 
-const GRID_SIZE: usize = 40;
+const GRID_SIZE: usize = 41;
```
cmyr, line 28: what's this about?

druid/examples/game_of_life.rs
```diff
@@ -12,125 +12,35 @@
...
-const C0: Color = Color::from_rgba32_u32(0xEBF1F7);
-const C1: Color = Color::from_rgba32_u32(0xA3FCF7);
-const C2: Color = Color::from_rgba32_u32(0xA2E3D8);
-const C3: Color = Color::from_rgba32_u32(0xF2E6F1);
-const C4: Color = Color::from_rgba32_u32(0xE0AFAF);
+const C0: Color = Color::from_rgba32_u32(0xEBF1F7); //Color::rgb(235, 241, 247)
+const C1: Color = Color::from_rgba32_u32(0xA3FCF7); //Color::rgb(162,252,247)
+const C2: Color = Color::from_rgba32_u32(0xA2E3D8); //Color::rgb(162,227,216)
+const C3: Color = Color::from_rgba32_u32(0xF2E6F1); //Color::rgb(242,230,241)
+const C4: Color = Color::from_rgba32_u32(0xE0AFAF); //Color::rgb(224,175,175)
```
cmyr, line 36: I agree that using `Color::rgb` would be better here.

druid/examples/game_of_life.rs
```diff
@@ -482,22 +464,10 @@ impl IndexMut<GridPos> for Grid {
...
-    fn eq(&self, other: &Self) -> bool {
-        for i in 0..POOL_SIZE {
-            if self.storage[i as usize] != other.storage[i as usize] {
-                return false;
-            }
-        }
-        true
-    }
-}
-
```
cmyr, line 495: hmm agree that this is weird, not sure what the rationale was.

druid/examples/game_of_life.rs
```diff
@@ -143,16 +53,15 @@ impl Grid {
...
-                    indices_to_mutate.push(pos);
-                }
+                match (self[pos], n_lives_around) {
+                    // death by overcrowding
+                    (true, x) if x > 3 => indices_to_mutate.push(pos),
+                    // death by loneliness
+                    (true, x) if x < 2 => indices_to_mutate.push(pos),
+                    // resurrection by life support
+                    (false, 3) => indices_to_mutate.push(pos),
+                    (_, _) => {}
```
cmyr, line 63: nittiest nit: the idiomatic way to write this would be as `_ => (),`. The rationale is that `()` (the [`unit` type](https://doc.rust-lang.org/std/primitive.unit.html) is what is returned by functions & expressions with no explicit return type, and we have to return the same type as the other branches of the `match`, so we should just do that explicitly.

The difference is that `{}` is an empty *expression*, which happens to evaluate to `()`; so it works fine either way, but the first is more explicit.
```suggestion
                    _ => (),
```

druid/examples/game_of_life.rs
```diff
@@ -210,6 +119,76 @@ impl Grid {
...
+struct ColorScheme {
+    colors: Vec<Color>,
+}
+
+#[derive(Clone, Lens, Data)]
+struct AppData {
+    grid: Grid,
+    drawing: bool,
+    paused: bool,
+    ups: f64,
```
cmyr, line 181: I would prefer some other name, I don't think it is obvious what `ups` means.

cmyr (commented):
Sorry for the wait! A few little questions but this looks good!

JAicewizard:
I thought I left replied for this ages ago!!! now they're all gone...

druid/examples/game_of_life.rs
JAicewizard: This is so that the background grid is not aligned. If it is aligned you get straight lines as cels which is a bit boring, and also doesn't highlight that it is actually a grid. This makes it so that neighboring cells (almost?? I dont know for sure anymore) always have a different colour.

druid/examples/game_of_life.rs
JAicewizard: yup, but sadly that isnt possible. I made them already but they're not const so decided to put them as comments.

druid/examples/game_of_life.rs
JAicewizard: no problem, I missed that you could just do `_`. rust is such a big language its easy to miss some things.
Althought I thought `{}` was always considered idiomatic? 
[the rust reference](https://doc.rust-lang.org/reference/expressions/match-expr.html) uses it too.

druid/examples/game_of_life.rs
JAicewizard: `updates_per_second`?

druid/examples/game_of_life.rs
cmyr: Oh right I meant `Color::rgb8`. That's the `const` one that takes u8s:

```rust
const C0: Color = Color::rgb8(0xEB, 0xF1, 0xF7); 
// same as:
const C0: Color = Color::from_rgba32_u32(0xEBF1F7);
```

druid/examples/game_of_life.rs
cmyr: sure!

druid/examples/game_of_life.rs
cmyr: I would also actually fix up these names while we're in here, like `BG` could be `BACKGROUND` and then I'd try and give more meaningful names to the other colors too?

JAicewizard: I just made it an array, much simpler.

cmyr (approved):
Thanks!

PR #1444, opened 2020-12-02
Lens docs rewrite

derekdreery:
This is a PR that rewrites the `Lens` page of the druid book. It aims to help avoid some of the misunderstandings I've seen on zulip. I've currently written it "green" without looking at the original page, and I intend to take another pass incorporating the best bits of both, but I think reviewing it in its current state would be a valuable endeavor. My main questions are

 1. Is it easy (or easy-ish) and enjoyable to read?
 2. Does it cover all the concepts required to use `Lens` effectively, and in sufficient detail?

Interested to hear what you think! Note that there are some technical things that need to happen if the PR is accepted:

 - [ ] move examples out of the file and think about which of them can be doctested
 - [ ] explain how the `lens` macro works, with an example of how code is expanded
 - [ ] think about if there is duplication in first 2 paragraphs

### DIFF
--- a/docs/src/lens.md
+++ b/docs/src/lens.md
@@ -1,10 +1,12 @@
 # Lenses and the `Lens` trait
 
-One of the key abstractions in `druid` along with `Data` is the `Lens` trait. This page explains what they are, and then how to use them. `Lens`es may seem complicated at first, but they are also very powerful, allowing you to write code that is reusable and concise, understandable and powerful (once you understand `Lens`es themselves).
+One of the key abstractions in `druid` along with `Data` is the `Lens` trait. This page explains what they are, and then how to use them. `Lens`es may seem complicated at first, but they are also very powerful, allowing you to write code that is reusable, concise, and understandable (once you understand `Lens`es themselves).
 
 ## Fundamentals: Definition and Implementation
 
-Like Rust itself, lenses are one of those things that require effort up front to learn, but are very fun and powerful to use once you understand them. This section represents the effort part of the equation. I promise if you stick with it you will reap the rewards.
+Like Rust itself, lenses are one of those things that require effort up front to learn, but are very fun and effective to use once you understand them. This section represents the effort part of the equation. I promise if you stick with it you will reap the rewards.
+
+### Definition
 
 Let's start with the definition of a `Lens`:
 
@@ -20,6 +22,8 @@ I've copied this definition from the `druid` source code, but then simplified it
 
 The first thing to notice is the generics on the `Lens` itself. There are 3 types involve in the lens: the lens itself, `T` and `U`. The two type parameters represent the mis-match that lenses solve: we have a function that operates on `U`, and an object of type `T`, so we need to transform `T` into `U` somehow.
 
+### Implementation
+
 Time for an example. Let's implement & use `Lens` manually so we can see what's going on.
 
 ```rust
@@ -32,7 +36,7 @@ struct Container {
 // it might, for example it might contain an index into a collection.
 struct InnerLens;
 
-// Our lens will apply functions that take a `u8` to a `Container`.
+// Our lens will apply functions that operate on a `String` to a `Container`.
 impl Lens<Container, String> for InnerLens {
     fn with<F: FnOnce(&String)>(&self, data: &Container, f: F) {
         f(&data.inner);
@@ -45,6 +49,8 @@ impl Lens<Container, String> for InnerLens {
 ```
 
 This is a very simple case. All we need to do is project the function onto the field. Notice that this isn't the only vaid lens from `Container` to `String` we could have made - we could also project from `Container` to `another`. We made the choice how to transform `Container` into `String` when we implemented `Lens`.
+
+> Side note: Actually we could project on to any string we have access to, including something in a global mutex, or a string that we create and discard in the lens. Lenses made like this are usually not what you want.
 
 You'll also notice that both methods take an immutable reference to `self`, even the `mut` variant. The lense itself should be thought of as a fixed thing that knows how to do the mapping. In the above case it contains no data, and will most likely not even be present in the final compiled/optimized code.
 
@@ -84,7 +90,7 @@ impl Lens<Container2, Name> for NameLens {
 
 > Side note: if you try doing this with `struct Name<'a> { first: &'a String, ...`, you'll find that it's not possible to be generic over the mutability of the fields in `Name`, so we can't make the `Name` struct borrow the data both mutably and immutably. Even if we could in this case, things quickly get very complicated. Also, sometimes `Widget`s need to keep a copy of the data around for use internally. For now the accepted best practice is to make `Clone`ing cheap and use that.
 
-Now as I'm sure you've realised, the above is very inefficient. Given that we will be traversing our data very often, we need it to be cheap. (This wasn't a problem before, because when we don't need to build the inner type, we can just use references. It also wouldn't be a problem if our data was cheap to copy/clone, for example any of the primitive number types `u8`, ... `f64`) Luckily, this is exactly the kind of thing that rust excels at. Let's rewrite the above example to be fast!
+Now as I'm sure you've realised, the above is very inefficient. Given that we will be traversing our data very often, we need it to be cheap. (This wasn't a problem before, because when we don't need to build the inner type, we can just use references. It also wouldn't be a problem if our data was cheap to copy/clone, for example any of the primitive number types `u8`, ... `f64`.) Luckily, this is exactly the kind of thing that rust excels at. Let's rewrite the above example to be fast!
 
 ```rust
 struct Container2 {
@@ -137,7 +143,7 @@ Right, now you understand how `Lens`es work. Congratulations, you've done the ha
 
 ### Bonus - The actual `Lens` definition
 
-The actual definition of `Lens` in `druid` allows the user to return values from the lens. This isn't strictly necessary, one could always use a value captured in the closure to store the result, but it is useful. Also, because the types `T` and `U` always appear behind pointers (`&` and `&mut`), we can relax the `Sized` requirement that is applied by default, meaning we can implement `Lens` for types like `[T]` (slice) and `str`.
+The actual definition of `Lens` in `druid` allows the user to return values from the lens. This isn't necessary for the core functioning of the lens, but it is useful. Also, because the types `T` and `U` always appear behind pointers (`&` and `&mut`), we can relax the `Sized` requirement that is applied by default, meaning we can implement `Lens` for types like `[T]` (slice) and `str`.
 
 Here is the real definition for completeness:
 
@@ -153,6 +159,8 @@ pub trait Lens<T: ?Sized, U: ?Sized> {
 
 Now on to the more fun bit: how we can use `Lens`es to get all those lovely qualities we talked about in the introduction. What you'll notice in this section is that we rarely have to build lenses ourself: we can often get what we want using the `Lens` proc macro, or through the functions in `LensExt`.
 
+### Deriving lenses
+
 Let's go back to the first example we looked at, with one of the fields removed for simplicity:
 
 ```rust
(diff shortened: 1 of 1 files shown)


derekdreery:
I'd welcome reviews even if it fails CI - the failures will be down to the examples, and I'd rather not invest time fixing them unless there was interest in the main content.

docs/src/lens.md
```diff
@@ -1,149 +1,251 @@
...
+struct NameLens;
+
+impl Lens<Container2, Name> for NameLens {
+    // .. identical to previous example
+}
 ```
 
-That seems pretty simple and fairly annoying to write, which is why you
-generally don't have to.
+Now in addition to almost free `Clone`s, we also have cheap incremental updates to the data itself. In the case of names, this isn't that important, but if the vector had `1_000_000_000` elements, we could still make changes in only *O(log(n))* time (in this case the difference between `1_000_000_000` and `30` - pretty big!).
```
Zarenor, line 140: I found this a little confusing - where does 30 come from?
I quickly checked for both the natural log and the base-10 log, and neither was 30 - maybe a typo?
```
ln(1E9)  = 20.72
log(1E9) = 9.00
```

Otherwise, this is an excellent point! It's a huuuuge savings on even moderately sized collections.

docs/src/lens.md
```diff
@@ -1,149 +1,251 @@
...
-mutation; see the main crate documentation for more.
+> Side note: Because `()` only has 1 value, it does actually respect mutations. It's just that mutations always result in the same value again (`()`).
 
-As your application gets more complicated, it will become likely that you want
-to use fancier sorts of lensing, and `map` and company can start to get out
-of hand; when that happens, you can always implement a lens by hand.
+### The `lens` macro
 
-### Getting something from a collection
+Finally, there is a macro for constructing lenses on the fly. It allows you to lens into fields of a struct you don't control (so you can't derive `Lens` for it), it also allows lensing into tuples and tuple structs, and lastly it will create index lenses into slices.
```
Zarenor, line 227: Definitely good to bring this up. I think we need to demonstrate how this macro invocation is written, though. Maybe a contrived example on a standard library type? (Length of a vec?)

docs/src/lens.md
derekdreery: It was from log2, it was really meant to be illustrative only - in reality you probably do better by having 32 elements per node (or even more).

cmyr:
@derekdreery where's this at? Should I give it a close read-through with the intention of merging?

derekdreery:
@cmyr There are some things that need tidying up, for example moving the code out into separate files and making sure everything matches the reset of the book. I would say that rather than a close review, you should first read it as a piece of text and see if it flows well & if it tells a coherent story. If you like it & think it is an improvement on what was there before, then we can worry about the details & getting it merged.

docs/src/lens.md
```diff
@@ -1,149 +1,251 @@
...
-Let's say we're building a todo list application, and we are designing the widget
-that will represent a single todo item. Our data model looks like this:
+One of the key abstractions in `druid` along with `Data` is the `Lens` trait. This page explains what they are, and then how to use them. `Lens`es may seem complicated at first, but they are also very powerful, allowing you to write code that is reusable, concise, and understandable (once you understand `Lens`es themselves).
 
-```rust,noplaypen
-{{#include ../book_examples/src/lens_md.rs:todo_item}}
+## Fundamentals: Definition and Implementation
+
+Like Rust itself, lenses are one of those things that require effort up front to learn, but are very fun and effective to use once you understand them. This section represents the effort part of the equation. I promise if you stick with it you will reap the rewards.
+
```
jneem, line 8: This feels to me like it's repeating the previous paragraph.

jneem (commented):
I like it, particularly the way that it introduces the lenses fully first, before explaining how they're integrated with druid.

docs/src/lens.md
```diff
@@ -1,149 +1,251 @@
 # Lenses and the `Lens` trait
 
-Let's say we're building a todo list application, and we are designing the widget
-that will represent a single todo item. Our data model looks like this:
+One of the key abstractions in `druid` along with `Data` is the `Lens` trait. This page explains what they are, and then how to use them. `Lens`es may seem complicated at first, but they are also very powerful, allowing you to write code that is reusable, concise, and understandable (once you understand `Lens`es themselves).
```
cmyr, line 3: We've decided on Druid as the standard spelling for the project, only doing `druid` if we're referring to the crate (as in, talking about `druid` vs `druid-shell`.

I'd also be fine with 'Lenses' once we're talking about the concept, and not the trait. 🤓

docs/src/lens.md
cmyr: agree, I'm reminded of the old rub, "I would've written less but I didn't have time". I find in this sort of thing it's easy to write too much, and one of the challenges is going back and cutting stuff out.

docs/src/lens.md
```diff
@@ -1,149 +1,251 @@
...
 ```
 
-We would like our widget to display the title of the item, and then below
-that to display two checkmarks that toggle the 'completed' and 'urgent' bools.
-`Checkbox` (a widget included in Druid) implements `Widget<bool>`.
-How do we use it with `TodoItem`? By using a `Lens`.
+I've copied this definition from the `druid` source code, but then simplified it a little, by removing the return types, as they are not fundamental to the way lenses work.
 
-## Conceptual
+The first thing to notice is the generics on the `Lens` itself. There are 3 types involve in the lens: the lens itself, `T` and `U`. The two type parameters represent the mis-match that lenses solve: we have a function that operates on `U`, and an object of type `T`, so we need to transform `T` into `U` somehow.
```
cmyr, line 23: In technical writing I follow the principle of spelling out numerals up to at least ten.

docs/src/lens.md
```diff
@@ -1,149 +1,251 @@
...
 
-You can think of a lens as a way of "focusing in" on one part of the data. You
-have a `TodoItem`, but you *want* a `bool`.
+### Implementation
 
-`Lens` is a trait for types that perform this "focusing in" (aka *lensing*).
-A simplified version of the `Lens` trait might look like this:
+Time for an example. Let's implement & use `Lens` manually so we can see what's going on.
```
cmyr, line 27: I'd cut this down to something like, "Let's look at an example."

docs/src/lens.md
```diff
@@ -1,149 +1,251 @@
...
+    }
+
+    fn with_mut<F: FnOnce(&mut String)>(&self, data: &mut Container, f: F) {
+        f(&mut data.inner);
+    }
+}
 ```
 
-That is, this type takes an instance of `In`, and returns an instance of `Out`.
+This is a very simple case. All we need to do is project the function onto the field. Notice that this isn't the only vaid lens from `Container` to `String` we could have made - we could also project from `Container` to `another`. We made the choice how to transform `Container` into `String` when we implemented `Lens`.
```
cmyr, line 51:
```suggestion
This is a very simple case. All we need to do is project the function onto the field. Notice that this isn't the only valid lens from `Container` to `String` we could have made - we could also project from `Container` to `another`. We made the choice how to transform `Container` into `String` when we implemented `Lens`.
```

I think I would just use a single field with the first example? I like the idea of keeping it maximally simple.

cmyr (commented):
Some little initial comments, but maybe it's more interesting to focus on the higher-level stuff. Firstly: I realized halfway through this that I'm really not the best audience, since I have a pretty firm understanding of what's going on, and it's hard for me to tell what bits and pieces are helpful and which bits are too much.

I think we could put off showing the trait itself even further; I think the real challenge is coming up with a simple way of expressing the overall pattern. The concept I personally find most helpful is the idea of a "two-way map"; a way to both get a value out of a type, and also get a value back in.

One problem with `Lens` is that the trait is so gross-looking. It's really almost an implementation detail that these things are called `with` instead of `get`; we only need `with_mut` to take a function so that we have an opportunity to inspect the data afterwards and update our base state if we've done something fancy. If all lenses were field access, I think we could just have the methods be `get` and `get_mut`? (I haven't thought this through  completely and it's possible that the closures also satisfy the borrow checker, but I don't *think* that's the case)..

So I think from a high-level the best way to explain this might just be as fancy getters and setters? I think a major part of why folks get confused is just the new name and the weird looking trait definition; the underlying concept just isn't that zany.



Anyway those are my high-level thoughts. I think that overall these docs are good, and I'd like to encourage you to fix them up and we'll get them merged.

docs/src/lens.md
```diff
@@ -1,149 +1,251 @@
...
+### Implementation
 
-`Lens` is a trait for types that perform this "focusing in" (aka *lensing*).
-A simplified version of the `Lens` trait might look like this:
+Time for an example. Let's implement & use `Lens` manually so we can see what's going on.
 
-```rust,noplaypen
-{{#include ../book_examples/src/lens_md.rs:simple_lens}}
+```rust
+struct Container {
```
arthmis, line 30: I'm of the mind of using more concrete example. The inner and outer fields are sort of vague to me, even though they are defined right there. When I read `another` later on, I thought you were talking about something else. I was thinking you can use something like `Contact` as the container and `name` and `email` as the fields. They would still be strings.

docs/src/lens.md
```diff
@@ -1,149 +1,251 @@
...
+    inner: String,
+    another: String,
+}
```
arthmis, line 33: I'm of the mind of using more concrete example. The inner and outer fields are sort of vague to me, even though they are defined right there. When I read another later on, I thought you were talking about something else. I was thinking you can use something like Contact as the container and name and email as the fields. They would still be strings.

docs/src/lens.md
```diff
@@ -1,149 +1,251 @@
...
+// Our lens will apply functions that operate on a `String` to a `Container`.
+impl Lens<Container, String> for InnerLens {
+    fn with<F: FnOnce(&String)>(&self, data: &Container, f: F) {
+        f(&data.inner);
+    }
+
+    fn with_mut<F: FnOnce(&mut String)>(&self, data: &mut Container, f: F) {
+        f(&mut data.inner);
+    }
+}
```
arthmis, line 48: So, now `InnerLens` can be renamed to `NameLens`. Part of the issue I had reading this initially, was that I thought InnerLens meant lens that was internal rather than a lens that lenses into `inner`. I just kept thinking inner was something else. This is my fault, but think it would be better to have more memorable field names for the example.

docs/src/lens.md
arthmis: I don't quite understand what "project the function onto the field" means here. I guess you're using project as a synonym for `map`? Like map a field to another field? I only just thought of this, but Lensing is like mapping from one thing to another right?

docs/src/lens.md
```diff
@@ -1,149 +1,251 @@
...
-function a closure that will *itself* be passed an `Out`; if our closure returns
-a result, that will be given back to us.
+The trade-off is that we introduce more complexity into the `Name` type: to make changes to the data we have to use `Rc::make_mut` to get mutable access to the `String`. (The code in the lens will ensure that the newer copy of the `Rc`d data is saved to the outer type.) This means the writing fast druid code requires knowledge of the Rust pointer types (`Rc`/`Arc`, and also potentially `RefCell`/`Mutex`).
 
-This is unnecessary in the case of non-mutable access, but it is important for
-mutable access, because in many circumstances (such as when using an `Rc` or
-`Arc`) accessing a field mutably is expensive even if you don't do any mutation.
+We can actually do even better than this. Suppose that we are working on a vector of data rather than a string. We can import the `im` crate to get collections that use *structural sharing*, meaning that even when the vector is mutated, we only *Clone* what we need to. Because `im` is so useful, it is included in `druid` (behind the `im` feature).
```
arthmis, line 118: Although the introduction of `im` is important, I don't think it should be introduced in these docs. Maybe another example using the List widget would be nice. If anything I think there should be a sort of druid recipe book that demonstrates different types of ideas. i.e, different `Lens` implementations, how to use them with lists and so on.

derekdreery: Yeah I had thought about having a cookbook-style list of examples at the end. I think examples aid learning concepts, as well as being useful when you have a particular problem to solve and use a particular similar example.

arthmis: Maybe the Lens page should become a Lens chapter? After you've introduced the `Lens`, you can continue with a plethora of different examples and demonstrate how they transform/map data. Even better if the examples are runnable.

docs/src/lens.md
arthmis: Also maybe there should be an explanation of the paramter `f: F`. What is that function doing under the hood?

docs/src/lens.md
cmyr: I agree that maybe 'project' is a bit too mathy? @arthmis you're correct, you can think of a lens as a two-way map; you can get one value from another, and then mutations in that value are also reflected in the original value.

docs/src/lens.md
cmyr: thanks for your comments @arthmis, it's super valuable to get feedback from someone who is less familiar with the subject matter!

docs/src/lens.md
```diff
@@ -1,149 +1,251 @@
...
 
-Field access is a very simple (and common, and *useful*) case, but lenses can do much more than that.
+`LensExt` contains a few more useful methods for handling things like negating a boolean, or auto-`Deref`ing a value.
 
-### `LensExt` and combinators
+There are also 3 special structs in `druid::lens`: `Constant`, `Identity` and `Unit`. `Constant` is a lens that always returns the same value, and always discards any changes, while `Identity` is a lens that does nothing. You might say "what is the point of a lens that does nothing", which would be a fair question. Well, there are some places where a lens is required, and having an identity allows the user to say act as if there was no lens. It's also used to begin a composition chain using the combinators like `then`. `Unit` is a special case of `Constant` where the constant in question is `()`.
```
xulien, line 221: I'm currently trying to understand how the lens works, and that's typically the sort of thing I'm looking to clear up: When to use Identity ? some idiomatic examples would be welcome. There is a case in "examples/list.rs" but it is not obvious to me as a novice.

derekdreery:
@arthmis the cookbook model could work well: when starting out just search for the similar example to move you on, but as you do this more and more you start to see the patterns that facilitate how `Lens` works in general.

derekdreery:
I'm going to do another draft, but I think even then letting it stew some more is useful. Or if it gets merged, directing people somewhere to comment on it.

arthmis:
@derekdreery I agree that maybe lens by themselves don't need cookbook examples. I will try and submit some basic examples of lens inplementations and lens usage to you to see if you would find any useful.

derekdreery:
@arthmis awesome thanks! :)

PoignardAzur (approved):
I don't think this is perfect, but the druid book is a work in progress anyway. I'm in favor of merging this for now, and improving later if need be.

PR #1445, opened 2020-12-05
Expose `request_focus` on `WidgetPod`

ForLoveOfCats:
It can be extremely useful to give a widget focus when instancing it. This patch allows that by exposing `request_focus` on `WidgetPod`. It was required to change `Window`'s behavior to handle focus requests after dispatching `WidgetAdded` if prompted to do so by a `children_changed` call otherwise the subject widget would be marked as focused according to `ctx.is_focused()` but would not get keyboard events (in addition to the logged warning).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,7 @@ You can find its changes [documented below](#060---2020-06-01).
 - `WidgetPod::is_initialized` to check if a widget has received `WidgetAdded`. ([#1259] by [@finnerale])
 - `TextBox::with_text_alignment` and `TextBox::set_text_alignment` ([#1371] by [@cmyr])
 - Add default minimum size to `WindowConfig`. ([#1438] by [@colinfruit])
+- Expose `request_focus` on `WidgetPod`. ([#????] by [@ForLoveOfCats])
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,7 +56,7 @@ You can find its changes [documented below](#060---2020-06-01).
 - `WidgetPod::is_initialized` to check if a widget has received `WidgetAdded`. ([#1259] by [@finnerale])
 - `TextBox::with_text_alignment` and `TextBox::set_text_alignment` ([#1371] by [@cmyr])
 - Add default minimum size to `WindowConfig`. ([#1438] by [@colinfruit])
-- Expose `request_focus` on `WidgetPod`. ([#????] by [@ForLoveOfCats])
+- Expose `request_focus` on `WidgetPod`. ([#1445] by [@ForLoveOfCats])
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


cmyr:
sorry, just explain it to me like I'm five... what's the use case here? When a container creates a child, and wants that child to get focus?

cmyr:
Also please feel free to open a thread on zulip or leave a comment in druid-chatter if something is waiting for review, it's easy to lose sight of the review queue these days 😬

ForLoveOfCats:
> what's the use case here? When a container creates a child, and wants that child to get focus?

Yes, among other potential usage cases. Say you have a complex search panel widget which houses a textbox, flex of entries, ect. When you construct it you would want it to auto-focus the textbox. With this patch you can pass the event ctx to the `new` fn and let it give its textbox focus. In fact this is the very usage case that drove the implementation of this patch which I have been using for the past few days. This is a bit clunky in the case of initial UI construction where that parameter would have to become optional and yada yada, more mess.

The one alternative I can think of would be to pass a command down to the widget to have it request focus which is fairly reasonable when you control the implementation of the widget such as I do with my custom text edit widget but if the widget is coming from Druid itself or some other library then the situation is markedly more painful.

Something to note is that in some specific situations, the widget automatically grabbing focus by itself on `WidgetAdded` might be a reasonable solution but this covers only a very specific set of cases and is also limited to which widgets you have source control over. Of course there is the fact that `LifeCycleCtx` does not expose `request_focus` but that is a separate issue that I also intend to propose exposing in a separate PR. (Requesting focus on `WindowConnected` in `event` works for widgets constructed during initial UI setup but not for dynamically added widgets)

If anyone has a nicer potential solution please do throw it out, especially if it cleans up the messiness around construction during initial UI setup vs in response to an event. I remain open to the possibility that there might be an existing mechanism that can already handle this, if so that would be great.

> Also please feel free to open a thread on zulip or leave a comment in druid-chatter if something is waiting for review, it's easy to lose sight of the review queue these days

I would if it sat for a week or two :) I don't like to shout for attention on PRs unless it really is about to slip through the cracks, especially as I personally don't do to many reviews myself.

cmyr:
Okay, so thoughts on this:

To me, the preferred way of handling this would be some sort of simple controller widget that wraps the text field and handles requesting focus. Another viable option would be to have a method that lets you request focus for a specific widget, using its widget id.

A bigger issue, I think, is just that the whole focus system needs some more thought and work. There is #1089, which was an ambitious attempt to change how focus works, but i'm not fully convinced by that approach; however one thing it does offer is a mechanism for setting which widget gets initial focus.

Here, I think my main hesitation is about *reaching down* into the widget pod to ask it to change some state. This isn't something we really do anywhere else, and so I'm reluctant to add API that introduces a new pattern unless there's really no easy way to achieve this otherwise, and I think a simple `TakeFocus` controller would be a reasonable alternative, and I could also definitely imagine a `EventCtx::set_focus` method that lets you pass an explicit `WidgetId`.

ForLoveOfCats:
That sounds like a reasonable path forward while using current Druid idioms which is generally a good thing to try to do :+1: All that's missing at the moment is the ability for a widget to request focus on `WidgetAdded` so I'm going to go poke at that. There may be blockers on doing that similar to the changes that this patch required, I'll see when I try.

cmyr:
@ForLoveOfCats I think that #1480 might be a good way to make working with `WidgetAdded` much easier.

PR #1446, opened 2020-12-07
small changes in hello example

JAicewizard:
Just some small adjustments to make it a bit simpler.


druid/examples/hello.rs
```diff
@@ -12,62 +12,54 @@
...
 
     // arrange the two widgets vertically, with some padding
     Flex::column()
+        .cross_axis_alignment(druid::widget::CrossAxisAlignment::Center)
+        .main_axis_alignment(druid::widget::MainAxisAlignment::Center)
         .with_child(label)
         .with_spacer(VERTICAL_WIDGET_SPACING)
         .with_child(textbox)
-        .align_vertical(UnitPoint::CENTER)
+        .expand()
```
cmyr, line 64: Is there a reason why you prefer expanding and then aligning internally? To me it feels more confusing and probably less idiomatic?

druid/examples/hello.rs
```diff
@@ -12,62 +12,54 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//! This is a very small example of how to setup a druid application.
+//! It does the almost bare minimum while still being usefull.
```
cmyr, line 16:
```suggestion
//! This is a very small example of how to setup a druid application.
```

I think short concise comments are good!

druid/examples/hello.rs
```diff
@@ -12,62 +12,54 @@
...
+
+use druid::widget::prelude::*;
```
cmyr, line 18: do we actually need the widget prelude? I normally only import this if I am implementing the `Widget` trait.

druid/examples/hello.rs
```diff
@@ -12,62 +12,54 @@
...
-    .align_horizontal(UnitPoint::CENTER);
+    .with_text_size(32.0);
 
     // a textbox that modifies `name`.
     let textbox = TextBox::new()
         .with_placeholder("Who are we greeting?")
         .with_text_size(18.0)
-        .fix_width(TEXT_BOX_WIDTH)
-        .align_horizontal(UnitPoint::CENTER)
-        .lens(HelloState::name);
```
cmyr, line 65: my initial thinking was to use a struct + a lens instead of just `String` in order to highlight the `Lens` trait, which is one of the central concepts in druid? I don't mind changing it though, simple is good..

cmyr (commented):
Okay, some comments, but happy to have you hacking away! :)

druid/examples/hello.rs
JAicewizard: I dont really remember? I was probably trying to remove the import, but it's probably more idiomatic to have it the old way indeed.

druid/examples/hello.rs
JAicewizard: it seems a bit complex for a first look. I think its more important to be able to see how it works first before trying to get people to understand lenses.

druid/examples/hello.rs
JAicewizard: I thought the idiomatic thing was to just always import the prelude if you need basically anything from it.
I think just always having it is a good way to avoid editors adding individual imports instead of just this prelude.

druid/examples/hello.rs
cmyr: I think that `Data` and `Lens` are such fundamental concepts in druid that it is worth introducing them immediately.

JAicewizard: I reverted that back so now its the HelloState again with a lens

cmyr (approved):
Thanks!

PR #1447, opened 2020-12-07
change Image example

JAicewizard:
This changes the image example to showcase FillStrat and Interpolation mode.


druid/examples/image.rs
```diff
@@ -12,44 +12,46 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! This example shows how to draw an png image.
+//! This example shows how to draw an image. Using the Image feature.
+//! Diferent image formats can be "unlocked" by adding their
+//! corresponding feature.
```
cmyr, line 17:
```suggestion
//! This example shows how to draw an image; using images requires
//! enabling the `image` feature in your Cargo.toml; you will also need
//! to specify the image formats you would like to use. See druid/Cargo.toml
//! to see all features.
```

druid/examples/image.rs
```diff
@@ -12,44 +12,46 @@
...
-    AppLauncher, Color, ImageBuf, Widget, WindowDesc,
-};
+use druid::piet::InterpolationMode;
+use druid::widget::prelude::*;
+use druid::widget::{FillStrat, Flex, Image, WidgetExt};
+use druid::{AppLauncher, Color, ImageBuf, WindowDesc};
 
 pub fn main() {
     let main_window = WindowDesc::new(ui_builder);
-    let data = 0_u32;
```
cmyr, line 24: although it might seem useless, the idea here is to pretend more like we have actual app data, and also to make it clear that it is the app data that is passed in with launch. I don't think it matters much either way, but I want to at least explain the thinking. :)

druid/examples/image.rs
```diff
@@ -12,44 +12,46 @@
...
-        .fill_mode(FillStrat::FitWidth)
-        .border(Color::WHITE, 1.0);
-    col.add_flex_child(otherimage, 1.0);
-    col
+    // We create 2 images, one not having any modifications and the other
+    // is set to a fixed width, a fill strategy and an interpolation mode.
+    // You can see how this affects the final result. You can play with the
+    // Interpolation mode to see hwo this affects things.
+    // Note that this image is already anti-aliased so NearestNeighbor looks
+    // weird
```
cmyr, line 40:
```suggestion
    // We create two images, one not having any modifications and the other
    // forced to a fixed width, a fill strategy and an interpolation mode.
    // You can see how this affects the final result. You can play with the
    // interpolation mode to see how this affects things. (Note that this image
    // is already anti-aliased so NearestNeighbor looks weird.)
```

Note on this: we use the same `FillStrat` in both examples, so I'm not sure it's correct to say that the one has "no modifications"? What do you think?

cmyr (commented):
Cool, I think showing more off here makes sense. If we wanted to get fancy we could also do something here like we do in `examples/flex.rs`, where we actually let the user choose the `FillStrat` and `InterpolationMode` from some radio buttons, and then rebuild the widget in response to that?

druid/examples/image.rs
JAicewizard: yeah that makes more sense, thanks for catching that!

JAicewizard: I just tested and the fillstrat doesn't actually matter when there is no fixed width.

JAicewizard:
I reworked it to be more like the flex, that was a great idea. You can now play around with width/height interpolation mode and fillstrat dynamically

druid/examples/image.rs
```diff
@@ -12,44 +12,319 @@
...
+//! This example shows how to draw an image; using images requires
+//! enabling the `image` feature in your Cargo.toml; you will also need
+//! to specify the image formats you would like to use. See druid/Cargo.toml
+//! to see all features.
 
-use druid::{
-    widget::{FillStrat, Flex, Image, WidgetExt},
-    AppLauncher, Color, ImageBuf, Widget, WindowDesc,
-};
+/* use druid::piet::InterpolationMode;
```
cmyr, line 20: let's delete this commented out code before we merge.

druid/examples/image.rs
```diff
@@ -12,44 +12,319 @@
...
+ */
+
+use druid::piet::InterpolationMode;
+use druid::widget::{prelude::*, FillStrat, Image};
+use druid::widget::{
+    Checkbox, CrossAxisAlignment, Flex, Label, RadioGroup, SizedBox, Stepper, TextBox, WidgetExt,
+};
+use druid::{AppLauncher, Color, Data, ImageBuf, Lens, LensExt, WindowDesc};
+
+const FILL_STRAT_OPTIONS: [(&str, FillStrat); 7] = [
```
cmyr, line 66: instead of having these be `const` arrays, I would have them be `static` slices; it takes up less space in the generated binary, and it doesn't require you to specify the length.

```suggestion
static FILL_STRAT_OPTIONS: &[(&str, FillStrat)] = &[
```

druid/examples/image.rs
```diff
@@ -12,44 +12,319 @@
...
+    ("Contain", FillStrat::Contain),
+    ("Cover", FillStrat::Cover),
+    ("Fill", FillStrat::Fill),
+    ("FitHeight", FillStrat::FitHeight),
+    ("FitWidth", FillStrat::FitWidth),
+    ("None", FillStrat::None),
+    ("ScaleDown", FillStrat::ScaleDown),
+];
+
+const INTERPOLATION_MODE_OPTIONS: [(&str, InterpolationModeData); 2] = [
```
cmyr, line 76: same here.

druid/examples/image.rs
```diff
@@ -12,44 +12,319 @@
...
+
+#[derive(Clone, PartialEq)]
+struct InterpolationModeData(InterpolationMode);
+
+impl InterpolationModeData {
+    const fn new(mode: InterpolationMode) -> InterpolationModeData {
+        InterpolationModeData(mode)
+    }
+}
+impl druid::Data for InterpolationModeData {
```
cmyr, line 120: we should be able to derive this, like:

```rust
#[derive(Clone, Data)]
 struct InterpolationModeData(
    #[data(same_fn="PartialEq::eq")]
    InterpolationMode
);
```

I would also be okay with us adding a `Data` impl to `InterpolationMode`, I did this for the various types in `examples/flex.rs` for this same reason.

druid/examples/image.rs
```diff
@@ -12,44 +12,319 @@
...
+fn make_width() -> impl Widget<Params> {
+    Flex::column()
+        .cross_axis_alignment(CrossAxisAlignment::Start)
+        .with_child(Label::new("width:"))
+        .with_default_spacer()
+        .with_child(
+            Flex::row()
+                .with_child(
+                    TextBox::new()
+                        .parse()
```
cmyr, line 237: I'd like to not use `parse()` in new code; instead you should call `with_formatter` and pass in a `ParseFormatter`.

druid/examples/image.rs
```diff
@@ -12,44 +12,319 @@
...
+fn make_height() -> impl Widget<Params> {
+    Flex::column()
+        .cross_axis_alignment(CrossAxisAlignment::Start)
+        .with_child(Label::new("height:"))
+        .with_default_spacer()
+        .with_child(
+            Flex::row()
+                .with_child(
+                    TextBox::new()
+                        .parse()
```
cmyr, line 259: ditto

cmyr (commented):
Cool, I think this is a great improvement.

Playing around with it: the default settings might be confusing, since changing many of the options doesn't actually change anything.

I think we can maybe make it a bit clearer? Here's what I would probably do:

- remove the 'debug layout' option, and just always paint the border of the image widget, which should make it easier to see what's going on.
- I would check 'fix width' and 'fix height' by default, and give 'height' and 'width' some non-zero starting values; maybe something like 200 & 320. (making them non-uniform will make it more obvious what's happening with the various fill strategies.


Other than that, the code looks good, just a couple of notes in-line.

Thanks!

druid/examples/image.rs
JAicewizard: does it really take up less space? but yes that sounds like a good idea, I didn't like the fixes lengths.

druid/examples/image.rs
JAicewizard: yes that would be great! I think it makes sense to have `InterpolationMode` in data for some things, its not something we want to prevent people from putting in there.

JAicewizard:
had to rebase off master but forgot to pull this first, there was a commit I hadn't pulled apparently

JAicewizard:
the failing ones are getrandom not being compatible with wasm(?)

cmyr:
that getrandom error is strange, since we don't see it on master and I don't think anything should have changed??

druid/examples/image.rs
cmyr: Did we make this change? If we did we can also probably get rid of this `InterpolationModeData` type, and just use `InterpolationMode` directly?

druid/examples/image.rs
```diff
@@ -12,44 +12,230 @@
...
+        bc: &BoxConstraints,
+        data: &AppState,
+        env: &Env,
+    ) -> Size {
+        self.inner.layout(ctx, bc, data, env)
+    }
+
+    fn paint(&mut self, ctx: &mut PaintCtx, data: &AppState, env: &Env) {
+        self.inner.paint(ctx, data, env)
+    }
```
cmyr, line 120: One gotcha here: if you write a widget that wraps another widget and does *not* use a `WidgetPod`, you should implement the following method; it will only matter in a small number of cases (if your inner widget has an explicit `WidgetId`) but it is strictly correct:

```suggestion
    }
    
    fn id(&self) -> Option<WidgetId> {
        self.inner.id()
    }
```

cmyr (approved):
Okay one fixup and one question about just using `InterpolationMode`, but other than that (and assuming we can figure out why CI is failing) this looks good. Thanks!

druid/examples/image.rs
cmyr: (maybe do this yourself instead of just merging my change unseen, because I haven't tested it and I might've screwed something up)

druid/examples/image.rs
JAicewizard: `InterpolationMode` does not have the data trait implemented for it, yet. It would be nice if that would be possible to do. 
I can create a PR for it later?

cmyr: @JAicewizard yes feel free to add it to this PR. Also if merge in the latest master it should fix that problem with CI failing.

cmyr:
Thanks!

PR #1453, opened 2020-12-11
Cursor setting fixups

cmyr:
This fixes two small issues:

The first is that cursors set in leaf widgets were not being
handled correctly, due to the logic in `merge_up` not being called
on leaves.

The second is that on macOS we were not receiving a MouseMoved
event when we first left the window; we fix this by also explicitly
handling MouseLeave.

cc @jneem


jneem (approved):
Looks good, thanks!

PR #1461, opened 2020-12-13
Various image improvements

derekdreery:
Currently blocked on a new release of piet.

This patch adds 3 features:

 1. Allow the user to specify a subrectangle of an image to display.
 2. Add a `Send + Sync` bound to `dyn Error`s from `image`. This means that, for example, you can load an image on a dedicated IO thread, and pass the error back to the UI thread.
 3. Improve support for rendering svg images in the `Svg` widget.

This allows, for example, a sprite sheet to be shared between multiple `Image` widgets. It also allows my motivating use case, which is an image viewer where the user can zoom in on part of the image.

I've made the changes so they are backwards compatible.

Update: this patch is growing a bit, but each commit is completely disjoint and can be reviewed/accepted separately.

Note: edited heavily.

### DIFF
--- a/druid-shell/Cargo.toml
+++ b/druid-shell/Cargo.toml
@@ -38,7 +38,7 @@ hdr = ["piet-common/hdr"]
 [dependencies]
 # NOTE: When changing the piet or kurbo versions, ensure that
 #       the kurbo version included in piet is compatible with the kurbo version specified here.
-piet-common = "=0.3.1"
+piet-common = "=0.3.2"
 kurbo = "0.7.1"
 
 log = "0.4.11"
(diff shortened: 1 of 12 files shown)


derekdreery:
I didn't realise that `ImageBuf` already uses reference counting, so `Image` doesn't need to.

cmyr:
@derekdreery any interest in getting this landed? I'd like to do a release shortly.

derekdreery:
Ooh I'll revisit.

derekdreery:
Currently working on this.

derekdreery:
Looks like the new rust version introduced some new clippy lints. I'll fix them in this PR.

derekdreery:
Ok after all my work on making empty images work, I think maybe we should disallow it altogether (by checking size > 0). A druid test is now breaking because we try to create an affine map from null space to non-null space (which is impossible) so we get NaNs.

derekdreery:
I will submit a piet patch.

derekdreery:
Ok with the `+Send+Sync` patch for `piet`, this will be read to merge.

derekdreery:
Woop clippy fixed :fireworks:

cmyr (approved):
Looks good!

PR #1465, opened 2020-12-14
Manually track click interval

### LINKED ISSUE #1464: click count is not correct on web
Playing around a bit with building for wasm, I notice that doubleclicks are not correctly reported.

cmyr:
I want this for the hyperbezier demo. 😁 

This lets us 'correctly' report click-count for mouse events on wasm.
Doing this 'actually correctly' would require finding some way
to determine what the user's current doubleclick delay is, and I
do not believe this is exposed by browsers, although I haven't looked
too hard and it wouldn't blow my mind?

- closes #1464

### DIFF
--- a/druid-shell/src/platform/web/window.rs
+++ b/druid-shell/src/platform/web/window.rs
@@ -19,6 +19,7 @@ use std::cell::{Cell, RefCell};
 use std::ffi::OsString;
 use std::rc::{Rc, Weak};
 use std::sync::{Arc, Mutex};
+use std::time::Duration;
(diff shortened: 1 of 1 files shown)


jneem (approved):
Looks good to me. It sounds like `ClickCounter` might be useful in other backends too; I remember there was some issue about having multi-clicks on various platforms.

cmyr:
good note, yea that sparks some memory for me, I think there's something in the GTK backend? I can look into reusing this there as a separate PR.

PR #1467, opened 2020-12-16
A tree widget

tirix:
Hi all, this is a simple implementation for a tree widget, with the associated example.

This is my very first piece of rust code, so please bear with me, I might be doing things in a not-so-orthodox way. Also, I tried not to aim too high. I know the following limitations, which I'd like to lift in the future:
- hot/hovering does not seem to work correctly, though I assumed WidgetPod would handle it for me,
- the node's labels are currently only "Label" widgets. It shall not be too hard to make it a generic type instead,
- animating the expand/collapse actions would be nice,
- the tree currently cannot be changed (nodes removed/added). That's maybe a harder task!
- very big trees: see below...

I would like this widget to handle possibly very big trees. The idea would be to only create the nodes currently visible on the screen, depending on the viewport position. Looking around, I saw that ~~two~~ three approaches exist for that:
- @rjwittams has [changes in his fork](https://github.com/rjwittams/druid/commit/80e8ec259fd51b6a0c350cb157bd8a413febbe3a) which allow for his table implementation to get events whenever the viewport changes (scroll events)
- the official repository has a ScrollComponent / Viewport mecanism which may allow me to access the viewport rectangle. 
- Jan's proposal #1365 Virtual list, which I understand also sends an event VIEWPORT_SIZE_CHANGED  whenever the viewport size changes

In any case discussions around how scrolling has to be handled might not have completely settled down yet, so I'll probably leave this part for later.

### DIFF
--- a/druid/examples/tree.rs
+++ b/druid/examples/tree.rs
@@ -17,123 +17,140 @@ use std::vec::Vec;
 
 //use druid::im::{vector, Vector};
 //use druid::lens::{self, LensExt};
-use druid::widget::{LabelText, Tree, TreeNode, Scroll};
-use druid::{
-    AppLauncher, Data, Lens, LocalizedString, Widget, WindowDesc,
-};
+use druid::widget::{LabelText, Scroll, Tree, TreeNode};
+use druid::{AppLauncher, Data, Lens, LocalizedString, Widget, WindowDesc};
 
 #[derive(Clone, Lens)]
 struct Taxonomy {
-	name: &'static str,
-	children: Vec<Taxonomy>,
+    name: &'static str,
+    children: Vec<Taxonomy>,
 }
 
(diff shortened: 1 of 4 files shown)


cmyr:
@tirix We're starting a new experiment for widget contributions, which is the [druid-widget-nursery](https://github.com/linebender/druid-widget-nursery) repo; the idea is to have a lower barrier for merging of new widgets and related experiments. Would you like to open this as a PR over there?

derekdreery:
@tirix you'd be adding the first widget, so don't worry about web - just put the example in. I'll then have a go at getting CI set up.

tirix:
I've added this as PR [#2](https://github.com/linebender/druid-widget-nursery/pull/2) in druid-widget-nursery, so I'm closing this PR here.

PR #1468, opened 2020-12-16
Share ClickCounter type between web & gtk

cmyr:
This is a pretty marginal win, but does remove some duplicate logic.


jneem (approved):
Looks good to me. Also, I found the earlier discussion (#859), which suggests that this might be useful for other platforms too.

cmyr:
Oooh, had completely forgotten about that. This implementation doesn't fully follow the design detailed there, but it could definitely be the basis of a unified design at some point in the future.

Issue #1449, opened 2020-12-08
Widget gallery example uses quite some CPU time due to spinner on Mac


kud1ing:
Weht the spinner it uses about 70% (and 25% WindowServer).
If one removes the spinner it drops to 1% CPU.

@cmyr hints:

> Colin Rofls: harder to guess at. My first step would just be to look at the animation timer durations
> 
> Colin Rofls: but then I might also play around with some other things that animate, like the switch, and see if it also shows similar performance
> 
> Colin Rofls: or even just the anim demo

kud1ing:
<img width="1385" alt="Bildschirmfoto 2020-12-08 um 20 28 30" src="https://user-images.githubusercontent.com/391975/101531914-1d08ef80-3994-11eb-8f98-a0c91e681978.png">

kud1ing:
> Joe Neeman 8:33 PM   
> yeah, the spinner is causing a repaint every frame, and since incremental painting isn't supported on mac yet, it's repainting everything on every frame

kud1ing:
@jpochyla seems to be working on this:
https://github.com/jpochyla/druid/commit/68d53a7c02e47644d7792dffa920a6b30999efb8

kud1ing:
Maybe also see https://github.com/linebender/druid/issues/1083

derekdreery:
Slightly OT, but glad that the widget gallery is exercising druid and surfacing issues! 😄

PR #1470, opened 2020-12-18
Experiment with partial paint for Mac windows

jpochyla:
Proof of concept, tries to avoid full repaint after every invalidation. Because `dirtyRect` in the `draw_rect` callback is (in my experiments) always the full window size, we keep the damage rects in `ViewState` and ignore the suggested rect while painting. Animation frames are now processed by `CVDisplayLink`, which is also kept in the state, and started from `request_anim_frame` and stopped in the `display_link_output` callback before requesting `viewWillDraw`.

Partial invalidation of usual painting seems to be working, but there are problems with invalidation in the animations. I feel this might need somebody actually knowledgable of Mac UI work.

https://user-images.githubusercontent.com/1055497/102645619-05cfbc00-4163-11eb-8a30-280998ece3ff.mov


druid-shell/src/platform/mac/window.rs
```diff
@@ -866,31 +884,55 @@ impl WindowHandle {
...
-            let () = msg_send![*self.nsview.load(), setNeedsDisplay: YES];
+            if let Some(view) = (*view).as_ref() {
+                let view_state: *mut c_void = *view.get_ivar("viewState");
+                (*(view_state as *mut ViewState))
+                    .invalid_region
+                    .lock()
+                    .unwrap()
+                    .add_rect(rect);
+            }
+            let () = msg_send![*view, setNeedsDisplayInRect: bounds];
```
jneem, line 915: Can't you just do `setNeedsDisplay` here instead of calculating the bounds yourself?

druid-shell/src/platform/mac/window.rs
```diff
@@ -725,23 +735,21 @@ extern "C" fn view_will_draw(this: &mut Object, _: Sel) {
...
+
         let context: id = msg_send![class![NSGraphicsContext], currentContext];
         //FIXME: when core_graphics is at 0.20, we should be able to use
         //core_graphics::sys::CGContextRef as our pointer type.
         let cgcontext_ptr: *mut <CGContextRef as ForeignTypeRef>::CType =
             msg_send![context, CGContext];
         let cgcontext_ref = CGContextRef::from_ptr_mut(cgcontext_ptr);
 
-        // FIXME: use the actual invalid region instead of just this bounding box.
-        // https://developer.apple.com/documentation/appkit/nsview/1483772-getrectsbeingdrawn?language=objc
```
jneem, line 736: Did you try using `getRectsBeingDrawn` instead of `dirtyRect`?

jpochyla: Yes, `getRectsBeingDrawn` were returning one rect in my tests, same as `dirtyRect`.

jneem: Hooray for accurate API docs :shrug:

druid-shell/src/platform/mac/window.rs
jpochyla: Yes that would have the same effect, but the rect would still need to get added to `invalid_region`, so `draw_rect` would repaint the whole screen.

jpochyla:
Update: after marking the view as `opaque`, the excessive invalidation disappears. Although I'm glad it works, and (reading the docs) `opaque` doesn't have any obvious downsides, I still don't really understand what's going on. This PR would need (at the very least) a lot of documentation, but it dramatically improves performance of the widget gallery example (#1449).

derekdreery:
I am in no way qualified to comment on this work, but thanks for doing it!

jneem:
Out of curiosity, does `opaque` have any effect on `getRectsBeingDrawn`?

jpochyla:
I've squashed the commits, added some docs and modified how the animation frames are sync'd, marking as ready for review.

> Out of curiosity, does opaque have any effect on getRectsBeingDrawn?

It doesn't, (un)fortunately :)

jpochyla:
I've been using this branch for a few weeks, and it works well. Only problem seem to be occasional invalidation artefacts around the edges of widgets, which I suspect is some rounding problem. I'll try to hit it with some `expand` / `trunc` combinations and report back. I think it's promising!

rjwittams:
I was experimenting a while back with metal and wgpu. (I added raw-window-handle). One thing that blocked it was that the normal paint events didn't work - invalidation etc? I think with the CAMetalLayer. I hacked it in some weird way or other (drawing on a timer maybe). I was going to look at CVDisplayLink so maybe this would help with that kind of situation.

jpochyla:
I think that CVDisplayLink is working fine now, and I correctly round the rects, but I've noticed that with multiple windows it goes haywire and invalidates too much.

jpochyla:
I've rebased this on top of master, and removed the `opaque` window setting, as it seems it's not needed anymore (after the latest invalidation changes on Mac).

jpochyla:
Approach from this PR does not work on Ventura, and Xilem/Vello is hopefully around the corner, so I think we can close this.

PR #1474, opened 2020-12-20
Use same codepath for dialog response as other events

cmyr:
These were previously calling dispatch_cmd directly, which means
commands submitted while handling these events would not be
dispatched, and also meant that update wasn't sent.

cc @jneem, I missed this in review but just hit the missing update call in the spline toy. :)


jneem (approved):
Ah, oops :)

PR #1475, opened 2020-12-20
Account for viewport_offset in children's LayoutCtx::mouse_pos

jpochyla:
`viewport_offset` needs to be included into the mouse position also in `WidgetPod::layout`. Fixes glitches, i.e. when some widgets inside `Scroll` are hot, with mouse hovering, and then an external layout pass switches them to cold because `WidgetPod::set_origin` thinks they have moved.

### DIFF
--- a/druid/src/core.rs
+++ b/druid/src/core.rs
@@ -537,7 +537,7 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
         self.state.is_expecting_set_origin_call = true;
 
         let child_mouse_pos = match ctx.mouse_pos {
-            Some(pos) => Some(pos - self.layout_rect().origin().to_vec2()),
+            Some(pos) => Some(pos - self.layout_rect().origin().to_vec2() + self.viewport_offset()),
             None => None,
         };
         let prev_size = self.state.size;


cmyr:
@jpochyla This makes sense, but I'm curious if you have a good way to reproduce the original problem so I can properly test?

jpochyla:
> @jpochyla This makes sense, but I'm curious if you have a good way to reproduce the original problem so I can properly test?

Sure, sorry about the delay! Here it is:

https://gist.github.com/jpochyla/7fecd1d1cbc99db146912dcce1a3124f

I run it from `druid/examples/scroll_offset.rs`. Attached video shows the situation without the fix applied:


https://user-images.githubusercontent.com/1055497/103438025-919b1b00-4c2e-11eb-8c81-62c4c905fa0b.mov

cmyr:
Great, thanks.

Interestingly I do still see an issue when scrolling with the mouse wheel, but I think this is a separate bug?
https://user-images.githubusercontent.com/3330916/103443417-15ec9e80-4c2d-11eb-9034-0506ab9e296f.mov

cmyr (approved):

jpochyla:
> Interestingly I do still see an issue when scrolling with the mouse wheel, but I think this is a separate bug?

I think this is a separate problem, yes. Unfortunately I don't have a mouse around to test :|

PR #1477, opened 2020-12-20
Add serde feature, and enable `im/serde` when active.

### LINKED ISSUE #1476: Implement serde::deserialize for Vector
This should be behind a feature and would improve ergonomics when ingesting data from any serialized data directly to `druid::Data`. This should be pretty straight forward as `Vector` already implements `FromIterator` and `FromVec`.

derekdreery:
This means that a user can (de)serialize im data structures using serde.

Closes #1476.

### DIFF
--- a/druid/Cargo.toml
+++ b/druid/Cargo.toml
@@ -26,6 +26,7 @@ image = ["druid-shell/image"]
 svg = ["usvg", "harfbuzz-sys"]
 x11 = ["druid-shell/x11"]
 crochet = []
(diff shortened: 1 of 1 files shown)


derekdreery:
Closes #1476 .

derekdreery:
TODO if other crates e.g. kurbo has serde features.

cmyr (approved):
Looks good, we can add the piet support on the next piet release.

PR #1479, opened 2020-12-21
Fix builtin command name collision

cmyr:
We were using 'menu-save-file-as' for two different commands.
This removes the 'menu' prefix from save-file and save-file-as,
since these are not actually sent by the menu.


derekdreery (approved):

derekdreery:
Would it be worth swapping to that naming convention for the other selector names?

cmyr:
I think in general we have, doing a *quick* skim all the ones I see with the `menu` prefix are things sent from menus?

PR #1481, opened 2020-12-21
Retain `Send` and `Sync` from type-erasing `image::Error`.

derekdreery:
This small patch makes handling image errors more flexible.

### DIFF
--- a/druid-shell/src/image.rs
+++ b/druid-shell/src/image.rs
@@ -196,7 +196,7 @@ impl ImageBuf {
     /// Attempt to load an image from raw bytes.
     ///
     /// If the image crate can't decode an image from the data an error will be returned.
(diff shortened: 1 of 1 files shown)


raphlinus (approved):
I'm not sure I understand *all* the implications, but seems pretty reasonable.

derekdreery:
So `image::Error` *is* `Send` and `Sync`, this change retains that information when the type is erased (as it's converted into a trait object).

derekdreery:
I remember some discussion about this in the [`failure`] years, where the fact that people were returning boxed errors that were not `Send` or `Sync` was a problem, because it limited what client libraries could do. I had a look for the conversation, but I fear it's lost to the great internet.

[`failure`]: https://github.com/rust-lang-nursery/failure

PR #1485, opened 2020-12-24
adding added controller widget

### LINKED ISSUE #1480: [feature request]: Create a method on `WidgetExt` that allows the user to run some code when the widget is added.
Follows on from https://github.com/linebender/druid/issues/1478.

Sometimes it's necessary to do something when a widget is created. One example would be to set the cursor when the mouse is hovered over the widget. Currently this would require creating a custom controller. This issue proposes the following function:

```rust
pub trait WidgetExt<T: Data>: Widget<T> + Sized + 'static {
  // .... after on_click
  fn on_added(self, f: impl Fn(&mut LifeCycleCtx, &T, &Env) + 'static) -> ControllerHost<Self, Added<T>>;
  // ...
}
```

(I might have got the function signature wrong). The function would require a new type be created (`Added<T>`) and `Controller` implemented for it. The controller would react to the `LifeCycle::WidgetAdded` lifecycle event by running the closure.

**EDIT** I did get the function signature wrong. Corrected now.

arthmis:
I had an issue with the documentation display. I followed the `Click` widget as an example, but I don't think the documentation is showing up as expected. Also there seems to be extra `'` in the `Click` widget documentation. Is there anything I missed in the implementation?

**EDIT** (Richard Dodd) Closes #1480.


druid/src/widget/added.rs
```diff
@@ -0,0 +1,61 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! A [`Controller`] widget that responds to [`LifeCycle::WidgetAdded`] event.
+//!
+//! [`Controller`]: struct.Controller.html
```
jneem, line 17: Now that intra-doc-links are stable, you should be able to just do
```
//! [`Controller`]: crate::widget::Controller
```
and you'll also benefit from broken link detection.

druid/src/widget/added.rs
```diff
@@ -0,0 +1,61 @@
...
+//! [`LifeCycle::WidgetAdded`]: enum.LifeCycle.html#variant.WidgetAdded
```
jneem, line 18:
```
//! [`LifeCycle::WidgetAdded`]: crate::LifeCycle::WidgetAdded
```

druid/src/widget/added.rs
```diff
@@ -0,0 +1,61 @@
...
+/// [`WidgetExt`]: ../trait.WidgetExt.html
+/// [`LifeCycle::WidgetAdded`]: enum.LifeCycle.html#variant.WidgetAdded
+pub struct Added<T> {
+    /// A closure that will be invoked when the child widget is added
+    /// to the widget tree
+    action: Box<dyn Fn(&mut LifeCycleCtx, &T, &Env)>,
+}
+
+impl<T: Data> Added<T> {
+    /// Create a new [`Controller`] widget to respond to widget added to tree event.
```
jneem, line 39: Splitting hairs, maybe, but a `Controller` isn't a widget (the `ControllerHost` is).

jneem (commented):
Looks good! I had some minor nitpicks with the docs.

druid/src/widget/added.rs
arthmis: Ok I'll get onto this then.

druid/src/widget/added.rs
arthmis: I think I was following the `Click` widget example. I don't remember if that's a full widget or a Controller type. So most of the docs I've written is paraphrasing `Click`'s docs. I was trying to be consistent with that.

arthmis: I don't know what would be preferred here. I'm leaving it as is. I think calling it a Controller widget is appropriate. Anything else might be too verbose. I guess you can just call it controller like, "Create a new controller to respond to widget added to tree event".

jneem:
Looks good to me! Unfortunately, clippy broke the build again and we can't merge until its fixed. I'll try to get to it later today if no one beats me to it.

derekdreery:
I've fixed up clippy for the rust 1.49 in https://github.com/linebender/druid/pull/1461. If you rebase on that after it's merged, the clippy warning should go away. :)

arthmis:
Hi, I see the label waiting on author. Is there something I'm supposed to do?

jneem:
You'll need to rebase your branch onto the master branch, to pull in the fixes for the broken CI.

jneem (approved):

arthmis:
Uhh did I do that correctly?

derekdreery:
Rebasing is fiddly! I will check out the PR and fix it up when I get chance.

cmyr:
it looks like something went a bit wrong and you brought in a bunch of old commits. I would just do `git fetch origin` and `git merge origin/master` and then we can clean it up here, when we merge. :)

arthmis:
Am i doing this from the current state?

druid/src/widget/widget_ext.rs
```diff
@@ -159,6 +161,19 @@ pub trait WidgetExt<T: Data>: Widget<T> + Sized + 'static {
         ControllerHost::new(self, controller)
     }
 
+    /// Control the [`LifeCycle::WidgetAdded`] event of this widget with an [`Added`] widget.
+    /// The function provided will be called when the widget is added to the
+    /// widget tree.
+    ///
+    /// [`Added`]: crate::widget::Added
+    /// [`LifeCycle::WidgetAdded`]: crate::LifeCycle::WidgetAdded
```
cmyr, line 169: I'm going to nit out here; I think our goal with the docs is to describe what this does and how to use it, and the particular details of the implementation are not important; in particular the user really never needs to know about the `Added` widget. I would go with,

```suggestion
    /// Provide a closure that will be called when this widget is added to the widget tree.
    ///
    /// You can use this to perform any initial setup.
    ///
    /// This is equivalent to handling the [`LifeCycle::WidgetAdded`] event in a
    /// custom [`Controller`].
    ///
    /// [`LifeCycle::WidgetAdded`]: crate::LifeCycle::WidgetAdded
```

druid/src/widget/added.rs
```diff
@@ -0,0 +1,61 @@
...
+// limitations under the License.
+
+//! A [`Controller`] widget that responds to [`LifeCycle::WidgetAdded`] event.
+//!
+//! [`Controller`]: crate::widget::Controller
+//! [`LifeCycle::WidgetAdded`]: crate::LifeCycle::WidgetAdded
+
+use crate::widget::Controller;
+use crate::{Data, Env, LifeCycleCtx, Widget};
+/// This [`Controller`] widget responds to [`LifeCycle::WidgetAdded`] event
```
cmyr, line 22:
```suggestion

/// This [`Controller`] widget responds to [`LifeCycle::WidgetAdded`] event
```

druid/src/widget/added.rs
```diff
@@ -0,0 +1,61 @@
...
+/// via [`WidgetExt`].
+///
+/// [`Controller`]: crate::widget::Controller
+/// [`ControllerHost`]: crate::widget::ControllerHost
+/// [`WidgetExt`]: crate::widget::WidgetExt
+/// [`LifeCycle::WidgetAdded`]: crate::LifeCycle::WidgetAdded
+pub struct Added<T> {
+    /// A closure that will be invoked when the child widget is added
+    /// to the widget tree
+    action: Box<dyn Fn(&mut LifeCycleCtx, &T, &Env)>,
```
cmyr, line 35: One thing I wonder: should this closure also be given access to the actual *widget*? Can we imagine, for instance, wanting to call some method on the widget from this closure?

cmyr (commented):
Just skimming this again: I have a few little doc tweaks, but I also have one larger question, which is whether or not we should also be passing in the widget itself to the closure, as like `&mut W`? I think this probably makes sense.

druid/src/widget/added.rs
arthmis: At the very least it brings it more in line with the `lifecycle` function in `Controller`. It would eliminate the need for the user to still implement `Controller` if they wanted to do something with the child when its added to the widget tree.

cmyr: yea, that's my thinking, I think we should just have the same signature as the method in `Controller`.

druid/src/widget/widget_ext.rs
```diff
@@ -159,6 +161,21 @@ pub trait WidgetExt<T: Data>: Widget<T> + Sized + 'static {
...
+    /// This is equivalent to handling the [`LifeCycle::WidgetAdded`] event in a
+    /// custom [`Controller`].
+    ///
+    /// [`LifeCycle::WidgetAdded`]: crate::LifeCycle::WidgetAdded
+    fn on_added<W: Widget<T>>(
+        self,
+        f: impl Fn(&mut W, &mut LifeCycleCtx, &T, &Env) + 'static,
+    ) -> ControllerHost<Self, Added<T, W>> {
+        ControllerHost::new(self, Added::new(f))
+    }
```
cmyr, line 177: This is a bit of a gotcha, but I think instead of `W` we want to use `Self`, which refers to the type of whichever widget we're using this method on, like:

```suggestion
    fn on_added(
        self,
        f: impl Fn(&mut Self, &mut LifeCycleCtx, &T, &Env) + 'static,
    ) -> ControllerHost<Self, Added<T, Self>> {
        ControllerHost::new(self, Added::new(f))
    }
```

I think that's the only remaining fixup, and then we can get this merged. Thanks!

derekdreery:
@cmyr are we OK to merge now? @arthmis thanks very much for going through the motions with this PR 😄

cmyr:
Yep squash sounds good :)

PR #1493, opened 2020-12-31
Iced-style StyleSheet concept

futurepaul:
This is a rough draft of doing stylesheets in a way (heavily) [inspired by Iced](https://github.com/hecrj/iced/blob/master/style/src/button.rs), while still using Env and KeyOrValue for what they're good for.

Ideally we'd be able to trim `theme.rs` down to a much more minimal set of values, and impl default stylesheets for each widget that use those values as appropriate for those use case. Users can then provide their own stylesheets as overrides, totally ignoring Env if they'd like to.

Any and all comments are welcome!


druid/src/widget/button.rs
```diff
@@ -14,19 +14,98 @@
...
+            background_color_a: theme::BUTTON_LIGHT.into(),
+            background_color_b: theme::BUTTON_DARK.into(),
+        }
+    }
+}
+
+/// The Button widget holds a StyleSheet which can resolve to multiple Styles
+pub trait StyleSheet {
+    /// How a button normally looks
+    fn normal(&self) -> Style;
```
cmyr, line 60: maybe we just have a `: Default` bound on the trait, and you use that for your 'normal' style? I admit it's a bit weird..

druid/src/widget/button.rs
```diff
@@ -14,19 +14,98 @@
...
+        Style::default()
+    }
+}
+
+impl std::default::Default for Box<dyn StyleSheet> {
+    fn default() -> Self {
+        Box::new(Default)
+    }
+}
+
```
cmyr, line 94: Okay I'm actually legit confused about all the default stuff here. Could we just not impl `Default`? what would that cost us?

cmyr (commented):
Overall I'd say this looks good, it feels like a nice simplification and standardization of what attributes can be styled for a given widget.

Galaxy brain moment: do we even need a `StyleSheet` trait? couldn't the actual details of a given widget's style be entirely internal to that widget? In this world some widgets could have different versions when hot/active, and some just wouldn't, and that would be totally okay?

futurepaul:
Before I was sort of blindly following Iced's way of doing things.

Doing this without a trait is:

- More straightforward to reason about
- Doesn't need any boxing
- Just as expressive as far as I can tell?

I don't think it's much more verbose from the user point of view. You do have to explicitly override all three button "states" for them to look different, there's no "cascade" of just overriding the normal button and getting hot / active ones for free (the StyleSheet had default impls).

It's probably possible to do some amount of "cascading" in this simpler struct-based setup, but also possibly could cause more mental overhead than it's worth. As it is right now it's very straightforward: if you want a state to be a certain style, give it that style.

druid/src/widget/button.rs
```diff
@@ -14,19 +14,81 @@
...
-use crate::widget::prelude::*;
 use crate::widget::{Click, ControllerHost, Label, LabelText};
-use crate::{theme, Affine, Data, Insets, LinearGradient, UnitPoint};
+use crate::{theme, Affine, Color, Data, Insets, LinearGradient, UnitPoint};
+use crate::{widget::prelude::*, KeyOrValue};
 
 // the minimum padding added to a button.
 // NOTE: these values are chosen to match the existing look of TextBox; these
 // should be reevaluated at some point.
 const LABEL_INSETS: Insets = Insets::uniform_xy(8., 2.);
```
futurepaul, line 24: Are we at the right point to reevaluate this yet? I've had some trouble getting a button / textbox / label to all have the same style. Would be nice if someone could use the same input values for each one and get the same results.

jpochyla: Insets in the `TextBox` have been moved to the env, so I'd consider this comment outdated, and would resolve the insets for `Button` from the env as well.

futurepaul:
I'm pretty happy with this as a way to do stylesheets... but it doesn't feel like the "win" here is large enough to justify the churn through the code base. This solves the discoverability problem for style values... but leaves a lot of other stuff unsolved (dynamic, state-based styling), and I worry it would be incompatible with solutions for the other stuff?

arthmis:
I think instead of defining the button like this
```rust
pub struct Button<T> {
    label: Label<T>,
    label_size: Size,
    style_normal: ButtonStyle,
    style_hot: ButtonStyle,
    style_active: ButtonStyle,
}
```
the button could have a single field, call it style for now, and then you could implement on the `Style` struct, the hot, active, and so on methods. 
```rust
pub struct Button<T> {
    label: Label<T>,
    label_size: Size,
    style: Style,
}

impl Style {
  fn hover() -> Style;
}
```
This is basically copied from how I see Iced doing it. I don't understand why they were using traits for each widget style.

Would state-based styling encompass changing themes? I feel like Env should be used for the cascading effect. There will probably be a need to define the set of styles, padding, and so on should be part of the default env. 

I actually would like to go through the churn of adding `Style` structs for the appropriate widgets. I liked the way Iced is doing it, minus the traits for each style. And it shouldn't break anything as far as I'm aware.

futurepaul:
So then the `Style` struct would have values for each possible state?
```
Style {
  border_width,
  hot_border_width,
  active_border_width,
  border_width,
  hot_border_width,
  active_border_width,
  background_is_gradient
  hot_background_is_gradient
  active_background_is_gradient
  ...
}
```

and the impl would just pull those colors as appropriate?

Just making sure this is as expressive. It's nice that from the user side you only need to impl one struct, although since there's no "inheritance" it would be something like this? Let me know if I'm misunderstanding.

`let my_style = button::Style { ..Default::default(), border_width: 0, hot_border_width: 0, active_border_width: 0 };`

arthmis:
Let me work up a proper example so it's easier to see what I'm saying. Don't want to have any misunderstandings!

PR #1498, opened 2021-01-02
Wayland

derekdreery:
A wayland backend for `druid-shell`.

Very much **not** ready for inclusion (might never be), but I'm opening the PR early so others can experiment if they like.

Currently some of the size calculations are broken, and there's no use input handling at all. But wayland is fun to hack on (trust me) so feel free to have a play. :smile: 

**Update**

You should be able to run the `empty_window` example without it crashing! It draws a gradient so we can check that the memory is all aligned nicely. I'm just beginning to work on input, currently incorporating `xkb` so I can translate scancodes to the nice things people expect. But its 11pm here so maybe tomorrow. I've pushed the current state.

No region invalidation yet so the invalidate example might trigger epilepsy.

**Update2**

I've got to start work next week, so I've done all I can for now. You can run druid examples directly on wayland and it works! (sort of). Definitely not ready for the prime time, but I'm going to try to get the branch ready for merging, so work can continue in-tree, when I (or someone else) next gets chance.

**Update3**

I've done some more work. I've refactored out the buffer management code, in an attempt to make things more modular. I've also moved to individual fields having interior mutability rather than the whole struct. A lot of fields can be `Cell` rather than `RefCell`, potentially saving a few nanoseconds. :) I've also moved to holding weak references most places, which matches how other backends work. Most druid examples will run now, although many will exhibit incorrect behavior.


derekdreery:
I need to decide how to solve the `impl Default for WindowHandle` (or not) problem.

derekdreery:
Right I'll try and get this in a state for merging. I wont have implemented everything: my target is to get some CI running on wayland and to make sure everything else builds.

cmyr:
Ooops I totally snoozed on this. I'm back in an environment where I can at least theoretically test, will add this to my queue!

derekdreery:
No worries I'm also busy 😄

derekdreery:
I will have another pass at tidying this soon.

derekdreery:
I've been thinking about this, and it might be worth using the smithay client toolkit. I think the authors mean it to be a zero-cost or low-cost abstraction to make working with wayland easier. I might investigate what an implementation would look like in it.

hyarsan:
Should druid render its own "native-feeling" window decorations? Wayland pretty much encourages client-side decoration, and I think it could make druid apps more appealing.

derekdreery:
> 
> 
> Should druid render its own "native-feeling" window decorations? Wayland pretty much encourages client-side decoration, and I think it could make druid apps more appealing.

This is an option. In my wm (sway) it would look a bit odd if all windows opted out of compositor decoration, but it could be useful for some windows. Part of the druid-shell API could expose an option not to decorate windows, but we'd have to think if all platforms supported it. This is (IMO) related to transparent windows, which includes non-rectangular windows because these use a larger rectangle and transparency to achieve the effect IIRC.

derekdreery:
I'm thinking that I will have to do a major refactor here. I've read the windows backend a bit more, and to match the semantics I need to do the following

 - `WindowHandle` is essentially a weak reference counted pointer to the actual data. That way, window handles won't keep a window alive after it has been closed. Methods will do nothing or return an error when the window doesn't exist. This also allows the window handle to exist before its corresponding window, which can be useful for ergonomics.
 - This PR takes a bit of a hodge-podge approach to re-entrancy and general sharing of data. The windows backend approach is essentailly `Rc<WindowData<Cell<fields>>>` so things that need access to the window data hold an `Rc<WindowData>` (`WindowHandle` holds a `Weak<WindowData>`). You can manipulate the fields of `WindowData` individually while holding a shared reference (they all use interior mutability. This would make everything much neater.
 - (Related to the last point) I also like the idea of having an enum of possible deferred jobs. That way you avoid re-entrancy by running any tasks that requires calling user functions after the current function has returned. You have a queue of jobs that need to run. Again this would simplify things.
 - As part of the cleaning up it makes sense to build abstractions to everything we can (e.g. pointer events) to keep the run loop code as clean as possible, because it is always going to be very complex.

derekdreery:
I'm also going to squash because otherwise rebasing is a big pain.

druid-shell/src/error.rs
```diff
@@ -24,6 +24,8 @@ use crate::platform::error as platform;
 pub enum Error {
     /// The Application instance has already been created.
     ApplicationAlreadyExists,
+    /// Tried to use the application after it had been dropped.
+    ApplicationDropped,
```
derekdreery, line 28: The behavior ok the wayland backend is to only hold weak handles to the application in the windows, so that the application might be dropped and then the window used. Not sure if this is correct.

derekdreery:
Done some more work and added another update.

james-lawrence:
I have a rebased and some improvements to this branch. how do we want to deal with potentially multiple branches related to wayland? ideally I'd like to see a wayland branch in druid we could merge into.

Maan2003:
I also have [some improvements and rebased it](https://github.com/Maan2003/druid/tree/wayland). I thought I have told here earlier but apparently didn't :/ . Wayland branch here will definitely help in less duplication of work

derekdreery:
Yes maybe the solution is to have a wayland branch in the repo that we can coordinate work on.

Maan2003:
Ok I will create wayland branch from [my wayland-just-rebase](https://github.com/Maan2003/druid/commits/wayland-just-rebase). Let me know if anyone has problems.

james-lawrence:
atm I have most of the rendering functionality working, and have layershell work started. (it renders just slightly improperly)

derekdreery:
@james-lawrence @Maan2003 sounds like there's some awesome work on this. I'm glad someone has picked it up - I've been meaning to but ...life.

I'm conscious that there might be duplicated effort, so I think the quicker we can get a branch in `druid` to share work, the better. I'm not sure which branch we should use as the base (to then merge any improvements from the other branch). Maybe if you could both look at each other's code to see who has made the most progress? If there is some way I can help with this let me know.

Does this sound OK as a way forward?

james-lawrence:
absolutely, will leave it up to you two. which branch we take here [mine](https://github.com/james-lawrence/druid/tree/wayland).

I did some major internal refactoring to support layershells cleanly. its meeting my needs atm for my applications. cursors and keyboards are the last two pieces I'm missing for it to be solid minus performance work to reduce idle cpu usage.

Maan2003:
I proposed to start from clean rebase of this pr and then create PRs based on our changes. because there are many changes in both branches I don't agree with[1]. 


[1] examples: all text input related changes, which should use the x11 backend's xkb, using Arc everywhere, my hacky attempts to get working request_anim_frame

derekdreery:
Yeah I think @Maan2003's idea is good. I'll try to contribute as well

One question I have is: should there be another layer between Wayland and `druid-shell`, something like [smithay-client-toolkit](https://crates.io/crates/smithay-client-toolkit)? I genuinely don't know the answer to this question, I chose the generated Wayland bindings originally because it would be easier to use new APIs as they get released/stabilized, since you just re-generate the code. But maybe it would be less error prone to use the intermediate layer? I don't know. On balance, I think it would be better to carry on using Wayland directly, but this is a weak opinion and I'm happy to go with `smithay-client-toolkit` if that's the consensus.

One thing I would like to do is to be much more systematic with how calls into `druid-shell` are handled. With this PR I started by trying to handle them directly, then realising that to avoid re-entrancy crashes you need to defer doing what the user requested until their callback has finished running (this is what all the backends do, I now realise). What I'd like to do is basically have an `enum` with one variant per method, and a queue. The methods push the corresponding variant to the queue, and after a user callback the next item is popped off the front of the queue and handled. This is repeated until the queue is empty, then you go into idle mode until the next event comes in.

This is basically how it works already, but I want to formalize it and tidy it up.

Maan2003:
Ok, [wayland branch](https://github.com/linebender/druid/tree/wayland) is setup. I hope to see some PRs

james-lawrence:
going to pass at this point, I've already expanded a fair amount of effort to make druid usuable for my use cases, additional unspecified refactoring work to get it back to usable from the current state doesn't seem like a good use of my time.

if the druid branch gets to feature parity with mine in the future i'll submit patches to improve it but current interests lie in actually building the stuff I want not spinning cycles to redo work what I've already done.

my branch is free for you to pull from for implementation details or if you have questions about how something is working I'll happily answer.

james-lawrence:
oh feel free to review my branch, and let me know which parts you'd be willing to accept and I'd probably be willing to generate individual patches for them depending how much effort it'd take to generate the patch.

derekdreery:
> 
> 
> oh feel free to review my branch, and let me know which parts you'd be willing to accept and I'd probably be willing to generate individual patches for them depending how much effort it'd take to generate the patch.
> 
> as it stands right now, my branch fully works for my use cases, minus some performance stuff due to animations, and mouse scrolling being slow.

No worries thanks for your contribution so far. Thanks for letting us know about your commitment, it means we can work on incorporating your patch into the branch without worrying about duplication. If you do decide to work on something, if you could put a comment on somewhere to make sure that someone else doesn't do the same thing.

Thanks again :)

james-lawrence:
absolutely, right now I think only thing left is getting popups to resize dynamically based on the context of the widget. but otherwise think my branch has stabilized. once I get my project working to my satisfaction I may revisit here and start offering patches. I made a bunch of changes I know won't be accepted just to make development easier. basically everything under the wayland backend should be mostly acceptable outside basic stylistic stuff.

Maan2003:
Superseded by #2079

PR #1500, opened 2021-01-03
Add MouseButtons.count()

jjl:


derekdreery (approved):
Looks good.

jjl:
Good spot, sorted.

derekdreery:
Ack, now clippy is complaining because the object is `Copy`. Would you change it from `&self` to `self`, then clippy will be happy.

jjl:
Done. Hazards of thinking your chamge is too small to upset clippy...

PR #1503, opened 2021-01-04
Fix nightly clippy lints

cmyr:
This includes some exciting changes to some atomics code, which always makes me want to go for a long walk. 🤕

### DIFF
--- a/druid-shell/src/application.rs
+++ b/druid-shell/src/application.rs
@@ -75,9 +75,9 @@ impl Application {
     ///
     /// [druid#771]: https://github.com/linebender/druid/issues/771
     pub fn new() -> Result<Application, Error> {
-        if APPLICATION_CREATED.compare_and_swap(false, true, Ordering::AcqRel) {
-            return Err(Error::ApplicationAlreadyExists);
-        }
+        APPLICATION_CREATED
+            .compare_exchange(false, true, Ordering::AcqRel, Ordering::Acquire)
+            .map_err(|_| Error::ApplicationAlreadyExists)?;
         util::claim_main_thread();
         let platform_app = platform::Application::new()?;
         let state = Rc::new(RefCell::new(State { running: false }));
--- a/druid-shell/src/platform/mac/window.rs
+++ b/druid-shell/src/platform/mac/window.rs
@@ -947,18 +947,14 @@ impl WindowHandle {
     }
 
     pub fn open_file(&mut self, options: FileDialogOptions) -> Option<FileDialogToken> {
-        self.open_save_impl(FileDialogType::Open, options)
+        Some(self.open_save_impl(FileDialogType::Open, options))
     }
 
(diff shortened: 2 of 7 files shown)


druid-shell/src/application.rs
```diff
@@ -75,9 +75,9 @@ impl Application {
     ///
     /// [druid#771]: https://github.com/linebender/druid/issues/771
     pub fn new() -> Result<Application, Error> {
-        if APPLICATION_CREATED.compare_and_swap(false, true, Ordering::AcqRel) {
-            return Err(Error::ApplicationAlreadyExists);
-        }
+        APPLICATION_CREATED
+            .compare_exchange(false, true, Ordering::SeqCst, Ordering::Acquire)
```
raphlinus, line 79: I think the success ordering should probably remain `AcqRel`; `SeqCst` is an unneeded strengthening.

druid-shell/src/util.rs
```diff
@@ -50,16 +50,16 @@ pub fn assert_main_thread() {
...
-        if old_thread_id == thread_id {
-            log::warn!("The main thread status was already claimed by the current thread.");
-        } else {
-            panic!(
-                "The main thread status has already been claimed by thread {}",
-                thread_id
-            );
-        }
+    let old_thread_id =
+        MAIN_THREAD_ID.compare_exchange(0, thread_id, Ordering::SeqCst, Ordering::Acquire);
```
raphlinus, line 54: Same.

druid-shell/src/platform/mac/window.rs
```diff
@@ -947,18 +947,14 @@ impl WindowHandle {
     }
 
     pub fn open_file(&mut self, options: FileDialogOptions) -> Option<FileDialogToken> {
-        self.open_save_impl(FileDialogType::Open, options)
+        Some(self.open_save_impl(FileDialogType::Open, options))
```
raphlinus, line 950: It looks like you ran clippy on mac. I get similar warnings on Windows in `vk_to_key` and `key_to_vk` (druid-shell/src/platform/windows/keyboard.rs), and `init` in platform/windows/application.rs. Wouldn't be surprised if there was a similar story on Linux.

raphlinus (commented):
I'm also getting a warning that we should have `From` instead of `Into` for `Option<Target>` in command.rs.

druid-shell/src/application.rs
cmyr: Sure, on your say-so. This was me getting frustrated and going nuclear.

druid-shell/src/platform/mac/window.rs
cmyr: ah good call.

cmyr: that looks to be a slightly different problem which I'd prefer to diagnose when I have a windows machine in front of me, since it also isn't showing up in ci. :/

Issue #1502, opened 2021-01-04
document steps involved in doing a release


cmyr:
This should go in `CONTRIBUTING`.

Off of the top of my head:

- update the changelog; give the current `[unreleased]` section a version, update the 'last release' note, add a new `[unreleased]` section
- update readme to point to most recent version (before publishing crate, so that the readme on docs.rs is up to date)
- bump versions of `druid`, `druid-shell` and `druid-derive` crates
- update our own dependencies to their most recent versions, where feasible
- publish to crates.io; for `druid-derive` we need to do `--no-verify` because it depends on `druid` which won't be released yet
- update docs to use new version in examples: grep for the current version. This will be in the book, but also in the readme and in various docstrings.
- update screenshots in readme: this means taking new screenshots, and pushing them directly to the `screenshots` branch, and then linking to that from the readme.
- push a new tag and do a release on github

... that's all I have off the top of my head.

PoignardAzur:
The [current CONTRIBUTING.md](https://github.com/linebender/druid/blob/4eaffd20712aee0284360cc03366fb702c4088a6/CONTRIBUTING.md#preparing-for-a-new-release) has most of this.

Currently missing:

- Publishing to crates.io (the `--no-verify` trick in particular)
- Updating version numbers.
- Updating screenshots.
- Add a tag and release on github.

PR #1504, opened 2021-01-04
Update version in various docs

cmyr:


raphlinus (approved):
Perhaps we're at the point we should script at least some of this.

cmyr:
yea, a good first step will be at least figuring out what all the changes are that need to be made, i'm starting to write that all down in #1502.

PR #1508, opened 2021-01-05
Change conditional use of svg to be on the feature svg instead of pla…

rjwittams:
…tform. This makes IntelliJ happier with the web examples.


derekdreery (approved):
Looks good

PR #1509, opened 2021-01-05
Scroll updates to support enabling particular bars and programmatical…

rjwittams:
* Enable individual scroll bars
* Allow scrolling one axis at a time programatically. 

These are the scroll changes needed by druid_table to synchronise scrolling (using druid_bindings). 
However they would also be useful for anyone else synchronising scroll areas.

### DIFF
--- a/druid/src/scroll_component.rs
+++ b/druid/src/scroll_component.rs
@@ -19,13 +19,45 @@ use std::time::Duration;
 
 use crate::kurbo::{Point, Rect, Vec2};
 use crate::theme;
-use crate::widget::Viewport;
+use crate::widget::{Axis, Viewport};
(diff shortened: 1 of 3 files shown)


rjwittams:
There are some logical followups to these changes - I think Axis should be moved as its not Flex specific, and over time attempts should be made to reduce the amount of 'axis duplicated' code.

ScrollbarsEnabled should probably be AxesEnabled or AxisSet, and also used in ClipBox.  
 
I did not want to make this PR into an epic to cover this - just trying to clear out what is absolutely required for druid_table to work with a stock druid.

jneem (approved):
I have to admit to not being 100% clear on why bindings need `pan_to_on_axis`, but it seems like a reasonable function to have anyway...

PR #1510, opened 2021-01-06
Paint window close buttons to avoid font reliance.

rjwittams:
+Fix type signature in Scope.


cmyr (approved):
Sure, makes sense to me.

PR #1513, opened 2021-01-07
Add simple markdown editor example

cmyr:
In the process of writing up a small `0.7.0` release post I noticed we didn't really have any example code that demonstrates rich text stuff at all. This adds a simple markdown editor demo that I had sitting around.

### DIFF
--- a/druid/Cargo.toml
+++ b/druid/Cargo.toml
@@ -77,6 +77,7 @@ console_log = "0.2.0"
 float-cmp = { version = "0.8.0", features = ["std"], default-features = false }
 tempfile = "3.1.0"
 piet-common = { version = "=0.3.2", features = ["png"] }
+pulldown-cmark = { version = "0.8", default-features = false }
 
 [[example]]
 name = "cursor"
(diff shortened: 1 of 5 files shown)


rjwittams:
Should it go in web examples? I think you have to do something explicit to include (lib.rs)  or exclude it (EXCEPTIONS in build.rs) or IntelliJ at least will get mildly angry

Dunno if it would be possible to have a check for this

rjwittams (approved):

alerque:
I understand the utility in having a rich text example here, but I do get a bit squeamish at the prospect of including a Markdown parser in the bargain. Markdown parsing is one of those things that *seems* simple at first blush and the ecosystem is cluttered with things that sort of work until they fall on their face. Kind of like Adobe and others insisting on writing their own shapers or people blindly assuming BiDi is easy. Not saying this can't go it, but maybe a comment highlighting the POC nature to be used for the display code not as a starting point for any real Rust based Markdown parsing would be in order.

rjwittams:
Do you mean the dev dependency on pulldown-cmark? Or something else

PR #1515, opened 2021-01-08
Made 'set_level' available from WindowDesc

Schaback:
There is no simple way of setting the WindowLevel prior to creation for a window.
This would allows tiling window managers like i3 to see the _NET_WM_WINDOW_TYPE_DIALOG flag on window creation
and mark the window as floating.
Since all other options of WindowConfig are exposed in WindowDesc, this seems like an oversight. :)

### DIFF
--- a/druid/src/app.rs
+++ b/druid/src/app.rs
@@ -428,6 +428,14 @@ impl<T: Data> WindowDesc<T> {
         self
     }
 
+    /// Sets the [`WindowLevel`] of the window
+    ///
+    /// [`WindowLevel`]: enum.WindowLevel.html
+    pub fn set_level(mut self, level: WindowLevel) -> Self {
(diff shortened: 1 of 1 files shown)


cmyr (approved):
This seems reasonable to me, thanks!

I dont' really like us using the prefix `set_` on these methods since it implies mutation and these are builder methods; but these particular methods are all over the place, so I think it's fine for now, and at some point in the future we can do a big 'api consistency' pass and fix this. 🤷

Schaback:
Thanks! Yea, same. In general i found it confusing as a noob, how the WindowConfig was supposed to be used, or rather why this layer of pass-trough with WindowDesc - hides -> WindowConfig was there. This interface could def. use some reworking when api consistency is looked at.

PR #1517, opened 2021-01-08
Add Env::try_set_raw

cmyr:
This also exposes the ValueTypeError at crate root;
both of these changes will be used for some theme
loading experiments.


Zarenor (approved):
Straightforward change exposing a raw API. LGTM.

PR #1519, opened 2021-01-09
Update to latest x11rb

psychon:
I wanted to see how bad the breakage would be and it wasn't too bad. So, here is an update to x11rb 0.8.0 "just because".


Maan2003 (approved):
Looks good to me, thanks!

raphlinus (approved):
I'll approve based on Manmeet's review, and also make sure to enable write access for that account as well, for the future.

PR #1521, opened 2021-01-09
Implement linear gradients for the svg widget.

derekdreery:
This PR expands svg support in the svg widget to include linear gradients.

### DIFF
--- a/druid/src/widget/svg.rs
+++ b/druid/src/widget/svg.rs
@@ -14,12 +14,14 @@
 
 //! An SVG widget.
 
-use std::error::Error;
-use std::str::FromStr;
-use std::sync::Arc;
+use std::{collections::HashMap, error::Error, rc::Rc, str::FromStr, sync::Arc};
 
 use crate::{
-    kurbo::BezPath, widget::common::FillStrat, widget::prelude::*, Affine, Color, Data, Rect,
+    kurbo::BezPath,
+    piet::{self, FixedLinearGradient, GradientStop},
+    widget::common::FillStrat,
+    widget::prelude::*,
+    Affine, Color, Data, Point, Rect,
 };
 
 /// A widget that renders a SVG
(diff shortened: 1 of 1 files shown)


druid/src/widget/svg.rs
```diff
@@ -192,91 +197,160 @@ impl FromStr for SvgData {
...
-    path.apply_affine(offset_matrix * transform_to_affine(p.transform));
-
-    match &p.fill {
-        Some(fill) => {
-            let brush = color_from_usvg(&fill.paint, fill.opacity);
-            ctx.fill(path.clone(), &brush);
-        }
-        None => {}
+    fn linear_gradient_def(&mut self, lg: &usvg::LinearGradient, ctx: &mut PaintCtx) {
+        // Get start and stop of gradient and transform them to image space (TODO check we need to
```
derekdreery, line 297: I've experimented and get the results I expect, so this is a "no".

druid/src/widget/svg.rs
```diff
@@ -192,91 +197,160 @@ impl FromStr for SvgData {
...
-            ctx.stroke(path, &brush, stroke.width.value());
+    fn brush_from_usvg(
+        &self,
+        paint: &usvg::Paint,
+        opacity: usvg::Opacity,
+        ctx: &mut PaintCtx,
+    ) -> Rc<piet::Brush> {
+        match paint {
+            usvg::Paint::Color(c) => {
+                // TODO I'm going to assume here that not retaining colors is OK.
```
derekdreery, line 326: Are color brushes always cheap to create?

cmyr: This is not a guarantee made by piet, but ideally I think that the caching of brushes, especially color brushes, should be something that is handled at the piet level. Gradient brushes are definitely not necessarily cheap to create.

cmyr (approved):
I'm happy when you're happy. 😅

derekdreery:
I'm going to merge this so it doesn't bitrot. There are a few comments that could be tidied up in another pass.

PR #1527, opened 2021-01-11
Add Key::raw

cmyr:
I want to use this in my theme loading stuff to check that multiple
key's don't have the same inner key value.


Finnerale (approved):
Approved.

PR #1528, opened 2021-01-11
Allow additional l10n resource in AppLauncher

edwin0cheng:
This PR added `localization_resource` in `AppLauncher`, which enabled additional resource to use in app.

### DIFF
--- a/druid/src/app.rs
+++ b/druid/src/app.rs
@@ -33,6 +33,7 @@ type EnvSetupFn<T> = dyn FnOnce(&mut Env, &T);
 pub struct AppLauncher<T> {
     windows: Vec<WindowDesc<T>>,
     env_setup: Option<Box<EnvSetupFn<T>>>,
+    l10n_resources: Option<(Vec<String>, String)>,
     delegate: Option<Box<dyn AppDelegate<T>>>,
     ext_event_host: ExtEventHost,
 }
@@ -107,6 +108,7 @@ impl<T: Data> AppLauncher<T> {
         AppLauncher {
             windows: vec![window],
             env_setup: None,
+            l10n_resources: None,
             delegate: None,
             ext_event_host: ExtEventHost::new(),
         }
@@ -146,6 +148,19 @@ impl<T: Data> AppLauncher<T> {
         self
     }
 
+    /// Use custom localization resource
+    ///
+    /// `resources` is a list of file names that contain strings. `base_dir`
+    /// is a path to a directory that includes per-locale subdirectories.
+    ///
+    /// This directory should be of the structure `base_dir/{locale}/{resource}`,
+    /// where '{locale}' is a valid BCP47 language tag, and {resource} is a `.ftl`
+    /// included in `resources`.
+    pub fn localization_resources(mut self, resources: Vec<String>, base_dir: String) -> Self {
+        self.l10n_resources = Some((resources, base_dir));
+        self
+    }
+
     /// Returns an [`ExtEventSink`] that can be moved between threads,
     /// and can be used to submit commands back to the application.
     ///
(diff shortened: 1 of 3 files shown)


edwin0cheng:
cc @cmyr

druid/src/app.rs
```diff
@@ -146,6 +148,12 @@ impl<T: Data> AppLauncher<T> {
         self
     }
 
+    /// Add additional localization resource
+    pub fn use_localization_resource(mut self, res: String) -> Self {
```
cmyr, line 152: I think this API should match the API of `L10nManager::new`; it should take a list of resources and a path explicitly, instead of us picking a hard-coded path.

I'd also include docs similar to those i n`L10nManager::new`.

druid/src/env.rs
```diff
@@ -461,7 +461,14 @@ impl Data for EnvImpl {
 
 impl Default for Env {
     fn default() -> Self {
-        let l10n = L10nManager::new(vec!["builtin.ftl".into()], "./resources/i18n/");
+        Env::with_l10n(vec![])
```
cmyr, line 464: I think we should keep this as is, since it at least means that the examples continue to work. This all obviously needs to be reworked, though...

druid/src/env.rs
```diff
@@ -461,7 +461,14 @@ impl Data for EnvImpl {
...
+    }
+}
+
+impl Env {
+    pub(crate) fn with_l10n(mut resources: Vec<String>) -> Self {
+        resources.insert(0, "builtin.ftl".into());
+        let l10n = L10nManager::new(resources, "./resources/i18n/");
```
cmyr, line 471: so here is where our existing limitations are exposed: basically this only works if all these files are in the same place, and that's not what we should expect. In the longer term we want to figure this out, and allow for both built-in and custom resources, but I think what we should do for now is just allow the user to *replace* the built-in resources with their own. This isn't perfect but at least lets the user do what they need to do; they can copy over the builtin resources as needed.

cmyr (commented):
I think this is the right general idea but I think I'd be less fancy; more details below.

druid/src/app.rs
edwin0cheng: Fixed

druid/src/env.rs
edwin0cheng: Fixed

druid/src/env.rs
edwin0cheng: Fixed.

druid/src/env.rs
```diff
@@ -467,37 +467,56 @@ impl Data for EnvImpl {
...
+
+impl Env {
+    pub(crate) fn with_i10n(resources: Vec<String>, base_dir: &str) -> Self {
+        let l10n = L10nManager::new(resources, base_dir);
 
         let inner = EnvImpl {
             l10n: Arc::new(l10n),
             map: HashMap::new(),
-            debug_colors,
+            debug_colors: DEBUG_COLOR.into(),
```
cmyr, line 519: You can write this as,

```rust
let inner = EnvImpl {
    l10n: Arc::new(l10n),
    ...Default::default()
};
```

and then the rest of the fields will be populated from the default impl.

Alternatively, maybe you have the `default` impl just call our `with_i10n` method, passing the default localization info?

edwin0cheng:
And I refactor out the DEBUG_COLORS in `Env::default`, I could roll it back if you want.

cmyr (commented):
Cool I think this looks right, one little comment but overall I'm happy.

Have you verified that this patch lets you do what you need to do?

druid/src/env.rs
edwin0cheng: `..Default::default()` will allocate `1l0n` manager twice so I didn't use it.

> Alternatively, maybe you have the default impl just call our with_i10n method, passing the default localization info?

I was thiniking you didn't want me to change the `Env::default` at all (https://github.com/linebender/druid/pull/1528#discussion_r555361473) , so I am happy to do that if you want :)

edwin0cheng: Fixed

edwin0cheng:
> Have you verified that this patch lets you do what you need to do?

Yes ! 

Screenshot from 2021-01-12 06-51-37

cmyr (approved):
Looks good, thanks!

PR #1532, opened 2021-01-13
Subwindow updates

rjwittams:
Expose WindowLevel, add get_content_insets, track widgets window relative coordinates.
Add WindowSizePolicy to allow windows to size themselves to their content.
Make scroll contained things re layout if the scroll offset changes. 

All of these are to let subwindows like dropdowns be laid out relative to their parent widget.

The drop down widget itself isn't here. It is currently in nursery locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ You can find its changes [documented below](#070---2021-01-01).
 - RichTextBuilder ([#1520] by [@Maan2003])
 - `get_external_handle` on `DelegateCtx` ([#1526] by [@Maan2003])
 - `AppLauncher::localization_resources` to use custom l10n resources. ([#1528] by [@edwin0cheng])
+- Shell: get_content_insets and mac implementation ([#XXXX] by [@rjwittams])
+- Contexts: to_window and to_screen (useful for relatively positioning sub windows) ([#XXXX] by [@rjwittams])
+- WindowSizePolicy: allow windows to be sized by their content ([#XXXX] by [@rjwittams])
 
 ### Changed
 
--- a/druid-shell/src/platform/gtk/window.rs
+++ b/druid-shell/src/platform/gtk/window.rs
@@ -818,8 +818,8 @@ impl WindowHandle {
         }
     }
 
-    pub fn get_content_insets(&self) -> Insets {
-        log::warn!("WindowHandle::get_content_insets unimplemented for GTK platforms.");
+    pub fn content_insets(&self) -> Insets {
+        log::warn!("WindowHandle::content_insets unimplemented for GTK platforms.");
         Insets::ZERO
     }
 
--- a/druid-shell/src/platform/gtk/window.rs
+++ b/druid-shell/src/platform/gtk/window.rs
@@ -32,7 +32,7 @@ use gio::ApplicationExt;
 use gtk::prelude::*;
 use gtk::{AccelGroup, ApplicationWindow, DrawingArea, SettingsExt};
 
-use crate::kurbo::{Point, Rect, Size, Vec2};
+use crate::kurbo::{Insets, Point, Rect, Size, Vec2};
 use crate::piet::{Piet, PietText, RenderContext};
 
 use crate::common_util::{ClickCounter, IdleCallback};
@@ -816,6 +816,11 @@ impl WindowHandle {
         } else {
             Point::new(0.0, 0.0)
         }
+    }
+
+    pub fn get_content_insets(&self) -> Insets {
+        log::warn!("WindowHandle::get_content_insets unimplemented for GTK platforms.");
+        Insets::ZERO
     }
 
     pub fn set_level(&self, level: WindowLevel) {
--- a/druid-shell/src/platform/mac/window.rs
+++ b/druid-shell/src/platform/mac/window.rs
@@ -785,8 +785,6 @@ fn set_size_deferred(this: &mut Object, _view_state: &mut ViewState, size: Size)
 
 fn set_position_deferred(this: &mut Object, _view_state: &mut ViewState, position: Point) {
     unsafe {
-
-
         let window: id = msg_send![this, window];
         let frame: NSRect = msg_send![window, frame];
 
@@ -1071,7 +1069,7 @@ impl WindowHandle {
         }
     }
 
-    pub fn get_content_insets(&self) -> Insets {
+    pub fn content_insets(&self) -> Insets {
         unsafe {
             let screen_height = crate::Screen::get_display_rect().height();
 
--- a/druid-shell/src/platform/mac/window.rs
+++ b/druid-shell/src/platform/mac/window.rs
@@ -40,7 +40,7 @@ use objc::rc::WeakPtr;
 use objc::runtime::{Class, Object, Sel};
 use objc::{class, msg_send, sel, sel_impl};
 
-use crate::kurbo::{Point, Rect, Size, Vec2};
+use crate::kurbo::{Insets, Point, Rect, Size, Vec2};
 use crate::piet::{Piet, PietText, RenderContext};
 
 use super::appkit::{
@@ -127,10 +127,17 @@ pub(crate) struct IdleHandle {
     idle_queue: Weak<Mutex<Vec<IdleKind>>>,
 }
 
+#[derive(Debug)]
+enum DeferredOp {
+    SetSize(Size),
+    SetPosition(Point),
+}
+
 /// This represents different Idle Callback Mechanism
 enum IdleKind {
     Callback(Box<dyn IdleCallback>),
     Token(IdleToken),
+    DeferredOp(DeferredOp),
 }
 
 /// This is the state associated with our custom NSView.
@@ -156,7 +163,7 @@ impl WindowBuilder {
             handler: None,
             title: String::new(),
             menu: None,
-            size: Size::new(500.0, 400.0),
+            size: Size::new(0., 0.),
             min_size: None,
             position: None,
             level: None,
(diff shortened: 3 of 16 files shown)


druid-shell/src/platform/gtk/window.rs
```diff
@@ -818,6 +818,11 @@ impl WindowHandle {
         }
     }
 
+    pub fn get_content_insets(&self) -> Insets {
```
cmyr, line 821: I'd like to just call this `content_insets` here and elsewhere, `get_` is considered 'not idiomatic'.

druid-shell/src/platform/mac/window.rs
```diff
@@ -758,6 +762,43 @@ extern "C" fn draw_rect(this: &mut Object, _: Sel, dirtyRect: NSRect) {
...
+    }
+}
+
+fn set_size_deferred(this: &mut Object, _view_state: &mut ViewState, size: Size) {
+    unsafe {
+        let window: id = msg_send![this, window];
+        let current_frame: NSRect = msg_send![window, frame];
+        let mut new_frame = current_frame;
+
+        // TODO: maintain druid origin (as mac origin is bottom left)
```
cmyr, line 778: Is there a change we want to make now?

druid-shell/src/platform/mac/window.rs
```diff
@@ -758,6 +762,43 @@ extern "C" fn draw_rect(this: &mut Object, _: Sel, dirtyRect: NSRect) {
...
+        new_frame.origin.y -= size.height - current_frame.size.height;
+        new_frame.size.width = size.width;
+        new_frame.size.height = size.height;
+        let () = msg_send![window, setFrame: new_frame display: YES];
+    }
+}
+
+fn set_position_deferred(this: &mut Object, _view_state: &mut ViewState, position: Point) {
+    unsafe {
+        // TODO this should be the max y in orig mac coords
```
cmyr, line 788: ditto, not sure if this is an issue?

druid-shell/src/window.rs
```diff
@@ -194,12 +194,18 @@ impl WindowHandle {
...
 
-    /// Returns the position of the window in [pixels](crate::Scale), relative to the origin of the
+    /// Returns the position of the top left corner of the window in [pixels](crate::Scale), relative to the origin of the
     /// virtual screen.
     pub fn get_position(&self) -> Point {
         self.0.get_position()
     }
 
+    /// Returns the insets of the window content from its position and size in [pixels](crate::Scale).
+    /// This is to account for any window system provided chrome, eg. title bars.
```
cmyr, line 204:
```suggestion
    /// Returns the insets of the window content from its position and size in [pixels](crate::Scale).
    ///
    /// This is to account for any window system provided chrome, eg. title bars.
```

druid/src/app.rs
```diff
@@ -38,11 +36,22 @@ pub struct AppLauncher<T> {
     ext_event_host: ExtEventHost,
 }
 
+/// Defines how a windows size should be determined
+#[derive(Copy, Clone, Debug)]
+pub enum WindowSizePolicy {
+    /// Use the content of the window to determine the size
+    Content,
+    /// Use the provided window size
+    User,
```
cmyr, line 45: Another option here would be to have this enum variant take a `Size`?

druid/src/contexts.rs
```diff
@@ -186,6 +186,25 @@ impl_context_method!(
             self.widget_state.size()
         }
 
+        /// The origin of the widget in window coordinates.
+        pub fn window_origin(&self) -> Point {
+            self.widget_state.window_origin()
+        }
+
+        /// Takes a point in widget coordinates and transforms it to
+        /// window coordinates (relative to the content area, excluding chrome)
```
cmyr, line 195:
```suggestion
        /// Convert a point from the widget's coordinate space to the window's.
        ///
        /// The returned point is relative to the content area; it excludes window chrome.
```

druid/src/contexts.rs
```diff
@@ -186,6 +186,25 @@ impl_context_method!(
...
+        pub fn to_window(&self, widget_point: Point) -> Point {
+            self.window_origin() + widget_point.to_vec2()
+        }
+
+        /// Takes a point in widget coordinates and transforms it to
+        /// screen coordinates
```
cmyr, line 201:
```suggestion
        /// Convert a point from the widget's coordinate space to the screen's.
```

Can we add a link to the screen module here? I've never really looked at that so not sure what would be best to link.

druid/src/core.rs
```diff
@@ -1177,6 +1190,10 @@ impl WidgetState {
     pub(crate) fn add_sub_window_host(&mut self, window_id: WindowId, host_id: WidgetId) {
         self.sub_window_hosts.push((window_id, host_id))
     }
+
+    pub(crate) fn window_origin(&self) -> Point {
+        self.parent_window_origin + self.origin.to_vec2() - self.viewport_offset
```
cmyr, line 1195: is this only used from the various contexts? Could we get away with just storing `parent_window_origin` once, in the `ContextState`, and doing this addition there?

druid/src/app.rs
```diff
@@ -38,11 +36,22 @@ pub struct AppLauncher<T> {
     ext_event_host: ExtEventHost,
 }
 
+/// Defines how a windows size should be determined
+#[derive(Copy, Clone, Debug)]
+pub enum WindowSizePolicy {
+    /// Use the content of the window to determine the size
```
cmyr, line 42:
```suggestion
    /// Use the content of the window to determine the size.
    ///
    /// If you use this option, your root widget will be passed infinite constraints;
    /// you are responsible for ensuring that your content picks an appropriate size.
```

This is my understanding, not sure if it's correct? But seems worth writing down.

cmyr (approved):
Looks great. a few little things inline but nothing major. Thanks!

druid-shell/src/platform/gtk/window.rs
rjwittams: Ok, was just matching the other methods here

druid-shell/src/platform/mac/window.rs
rjwittams: It is if there is a weird monitor layout. Secondary monitors above And below the primary or something like that

druid/src/app.rs
rjwittams: I tried that and it got really tangled up with the events from resize.

druid/src/core.rs
rjwittams: I think it’s much easier to communicate this type of thing in a follow up PR

druid-shell/src/platform/mac/window.rs
rjwittams: This was a TODO I actually fixed

druid-shell/src/platform/mac/window.rs
rjwittams: I made the comment more explicit. I'll fix it when I can set up this weird monitor arrangement to test it or someone else can.

druid/src/app.rs
rjwittams: I think that the whole sizing options on WindowConfig need some thought (eg what gets pushed down to the platform, is that the same everywhere?) but I don't want to get bogged down in it here. Its definitely not in its final form, perfect is the enemy of the good.

druid/src/core.rs
rjwittams: I don't understand how this would work tbh (wouldn't it end up doing calcs that it doesn't need in lots of contexts?) so I'll leave it for now. I did previously store the window origin but that became annoying for viewport_offset updates and had to fiddle with the internal event on the way down.... I'm sure there are ways to rationalise this later when its in place. 

I think viewport_offset in the end wants to be a general invertible transformation (possibly more general than Affine, eg if you wanted perspective or 3d projection etc). Anyway off topic for this...

CHANGELOG.md
```diff
@@ -13,6 +13,9 @@ You can find its changes [documented below](#070---2021-01-01).
 - RichTextBuilder ([#1520] by [@Maan2003])
 - `get_external_handle` on `DelegateCtx` ([#1526] by [@Maan2003])
 - `AppLauncher::localization_resources` to use custom l10n resources. ([#1528] by [@edwin0cheng])
+- Shell: get_content_insets and mac implementation ([#XXXX] by [@rjwittams])
+- Contexts: to_window and to_screen (useful for relatively positioning sub windows) ([#XXXX] by [@rjwittams])
+- WindowSizePolicy: allow windows to be sized by their content ([#XXXX] by [@rjwittams])
```
cmyr, line 18: @rjwittams might want to pr something to fix these up?

druid-shell/src/platform/mac/window.rs
```diff
@@ -156,7 +163,7 @@ impl WindowBuilder {
             handler: None,
             title: String::new(),
             menu: None,
-            size: Size::new(500.0, 400.0),
+            size: Size::new(0., 0.),
```
tirix, line 166: See issue #1543: because of this change, windows don't appear at all on Mac if their size are not explicitly set.

cmyr:
@rjwittams I am hitting a mysterious exception thrown in a cocoa paint method that bisect has tracked down to de5f88ad946d052e377cffa8347ad2cf2c3662dd.

As far as I've tested this only happens on Catalina, although I can reproduce in runebender consistently: run the application, and then open two glyph windows; the app aborts when you try to open the second window.

I'm curious to know more about the rationale of the deferring of the size & position setting; that might help me track this down a bit further.

rjwittams:
This was deferred in order to avoid reentrancy issues (wrt borrowing the app state I think? Due to the window handler going back into druid). 
I was following the patterns in the Windows backend on this (@jneem pointed it out the pattern I think). However I think it might be dumb as it is, and maybe what we should do is defer the notifications, but do the cocoa stuff? I will take another look at it tomorrow.

PR #1533, opened 2021-01-14
Warn on Unhandled commands

Maan2003:


cmyr (commented):
Interesting.

I don't dislike this, but it doesn't totally fit how I use commands myself.

Basically: `is_handled` is way to optionally suppress the continued delivery of an event. I do not expect that all events are handled; and I wouldn't expect a widget to bother calling 'set_handled' in a situation where it was the only target of a command, for instance.

It's possible that we might want to rethink this, though? I'm certainly open to discussing, and this patch seems reasonable on its own...

Maan2003:
Yeah, I also intended this for more of a discussion but patch was simple enough that I submitted the PR.

> I wouldn't expect a widget to bother calling 'set_handled' in a situation where it was the only target of a command, for instance.

I agree but do you think it is a good practice to `set_handled` if a widget is a sole target.

I have also thought of making warning opt in, i.e. replace `@@can-ignore` which will be `@@must-use`. So that it is less annoying. It is a logic error to ignore `OPEN_FILE`, for example

rjwittams:
I feel like if we are going to have options on the selector to control them they should be proper fields and not bits of a string?

cmyr:
I think that having this be opt-in instead of opt-out does definitely make sense, I could see really wanting to log if *certain* events aren't handled.

I agree that stringly-typing here isn't great; a reasonable alternative would be adding a field to `Command`? Although if this is for internal use only I don't mind it being part of the string 🤷

Maan2003:
>  a reasonable alternative would be adding a field to Command?

maybe adding field to `Selector` or `SelectorSymbol`(making it a struct).

cmyr:
I guess it feels like, semantically, whether or not a given command needs to be handled is something related to the command, not to the selector? As in I could at least theoretically imagine a situation where I want the same selector to be used in cases where it is both expected to be handled and not. Although when I actually write this it sounds bad, so maybe not!

rjwittams:
I guess this is a use site vs declaration site dichotomy. I'd normally go for declaration site so most users don't have to worry about it.

Maan2003:
Ready for Review

druid/src/command.rs
```diff
@@ -384,6 +405,11 @@ impl Command {
         .default_to(Target::Global)
     }
 
+    /// Checks if this command must be used.
+    pub fn is_must_use(&self) -> bool {
```
cmyr, line 409: hmm, I find this name awkward, but the name I'd propose ("must_use") is used by the constructor. We could call the construct `new_must_use`, maybe, although that doesn't feel great either? We could also call this `must_be_used` which at least flows a bit better grammatically. I don't love any of these ideas... 😕

druid/src/command.rs
```diff
@@ -473,8 +507,8 @@ impl Command {
         self.get(selector).unwrap_or_else(|| {
             panic!(
                 "Expected selector \"{}\" but the command was \"{}\".",
-                selector.symbol(),
-                self.symbol
+                selector.symbol().str,
+                self.symbol.str
```
cmyr, line 511: instead of these changes I would manually implement `Display` for `SelectorSymbol`. Something like,


```rust
fn fmt(&self, f: Formatter) -> Result<_, _> {
    let must_use = if self.must_use {" (must use" } else { "" };
write!(f, "{}{}", self.str, must_use)
}
```

cmyr (approved):
Looks good, a couple little comments inline :)

druid/src/command.rs
Maan2003: I followed the second one because I expect `Selector::must_use("..")` to used much more than `Command::must_be_used`

druid/src/command.rs
Maan2003: I manually implemented `Debug` now. This output is not meant for end user

cmyr (approved):
Great, thanks!

PR #1537, opened 2021-01-14
Add the changelog

derekdreery:
I'm really bad at doing changelogs. Must. try. harder.


Maan2003 (approved):
Happens to me as well 😄

Issue #1462, opened 2020-12-13
Prevent background redraw for transparent widget (embedding wgpu inside druid)


djeedai:
Hi,

I am trying to write a small proof of concept of embedding a `wgpu` surface inside a druid app, as a custom widget. I've started druid yesterday and I am really liking this project, but would like to make sure it is possible to render a 3D view inside a desktop app. The typical use case is a 3D viewport with some 2D UI around it (think something like a game editor / Unity3D / UnrealEngine, or any CAD application, or some 3D modeller like Maya or Max).

I've started from the `multiwin` example and _lightly_ modified druid for `WindowHandle` to implement the `HasRawWindowHandle` trait, which allow creating a `wgpu` instance from it (is that of interest for #891 by the way? the patch is 10 lines (Windows only)). Then I more or less copy-pasted the "cube" example of `wgpu-rs` into the `multiwin` example of druid.

Now it *almost* works, but it seems that some background redraw is still active, so I cannot get a stable image. When the mouse moves out and in of the window borders, or other repaint is triggered, I briefly see the 3D cube, but then immediately get back the normal 2D druid UI on top of it.

Does someone has any idea how I could fix this, or could point me in the direction of where widgets are cleared? I don't do that background repaint myself. I've tried with `PaintCtx::paint_with_z_index(9999)` but that doesn't work either.

Many thanks for any pointer! :)

djeedai:
I've just seen #1249 and changed `theme::WINDOW_BACKGROUND_COLOR` to (0,0,0,0) but this doesn't do anything in my case, presumably because the 3D clear from `wgpu` is clearing the window anyway. But it seems druid is still painting it.

jneem:
The background is drawn at `druid/src/window.rs:354`. Does commenting it out help at all?

djeedai:
Thanks @jneem; no not really, I see a change (it flickers to black instead of dark gray) but there's still some bad interaction between `wgpu` and `druid`. So it's not a problem with the background. I need to dig into the `wgpu` rendering and the `wgpu::Queue` to understand what is being submitted when.

djeedai:
Sadly I think this is a composition issue. The way I patched up things, wgpu reuses the same HWND created by Druid:
- wgpu creates its own composition pipeline (I don't know the details) from the HWND
[...]

PR #1539, opened 2021-01-16
Add partial support for embedded wgpu on Windows

djeedai:
This change adds partial support for embedding a wgpu rendering pipeline
inside a widget hierarchy via a new `WgpuView` widget backed by a native
platform window child of the main top-level window.

The following set of changes allow embedding the wgpu pipeline:
- Add the `LifeCycleCtx::request_native_window()` method to associate a
  platform window to a widget during the `WidgetAdded` event.
- Add a new event `Event::NativeWindowConnected` raised when the native
  platform window has been created, to allow a widget to initialize
  platform resources requiring that child window handle. The event
  receives a new `NativeWindowHandle` object which allows accessing that
  child platform window.
- Add a new field `PaintCtx::native_origin` which describes the position
  of the widget relative to its inner-most native platform window, which
  is almost always the top-level window (`Window<T>`) or can be a child
  native window when using the previous mechanism.
- Add a new method `Widget::post_render()` invoked after the entire
  widget hierarchy completed a `Widget::paint()` traversal and the
  rendering back-end submitted the associated rendering commands to the
  GPU. This allows `WgpuView` or any other similar future widget to
  submit further draw commands to the GPU (via `wgpu` in the current
  case) without the risk of a race condition with the normal druid
  back-end. This mechanism is a workaround for the lack of explicit
  synchronization or unified rendering pipeline.
- Implement the `HasRawWindowHandle` trait for `WindowHandle` to allow
  native interop to any crate supporting that trait, without explicit
  dependency to said crate.

The new `WgpuView` widget uses those new features to provide a framework
to render arbitrary 3D content inside that widget via a new
`WgpuRenderer` trait. This widget is only enabled with the new
`wgpu_view` feature, because it requires some additional dependencies
that users might not be interested in always pulling into their project
if not using that widget.

A new `wgpu_view` example demonstrates how to use that new widget by
embedding a `WgpuView` widget inside multiple `Split` widgets in a
configuration similar to common desktop 3D applications (toolbars on the
side and top, and 3D view in the center). The sample reuse the cube
sample from the `wgpu-rs` project.

Known limitations:
- The change is limited to support for the Windows platform, although it
  lays some architecture that should make it easy to add support for
  other platforms.
- This change only add support for single-depth native child windows,
  that is platform windows which are a direct child of a top-level
  window. A platform child window cannot be parented to another platform
  child window.
- Interactive animation of the `WgpuView` content (that is, painting
  outside the `post_render()` callback, for example at a given
  framerate) is not addressed nor even explored by this change, and may
  require future work. It is assumed here that the 3D content is static
  and rendering is entirely druid-event-based (mouse click, window
  resize, etc.).

Bug: #1462


djeedai:
Note fore reviewers:
- It seems there's a new "sub-window" feature since I last talked about implementing this feature a few weeks ago, but I have no context/idea what this is (I just noticed it while rebasing before opening that PR) nor if this is duplicate work that was pushed while I was also working on a similar implementation, or if this is orthogonal/unrelated. I am opening that PR anyway to show the current work as is, so please advise.
- I am not sure how we can easily implement the remaining platforms. For example I do not have readily access to Linux, so it obviously fails CI. Please advise on this too, if someone can have a look, if I need to do a "blind implementation" that just compiles, ...?

djeedai:
Ping @jneem @raphlinus can someone have a look at the questions above please to unblock this PR?

rjwittams:
Hi, this is really interesting! I did the subwindows stuff.  
Just to clarify - the subwindows are top level native windows, and the 'sub' part of them is their 'data' / hierarchy relationship - they get sent the data/env of their widgetpod 'parent'.
 
I also did hack raw window handle and wgpu support into a branch a while back, but it looks like you've taken it pretty far!

I haven't looked at the code yet but will try to at some point and see if I can get it compiling/working on the mac. 

To actually get this landed I think you will need to take the same approach I did: 
* Work out what are the minimal changes to druid proper to get this to work. At a guess, this is the exposure of the window handle and the shell changes to 'parent' one window to another? Could all of the wgpu stuff be 'out of tree' in a seperate repo? This could probably go to nursery for a while, or live separately for a decent stretch?
* Split up that minimal set of changes into a series of smaller pull requests that add a single feature. Eg. raw-window-handle in one, set parent with another.   
* At least stub everything out with warnings for other platforms so it compiles...

In general I would say get used to making a bunch of branches/PRs!

I think it also might need some minimal reasoning to show how this api will work on the platforms we have - ie are we making any assumptions that work for windows but are not correct for the mac. 
(As far as I know, on the mac an embedded native window is really the same NSWindow, with another NSView positioned inside it)

rjwittams:
I got this working on the mac, see https://github.com/djeedai/druid/pull/1

rjwittams:
So from getting that working on the mac... 

* post_render 
    * Could be a default do nothing impl in the trait, so it is only wrappers that need to deal with it?
    * I have hacked a further followup that uses paint_with_z_order in the wgpu view instead of post render, did you consider using this instead of adding the post_render method? (To do this it does need to use interior mutability for the closure to have the static lifetime (ie Rc<RefCell<>> of the guts of the WGPUView. But it does mean that a new widget method is not required.)
    * If that doesn't seem right, it feels like it should be a lifecycle event that widgets request in paint. Probably with an internal routing event. 
    
Right now is there any way for events from the native window to be used? I think they need to get delivered to the owner of the window in some way.

druid-shell/src/window.rs
```diff
@@ -224,6 +226,13 @@ impl WindowHandle {
         self.0.get_size()
     }
 
+    /// Set the position and/or size of the native platform window that this window manages.
+    /// This is only valid if the window is a child window. Do not use this to move a top-level
+    /// window `Window<T>`.
+    pub fn set_native_layout(&self, position: Option<Point>, size: Option<Size>) {
```
rjwittams, line 232: Do we envisage these being separately set? ie why not a Rect

druid/src/core.rs
```diff
@@ -468,7 +495,9 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
             let mut visible = ctx.region().clone();
             visible.intersect_with(self.state.paint_rect());
             visible -= layout_origin;
+            ctx.native_origin += layout_origin;
             ctx.with_child_ctx(visible, |ctx| self.paint_raw(ctx, data, env));
+            ctx.native_origin -= layout_origin;
```
rjwittams, line 500: Wouldn't it make more sense to modify this inside the closure to with_child_ctx so we don't need to back it out?

druid/src/win_handler.rs
```diff
@@ -336,6 +389,25 @@ impl<T: Data> Inner<T> {
...
     }
 
+    fn post_render(&mut self, window_id: WindowId) {
+        // Pre-order traversal, so that children render on top of parents
+        if let Some(win) = self.windows.get_mut(window_id) {
+            // Render self first
+            win.do_post_render();
+
+            // Render children next
+            let mut child_ids = vec![];
```
rjwittams, line 399: I don't get what this vec is for

druid/src/widget/wgpu_view.rs
```diff
@@ -0,0 +1,238 @@
...
+    /// Callback invoked when the wgpu view needs to be repaint.
+    ///
+    /// This is where rendering must be submitted by the painter to the given queue.
+    fn render(
+        &mut self,
+        frame: &wgpu::SwapChainTexture,
+        device: &wgpu::Device,
+        queue: &wgpu::Queue,
+    );
+}
```
rjwittams, line 60: This trait seems like it should have event, lifecycle and update methods (that pass along the references to data and env) to allow interactivity and reactivity.

druid/src/widget/wgpu_view.rs
```diff
@@ -0,0 +1,238 @@
...
+
+use crate::widget::prelude::*;
+use crate::{Data, NativeWindowHandle};
+
+use log::{debug, info};
+
+/// A trait for rendering custom content into a `WgpuView` by accessing its internal wgpu pipeline.
+pub trait WgpuRenderer {
+    /// Callback invoked to initilize the painter once the wgpu device and swap chain are up.
+    fn init(
```
rjwittams, line 34: I *think* this would be better if we passed a function type that can construct the WGpuRenderer itself. 
Otherwise every implementer of the trait needs to handle the 'unitialised' state, by being an enum or containing an option.

rjwittams:
I can see there being a case for an option to use the main window as the window handle (i.e not request a new native child window). 
Eg. in a popup, where you know that you are the only widget in the top level window.

djeedai:
> post_render
> Could be a default do nothing impl in the trait, so it is only wrappers that need to deal with it?

I have thought about it, that was one of the open questions I had. Advantage is simplicity. Inconvenient is that any widget pod with children which forgets to implement it will break all its child widgets, and this is not immediately visible until you put a `WgpuView` as a child. I'm happy for the answer to be "testing should ensure we catch that" though.

> I have hacked a further followup that uses paint_with_z_order in the wgpu view instead of post render, did you consider using this instead of adding the post_render method? (To do this it does need to use interior mutability for the closure to have the static lifetime (ie Rc<RefCell<>> of the guts of the WGPUView. But it does mean that a new widget method is not required.)

The problem with `paint_with_z_order()` is that it's still using the same Piet frame to render than the rest of the painting. As far as I understand it works like that:
1. Widget calls `paint_with_z_order()` to delay some painting
2. Core druid painter finishes the normal widget hierarchy traversal pass, then goes to render delayed paintings
3. The Piet rendering is finalized, GPU commands created and sent to the GPU
4. Some time later the GPU renders everything

But since `wgpu` has its own backend, which also submits to the GPU around some time between 1. and 2., the `wgpu` GPU commands are almost always sent first to the GPU and rendered/presented first too. Then later the Piet ones arrive and are rendered/presented, on top of the `wgpu` ones. So at best you see flickering, and generally you don't see any `wgpu` rendering at all. Maybe it works different on macOS but on Windows with DX11 that's what you get. `paint_with_z_order()` was my first approach and it never worked.

> If that doesn't seem right, it feels like it should be a lifecycle event that widgets request in paint. Probably with an internal routing event.

That could work too. But they would probably request that repaint during each `paint()` call anyway. The advantage maybe is that a request mechanism could help support simple animation, something that is impossible with the current design since it's locked to the normal Druid paint logic (at least I think; never looked into Druid animations).

> Right now is there any way for events from the native window to be used? I think they need to get delivered to the owner of the window in some way.

Are you talking about things like mouse clicks etc.? I didn't try that at all, that was kind of out of scope for the first pass, which is already decently large and complex. But yes that is a concern that needs to be addressed.

druid/src/core.rs
djeedai: Do you mean add `native_origin` as a parameter to `with_child_ctx()`, and perform the modification after saving the context?

druid/src/widget/wgpu_view.rs
djeedai: That makes sense a priori. I didn't touch that code for a few weeks so I'm not sure I fully grasp that but I can give it a try for sure.

druid/src/widget/wgpu_view.rs
djeedai: You're talking about the native window events (mouse click _etc._) for example? That makes sense I guess.

druid/src/win_handler.rs
djeedai: Probably just me being a novice in Rust, I couldn't manage to write the second `for` loop below without the borrow checker complaining about a mutable reference already in use, so I had to first extract the child IDs then release the mutable reference to `win` I think, then finally do the loop with the `do_post_render()` call. I'm a bit fuzzy on details. Happy to be taught a better way here.

druid-shell/src/window.rs
djeedai: That was the idea originally yes, because position and size have different semantic when it comes to repainting: position might just move the window without invalidating its content, size almost surely _will_ invalidate the content.

But as I remember it turned out that by the time the PR was opened 1. the native position was trickling down via the `PaintCtx` and leaking into the widget implementation of `paint()`, instead of being handled purely internally, and 2. the widget itself probably had little knowledge of when is best to set the size and when not. So the PR version ended up setting both at once always. I think this is more or less what happened here.

We can probably use a `Rect` now.

rjwittams:
Actually that is what I did first before I was calling post_render, just used request_animation_frame and painted from the AnimFrame event. That worked ok too. I think that the CALayers on the mac make this all magically work no matter what you do though so windows or another platform may be the limiting factor.

djeedai:
Sounds like it indeed. I am pretty sure there's 0 synchronization between Piet and wgpu backends. So the idea was just to do the less bad thing to render Piet first and wgpu second, assuming `WgpuView` will never need to nor be able to be overlaid by any druid rendering. This is for sure a strong limitation, but in the inverse order nothing works.

druid/src/win_handler.rs
rjwittams: Ah I see. No I don't know a way to avoid that, disjoint borrows of collections is a bit of a nightmare. 

It could be written as and maybe a comment saying we clone into a vector to satisfy the borrow checker. 

for child_id in win.children.iter().cloned().collect::<Vec<_>>(){
   ... stuff
}

You could avoid the vec allocation if we limit the number of child windows and put it in a fixed length slice on the stack, or fail over to the vec version when the number goes over some limit.. But probably not worth it....

druid/src/widget/wgpu_view.rs
rjwittams: I mean druid events, so someone can effectively write a widget here all in one place. I think in the end probably WgpuRenderer is a WgpuWidgetHost, ie it adapts a WgpuWidget into a druid one. I can put a proposal together as another PR on your repo, might be easier to explain what I mean in code...

rjwittams:
Yeah so I think for now a LifeCycle event is going to be the least invasive way to achieve this post render behaviour with the semantics you need to have on windows. LifeCycle events don't have a fixed time when they happen (it can be different for each one) so that probably makes the most sense.

djeedai:
Split `HasRawWindowHandle` trait implementation on #1586, including your imlementation for mac @rjwittams.

druid/src/win_handler.rs
djeedai: Actually that's a good point:
- Currently the change only support 1 level of parenting depth. I am not sure if we can make things work for a random parenting depth, but the vector is clearly overkill for now.
- Even if that is implemented, it is unlikely we will see deep hierarchies of native windows. So a fast no-alloc handling for a smaller number is probably a good thing to have for 99% of the cases.
- Painting happens a lot more often than parenting, so we could keep the vector around allocated, and only try to resize it, which would be a no-op almost all of the time.

djeedai:
> I can see there being a case for an option to use the main window as the window handle (i.e not request a new native child window). Eg. in a popup, where you know that you are the only widget in the top level window.

Clarified on Zulip : make requesting a child window optional, for example if `WgpuView` is alone at the root of a (sub)window and therefore can use it directly without the need for its own native child window (which would be at the same position and have the same size anyway).

Eschryn:
If I understood the PR correctly, this is a bad idea as is and the WGPU integration would have a better place in the nursery for now.
From what I've understood this PR fixes the fighting between both renderers by drawing the WGPU content after everything.
This problem also persists in WPF on windows. Implementing it like that introduces the airspace problem though. Now nothing can be drawn above the WGPU render target, which completely puts this control out of the widget hierarchy.

What we actually want is interop with piet. We want to create a texture that acts as render target and gets shared with your favorite graphics api. Your graphics api needs to be able to use that texture as render target. So basically we want offscreen rendering. druid would then just draw this texture on the usual render and other stuff can be drawn above this. Airspace problem fixed. But, that being said wgpu still doesnt support external shared memory as far as im aware. What we wan't is the GPUExternalTexture from the WebGPU specification.

Here a control that does this in WPF: (it uses the OpenGL NV_DX_interop extension to directly interop with DirectX (the only backend WPF supports))
https://github.com/opentk/GLWpfControl

xarvic:
I am not sure if this would work, but since we now support transparent windows, we could show child-window behind the main window and clear the corresponding area in the main window. This would allow us to draw on top of the child window and receive events for the area.

druid/src/widget/widget.rs
```diff
@@ -184,6 +184,16 @@ pub trait Widget<T> {
...
 
+    /// Utility callback invoked after all painting is done and submitted, to perform
+    /// post-painting actions.
+    ///
+    /// This is typically for advanced use only. Most widgets have no use for this. This
+    /// is currently only used for wgpu integration, to allow wgpu-based rendering to
+    /// reliably occur after all druid native rendering finished, to avoid non-deterministic
+    /// out-of-order rendering (parent covering child) between those two unsynchronized
+    /// rendering pipelines.
+    fn post_render(&mut self);
```
xarvic, line 195: It would probably be easier to create a `Lifecycle::PostRender` variant instead of a new method.

PR #1540, opened 2021-01-17
Tuple optional

derekdreery:
This patch adds 2 features. Special care is needed with the `Optional` widget. While I'm sure it is very useful, I'm not sure I've implemented it correctly.


Maan2003 (commented):
Tuple lens looks good! For `Optional`, I recommend moving [Runebender's `Maybe`](https://github.com/linebender/runebender/blob/f38c736/runebender-lib/src/widgets/maybe.rs) to druid

cmyr:
I have similar doubts about the correctness of my `Maybe` widget, but I've worked out at least some of the bugs, and it is certainly useful.

derekdreery:
@cmyr could you comment on the difference between this `Optional` and runebender's `Maybe`? I didn't include a `None` option here, since you can get that functionality using `Either`, but maybe it makes more sense to include it. Also, how to handle events where the event should go to the widget (`should_go_to_hidden`, or something like that, and lifecycle events), but we don't have any data to send to the widget? I handled it by just sending the default value of the `T` of `Option<T>` to the some widget when the `data` was `None`, but this seems suboptimal.

derekdreery:
And another question, if we have a widget for the `None` case, could we include a `Null` widget that lays itself out as `bc.min()` and does nothing else? It would go nicely with `Maybe`/`Optional`.

cmyr:
in no order:

- there sort of is a `Null` widget, in the form of `SizedBox::empty`.
- I like letting there be a `None` option, in case you want a custom placeholder.
- I don't have any good answer about `should_go_to_hidden`. I think the solution I would advocate to not have the widget be hidden, but instead to create and destroy it as needed.

derekdreery:
About the `SizedBox::empty` - it looks to be equivalent to a theoretical `Null` widget. My question is: how do we make this use of `SizedBox` discoverable?

cmyr:
I'm open to suggestions!

derekdreery:
@cmyr I've swapped out the `Optional` for runebender's `Maybe`, and added some intra-crate doc links.

derekdreery:
Ready for review :)

druid/src/widget/sized_box.rs
```diff
@@ -45,6 +45,10 @@ impl<T> SizedBox<T> {
     }
 
     /// Construct container without child, and both width and height not set.
+    ///
+    /// If the widget is unchanged, it will do nothing, which can be useful if you want to draw a
+    /// widget some of the time (for example, it is used to implement
+    /// [`Maybe`][crate::widget::Maybe]).
     pub fn empty() -> Self {
```
Maan2003, line 52: adding `#[doc(alias = "null")]` will help discoverability

derekdreery: Ack I was genuinely going to do that but forgot. Thanks!

druid/src/lens/lens.rs
```diff
@@ -567,3 +567,31 @@ impl<A, B: Clone> Lens<A, B> for Constant<B> {
         f(&mut tmp)
     }
 }
+
+/// A lens that combines two lenses into a tuple.
+#[derive(Debug, Copy, Clone)]
+pub struct Tuple<L1, L2>(pub L1, pub L2);
```
rjwittams, line 573: Wonder if this should be Tuple2 or Pair, as it doesn't apply to all tuples

cmyr:
Do we need both `Maybe` *and* `Optional`? In my use, `Maybe` covers the optional category well enough.

derekdreery:
> 
> 
> Do we need both `Maybe` _and_ `Optional`? In my use, `Maybe` covers the optional category well enough.

Oh I thought I had removed `Optional`. Will do.

derekdreery:
Ready for review again.

cmyr (approved):
I won't make strong guarantees about covering all the corner cases in `Maybe`, but I haven't seen a crash in runebender in a while? 😬

Issue #1543, opened 2021-01-21
Windows don't open on Mac if size has not been set


tirix:
If an application does not call explicitly the `window_size` method of `WindowDesc`, the window does not appear on Mac.
This is the case for some of the examples (anim, custom_widget). Others which actually call that method (like calc) don't have the issue.

tirix:
It seems the issue is with [line 166 of druid-shell's window.rs for mac](https://github.com/linebender/druid/blob/953bd0b2bcd0e4a976facc0d82abfb748a027d8d/druid-shell/src/platform/mac/window.rs#L166).
The PR #1532 changed that line to set a default windows size to (0,0). If I manually set it back to (500, 400), the issue disappears. However, when I run a dropdown example (still on mac), the drop-down sub-windows briefly appears with a bigger size (most probably 500x400) when opened. 
I believe that artefact is less an issue than the fact that some examples seemingly don't work at all. I'd suggest reverting that line and opening a new issue to track the very brief artefact. Ultimately the issue will probably be with finding out the size of the sub-window at the time they are opened.

PR #1547, opened 2021-01-22
Fix window size problem in #1543

rjwittams:
Not convinced this is the ultimate solution but undoes the regression in behaviour


cmyr (approved):
Seems reasonable as a spot fix. Do you have a sense of what the better solution might be?

PR #1551, opened 2021-01-24
Add a CrossAxisAlignment to Flex to fill the available cross space

tirix:
This is for introducing a new enum value `Fill` to `Flex`'s `CrossAxisAlignment`.
Children with the `Fill` cross axis alignment are laid out to fill the entire available cross space, resulting in such layouts:
image
The same layout using `Start`looks like this:
image

### DIFF
--- a/druid/src/widget/flex.rs
+++ b/druid/src/widget/flex.rs
@@ -306,6 +306,9 @@ pub enum CrossAxisAlignment {
     ///
     /// The calculated baseline is the maximum baseline offset of the children.
     Baseline,
+    /// Fill the available space.
+    ///
+    Fill,
 }
 
 /// Arrangement of children on the main axis.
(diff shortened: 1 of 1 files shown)


druid/src/widget/flex.rs
```diff
@@ -306,6 +306,9 @@ pub enum CrossAxisAlignment {
     ///
     /// The calculated baseline is the maximum baseline offset of the children.
     Baseline,
+    /// Fill the available space.
+    ///
```
cmyr, line 310:
```suggestion
    /// Fill the available space.
    ///
    /// The size on this axis is the size of the largest widget;
    /// other widgets must fill that space.
```

cmyr (approved):
This makes sense, and I think it's worth trying, although there are some possible gotchas: for instance the user might expect to be able to use `SizedBox::expand` here in order to use the available space but that will end up in calculating an infinite size in the first layout pass.

I think it's worth getting this in though and seeing if that ends up being a problem in practice, I can see the real benefit here. :)

Thanks!

PR #1557, opened 2021-01-28
Make Index lens a tuple struct

Maan2003:


koutoftimer:
1. You have made inner data public.
2. Druid is still "work in progress" and verbose way of doing things wanted.

I have doubts this pull request makes any sense.

Maan2003:
1. I don't see any reason not to make the index public. (Same in [`lens::Constant`](https://github.com/linebender/druid/blob/4c865b959f6411e464cf89f4c23e81fd9e50081a/druid/src/lens/lens.rs#L559))
2. But people(like me) still want to build stuff with it; reducing verbosity doesn't hurt.

cmyr:
If it were already a tuple struct I think that would be fine, but I don't see any advantage to changing it now. I get that `Index(5)` might feel better than `Index::new(5)` but I also think it's less idiomatic, and I don't expect this to actually matter much in practice.

in any case thanks for the PR, I do appreciate the interest in small improvements, I'm just not quite sold in this case. 😄

I'm going to close this; if you think I've missed something and there's a strong rationale behind this feel free to reopen.

Issue #1554, opened 2021-01-26
Allow container's background and border to be removed


sjoshid:
Once a background and border is added to a `Container` there is no way to remove it. 

Removing them is necessary when, suppose, you have a collection of `Container`s and you want only of them to be painted/highlighted. Clicking on a different `Container` must remove previously selected `Container` background and border.

PR #1558, opened 2021-01-29
Allowing Container's background and border to be removed.

sjoshid:
Closes https://github.com/linebender/druid/issues/1554

### DIFF
--- a/druid/src/widget/container.rs
+++ b/druid/src/widget/container.rs
@@ -74,8 +74,8 @@ impl<T: Data> Container<T> {
         self.background = Some(brush.into());
     }
 
-    /// Removes background.
-    pub fn remove_background(&mut self) {
+    /// Clears background.
+    pub fn clear_background(&mut self) {
         self.background = None;
     }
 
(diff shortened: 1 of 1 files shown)


druid/src/widget/container.rs
```diff
@@ -74,6 +74,11 @@ impl<T: Data> Container<T> {
         self.background = Some(brush.into());
     }
 
+    /// Removes background.
+    pub fn remove_background(&mut self) {
```
tirix, line 78: I'm used to the 'set/clear' and 'add/remove' verbs used in pairs. This would be `set/remove`, so would the right verb rather be 'clear_background'?

tirix (commented):
Technically nothing to say! 
I only have a remark on the naming.

cmyr (approved):
+1, I would go with clear here. otherwise looks good!

druid/src/widget/container.rs
sjoshid: I agree. Made the change.

PR #1559, opened 2021-01-31
Take root widget directly in WindowDesc::new

lassipulkkinen:
`WindowDesc::new` currently accepts the root widget as a closure that returns `W: Widget<_>`, which made sense before the changes in #519. This PR changes that to take `W` directly instead, as using a closure serves no purpose anymore.

EDIT: Sorry for the force push mess; I forgot to `cargo fmt`. Also, reverted AUTHORS change on second thought, as this is probably not eligible for copyright.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ You can find its changes [documented below](#070---2021-01-01).
 ### Changed
 
 - Warn on unhandled Commands ([#1533] by [@Maan2003])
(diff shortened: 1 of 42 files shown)


cmyr (approved):
ha funny, I'd forgotten why this API was like this, but I'm pretty sure you're right.

I do think that this is a simplification, and I'm happy to merge, once we figure out what's up with CI.

PR #1562, opened 2021-02-03
Switch to trace-based logging

PoignardAzur:
* Add dependency to `tracing`
* Convert `log::foobar` calls to `tracing::foobar`
* Remove default dependency to `log`
* Add `default-logger` feature
* Add dev-dependency to `test-env-log`
* Replace `console_log` with `tracing-wasm`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ You can find its changes [documented below](#070---2021-01-01).
 
 - Warn on unhandled Commands ([#1533] by [@Maan2003])
 - `WindowDesc::new` takes the root widget directly instead of a closure ([#1559] by [@lassipulkkinen])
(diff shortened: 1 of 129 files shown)


PoignardAzur:
Quick note: I haven't actually tested the changes on a Windows, MacOs or wasm32 platform.

I *have* ran `cargo check` on the above targets, which helped me find a lot of small errors; rustc is pretty good with this.

PoignardAzur:
@raphlinus I get two

```sh
Error: failed to run custom build command for `pango-sys v0.10.0`
```

Any idea what that means?

raphlinus:
Unfortunately I do not. It seems likely unrelated to your trace PR, but it would take some digging to figure out what's going wrong here.

PR #1563, opened 2021-02-04
Use inferred as-cast instead of isize in SetWindowLongPtrW call

crumblingstatue:
This allows it to be a valid cast for i686-pc-windows-gnu target, where
the style argument is i32 instead of isize.

Ref: #9


raphlinus (approved):
This seems reasonable to me. Thanks!

Issue #1560, opened 2021-02-01
build fails cause No package 'gdk-3.0' found.


astrale-sharp:
Hi there :)

couldn't build the example project cause `No package 'gdk-3.0' found`.

I already have the packages `libgtk-3-bin` , `libgtk-3-0` so Im not sure why its not working

```
❯ screenfetch
                                       astrale@Sharp
 MMMMMMMMMMMMMMMMMMMMMMMMMmds+.        OS: Linuxmint 20.1 ulyssa
 MMm----::-://////////////oymNMd+`     Kernel: x86_64 Linux 5.4.0-65-generic
 MMd      /++                -sNMd:    Uptime: 5h 16m
 MMNso/`  dMM    `.::-. .-::.` .hMN:   Packages: 3878
 ddddMMh  dMM   :hNMNMNhNMNMNh: `NMm   Shell: fish 3.1.0
     NMm  dMM  .NMN/-+MMM+-/NMN` dMM   Resolution: 1920x1080
     NMm  dMM  -MMm  `MMM   dMM. dMM   DE: Cinnamon 4.8.6
     NMm  dMM  -MMm  `MMM   dMM. dMM   WM: Muffin
     NMm  dMM  .mmd  `mmm   yMM. dMM   WM Theme: Mint-Y-Dark-Blue (Mint-Y)
     NMm  dMM`  ..`   ...   ydm. dMM   GTK Theme: Mint-Y-Dark-Red [GTK2/3]
     hMM- +MMd/-------...-:sdds  dMM   Icon Theme: breeze-dark
     -NMm- :hNMNNNmdddddddddy/`  dMM   Font: Ubuntu Condensed, 24
      -dMNs-``-::::-------.``    dMM   Disk: 710G / 1,1T (73%)
       `/dMNmy+/:-------------:/yMMM   CPU: Intel Core i3-2330M @ 4x 2.2GHz [63.0°C]
          ./ydNMMMMMMMMMMMMMMMMMMMMM   GPU: GeForce GT 630M
             \.MMMMMMMMMMMMMMMMMMM     RAM: 2241MiB / 3844MiB
                                      

```

Cargo.toml :
```Toml
[package]
name = "druid"
version = "0.1.0"
authors = ["Astrale"]
edition = "2018"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
druid = { git = "https://github.com/linebender/druid.git" }
```

src/main.rs : 
```rust
[...]

PR #1565, opened 2021-02-06
adds the command for installing gtk to README.md 

astrale-sharp:
... for ubuntu-based distributions, see #1560 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,9 @@ druid = { git = "https://github.com/linebender/druid.git" }
 
 #### Linux
 
-On Linux, Druid requires gtk+3; see [GTK installation page].
+On Linux, Druid requires gtk+3; see [GTK installation page]. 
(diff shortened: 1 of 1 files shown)


jneem (approved):
Thanks, this definitely looks like an improvement so I'll go ahead and merge it.

I do think we could do even better here in the future:
- is the link to the GTK installation page useful? I suspect 99%+ of linux users will want their distro packages
- are there other common distributions that require a development package of GTK? A quick google suggests that Fedora does, I don't know about any others...

jneem (approved):
Thanks, this information is definitely worth having. I'll go ahead and merge this, although I think this section could be improved still further:
- do we want the link to the GTK installation page? I suspect 99%+ of users will want their distro packages
- are there other common distros that need a separate development package? A quick google suggests that Fedora does, but maybe there are others...

PR #1574, opened 2021-02-09
Ctrl+Backspace delete left word not working in Windows 10

Ciantic:
I tested all CTRL shortcuts #615 and found out that Ctrl+Backspace was not working in Windows 10, turns out that character 7f is ctrl+backspace. This fixes the behavior in Windows 10.

If you search how other have handled the 0x7f, [WM_CHAR it seems to be common exception](https://github.com/eval1749/evita-common-lisp/blob/cb2cbd3cc541878d25dcedb4abd88cf8e2c881d9/listener/work/NOCMD__vi_edit_pane.cpp#L267).


druid-shell/src/platform/windows/keyboard.rs
```diff
@@ -610,6 +610,8 @@ impl KeyboardState {
                     let code = scan_to_code(scan_code);
                     let key = if self.stash_utf16.is_empty() && wparam < 0x20 {
                         vk_to_key(vk).unwrap_or_else(|| self.get_base_key(vk, mods))
+                    } else if wparam == 0x7f {
```
raphlinus, line 613: I think I would prefer `(wparam < 0x20 || wparam == 0x7f)` two lines up instead of this, as we already have the mapping and that would make it more likely that the non-ctrl and ctrl keys map to the same thing, in edge cases. But yes, this is the right place to fix it.

Ciantic: @raphlinus I incorporated this, should I squash these commits?

raphlinus (approved):
Thanks! And yeah, squashing commits is probably cleaner here.

Ciantic:
Ok, I think it could be now merged, I think Github has a button to Merge+Squash, I haven't done githubs pull request too often.

PR #1576, opened 2021-02-09
Fix doc links, formatting and clippy warnings

Maan2003:


druid-shell/src/platform/windows/window.rs
```diff
@@ -95,6 +95,7 @@ pub(crate) struct WindowBuilder {
 /// It's very tricky to get smooth dynamics (especially resizing) and
 /// good performance on Windows. This setting lets clients experiment
 /// with different strategies.
+#[allow(dead_code)]
```
cmyr, line 98: I think this should probably be removed, but this is okay for now.

cmyr (approved):
Thanks, cleanup is always appreciated.

PR #1580, opened 2021-02-11
Updates for rustc 1.50

cmyr:
Suppresses a few new clippy lints and move to using the
std f64::clamp method.


raphlinus (approved):
Works for me! Hopefully this will unblock other changes.

Issue #1577, opened 2021-02-10
Overlay Widgets


raymanfx:
One thing I'm missing in Druid right now is overlay type widgets. Common widgets that fit into this category are picklists, but also modal widgets. As a visual example, this is the `material.io` example overlay widget [1]:

[<img src="https://lh3.googleusercontent.com/djK1rThQ30yU32azhWve0J0rMUkJLzPnitUbEwYf5Hnf5MZgmmFlY-LbGj6n51UeyroPUywoqirccgyzAPAaNNCl0gKMqmqCsnA5Dg=w1064-v0">](material.io)

Now I guess https://github.com/linebender/druid/issues/429 (modal widgets) is related, but I'm not entirely sure whether it's the same thing. A full blown modal is a window that requires immediate attention and usually has buttons or other methods of control flow alteration (OK/Cancel/..). The example I linked in [1] is more of a 'simple' popover widget: you press a button and the list pops up. I guess it's technically just a `Scroll<List<Button>>`, since each element can be clicked to provoke an action.

Please let me know what you think.

**EDIT 1**
Some work for dropdown menus, modal widgets etc. has been done here: https://github.com/rjwittams/druid/commits/sub-window, but not merged back into upstream Druid (this repo). The current state of affairs is unclear, but I guess it would require some amount of work to rebase it onto the current HEAD.

[1] https://material.io/components/menus#dropdown-menu

Ciantic:
Have you tried the dropdown in nursery? https://github.com/linebender/druid-widget-nursery

Before websites we just did these with separate windows, which I think is a nicer way to do these than overlay on others. But I can see Overlay Widgets as useful too if they are not out of scope.

rjwittams:
That branch of mine is all merged from a separate squashed branch. 
I'm not sure if this issue really makes much sense given that subwindows do work in master, but specific widgets need to be written to use them. 
I'm sure there will be improvements to make once more experience is gathered from using subwindows.

Issue #1249, opened 2020-09-21
Transparent Window Background Color


Maan2003:
setting `theme::WINDOW_BACKGROUND_COLOR` to a transparent/translucent color leads to some rendering bugs.
examples: 

![image](https://user-images.githubusercontent.com/49202620/93801075-186b4f80-fc5f-11ea-9647-07dd836812eb.png)
![image](https://user-images.githubusercontent.com/49202620/93801135-320c9700-fc5f-11ea-9603-37c9fcf53ab5.png)

derekdreery:
IIRC transparent backgrounds is something that is *very* difficult to get right, because of the cross-platform nature of druid. Because druid will write over parts of the buffer that have changed, if those bits are not opaque you will see the content that was behind them before. Also, if you are writing to some OS-supplied buffer, then it will essentially be uninitialized data, i.e. it will have random stuff that was there before (which looks like the first image).

derekdreery:
Out of interest, are you on windows?

derekdreery:
This may be relevant: https://stackoverflow.com/questions/3970066/creating-a-transparent-window-in-c-win32

raphlinus:
Pasted from Zulip: Doing transparent window backgrounds entails a bunch of changes. You have to do some settings on the swapchain. I think you have to disable the redirection surface on window creation, but I'm not 100% sure about that. If you do that, I believe you lose menus unless you do some more magic to bring that back.

Maan2003:
This is reproducible without any changes to druid, just changing the `WINDOW_BACKGROUND_COLOR`.  

most of the bugs occur on resize. but some like textbox caret leaving trails don't require resize ( trails fade away after sometime )

> Out of interest, are you on windows? 
yes.

cmyr:
my feeling is that we should ignore alpha for the background channel. If at some point we want to support transparent backgrounds as a platform-specific feature I'm open to that, but for the time being i'd consider this a bug.

Ciantic:
I have a fork, where transparency now works:

<img src="https://user-images.githubusercontent.com/64731/107764574-16eb1280-6d39-11eb-9ac0-c2fb6473f1c8.png" width="500" />

Resizing too seems to work:

<img src="https://user-images.githubusercontent.com/64731/107764639-35510e00-6d39-11eb-9f06-da65f9231828.png" width="500" />

Test out the example "transparency.rs" in my fork #1583.

PR #1583, opened 2021-02-11
Transparent background for tooltips, dropdowns etc.

Ciantic:
Goal is to allow create overlay windows, such as tooltips, dropdowns etc. by simply setting the backround of the window to transparent. It also allowes to do windows without decorations (#427) and would be better way for creating overlay widgets (#1577). Additionally #1249.

https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/Semi-transparent.20windows.20without.20titlebars

For instance transparency is required to make tooltip shaped window:

Or dropdown window with custom shadows:

Or if you want to make your UI go full WinAmp style from 90's:

---

New API for window builder `transparent`:

```rust
    let main_window = WindowDesc::new(build_root_widget())
        .title("Hello World!")
        .transparent(true) // <-- NEW SETTING HERE
        .window_size((400.0, 400.0));
```

To get fully transparent background, hide also the title bar with `show_titlebar(false)`.

---

Changes include a simple example in `examples/transparency.rs`, currently it renders a test widget like this on a transparent background:

---

Note about Windows 10 code changes: This feature is not supposed to change any behavior of non-transparent windows at all, the opaque windows still should use the same code as before.

The method I chose follows this tutorial: ["Windows with C++ : High-Performance Window Layering Using the Windows Composition Engine"](https://docs.microsoft.com/en-us/archive/msdn-magazine/2014/june/windows-with-c-high-performance-window-layering-using-the-windows-composition-engine), it's the most recent, I asked about that in [window-rs repo](https://github.com/microsoft/windows-rs/issues/541). 

Included are Mac support by @rjwittams, and Linux (GTK) support by @JAicewizard. X11 and web support is stubbed.

### DIFF
--- a/druid-shell/src/platform/windows/window.rs
+++ b/druid-shell/src/platform/windows/window.rs
@@ -272,6 +272,8 @@ impl Drop for HCursor {
 const DS_RUN_IDLE: UINT = WM_USER;
 
 /// Transparent bg clearing color
+///
+/// FIXME: Replace usage with Color::TRANSPARENT on next Piet release
 const TRANSPARENT: Color = Color::rgba8(0, 0, 0, 0);
 
 /// Message relaying a request to destroy the window.
@@ -398,15 +400,12 @@ impl WndState {
             // Piet is missing alpha blending setting, so we have to call
             // ID2D1DeviceContext::SetPrimitiveBlend() manually to clear just
             // the required pixels
-            let dc_for_transparency: Option<&ComPtr<ID2D1DeviceContext>> = if self.transparent {
-                Some(unsafe {
+            let dc_for_transparency: Option<&ComPtr<ID2D1DeviceContext>> =
+                self.transparent.then(|| unsafe {
                     (rt as *mut _ as *mut ComPtr<ID2D1DeviceContext>)
                         .as_ref()
                         .unwrap()
-                })
-            } else {
-                None
-            };
+                });
 
             let mut piet_ctx = Piet::new(d2d, dw.clone(), rt);
 
--- a/druid-shell/src/platform/windows/window.rs
+++ b/druid-shell/src/platform/windows/window.rs
@@ -35,6 +35,11 @@ use winapi::shared::dxgitype::*;
 use winapi::shared::minwindef::*;
 use winapi::shared::windef::*;
 use winapi::shared::winerror::*;
+use winapi::um::d2d1_1::ID2D1DeviceContext;
+use winapi::um::d2d1_1::D2D1_PRIMITIVE_BLEND_COPY;
+use winapi::um::dcomp::{
+    DCompositionCreateDevice, IDCompositionDevice, IDCompositionTarget, IDCompositionVisual,
+};
 use winapi::um::dwmapi::DwmExtendFrameIntoClientArea;
 use winapi::um::errhandlingapi::GetLastError;
 use winapi::um::shellscalingapi::MDT_EFFECTIVE_DPI;
@@ -43,12 +48,14 @@ use winapi::um::uxtheme::*;
 use winapi::um::wingdi::*;
 use winapi::um::winnt::*;
 use winapi::um::winuser::*;
+use winapi::Interface;
+use wio::com::ComPtr;
 
 use piet_common::d2d::{D2DFactory, DeviceContext};
 use piet_common::dwrite::DwriteFactory;
 
 use crate::kurbo::{Insets, Point, Rect, Size, Vec2};
-use crate::piet::{Piet, PietText, RenderContext};
+use crate::piet::{Color, Piet, PietText, RenderContext};
 
 use super::accels::register_accel;
 use super::application::Application;
@@ -86,6 +93,7 @@ pub(crate) struct WindowBuilder {
     resizable: bool,
     show_titlebar: bool,
     size: Option<Size>,
+    transparent: bool,
     min_size: Option<Size>,
     position: Option<Point>,
     state: window::WindowState,
@@ -188,6 +196,7 @@ struct WindowState {
     timers: Arc<Mutex<TimerSlots>>,
     deferred_queue: RefCell<Vec<DeferredOp>>,
     has_titlebar: Cell<bool>,
+    is_transparent: Cell<bool>,
     // For resizable borders, window can still be resized with code.
     is_resizable: Cell<bool>,
     handle_titlebar: Cell<bool>,
@@ -225,6 +234,7 @@ struct WndState {
     // capture. When the first mouse button is down on our window we enter
     // capture, and we hold it until the last mouse button is up.
     captured_mouse_buttons: MouseButtons,
+    transparent: bool,
     // Is this window the topmost window under the mouse cursor
     has_mouse_focus: bool,
     //TODO: track surrogate orphan
@@ -236,6 +246,14 @@ struct WndState {
 /// State for DXGI swapchains.
 struct DxgiState {
     swap_chain: *mut IDXGISwapChain1,
+
+    // These ComPtrs must live as long as the window
+    #[allow(dead_code)]
+    composition_device: Option<ComPtr<IDCompositionDevice>>,
+    #[allow(dead_code)]
+    composition_target: Option<ComPtr<IDCompositionTarget>>,
+    #[allow(dead_code)]
+    composition_visual: Option<ComPtr<IDCompositionVisual>>,
 }
 
 #[derive(Clone, PartialEq)]
@@ -254,6 +272,9 @@ impl Drop for HCursor {
 
 /// Message indicating there are idle tasks to run.
 const DS_RUN_IDLE: UINT = WM_USER;
+
+/// Transparent bg clearing color
+const TRANSPARENT: Color = Color::rgba8(0, 0, 0, 0);
 
 /// Message relaying a request to destroy the window.
 ///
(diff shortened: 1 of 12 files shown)


druid-shell/src/platform/windows/keyboard.rs
```diff
@@ -608,7 +608,7 @@ impl KeyboardState {
                     let scan_code = ((lparam & SCAN_MASK) >> 16) as u32;
                     let vk = self.refine_vk(stash_vk.unwrap_or(0), scan_code);
                     let code = scan_to_code(scan_code);
-                    let key = if self.stash_utf16.is_empty() && wparam < 0x20 {
+                    let key = if self.stash_utf16.is_empty() && (wparam < 0x20 || wparam == 0x7f) {
```
Ciantic, line 611: This was already commited to master, not sure why it appears here again, I might need to do a new pull request to remove this line.

druid-shell/src/platform/windows/window.rs
```diff
@@ -670,7 +676,7 @@ impl WndProc for MyWndProc {
             WM_ACTIVATE => {
                 if LOWORD(wparam as u32) as u32 != 0 {
                     unsafe {
-                        if !self.has_titlebar() {
+                        if !self.has_titlebar() && !self.is_transparent() {
```
JAicewizard, line 679: why should it necessarily be transparent to ignore drop shadows? This also depends on the application, if its 100% transparent shadows would look weird but if its only 10% transparent it would still be nice to have shadows.

druid/examples/transparency.rs
```diff
@@ -0,0 +1,89 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! An example of a transparent window background.
+//! Useful for dropdowns, tooltips and other overlay windows.
+
+use druid::{kurbo::{Circle}, widget::{Button, Flex}};
```
JAicewizard: I dont know the druid guidelines around this, but I think its preferred to not have nested imports like this?

druid/examples/transparency.rs
```diff
@@ -0,0 +1,89 @@
...
+            bc.constrain(size)
+        } else {
+            bc.max()
+        }
+    }
+
+    // The paint method gets called last, after an event flow.
+    // It goes event -> update -> layout -> paint, and each method can influence the next.
+    // Basically, anything that changes the appearance of a widget causes a paint.
+    fn paint(&mut self, ctx: &mut PaintCtx, data: &String, env: &Env) {
```
JAicewizard: put underscores before unused parameters

druid/examples/transparency.rs
```diff
@@ -0,0 +1,89 @@
...
+
+pub fn main() {
+    let btn = Button::new("Hello there");
+    let example = Flex::column().with_child(CustomWidget {}).with_child(btn);
+    let window = WindowDesc::new(example)
+        .show_titlebar(false)
+        .window_size((823., 823.))
+        .transparent(true)
+        .resizable(true)
+        .title(LocalizedString::new("Fancy Colors"));
```
JAicewizard: just use a string instead of `LocalizedString`

druid/src/app.rs
```diff
@@ -57,6 +57,7 @@ pub struct WindowConfig {
     pub(crate) min_size: Option<Size>,
     pub(crate) position: Option<Point>,
     pub(crate) resizable: Option<bool>,
+    pub(crate) transparent: Option<bool>,
```
JAicewizard, line 60: I am not really sure why resizable is an option either, but to me it would make sense for this to not be a option. its either transparent or not. (defaulting to false of course)

JAicewizard:
you should also ad stubs for any unsupported platform

druid-shell/src/platform/windows/window.rs
Ciantic: Yeah, I think it's just not possible to run `DwmExtendFrameIntoClientArea` twice. To get the transparency it's already executed above with negative margins.

druid/examples/transparency.rs
Ciantic: This was patched by Robert.

druid-shell/src/platform/windows/window.rs
```diff
@@ -1293,7 +1305,11 @@ impl WindowBuilder {
             };
             win.wndproc.connect(&handle, state);
 
-            let mut dwStyle = WS_OVERLAPPEDWINDOW;
+            let mut dwStyle = if self.transparent {
+                WS_POPUP
```
JAicewizard, line 1309: not every transparent window is a popup!!

druid/examples/transparency.rs
Ciantic: Yeah, I removed that, VSCode does these automatically, thanks!

druid/examples/transparency.rs
Ciantic: Changed to String

druid-shell/src/platform/windows/window.rs
Ciantic: I removed that, new way of doing transparency doesn't require WS_POPUP, old way required that.

druid/src/app.rs
Ciantic: Yeah I agree, however I think the reason it and the others are Option, is probably because it's a builder? In builder pattern I think it's usual to have Option so that to not force the defaults in builder.

Ciantic:
This is now ready for review 🚢, only thing I'm annoyed is the amount of commits there, I think we can't squash this as there are contributions from others, and GitHub doesn't allow squashing only some from the older commits?

JAicewizard:
local squash + force push ;)

Ciantic:
I'm rebasing this, currently a total mess.

Ciantic:
Now I have rebased / squashed this

Ciantic:
Note for posterity, we removed two spurious `println!` lines with force push right there. The Github log doesn't make this very clear if we don't do real commits.

Ciantic:
@cmyr I think this is ready for review.

druid/src/app.rs
JAicewizard: I cant resolve this for some reason, you'd think it makes sense for the reviewer to be able to mark their own comments are resolved right...
Could you mark it as resolved it for me?

raphlinus:
This causes failure on Windows 7, STATUS_DLL_NOT_FOUND in particular. I haven't investigated in detail, but it's most likely the dependency on DirectComposition. Since it's not supported on Windows 7, it needs to be loaded as optional functions. That existed for a long while but was removed in #1191, specifically commit 290f6e4ffa52e153cf27a25d281a0ff9c00b06d2.

Ciantic:
> it needs to be loaded as optional functions

What do you mean by this? Should we make the `transparent()` function optional somehow?

Ciantic:
It's all green again 🟢. I don't have Windows 7 to test, and looks like the CI doesn't test that either, but when you've got time give it a test on Win 7. 

If it doesn't work, I may have to get Win 7 running in vbox, I don't want to bother you to test the compilation after my changes.

raphlinus:
It works as expected - hello works fine, and the "transparency" example isn't transparent but is otherwise functional. You might consider logging a warning in that case. So this is unblocked on me, but still requires someone to review in all the PR's glorious detail.

Ciantic:
I added the warning if user is using Win 7 and tries transparent, and squashed my win 7 fix changes.

druid-shell/src/platform/windows/window.rs
```diff
@@ -255,6 +271,9 @@ impl Drop for HCursor {
 /// Message indicating there are idle tasks to run.
 const DS_RUN_IDLE: UINT = WM_USER;
 
+/// Transparent bg clearing color
+const TRANSPARENT: Color = Color::rgba8(0, 0, 0, 0);
```
cmyr, line 275: Let's add a FIXME here to use the `Color::TRANSPARENT` const in piet whenever we do the next release there.

druid-shell/src/platform/windows/window.rs
```diff
@@ -373,9 +392,38 @@ impl WndState {
     // Renders but does not present.
     fn render(&mut self, d2d: &D2DFactory, dw: &DwriteFactory, invalid: &Region) {
         let rt = self.render_target.as_mut().unwrap();
+
         rt.begin_draw();
         {
+            // Piet is missing alpha blending setting, so we have to call
+            // ID2D1DeviceContext::SetPrimitiveBlend() manually to clear just
+            // the required pixels
+            let dc_for_transparency: Option<&ComPtr<ID2D1DeviceContext>> = if self.transparent {
```
cmyr, line 401: As of 1.50 there is the very helpful [`bool::then`](http://doc.rust-lang.org/1.50.0/std/primitive.bool.html#method.then) for this sort of situation...

druid/examples/transparency.rs
```diff
@@ -0,0 +1,88 @@
...
+//! Useful for dropdowns, tooltips and other overlay windows.
+
+use druid::kurbo::Circle;
+use druid::widget::prelude::*;
+use druid::widget::{Button, Flex};
+use druid::{AppLauncher, Color, Rect, WindowDesc};
+
+struct CustomWidget;
+
+impl Widget<String> for CustomWidget {
```
cmyr, line 25: you probably don't need to implement all of `Widget` here; you can maybe just use a `Painter` widget?

cmyr (approved):
This looks good to me, a few little observations inline but nothing blocking.

I'm curious if this has been tested on multiple monitors, at all; that seems like a possible source of complications. I'm also slightly scared of macOS, and I'll test again on my laptop with a second monitor when I get home later tonight.

Thanks in any case, this should unblock a lot of other work!

Ciantic:
I pushed some changes.

cmyr:
Okay, this looks good to me, thanks again!

PR #1584, opened 2021-02-12
dont use widgets for padding in flex

### LINKED ISSUE #1566: Flex spacers should not be implemented as widgets.
Currently spacers are inefficiently implemented via normal widgets, this is not really the right tool for the job.
If spacers are just widgets, every time they are layed out they have to go through a `WidgetPod`, which doesn't actually do much in this case.

If instead we put some spacer data with the widgets, we can remove the unneeded layer of abstraction.
I used the `flex` showcase as a benchmark, timing the amount of time a layout each time, and averaging the values over time. (and a separate thread to force a rebuild and thus reload)
I implemented a simple system to use this method, which (at least looks like) it produces the exact same results on screen as before.
These are the results:
| spacer setting\time(µs)  | before | after |
| ------------- | ------------- |------------- |
| none  | 100  | 104  |
| default  | 128  | 112  |
| flex  | 116  | 104  |
| fixed  | 112  | 104  |

It is expected to see values go up with none since it doesn't benefit, while still having to carry the extra weight.

Implementation aside, is this something druid wants?

EDIT:
https://github.com/JAicewizard/druid/tree/non_widget_padding

JAicewizard:
Having a special purpose widget just for padding in flexes is a bit too flexible ( ;) ), and not needed. It really doesnt need to be a widget, if we just store the amount of padding (px/flex) we can easily calculate its width and layout.

fixes #1566


druid/src/widget/flex.rs
```diff
@@ -583,39 +571,49 @@ impl<T: Data> Flex<T> {
...
+        let new_child = Child::FlexedSpacer(flex, 0.0);
+        self.children.push(new_child);
     }
 }
 
 impl<T: Data> Widget<T> for Flex<T> {
     fn event(&mut self, ctx: &mut EventCtx, event: &Event, data: &mut T, env: &Env) {
         for child in &mut self.children {
-            child.widget.event(ctx, event, data, env);
+            match child {
```
cmyr, line 589: I would personally be tempted to have a method on `Child`, like:

```rust
impl<T> Child<T> {
    fn widget_mut(&mut self) -> Option<&mut WidgetPod<T, Box<dyn Widget<T>>>;
}
```

and then this stuff cleans up a little bit. Just a thought...

druid/src/widget/flex.rs
```diff
@@ -927,6 +948,17 @@ impl From<f64> for FlexParams {
     }
 }
 
+enum Child<T> {
+    Fixed(WidgetPod<T, Box<dyn Widget<T>>>, Option<CrossAxisAlignment>),
```
cmyr, line 952: I personally find that once I hit two or so fields I really prefer struct variants over enum variants.

cmyr (approved):
Okay, this looks good. I was a little concerned it would be harder to understand, but I don't really think that's the case.


I have a few notes inline, but nothing blocking, will give you a chance to address what you want and then merge.

Thanks!

druid/src/widget/flex.rs
JAicewizard: oh wow why do people use tuple variants when struct variants exist!!! this is so much better!!!!!!!!

JAicewizard: I remember, you cant easily ignore fields...

druid/src/widget/flex.rs
```diff
@@ -585,34 +588,25 @@ impl<T: Data> Flex<T> {
...
 impl<T: Data> Widget<T> for Flex<T> {
     fn event(&mut self, ctx: &mut EventCtx, event: &Event, data: &mut T, env: &Env) {
-        for child in &mut self.children {
-            match child {
-                Child::Fixed(widget, _) | Child::Flex(widget, _, _) => {
-                    widget.event(ctx, event, data, env)
-                }
-                _ => {}
+        for child in self.children.iter_mut().map(|x| x.widget_mut()) {
+            if let Some(widget) = child {
```
cmyr, line 592: just a note, if you used `flat_map` instead of `map` you wouldn't need the `if let Some(..)` business.

druid/src/widget/flex.rs
```diff
@@ -661,7 +655,12 @@ impl<T: Data> Widget<T> for Flex<T> {
                     *calculated_siz = kv.resolve(env);
                     major_non_flex += *calculated_siz;
                 }
-                Child::Flex(_, _, flex) | Child::FlexedSpacer(flex, _) => flex_sum += *flex,
+                Child::Flex {
```
cmyr, line 658: you can write this as, `Child::Flex { flex, .. } | Child::FlexedSpacer(flex, _) =>`.

druid/src/widget/flex.rs
```diff
@@ -674,7 +673,11 @@ impl<T: Data> Widget<T> for Flex<T> {
         // Measure flex children.
         for child in &mut self.children {
             match child {
-                Child::Flex(widget, _, flex) => {
+                Child::Flex {
```
cmyr, line 676: ditto, just `Child::Flex { widget, flex, ..}`

with named arguments order doesn't matter.

druid/src/widget/flex.rs
```diff
@@ -724,7 +727,12 @@ impl<T: Data> Widget<T> for Flex<T> {
 
         for child in &mut self.children {
             match child {
-                Child::Fixed(widget, alignment) | Child::Flex(widget, alignment, _) => {
+                Child::Fixed { widget, alignment }
+                | Child::Flex {
```
cmyr, line 731: ditto, just `Child::Flex { widget, alignment, .. }`

druid/src/widget/flex.rs
```diff
@@ -949,16 +957,50 @@ impl From<f64> for FlexParams {
...
+                widget,
+                alignment: _,
+            }
+            | Child::Flex {
+                widget,
+                alignment: _,
+                flex: _,
+            } => Some(widget),
+            _ => None,
+        }
```
cmyr, line 986:
```suggestion
            Child::Fixed { widget, .. } | Child::Flex { widget, .. } => Some(widget),
            _ => None,
        }
```

etc etc 😇

cmyr (commented):
Was initially going to just make some notes and merge this but I think you can clean it up by quite a bit by using `flat_map` instead of `map`, and by using more concise destructuring syntax.

Sorry for the nits, will be nice to have this all cleaned up though. :)

druid/src/widget/flex.rs
JAicewizard: oh rust, the language with infinite hidden wonders... :)
This is where it shows that I dont program rust, I google syntax.

druid/src/widget/flex.rs
JAicewizard: I got you one better, google tells me `filter_map` is even better!!!

druid/src/widget/flex.rs
```diff
@@ -583,39 +574,34 @@ impl<T: Data> Flex<T> {
...
+        self.children.push(new_child);
     }
 }
 
 impl<T: Data> Widget<T> for Flex<T> {
     fn event(&mut self, ctx: &mut EventCtx, event: &Event, data: &mut T, env: &Env) {
-        for child in &mut self.children {
-            child.widget.event(ctx, event, data, env);
+        for child in self.children.iter_mut().filter_map(|x| x.widget_mut()) {
+            child.event(ctx, event, data, env)
```
cmyr, line 592: doesn't matter except that it can be slightly visually confusing.

```suggestion
            child.event(ctx, event, data, env);
```

druid/src/widget/flex.rs
```diff
@@ -583,39 +574,34 @@ impl<T: Data> Flex<T> {
...
         }
     }
 
     fn lifecycle(&mut self, ctx: &mut LifeCycleCtx, event: &LifeCycle, data: &T, env: &Env) {
-        for child in &mut self.children {
-            child.widget.lifecycle(ctx, event, data, env);
+        for child in self.children.iter_mut().filter_map(|x| x.widget_mut()) {
+            child.lifecycle(ctx, event, data, env)
```
cmyr, line 598:
```suggestion
            child.lifecycle(ctx, event, data, env);
```

druid/src/widget/flex.rs
```diff
@@ -583,39 +574,34 @@ impl<T: Data> Flex<T> {
...
         }
     }
 
     fn update(&mut self, ctx: &mut UpdateCtx, _old_data: &T, data: &T, env: &Env) {
-        for child in &mut self.children {
-            child.widget.update(ctx, data, env);
+        for child in self.children.iter_mut().filter_map(|x| x.widget_mut()) {
+            child.update(ctx, data, env)
```
cmyr, line 604:
```suggestion
            child.update(ctx, data, env);
```

druid/src/widget/flex.rs
```diff
@@ -782,8 +803,8 @@ impl<T: Data> Widget<T> for Flex<T> {
     }
 
     fn paint(&mut self, ctx: &mut PaintCtx, data: &T, env: &Env) {
-        for child in &mut self.children {
-            child.widget.paint(ctx, data, env);
+        for child in self.children.filter_map().map(|x| x.widget_mut()) {
```
cmyr, line 806:
```suggestion
        for child in self.children.iter_mut().filter_map(|x| x.widget_mut()) {
```

cmyr (commented):
Basically there :)

druid/src/widget/flex.rs
JAicewizard: weird that it removed it?? I didnt change anything of the lines besides pasteing child in place of widget. 
VScode also changed indentation, so maybe it also removed the `;` for me.

druid/src/widget/flex.rs
```diff
@@ -782,8 +803,8 @@ impl<T: Data> Widget<T> for Flex<T> {
     }
 
     fn paint(&mut self, ctx: &mut PaintCtx, data: &T, env: &Env) {
-        for child in &mut self.children {
-            child.widget.paint(ctx, data, env);
+        for child in self.children.iter_mut().filter_map(|x| x.widget_mut()) {
+            child.paint(ctx, data, env)
```
cmyr, line 807:
```suggestion
            child.paint(ctx, data, env);
```

cmyr (approved):
Great, thanks!

PR #1586, opened 2021-02-13
Implement HasRawWindowHandle trait for win/mac

djeedai:
Implement the HasRawWindowHandle trait for Windows and macOS platforms,
to allow access to native window handles for interop.

### DIFF
--- a/druid-shell/Cargo.toml
+++ b/druid-shell/Cargo.toml
@@ -18,7 +18,7 @@ default = ["gtk"]
 gtk = ["gio", "gdk", "gdk-sys", "glib", "glib-sys", "gtk-sys", "gtk-rs", "gdk-pixbuf"]
 x11 = ["x11rb", "nix", "cairo-sys-rs"]
 # Implement HasRawWindowHandle for WindowHandle 
-raw-win-handle = ["libc", "raw-window-handle"]
+raw-win-handle = ["raw-window-handle"]
 
 # passing on all the image features. AVIF is not supported because it does not
 # support decoding, and that's all we use `Image` for.
@@ -54,7 +54,6 @@ keyboard-types = { version = "0.5.0", default_features = false }
 # Optional dependencies
 image = { version = "0.23.12", optional = true, default_features = false }
 raw-window-handle = { version = "0.3.3", optional = true, default_features = false }
-libc = { version = "0.2.81", optional = true, default_features = false }
 
 [target.'cfg(target_os="windows")'.dependencies]
 scopeguard = "1.1.0"
--- a/druid-shell/Cargo.toml
+++ b/druid-shell/Cargo.toml
@@ -17,6 +17,8 @@ default-target = "x86_64-pc-windows-msvc"
 default = ["gtk"]
 gtk = ["gio", "gdk", "gdk-sys", "glib", "glib-sys", "gtk-sys", "gtk-rs", "gdk-pixbuf"]
 x11 = ["x11rb", "nix", "cairo-sys-rs"]
+# Implement HasRawWindowHandle for WindowHandle 
+raw-win-handle = ["libc", "raw-window-handle"]
 
 # passing on all the image features. AVIF is not supported because it does not
 # support decoding, and that's all we use `Image` for.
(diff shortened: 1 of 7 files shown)


druid-shell/src/platform/windows/window.rs
```diff
@@ -159,6 +162,25 @@ pub struct WindowHandle {
...
+                hinstance: unsafe {
+                    winapi::um::libloaderapi::GetModuleHandleW(0 as winapi::um::winnt::LPCWSTR)
+                        as *mut libc::c_void
+                },
+                ..WindowsHandle::empty()
+            };
+            RawWindowHandle::Windows(handle)
+        } else {
+            panic!("Cannot retrieved HWMD for window.");
+        }
```
rjwittams, line 180: Isn’t it HWND

rjwittams: I’m not convinced panicking is right for any of these. I think they should return an empty window handle and maybe log a warning. Given that the raw window handle can represent empty handles.

cmyr: If we're okay with the idea that a `RawWindowHandle` can potentially be empty then I agree, otherwise we could consider making this an `Option<RawWindowHandle>`.

rjwittams: This is an external crate, that is depended on by various things like wgpu and video libraries (and some other gui/rendering libraries). The RawWindowHandles all explicitly have an empty state and consumers have to handle that.

druid-shell/Cargo.toml
```diff
@@ -51,6 +53,8 @@ keyboard-types = { version = "0.5.0", default_features = false }
 
 # Optional dependencies
 image = { version = "0.23.12", optional = true, default_features = false }
+raw-window-handle = { version = "0.3.3", optional = true, default_features = false }
+libc = { version = "0.2.81", optional = true, default_features = false }
```
rjwittams, line 57: https://github.com/rust-windowing/raw-window-handle/commit/73a41ea28907b0f261816136b1b3f39cfdd791f3

Looks like the dependency on libc was removed a while back?

rjwittams: Hm, no cargo release since 2019.

rjwittams: https://github.com/rust-windowing/raw-window-handle/pull/67 getting ready for one though.

druid-shell/src/platform/windows/window.rs
djeedai: Right, typo.

druid-shell/Cargo.toml
djeedai: Right, but even 3.3 doesn't need `libc`. Removed.

druid-shell/src/platform/windows/window.rs
djeedai: I think a `panic!` is justified because if we have a `WindowHandle` without a native `HWND` I'm assuming we're in serious trouble. Or is there a valid state where that happens?

cmyr: It might be possible when the window is in the process of being set up or torn down? Elsewhere in the windows backend we log when we can't get the `HWND`, so I might just keep doing that. 🤷

rjwittams: I think it is better to log. These handles are meant to represent empty states and users of them are expected to handle those states (its better to have an error on screen than a program crash!)

cmyr:
@rjwittams are you satisfied? feel free to approve & merge if so. :)

cmyr (changes requested):
Okay happy to get this in with the `panic` replaced by a `tracing::error!()` message.

Is there a reason we don't want to add this for x11 & gtk as well?

rjwittams:
I am happy too modulo the panics. 
I think that at a minimum the other shell platforms should return the empty version and log that it is unimplemented.
This is the pattern we have for other shell features.

cmyr:
I *could* imagine this just not being implemented on those platforms, since it's defined in a trait, and if you're using a raw window handle I would expect you to be compiling or `cfg`ing to a specific platform? But curious to hear if this was a decision or an omission.

djeedai:
> curious to hear if this was a decision or an omission.

This was discussed. I have no access to non-Windows platforms and little knowledge about them, so rather than doing a blind implementation likely to be buggy, implementations on those missing platforms is left for a future change by whoever can contribute it. I did Windows and @rjwittams did macOS.

rjwittams:
Looks good. Could you get it up to date with master (should be a quick merge or rebase, whichever you prefer as we'll squash it anyway), as I'm not sure if the log vs tracing changes may confound Githubs 'this is ok to merge' detection...

cmyr (approved):
Okay I'm also happy once CI is.

Let's open an issue about the missing impl for x11.

PR #1591, opened 2021-02-15
Add type parameters to generated lenses of generic structs.

rjwittams:
This gives type inference something to grab hold of when chaining
LensExt methods or using WidgetExt::lens.
It removes the need for providing annoying and hard to work out type arguments to some method calls (removed one in Tabs).

This changes generated lenses from unit structs to field structs to hold PhantomDatas, so construction is more involved. It also adds generic parameters to these structs, so their types are changed. 
A new method wasn't generated and used for the const fields in the targets impl, as lifetime parameters on const functions are unstable.

### DIFF
--- a/druid-derive/src/lens.rs
+++ b/druid-derive/src/lens.rs
@@ -62,6 +62,23 @@ fn derive_struct(input: &syn::DeriveInput) -> Result<proc_macro2::TokenStream, s
             "Lens implementations can only be derived from CamelCase types",
         ));
     };
+    let (impl_generics, ty_generics, where_clause) = input.generics.split_for_impl();
+
+    let mut lens_ty_idents = Vec::new();
+    let mut phantom_decls = Vec::new();
+    let mut phantom_inits = Vec::new();
+
+    for gp in input.generics.params.iter() {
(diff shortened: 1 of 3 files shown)


rjwittams:
I realised I could use tuple structs to avoid making up idents for PhantomDatas, and I am skipping lifetime parameters for these anyway as I don't think they are needed - so have added and used the const fn new .

It would be possible to make the generated structs identical to their current form if they have no args. 
I doubt that anyone is relying on this, and it would make this code a bit more complex (a few more conditionals etc).

cmyr (approved):
> It would be possible to make the generated structs identical to their current form if they have no args.
> I doubt that anyone is relying on this, and it would make this code a bit more complex (a few more conditionals etc).

Not sure I totally understand, do you mean possible to keep them unit structs?

iirc I have maybe once needed to go find the actual type of a generated lens, but I don't think we should expect anyone to ever use these directly; I think if they need to it's because we've done something wrong.

In any case I think this looks good and definitely improves ergonomics.

rjwittams:
Yeah the only case it would matter is if someone is referencing the type directly as a value. Which seems unlikely, as we provide the const fields for that purpose.

Issue #1589, opened 2021-02-13
unimplemented function


Zenthae:
Can't use `WindowSizePolicy::Content` on windows until https://github.com/linebender/druid/blob/master/druid-shell/src/platform/windows/window.rs#L1661-L1664 is implemented

HoNile:
I think it's better to continue here than in the merge PR.

Reminder:

> 
> 
> > @Zenthae did you have an example to reproduce this ? Myself I test using WindowSizePolicy::Content, with no size set on the tree example from nursery and it work fine.
> 
> it's the beginning of a new app i'm doing. i used this as a test
> 
> ```rust
> use druid::{
>     widget::{Label, Padding},
>     AppLauncher, Data, Lens, Widget, WindowDesc, WindowSizePolicy,
> };
> 
> #[derive(Debug, Default, Data, Clone, Lens)]
> struct AppState {}
> 
> fn ui_builder() -> impl Widget<AppState> {
>     Padding::new(5., Label::new("Hello Bluetooth monitor"))
> }
> 
> fn main() -> windows::Result<()> {
>     let main_window = WindowDesc::new(ui_builder())
>         .title("Bluetooth Monitor")
>         // .window_size((100.0, 40.0))
>         .window_size_policy(WindowSizePolicy::Content);
> 
>     AppLauncher::with_window(main_window)
>         .use_simple_logger()
>         .launch(AppState::default())
>         .expect("Failed to start the app");
> 
>     Ok(())
> }
> ```

So it looks like that if the window has no initial size, there is an issue, but I don't think `content_insets(&self)` should  do more than what it does I think this should be solved elsewhere. Still an easy workaround is to just set a temporary window_size.

Ex:

```
    WindowDesc::new(Padding::new(5., Label::new("Hello Bluetooth monitor")))
        .window_size((1., 1.))
        .window_size_policy(WindowSizePolicy::Content)
        .resizable(false)
```

This I think did what you want.

Zenthae:
indeed, i just find it "sad" to have to define a size even when it should take the Content size by default.

HoNile:
I think an okay solution could be to change `window_size_policy` on windows, to set the size to (1., 1.)  `if size_policy == WindowSizePolicy::Content`. (@cmyr Is it okay for you ?)

Or just handle the thing in the windows druid-shell but set the size is a `DeferredOp` so I don't know how this could be done.

I know @Ciantic want to do some refactoring on the windows druid-shell he may have a idea.

[...]

PR #1592, opened 2021-02-16
Shell: windows implementation from content_insets

HoNile:
With this dropdown should be better on windows.

There is still an issue with HiDPI screen, it doesn't look like using the scale factor in this function helps.

### DIFF
--- a/druid-shell/src/platform/windows/window.rs
+++ b/druid-shell/src/platform/windows/window.rs
@@ -1659,7 +1659,32 @@ impl WindowHandle {
     }
 
     pub fn content_insets(&self) -> Insets {
-        warn!("WindowHandle::content_insets unimplemented for windows.");
+        if let Some(w) = self.state.upgrade() {
+            let hwnd = w.hwnd.get();
+            unsafe {
+                let mut info: WINDOWINFO = mem::zeroed();
+                info.cbSize = mem::size_of::<WINDOWINFO>() as u32;
+
+                if GetWindowInfo(hwnd, &mut info) == 0 {
+                    warn!(
+                        "failed to get window info: {}",
+                        Error::Hr(HRESULT_FROM_WIN32(GetLastError()))
+                    );
+                };
+
+                let window_frame = Rect::from_points(
+                    (info.rcWindow.left as f64, info.rcWindow.top as f64),
+                    (info.rcWindow.right as f64, info.rcWindow.bottom as f64),
+                );
+                let content_frame = Rect::from_points(
+                    (info.rcClient.left as f64, info.rcClient.top as f64),
+                    (info.rcClient.right as f64, info.rcClient.bottom as f64),
+                );
+
+                return window_frame - content_frame;
+            }
+        }
+
         Insets::ZERO
     }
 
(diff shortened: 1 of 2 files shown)


Zenthae:
longing for it ^^, #1589

cmyr:
I won't be able to test this until tomorrow afternoon at the earliest. If anyone else on windows is able to do a review I'm happy to delegate approval to them :)

Zenthae:
i don't know if i should post it here but, `WindowSizePolicy::Content` behave strangely when no size is set, it start the window with a 0, 0 size and you need to resize it manualy to make it take the policy into account, and well, it wouldn't work at all if resize is set to false

cmyr:
@rjwittams any thoughts on this size business?

rjwittams:
I don't know off the top of my head. I do think that without the content offsets we could end up iteratively shrinking the window. I don't have windows currently to test with.

HoNile:
@Zenthae did you have an example to reproduce this ? Myself I test using WindowSizePolicy::Content, with no size set on the tree example from nursery and it work fine.

cmyr:
Okay I'm in front of a computer where I could test this, if anyone has any suggestions for what I should be testing?

HoNile:
During my tests I checked the dropdown example from nursery, previously the sub_windows didn't pop where there should (and it's still the case if you use HiDPI monitor). And I tested WindowSizePolicy::Content on tree as explain, I don't know if there is better tests to do.

druid-shell/src/platform/windows/window.rs
```diff
@@ -1659,7 +1659,51 @@ impl WindowHandle {
...
+                        bottom: 0,
+                    },
+                    dwStyle: 0,
+                    dwExStyle: 0,
+                    dwWindowStatus: 0,
+                    cxWindowBorders: 0,
+                    cyWindowBorders: 0,
+                    atomWindowType: 0,
+                    wCreatorVersion: 0,
+                };
```
cmyr, line 1686: You could write this all as,

```rust
let mut info: WINDOWINFO = mem::zeroed();
info.cbSize = mem::size_of::<WINDOWINFO>() as u32;
```

cmyr (approved):
This looks correct to me; `content_insets` is supposed to return a pixel value, so it's possible that elsewhere there is something that is working in logical units and needs to be converted to pixels?

This is actually a bit iffy, though, I would expect window sizes to be in logical units, and not be display dependant, but this does at least do what the docs say it does.

Thanks!

druid-shell/src/platform/windows/window.rs
HoNile: Yes indeed that's better ! I've done the changes.

cmyr:
Thanks!

Zenthae:
> @Zenthae did you have an example to reproduce this ? Myself I test using WindowSizePolicy::Content, with no size set on the tree example from nursery and it work fine.

it's the beginning of a new app i'm doing. i used this as a test
```rust
use druid::{
    widget::{Label, Padding},
    AppLauncher, Data, Lens, Widget, WindowDesc, WindowSizePolicy,
};

#[derive(Debug, Default, Data, Clone, Lens)]
struct AppState {}

fn ui_builder() -> impl Widget<AppState> {
    Padding::new(5., Label::new("Hello Bluetooth monitor"))
}

fn main() -> windows::Result<()> {
    let main_window = WindowDesc::new(ui_builder())
        .title("Bluetooth Monitor")
        // .window_size((100.0, 40.0))
        .window_size_policy(WindowSizePolicy::Content);

        .use_simple_logger()
        .launch(AppState::default())
        .expect("Failed to start the app");

    Ok(())
}
```

Issue #1593, opened 2021-02-16
Issues with invalidation when multiple windows are open


cmyr:
*This issue was me attempting to talk through something I'm still confused by*

This is a strange issue that I do not yet fully understand. I ran into it while testing some recent runebender work on my laptop (running macOS 10.15) and I did not encounter it on my desktop, running macOS 10.14. I haven't had the chance to test elsewhere.

The problem itself is fairly simple: with multiple windows open, certain views are failing to redraw correctly; they either do not redraw at all, or they redraw partially, leaving visual artifacts:

https://user-images.githubusercontent.com/3330916/108091795-0c3ac100-704a-11eb-8f12-272ef7852a7b.mov

There is nothing *obviously* wrong; the correct paint methods are being called, for instance.

(*edit*: after more experimenting, it looks like invalidation is behaving differently when there are multiple windows open?)

After a bunch of experimenting, I've tracked this down to being some consequence of moving `AnimFrame` to be an event, and how that interacts with the normal event/update cycle. Basically: when a window receives `prepare_paint`, it dispatches (if needed) the `AnimFrame` event, and then (as with all events) does an update pass (essentially does the `Widget::update` method) with *all* windows.

There are two things I've tried that at least make the symptoms go away: the first thing is to only do `update` on the current window after doing `prepare_paint`, and the second is to move to doing our `layout` pass during `prepare_paint`, instead of during `paint`.

My general understanding of the problem is that `update` is being called on windows at an unexpected time, and this is causing weird side-effects. I think there are multiple bugs being triggered here, and in trying to track this down I've come to suspect that the current mechanism for handling animation is broken.

Some things that feel wrong:

- by sending `update` to other windows after `prepare_paint`, those windows may request additional animation frames, because they only clear their own `anim_frame` request flag in their *own* prepare_paint methods.
- we call `request_anim_frame` in two places: in `post_event_processing` (which is itself called after `layout`, `lifecycle`, and `update`) as well as at the end of `paint`.

Looking through `window.rs`, there are a number of other things that seem a bit iffy; it's odd that we do `post_event_processing` in `lifecycle`, for instance.

Okay, I still don't know exactly what's going on. I've thought I did, several different times. I think there are various moving parts.

My best guess, so far, is that the extra `update` call means that `invalidate_and_finalize` is always invalidating everything (because we haven't done layout yet, and `needs_layout`) is true) which means that we aren't correctly invalidating *regions*, and then *hand wave* somehow we aren't repainting everything?

[...]

PR #1594, opened 2021-02-16
Always clear invalid regions if performing layout

cmyr:
It has taken me a large number of hours to decide on this particular
one-line fix, and I'm still not 100% sure what's going on.

That said, the logic now feels more correct: if we're invalidating
everything then that necessarily means we're invalidating whatever
rects were in the list, and we can ignore them.

More concretely, I think this was a bug; by leaving the invalid
rects in place, a subsequent call to `invalidate_and_finalize` would
see the left over rects and invalidate again, unnecessarily.

I do *not* know why this bug caused (or was a factor in) #1593.

My best guess is that there was a change to how macOS handles
setNeedsDisplayInRect: when it is called from viewWillDraw:; my
guess is that if you call this method after having previously called
setNeedsDisplay: it will assume that the new rect is more accurate
than the previous blanket invalidation, and will only redraw the
new rect.

I'm not convinced of this explanation, though. In particular,
it looks like our drawRect call is still being passed the expected
rect, the pixels just end up not the ones we expect.

Okay, here's my *revised* theory; the additional invalidation
is triggering a second paint call before the frame deadline, and
so the initial paint work, which covered the entire window, is
discarded; the second paint call does not update the regions
that weren't explicitly invalidated, and so we end up failing
to repaint portions of the view.

I think that might be it.

---- 
review:

I think the actual manifestation of the bug that motivated this change is os dependent; I've only seen it on macOS 10.15.x. I think it should be possible to bench check the logic, though, and some println debugging shows, at least in my case, that I'm now doing one less paint call.

### DIFF
--- a/druid/src/window.rs
+++ b/druid/src/window.rs
@@ -318,8 +318,8 @@ impl<T: Data> Window<T> {
             for rect in self.invalid.rects() {
                 self.handle.invalidate_rect(*rect);
             }
-            self.invalid.clear();
         }
+        self.invalid.clear();
     }
 
     #[cfg(test)]


jneem (approved):
This change looks clearly correct to me.

I've worried about your comment, though, because I was working under the assumption that extra invalidations are always fine. Like, in general can't we call `invalidate_and_finalize` multiple times per frame? What if it gets called once with a bunch of invalid rects, and then gets called a second time with a layout change? Wouldn't that give the same behavior that you're trying to avoid here?

cmyr:
> I've worried about your comment, though, because I was working under the assumption that extra invalidations are always fine. Like, in general can't we call `invalidate_and_finalize` multiple times per frame? What if it gets called once with a bunch of invalid rects, and then gets called a second time with a layout change? Wouldn't that give the same behavior that you're trying to avoid here?

Honestly, I'm not sure.

I do think that your expectations are right, and you should be able to handle an arbitrary number of events, (and do an arbitrary amount of invalidation) between frames. 

I think the particular problem I'm dealing with is where damage regions persist in druid between paint calls, or otherwise causing two paint calls to happen before the frame deadling.

That said, it may be that we have not clearly defined, at the `druid-shell` level, what the expections of `invalidate` and `invalidate_rect` are, and it's possible that the platform might change how these behave, or that their behaviour may be stateful based on things internal to the system runloop.

I think this problem is basically not solvable with the given architecture.

The long-term solution, I think, is to always use some sort of system display-timing/vsync function to determine when we should paint each frame, and then at that point, all at once, go and determine the damage regions and do the invalidation.

Whether or not this is possible on each platform is unclear to me, but it should be the goal. Ideally there would be some sort of display-sync API that would give us the *opportunity* to paint, but wouldn't *require* us to paint. I'm not sure if this exists or not.

I also think that we should look at rethinking animation flow; in particular `invalidate_and_finalize` currently has two distinct roles: it is responsible for determining *if we should paint at all* (when called after update, after some 'normal' event) and then it is also responsible for specifying damage regions for animation, which occurs after paint has already been requested; I think that ultimately I would prefer to find some more principled design, where animation has its own code path.


In any case, I still don't feel like I have a great sense of what's going on here, but these are my current thoughts. 🤷

jneem:
IIRC the X11 backend currently works like you describe, meaning that it should be possible to support within the current druid <-> druid-shell interface. Basically, all that `WindowHandle::invalidate_rect` does is internally stash the invalid rect and then call `WindowHandle::request_anim_frame`. Then at vsync time the invalid rects are collected and communicated to X11. Could the mac backend do something similar?

I do agree that there are some simplifications to be had. Note, though, that `invalidate_and_finalize`'s two roles aren't exactly orthogonal, because having a damage region implies that we should paint.

cmyr:
> IIRC the X11 backend currently works like you describe, meaning that it should be possible to support within the current druid <-> druid-shell interface. Basically, all that `WindowHandle::invalidate_rect` does is internally stash the invalid rect and then call `WindowHandle::request_anim_frame`. Then at vsync time the invalid rects are collected and communicated to X11. Could the mac backend do something similar?

This is what I'm unsure of. Mac has the [`CVDisplayLink`](https://developer.apple.com/documentation/corevideo/cvdisplaylink?language=objc#) API, which is what we'd want to use for this; I'm just not sure exactly how it interacts with invalidation. I'm sure this wouldn't be too hard to figure out.


> I do agree that there are some simplifications to be had. Note, though, that `invalidate_and_finalize`'s two roles aren't exactly orthogonal, because having a damage region implies that we should paint.

Totally; it's this sort-of-almost-alignment that I find a little scary. For instance: there should never be a reason to call `handle.invalidate()` in the case where we are being called from `prepare_paint`; we know a paint call is imminent. Similarly, exactly what we should expect from `invalidate_rect` is not clear: should any calls to this made during `prepare_paint` be unioned with the existing invalid region, or supersede it? I would certainly expect the first, but we don't actually control this implementation, necessarily, and I don't *think* we have the expectations written down.

jneem:
Oh, I hadn't realized but you're exactly right: during `prepare_paint` we do want to invalidate rects without causing another paint call. So it seems like the druid-shell API should have `invalidate_rect` (which adds an invalid rect but doesn't cause a repaint) and `request_paint` (which requests a new paint call, even if it's called during `prepare_paint` -- because that's how requesting a new anim frame from the current frame will be implemented).

cmyr:
> Oh, I hadn't realized but you're exactly right: during `prepare_paint` we do want to invalidate rects without causing another paint call. So it seems like the druid-shell API should have `invalidate_rect` (which adds an invalid rect but doesn't cause a repaint) and `request_paint` (which requests a new paint call, even if it's called during `prepare_paint` -- because that's how requesting a new anim frame from the current frame will be implemented).

Yea, so I think the solution is for us to just take total control of the whole invalidation process; currently (at least on macOS) `invalidate_rect` maps directly to `NSView.needsDisplayInRect:`, which, when called while handling `viewWillDraw:` is not *supposed* to cause a new paint call, but this stuff is fiddily, and is going to be easy to break in the future.

jpochyla:
Hi! This is very similar to the issues I had in #1470. I managed to get CVDisplayLink working, and It's _almost_ correct, but with multiple windows the invalidation completely falls apart, relevant to #1593.

cmyr:
@jpochyla thanks for the reminder, I didn't follow that work carefully but it sounds like it really might be relevant. :)

PR #1596, opened 2021-02-17
write! for RichTextBuilder

Maan2003:


druid/src/text/rich_text.rs
```diff
@@ -131,6 +135,17 @@ impl RichTextBuilder {
         self.add_attributes_for_range(range)
     }
 
+    /// Glue for usage of the write! macro.
+    ///
+    /// This method should generally not be invoked manually, but rather through the write! macro itself.
+    pub fn write_fmt(&mut self, fmt: std::fmt::Arguments<'_>) -> AttributesAdder {
```
cmyr, line 141: Maybe add `#[doc(hidden)]` on this, since it's really an implementation detail and doesn't need to show up in public docs?

Hmm, should we be implementing the `Write` trait directly? Is this depending on an implementation detail of the macro, where it calls `write_fmt` with no qualifiers, instead of doing like `Write::write_fmt(writer, args)`?

cmyr (approved):
Interesting. I trust you've found this useful for something and it's not very intrusive, so I'm happy to merge. A few little questions inline, but nothing blocking.

druid/src/text/rich_text.rs
Maan2003: I didn't implement `Write` because it returns a Result, I wanted it to feel more like `format!`.

> Is this depending on an implementation detail of the macro?

the [docs](https://doc.rust-lang.org/stable/std/macro.write.html) say it just needs something with `write_fmt` method.

> The writer may be any value with a write_fmt method; generally this comes from an implementation of either the fmt::Write or the io::Write trait.

Maan2003:
> you've found this useful for something and it's not very intrusive.

I found myself doing a lot of `builder.push(&format!(....))`. The first thought was to make `push` take `impl Display`.
which will be make it `builder.push(format_args!(...))`. `write!` is just a cleaner interface.

cmyr:
It's possible this also saves us an intermediate allocation, although I'm not sure about this part.

PR #1597, opened 2021-02-17
Resolve ISO_Left_Tab key to tab

### LINKED ISSUE #1595: Focus previous doesnt work on Linux
If i press Tab inside a Textbox it focuses the next one as expected, but shift + tab does nothing.

The `KeyEvent` in `Event::KeyDown` for Tab:
```
[druid/examples/focus.rs:20] ke = KeyEvent {
    state: Down,
    key: Tab,
    code: Tab,
    location: Standard,
    mods: Modifiers(
        (empty),
    ),
    repeat: false,
    is_composing: false,
}
```

The `KeyEvent` in `Event::KeyDown` for Shift + Tab:
```
[druid/examples/focus.rs:20] ke = KeyEvent {
    state: Down,
    key: Unidentified,
    code: Tab,
    location: Standard,
    mods: Modifiers(
        SHIFT,
    ),
    repeat: false,
    is_composing: false,
}
```

I think the problem is that `HotKey::matches` uses the `key` field of `KeyEvent`, which is Unidentified if Shift is pressed.

cmyr:
This is sent by gtk when tab is pressed with shift down.

- closes #1595

@xarvic, this should fix your problem, you're welcome to double-check though!

### DIFF
--- a/druid-shell/src/platform/gtk/keycodes.rs
+++ b/druid-shell/src/platform/gtk/keycodes.rs
@@ -26,7 +26,7 @@ pub fn raw_key_to_key(raw: RawKey) -> Option<Key> {
     Some(match raw {
         Escape => Key::Escape,
         BackSpace => Key::Backspace,
-        Tab => Key::Tab,
+        Tab | ISO_Left_Tab => Key::Tab,
         Return => Key::Enter,
         Control_L | Control_R => Key::Control,
         Alt_L | Alt_R => Key::Alt,


raphlinus (approved):
Looks good. Doing a bit of quick search, I think this change in mapping is more recent. In any case, thanks!

xarvic:
> @xarvic, this should fix your problem, you're welcome to double-check though!

I have tested it. It works now, thanks :)

PR #1600, opened 2021-02-18
Add scroll() method in WidgetExt

### LINKED ISSUE #1544: scroll method in WidgetExt
From my limited experience, I sometimes wished I could just add a `.scroll` to the end of a widget definition. Wrapping a widget in a scroll isn't too annoying but it does, at least to me, make things a bit more difficult to read. The function definition would look like this

```rust
// I don't know if this definition is the correct one, specifically for the return type
pub fn scroll(self, direction: ScrollDirection) -> Scroll<T, Self> {
 // within here there can be some matching on the `ScrollDirection` which would 
// create the Scroll appropriately
}
pub enum ScrollDirection {
    Horizontal,
    Vertical,
    Both,
```
Would this sort of change be welcome in `WidgetExt`?

totsteps:
closes #1544

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ You can find its changes [documented below](#070---2021-01-01).
 ### Highlights
 
 ### Added
+- Add `scroll()` method in WidgetExt ([#1600] by [@totsteps])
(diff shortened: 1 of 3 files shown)


cmyr (approved):
Looks good!

PR #1602, opened 2021-02-18
Don't deliver InternalLifecycle::ParentWindowOrigin to hidden widgets

rjwittams:
… (e.g in Either). Because they aren't laid out, the needs_layout flag of their widget pod gets merged up and eventually causes over invalidation of the window handle. On macOS 10.15, this is inexplicably segfault worthy.

### DIFF
--- a/druid/src/event.rs
+++ b/druid/src/event.rs
@@ -394,8 +394,25 @@ impl LifeCycle {
     /// (for example the hidden tabs in a tabs widget).
     pub fn should_propagate_to_hidden(&self) -> bool {
         match self {
-            LifeCycle::WidgetAdded | LifeCycle::Internal(_) => true,
+            LifeCycle::Internal(internal) => internal.should_propagate_to_hidden(),
+            LifeCycle::WidgetAdded => true,
             LifeCycle::Size(_) | LifeCycle::HotChanged(_) | LifeCycle::FocusChanged(_) => false,
+        }
+    }
+}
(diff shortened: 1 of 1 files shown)


druid/src/window.rs
```diff
@@ -402,14 +402,14 @@ impl<T: Data> Window<T> {
...
         self.root
             .set_origin(&mut layout_ctx, data, env, Point::ORIGIN);
+        // This lifecycle call will call post_event_processing for us.
         self.lifecycle(
             queue,
             &LifeCycle::Internal(InternalLifeCycle::ParentWindowOrigin),
             data,
             env,
-            false,
+            true,
```
cmyr, line 411: is this related to the fix, or is this just you doing some cleanup while you're in here?

rjwittams: It was cleanup because it was calling post_event_processing twice (once inside the lifecycle, once outside). 
I don't think it is necessary for the fix, I just noticed it.

cmyr: @rjwittams can we revert this, and have this patch only include the fix? I'd like to have as few moving parts as possible, so we can be confident about the root cause.

I will open an issue about improving post-event stuff.

rjwittams: Done

cmyr (approved):
Thanks for tracking this down, I'm not seeing the crash with this patch applied.

PR #1606, opened 2021-02-20
Use correct fill rule when rendering SVG paths

SecondFlight:
SVG paths can specify a fill rule, either `"evenodd"` or `"nonzero"`. This PR updates the SVG widget so it respects this property.

Before:
image

After:
image


Maan2003 (approved):
Looks Good. Thanks!

PR #1611, opened 2021-02-20
Convenience traits for code generic over contexts.

rjwittams:
Widgets sometimes have code they wish to call from one or more Widget methods, e.g calling commands, or doing similar things in LifeCycle::WidgetAdded and update(). Currently its required to pass lambdas or wrap various contexts up in order to achieve this. 

This allows abstraction over contexts according to the sets of methods they provide.
It does not remove or remove any inherent methods, just forwards to them from the trait impls. Code that does not need these facilities does not need to import the traits and should remain unaware of them.

Open questions:
- Naming of the traits - LaidOutCtx seems a bit forced... LayoutInfoCtx? RequestCtx will probably make people think of http frameworks. Open to discussion and bikeshedding. Also they could be public in a module rather than the root, e.g druid::context_traits.
- It would probably be possible to have a lot less duplication/risk of things coming out of sync at the expense of (potentially extreme) macro complexity: it would probably mean going a proc macro route to be comprehensible. I didn't want to go down that route if unwarranted.
- There are a bunch of different schemes that could be done here (in terms of where methods live, which direction they are forwarded etc.) This was the one that @cmyr thought (several months back) would be least disruptive and leads to still discoverable docs (via the inherent trait methods). 
-  I avoided copying swathes of docs and just copied the first line and referenced the inherent impl on EventCtx.

### DIFF
--- a/druid/src/contexts.rs
+++ b/druid/src/contexts.rs
@@ -45,6 +45,16 @@ macro_rules! impl_context_method {
     };
 }
 
+macro_rules! impl_context_trait{
+    ($tr:ty => $ty:ty,  { $($method:item)+ } ) => {
+        impl $tr for $ty { $($method)+ }
+    };
+    ($tr:ty => $ty:ty, $($more:ty),+, { $($method:item)+ } ) => {
+        impl_context_trait!($tr => $ty, { $($method)+ });
+        impl_context_trait!($tr => $($more),+, { $($method)+ });
+    };
+}
+
 /// Static state that is shared between most contexts.
 pub(crate) struct ContextState<'a> {
     pub(crate) command_queue: &'a mut CommandQueue,
@@ -166,6 +176,42 @@ impl_context_method!(
     }
 );
 
+/// Convenience trait for code generic over contexts. Methods provided by all contexts.
+pub trait AnyCtx {
+    /// get the `WidgetId` of the current widget.
+    fn widget_id(&self) -> WidgetId;
+
+    /// Returns a reference to the current `WindowHandle`.
+    fn window(&self) -> &WindowHandle;
+
+    /// Get the `WindowId` of the current window.
+    fn window_id(&self) -> WindowId;
+
+    /// Get an object which can create text layouts.
+    fn text(&mut self) -> &mut PietText;
+}
+
+impl_context_trait!(
+   AnyCtx => EventCtx<'_, '_>, UpdateCtx<'_, '_>, LifeCycleCtx<'_, '_>,PaintCtx<'_, '_, '_>, LayoutCtx<'_, '_>,
+   {
+        fn widget_id(&self) -> WidgetId {
+            Self::widget_id(self)
+        }
+
+        fn window(&self) -> &WindowHandle {
+            Self::window(self)
+        }
+
+        fn window_id(&self) -> WindowId {
+            Self::window_id(self)
+        }
+
+        fn text(&mut self) -> &mut PietText {
+            Self::text(self)
+        }
+   }
+);
+
 // methods on everyone but layoutctx
 impl_context_method!(
     EventCtx<'_, '_>,
@@ -276,6 +322,79 @@ impl_context_method!(
     }
 );
 
+/// Convenience trait for code generic over contexts. Methods providing layout information: available on all but LayoutCtx.
+pub trait LaidOutCtx {
+    /// The layout size.
+    ///
+    /// ['size']: EventCtx::size
+    fn size(&self) -> Size;
+    /// The origin of the widget in window coordinates. See ['window_origin']
+    ///
+    /// [window_origin]: EventCtx::window_origin
+    fn window_origin(&self) -> Point;
+    /// Convert a point from the widget's coordinate space to the window's. See ['to_window']
+    ///
+    /// [to_window]: EventCtx::to_window
+    fn to_window(&self, widget_point: Point) -> Point;
+    /// Convert a point from the widget's coordinate space to the screen's. See ['to_screen']
+    ///
+    /// ['to_screen']: EventCtx::to_screen
+    fn to_screen(&self, widget_point: Point) -> Point;
+    /// The "hot" status of a widget. See ['is_hot']
+    ///
+    /// ['is_hot']: EventCtx::is_hot
+    fn is_hot(&self) -> bool;
+    /// The active status of a widget. See ['is_active']
+    ///
+    /// ['is_active']: EventCtx::is_active
+    fn is_active(&self) -> bool;
+    /// The focus status of a widget. See ['is_focused']
+    ///
+    /// ['is_focused']: EventCtx::is_focused
+    fn is_focused(&self) -> bool;
+    /// The focus status of a widget or any of its descendents. See ['has_focus']
+    ///
+    /// ['has_focus']: EventCtx::has_focus
+    fn has_focus(&self) -> bool;
+}
+
+impl_context_trait!(
+   LaidOutCtx => EventCtx<'_, '_>, UpdateCtx<'_, '_>, LifeCycleCtx<'_, '_>,PaintCtx<'_, '_, '_>,
+   {
+        fn size(&self) -> Size{
+            Self::size(self)
+        }
+
+        fn window_origin(&self) -> Point{
+            Self::window_origin(self)
+        }
+
+        fn to_window(&self, widget_point: Point) -> Point{
+            Self::to_window(self, widget_point)
+        }
+
+        fn to_screen(&self, widget_point: Point) -> Point{
+            Self::to_screen(self, widget_point)
+        }
+
+        fn is_hot(&self) -> bool{
+            Self::is_hot(self)
+        }
+
+        fn is_active(&self) -> bool{
+            Self::is_active(self)
+        }
+
+        fn is_focused(&self) -> bool{
+            Self::is_focused(self)
+        }
+
+        fn has_focus(&self) -> bool{
+            Self::has_focus(self)
+        }
+   }
+);
+
 impl_context_method!(EventCtx<'_, '_>, UpdateCtx<'_, '_>, {
     /// Set the cursor icon.
     ///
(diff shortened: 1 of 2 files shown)


cmyr:
Cool, thanks for jumping on this.

Looking through the diff, I really like the *functional* names, like `CommandCtx`, `CursorCtx`, more than I like the sort of... I don't know what to call the other ones, so let's jut say the *other* names.

I also think that we can maybe be a bit more focused here. There are only a handful of things I've ever really wanted to do from multiple contexts. I can think of timers, commands and text offhand; there may have been others.

For `AnyContext`, for instance: if you need a `WidgetId`, a `WindowHandle` or a `WindowId`, I think you should just have your method take that type. I might just make a `TextContext`?

Maybe in general we should take a more minimal approach, and just add traits for behaviour we've needed, knowing that we can always add more later?


Another way that I'm thinking about this, right now: methods that have return values are less important here, because you can just call them and pass their return values to wherever they're needed. The methods that we most want these sorts of traits for are methods that mutate something, like invalidation or adding a timer.

Maybe also having fewer methods will make it easier for us to keep all of these in sync?

rjwittams:
I think that the traits could be split up a bit more to have better names. I was trying to match the method impl blocks we had. Those could be split up to match to make things line up. 

The other issue is "actual" generic code. So e.g in bindings, on "Property", which represents an arbitrary aspect of a widget,  I have a method called 'append_changes' which really is 'look at this widget and work out if anything changed'. This gets called from lets say update, lifecycle and event. The caller, BindingHost, doesn't know what any of the implementations of Property (embedded into a compound binding of multiple properties) might need. E.g you can imagine generic read only bindings for Size and WidgetId, things that aren't on the widget itself, but can be grabbed out of the context, and 'magically' end up in AppData for another part of the 
app to use. I'm not sure that it will be the complete solution for bindings as I haven't tried it yet (it might need to be more involved to deal with things the contexts don't share).
 
Anyway, the case I'm talking about is one generic thing (a widget) calling another one (a property). Neither side can make any assumptions about what the other side needs other than 'here is what I've got'. 

I'm slightly worried that we end up in a nickel and diming situation, where every method on a trait needs to be individually justified (with the attendant activation energy working against change) rather than exposing shared methods as shared being the default. IE people will assume there is some deep reasoning behind which methods are exposed like this and which aren't.

cmyr:
If you have a concrete case for wanting to expose everything I'm definitely open to that, my inclination towards "less is more" is mostly just motivated by liking A) less code, B) less maintenance, and C) fewer types in the public docs.

xarvic:
Ups, that was not intentional :)

xarvic:
Is there currently progress on this PR and if not, is there some way i could help? Its a cool feature and i would like to use it :)

cmyr:
I'm happy to get this merged, we just need to fix up the broken links. I'll see if I can do that inline...

druid/src/contexts.rs
```diff
@@ -166,6 +176,42 @@ impl_context_method!(
     }
 );
 
+/// Convenience trait for code generic over contexts. Methods provided by all contexts.
```
cmyr, line 179:
```suggestion
/// Convenience trait for methods available on all contexts.
```

cmyr (approved):
Okay, I think this is worth getting merged, I've taken the liberty of cleaning up some docs, so assuming I didn't fat finger any links we can get this in when it passes CI. :)

druid/src/contexts.rs
```diff
@@ -399,6 +549,79 @@ impl_context_method!(EventCtx<'_, '_>, UpdateCtx<'_, '_>, LifeCycleCtx<'_, '_>,
...
+    ///
+    /// ['request_paint']: EventCtx::request_paint
+    fn request_paint(&mut self);
+    /// Request a [`paint`] pass for redrawing a rectangle. See ['request_paint_rect']
+    ///
+    /// ['request_paint_rect']: EventCtx::request_paint_rect
+    fn request_paint_rect(&mut self, rect: Rect);
+    /// Request a layout pass. See ['request_layout']
+    ///
+    /// ['request_layout']: EventCtx::request_layout
```
cmyr, line 566:
```suggestion
    /// [`request_layout`]: EventCtx::request_layout
```

druid/src/contexts.rs
```diff
@@ -399,6 +549,79 @@ impl_context_method!(EventCtx<'_, '_>, UpdateCtx<'_, '_>, LifeCycleCtx<'_, '_>,
...
+    fn request_layout(&mut self);
+    /// Request an animation frame. See ['request_anim_frame']
+    ///
+    /// ['request_anim_frame']: EventCtx::request_anim_frame
```
cmyr, line 570:
```suggestion
    /// [`request_anim_frame`]: EventCtx::request_anim_frame
```

druid/src/contexts.rs
```diff
@@ -399,6 +549,79 @@ impl_context_method!(EventCtx<'_, '_>, UpdateCtx<'_, '_>, LifeCycleCtx<'_, '_>,
...
+    fn request_anim_frame(&mut self);
+    /// Indicate that your children have changed. See ['children_changed']
+    ///
+    /// ['children_changed']: EventCtx::children_changed
```
cmyr, line 574:
```suggestion
    /// [`children_changed`]: EventCtx::children_changed
```

druid/src/contexts.rs
```diff
@@ -399,6 +549,79 @@ impl_context_method!(EventCtx<'_, '_>, UpdateCtx<'_, '_>, LifeCycleCtx<'_, '_>,
...
+    fn children_changed(&mut self);
+    /// Set the menu of the window containing the current widget. See ['set_menu']
+    ///
+    /// ['set_menu']: EventCtx::set_menu
+    fn set_menu<T: Any>(&mut self, menu: MenuDesc<T>);
+    /// Create a new sub-window. See ['new_sub_window']
+    ///
+    /// ['new_sub_window']: EventCtx::new_sub_window
```
cmyr, line 582:
```suggestion
    /// [`new_sub_window`]: EventCtx::new_sub_window
```

druid/src/contexts.rs
```diff
@@ -439,6 +662,43 @@ impl_context_method!(
...
 );
 
+/// Convenience trait for code generic over contexts.
+///
+/// Methods to do with commands and timers.
+/// Available to all contexts but PaintCtx.
+pub trait CommandCtx {
+    /// Submit a [`Command`] to be run after this event is handled. See ['submit_command']
+    ///
+    /// ['submit_command']: EventCtx::submit_command
```
cmyr, line 672:
```suggestion
    /// [`submit_command`]: EventCtx::submit_command
```

druid/src/contexts.rs
```diff
@@ -399,6 +549,79 @@ impl_context_method!(EventCtx<'_, '_>, UpdateCtx<'_, '_>, LifeCycleCtx<'_, '_>,
...
+    ///
+    /// ['request_anim_frame']: EventCtx::request_anim_frame
+    fn request_anim_frame(&mut self);
+    /// Indicate that your children have changed. See ['children_changed']
+    ///
+    /// ['children_changed']: EventCtx::children_changed
+    fn children_changed(&mut self);
+    /// Set the menu of the window containing the current widget. See ['set_menu']
+    ///
+    /// ['set_menu']: EventCtx::set_menu
```
cmyr, line 578:
```suggestion
    /// [`set_menu`]: EventCtx::set_menu
```

druid/src/contexts.rs
```diff
@@ -439,6 +662,43 @@ impl_context_method!(
...
+/// Methods to do with commands and timers.
+/// Available to all contexts but PaintCtx.
+pub trait CommandCtx {
+    /// Submit a [`Command`] to be run after this event is handled. See ['submit_command']
+    ///
+    /// ['submit_command']: EventCtx::submit_command
+    fn submit_command(&mut self, cmd: impl Into<Command>);
+    /// Returns an [`ExtEventSink`] for submitting commands from other threads. See ['get_external_handle']
+    ///
+    /// ['get_external_handle']: EventCtx::get_external_handle
```
cmyr, line 676:
```suggestion
    /// [`get_external_handle`]: EventCtx::get_external_handle
```

druid/src/contexts.rs
```diff
@@ -439,6 +662,43 @@ impl_context_method!(
...
+    fn get_external_handle(&self) -> ExtEventSink;
+    /// Request a timer event. See ['request_timer']
+    ///
+    /// ['request_timer']: EventCtx::request_timer
```
cmyr, line 680:
```suggestion
    /// Request a timer event. See [`request_timer`]
    ///
    /// [`request_timer`]: EventCtx::request_timer
```

cmyr (commented):
missed a few more 😓

cmyr:
Oops, turns out I can't actually do the cleanup here, because of the changes in #1625; some menu methods have been removed, and `MenuDesc` is renamed to `Menu`.

xarvic:
Since this PR is older than #1632 in `RequestCtx` `set_disabled()` and `is_disabled()` are missing

PoignardAzur:
Do we want to merge this PR?

I agree with Colin's early assessment, that this seems like over-engineering. I'm not really seeing a use-case where you'd need those traits and wouldn't already have access to inherent methods doing the same thing.

Can someone provide an example where the added indirection brings something? An example of [semantic compression](https://caseymuratori.com/blog_0015), to be specific?

If we still want to merge this, someone needs to rebase this branch. I can do it if the original author is no longer available. Otherwise, we should probably close this PR.

Maan2003:
I am in favor of keeping this in nursery for a while, see how it works out. I had many situation where this will be useful

cmyr:
@Maan2003 I'm not sure if this would work in nursery? If so you are free to move it over there; I'm going to close this PR, and robert can reopen if he wishes to pick this up at some point.

Maan2003:
my understanding: this is just implementing a bunch of traits on different contexts. nursery can implement the traits, if traits are defined in nursery itself.

PR #1612, opened 2021-02-21
More Cursors

Maan2003:
This adds pointer and busy cursors. I have only tested this on linux(gtk).

### DIFF
--- a/druid-shell/src/platform/windows/window.rs
+++ b/druid-shell/src/platform/windows/window.rs
@@ -1676,13 +1676,16 @@ unsafe fn create_window(
 
 impl Cursor {
     fn get_hcursor(&self) -> HCURSOR {
+        #[allow(deprecated)]
         let name = match self {
             Cursor::Arrow => IDC_ARROW,
             Cursor::IBeam => IDC_IBEAM,
             Cursor::Pointer => IDC_HAND,
             Cursor::Crosshair => IDC_CROSS,
-            // not correct, maybe polyfill this
-            Cursor::OpenHand => IDC_HAND,
+            Cursor::OpenHand => {
+                warn!("Cursor::OpenHand not available on windows");
+                IDC_ARROW
+            }
             Cursor::NotAllowed => IDC_NO,
             Cursor::ResizeLeftRight => IDC_SIZEWE,
             Cursor::ResizeUpDown => IDC_SIZENS,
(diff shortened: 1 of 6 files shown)


cmyr:
The cursors that are currently available represent the set of cursors that are implemented on all of our major platforms. I'm not sure how best to expose cursors that are only available on *some* platforms. I do agree that this would be nice, but I would like to be principled about it.

One option might be to have separate `PlatformCursor` enums that describe cursors that are platform specific, and then maybe have a `Cursor::Platform` variant in the main `Cursor` enum? `PlatformCursor` would only be exported if the appropriate `target_os`/`feature`, but this is sort of fragile because of gtk/x11.

what we really want is some way to do like `if `cfg(target_backend = "gtk")`, but that doesn't exist. I'm not sure what to suggest, otherwise...

Maan2003:
So, adding `Cursor::Pointer` should be Ok. Also, I think `Cursor::OpenHand` is not available on windows. 

`Cursor::Busy` is a fundamental cursor IMO :shrug: . How do browsers handle it on Mac OS?

cmyr:
I think macOS only shows the spinner when the main thread is blocked or something?

cmyr:
As far as I can tell there is no pointing hand cursor on windows: ([see docs here?](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-loadcursora))

Maan2003:
> As far as I can tell there is no pointing hand cursor on windows: ([see docs here?](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-loadcursora))

I think `IDC_HAND` is the pointing hand cursor, 
see https://docs.microsoft.com/en-us/windows/win32/menurc/about-cursors#the-mouse-and-the-cursor
> For example, when the cursor crosses over a hyperlink, the system changes the cursor from an arrow to a hand.
> image

cmyr:
Hmm okay, so there's no open hand, for dragging?

Maan2003:
yeah, no open hand :face_with_head_bandage:

Maan2003:
Ready for review. I have removed the busy cursor. Also, I am not sure about CI failure.

totsteps:
Tested on macOS and Windows 10. Works like charm.

druid-shell/src/platform/windows/window.rs
```diff
@@ -1679,7 +1679,9 @@ impl Cursor {
         let name = match self {
             Cursor::Arrow => IDC_ARROW,
             Cursor::IBeam => IDC_IBEAM,
+            Cursor::Pointer => IDC_HAND,
             Cursor::Crosshair => IDC_CROSS,
+            // not correct, maybe polyfill this
```
cmyr, line 1684: yea, can we open an issue here? Not sure what the best solution is, I'd probably prefer to deprecate the open hand from this enum. 😢

cmyr (approved):
Looks good!

PR #1613, opened 2021-02-21
Cleaning tracing deps

sjoshid:
This PR uses `tracing` crate for console logs. It:
* Removes `tracing-subscriber` as a dependency.
* Uses `log` feature of `tracing` to get console logs.

Note on how to add console logs:
* You'll need to use `tracing`'s non-span macros(`debug!()`, `info!()`, etc.). You can use span macros (`debug_span!()`, `info_span!()`, etc.) but your console log will be polluted.

Tests specific notes:
* All tests use `env_logger` instead of `simpler_logger`. This is because `test-env-log` doesnt support `simple_logger`. So to enable console logs while running tests, set env var `RUST_LOG` to a desired log level.
* `env_logger` is initialized differently than `simple_logger`. Check examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ fn main() -> Result<(), PlatformError> {
     let main_window = WindowDesc::new(ui_builder());
     let data = 0_u32;
     AppLauncher::with_window(main_window)
-        .use_env_tracing()
+        .use_simple_logger()
         .launch(data)
 }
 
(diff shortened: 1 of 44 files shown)


cmyr:
@sjoshid Is this still in progress, or can we close?

sjoshid:
This can be closed. What I had planned to do in this PR will be done in the other PR. 

Thanks for asking. 😀😀

PR #1619, opened 2021-02-23
druid-shell IME api, take two

cmyr:
This supersedes #1378; I'm going to use this branch to make some tweaks to the API, and eventually to integrate this into the druid textbox.

cc @lord, in case you'd like to follow along with some of this stuff.

The main change I'm envisioning currently is to have separate `acquire` and `release` methods for the locking, because druid needs some callback to know if it has to update the app data.

### DIFF
--- a/druid-shell/src/text.rs
+++ b/druid-shell/src/text.rs
@@ -228,7 +228,7 @@ pub trait InputHandler {
 
     /// Returns the contents of some range of the document.
     /// If `range.start` or `range.end` do not fall on a code point sequence boundary, this method may panic.
-    fn slice<'a>(&'a mut self, range: Range<usize>) -> Cow<'a, str>;
+    fn slice(&mut self, range: Range<usize>) -> Cow<str>;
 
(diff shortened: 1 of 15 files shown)


cmyr:
@lord I've pushed a few more changes here that made sense in the process of doing #1636. Happy to have you take a look if you like, or else I'll self-review and do any further cleanup from there. :)

druid-shell/src/text.rs
```diff
@@ -403,6 +403,7 @@ pub fn simulate_input<H: WinHandler + ?Sized>(
     event: KeyEvent,
 ) -> bool {
     if handler.key_down(event.clone()) {
+        eprintln!("key handled, returning early");
```
cmyr, line 406: oops

cmyr:
Okay I'm going to take a self-review on this, it's mostly the same code as #1378, which has already been reviewed.

lord:
Sorry, just saw this! It looks great to me!

Issue #746, opened 2020-03-23
Add 'disabled' state to widgets.


cmyr:
This is a way to communicate to control widgets that they should not respond to user input, and that they should paint themselves as 'disabled'. The 'not responding' bit might even be handled in the `WidgetPod`; it could just not send certain events on to the widget when this is set?

We may also want lifecycle events for `DisabledChanged`? That's sort of funky because it is missing concerns a little bit. This probably needs a bit more thought.

derekdreery:
Flutter uses "null" to indicate disabled for things like text boxes.

cmyr:
The big question to me is whether this should be something that is set in the `Env` or if it is something that is part of `WidgetState`. Env definitely has some nice properties, but it weirds up the API a bit; things like `is_focused` and `is_active` are all exposed via the `ctx`s, and it feels weird if `is_disabled` is something you get out of the `Env`?

I do think there might be some problems with putting it in `WidgetState`, though. In particular, we'd want to be cascading this *down* the tree; if a parent is disabled, then so are the children; and if the children are disabled and the parent is enabled, we want to still be disabled. This is doable, but it might be some slightly finicky book-keeping?

In any case I *do* think this should be added, and should be used in built-in widgets.

xarvic:
I'm not sure if it would be enough to set a flag in `Env`, since the widget could still get focused, but since it does not receive events, it can not resign its focus. Therefore switching focus via tab would be broken. This happend to me: [https://github.com/linebender/druid-widget-nursery/blob/master/examples/multi_value.rs]().

> I do think there might be some problems with putting it in WidgetState, though. In particular, we'd want to be cascading this down the tree; if a parent is disabled, then so are the children; and if the children are disabled and the parent is enabled, we want to still be disabled. This is doable, but it might be some slightly finicky book-keeping?

Maybe the `WidgetPod` could contain 2 flags: `widget_manually_disabled` and `parent_widget_disabled`. The `is_disabled` method on ctx would return `widget_manually_disabled | parent_widget_disabled`.

cmyr:
I agree that the env is not enough for this; 'disabled' is a special state that needs to be known to the framework.

xarvic:
I would like to try to implement this, if no one is already doing it.

cmyr:
[...]

PR #1620, opened 2021-02-23
Disable widgets

xarvic:
An implementation of a disabled state for widgets (issue: [#746 ](https://github.com/linebender/druid/issues/746)).
It adds the following:

- `set_enabled(bool)` for `EventCtx` and `UpdateCtx` which changes if the state of the current widget is "manually disabled".
- `is_enabled() -> bool` for all Context which returns false if this widget or any of its ancestors is manually disabled".
- `LifeCycle::EnabledChanged(bool)` which is called every time when `is_enabled()` would return a different value than before.

The behaviour of focusing widgets was changed. The `WidgetPod` now has field `can_auto_focus` that is used to determine if this widget should get focused as a result to `focus_pref` or `focus_next`. If the widget is focused when it gets disabled, it resigns the focus and is`nt able to request the focus until it is enabled again.

While disabled, a widget recieves only some of the event. MouseMove, MouseDown, MouseUp, Zoom and MouseWheel are blocked. But i`m not sure if this is desirable, maybe a scroll widget should still be able to scroll its content.

**Problems:**

- widgets are always enabled until the first call to update or event, therefore for the moment using EnabledIf is not a reliable safety mechanism to prevent the user from doing illegal operations. 

**Todo:**

- make widget inactive when disabled


xarvic:
This was way to complicated and didn't work as well as expected. I'm closing this Issue for now.

cmyr:
@xarvic this is a definitely a harder problem than it might seem like at first, happy to talk about it sometime if you'd like.

PR #1621, opened 2021-02-24
Add some tracing and logging

PoignardAzur:
Copy-pasting from Zulip:

I'm still using `#[instrument]` because, as I've said earlier, this is far from the biggest compile bottleneck right now; (I'm going to do some tests again to measure the impact it has). Also, while I don't like the syntax instrument has right now (having to add skip(...) everywhere), apparently it's due to be changed in the next version of tracing.

Right now I'm trying to address the use-case that made me want to add more logging: being able to tell when an internal event (warning, something being redrawn, etc) happens and why.

Part of that is adding spans everywhere; since druid, like basically every GUI, uses a variation on the visitor pattern, that means most spans are going be named after the Self type of the function they instrument; except for the root span. For instance, if you have Window::event calling Flex::event calling Button::event, the context chain is going to look like event:Flex:Button.

I feel like I should document that last part somewhere.

### DIFF
--- a/druid/src/app.rs
+++ b/druid/src/app.rs
@@ -183,7 +183,7 @@ impl<T: Data> AppLauncher<T> {
             let filter_layer = tracing_subscriber::filter::LevelFilter::DEBUG;
             let fmt_layer = tracing_subscriber::fmt::layer()
                 // Display target (eg "my_crate::some_mod::submod") with logs
-                .with_target(false);
+                .with_target(true);
 
             tracing_subscriber::registry()
                 .with(filter_layer)
--- a/druid/src/app.rs
+++ b/druid/src/app.rs
@@ -162,16 +162,9 @@ impl<T: Data> AppLauncher<T> {
     /// # Panics
     ///
     /// Panics if the logger fails to initialize.
-    #[deprecated(since = "0.7.0", note = "Use use_env_tracing instead")]
+    #[deprecated(since = "0.7.0", note = "Use log_to_console instead")]
     pub fn use_simple_logger(self) -> Self {
-        #[cfg(not(target_arch = "wasm32"))]
-        simple_logger::SimpleLogger::new()
-            .with_level(log::LevelFilter::Debug)
-            .init()
-            .expect("Failed to initialize logger.");
-        #[cfg(target_arch = "wasm32")]
-        console_log::init_with_level(log::Level::Debug).expect("Failed to initialize logger.");
-        self
+        self.log_to_console()
     }
 
     /// Initialize a minimal tracing subscriber with DEBUG max level for printing logs out to
@@ -183,7 +176,7 @@ impl<T: Data> AppLauncher<T> {
     /// # Panics
     ///
     /// Panics if the subscriber fails to initialize.
-    pub fn use_env_tracing(self) -> Self {
+    pub fn log_to_console(self) -> Self {
         #[cfg(not(target_arch = "wasm32"))]
         {
             use tracing_subscriber::prelude::*;
--- a/druid/src/app.rs
+++ b/druid/src/app.rs
@@ -187,10 +187,10 @@ impl<T: Data> AppLauncher<T> {
         #[cfg(not(target_arch = "wasm32"))]
         {
             use tracing_subscriber::prelude::*;
-            let fmt_layer = tracing_subscriber::fmt::layer().with_target(true);
-            let filter_layer = tracing_subscriber::EnvFilter::try_from_default_env()
-                .or_else(|_| tracing_subscriber::EnvFilter::try_new("debug"))
-                .expect("Failed to initialize tracing subscriber");
+            let filter_layer = tracing_subscriber::filter::LevelFilter::DEBUG;
+            let fmt_layer = tracing_subscriber::fmt::layer()
+                // Display target (eg "my_crate::some_mod::submod") with logs
+                .with_target(false);
 
             tracing_subscriber::registry()
                 .with(filter_layer)
--- a/druid/src/app.rs
+++ b/druid/src/app.rs
@@ -168,7 +168,7 @@ impl<T: Data> AppLauncher<T> {
     }
 
     /// Initialize a minimal tracing subscriber with DEBUG max level for printing logs out to
-    /// stderr, controlled by ENV variables.
+    /// stderr.
     ///
     /// This is meant for quick-and-dirty debugging. If you want more serious trace handling,
     /// it's probably better to implement it yourself.
@@ -191,7 +191,7 @@ impl<T: Data> AppLauncher<T> {
                 .init();
         }
         // Note - tracing-wasm might not work in headless Node.js. Probably doesn't matter anyway,
-        // because wasm targets will virtually always be browsers.
+        // because this is a GUI framework, so wasm targets will virtually always be browsers.
         #[cfg(target_arch = "wasm32")]
         {
             console_error_panic_hook::set_once();
--- a/druid/src/contexts.rs
+++ b/druid/src/contexts.rs
@@ -20,6 +20,7 @@ use std::{
     ops::{Deref, DerefMut},
     time::Duration,
 };
+use tracing::{error, trace, warn};
 
 use crate::core::{CommandQueue, CursorChange, FocusChange, WidgetState};
 use crate::env::KeyLike;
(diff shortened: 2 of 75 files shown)


druid/src/contexts.rs
```diff
@@ -311,6 +314,7 @@ impl_context_method!(EventCtx<'_, '_>, UpdateCtx<'_, '_>, {
     /// [`override_cursor`]: EventCtx::override_cursor
     /// [`set_cursor`]: EventCtx::set_cursor
     pub fn clear_cursor(&mut self) {
+        trace!("override_cursor");
```
totsteps, line 317: `trace!("clear_cursor");`?

totsteps (changes requested):

druid/src/app.rs
```diff
@@ -187,10 +187,10 @@ impl<T: Data> AppLauncher<T> {
         #[cfg(not(target_arch = "wasm32"))]
         {
             use tracing_subscriber::prelude::*;
-            let fmt_layer = tracing_subscriber::fmt::layer().with_target(true);
-            let filter_layer = tracing_subscriber::EnvFilter::try_from_default_env()
-                .or_else(|_| tracing_subscriber::EnvFilter::try_new("debug"))
-                .expect("Failed to initialize tracing subscriber");
+            let filter_layer = tracing_subscriber::filter::LevelFilter::DEBUG;
```
JAicewizard, line 190: I would disable the filter all together. This can be done via feature flags on the application side.

druid/src/widget/button.rs
```diff
@@ -110,33 +111,40 @@ impl<T: Data> Button<T> {
 }
 
 impl<T: Data> Widget<T> for Button<T> {
+    #[instrument(name = "Button", level = "trace", skip(self, ctx, event, _data, _env))]
     fn event(&mut self, ctx: &mut EventCtx, event: &Event, _data: &mut T, _env: &Env) {
         match event {
             Event::MouseDown(_) => {
                 ctx.set_active(true);
                 ctx.request_paint();
+                use tracing::info;
```
JAicewizard, line 120: add this to the top of the file

druid/src/widget/clip_box.rs
```diff
@@ -315,9 +324,11 @@ impl<T: Data, W: Widget<T>> Widget<T> for ClipBox<T, W> {
         self.port.rect = self.port.rect.with_size(bc.constrain(content_size));
         let new_offset = self.port.clamp_view_origin(self.viewport_origin());
         self.pan_to(new_offset);
+        trace!("Computed sized: {}", self.viewport_size());
```
JAicewizard, line 327: only call this once and reuse the value

druid/src/window.rs
```diff
@@ -226,11 +227,15 @@ impl<T: Data> Window<T> {
                 is_root: true,
             };
 
+            let _span = info_span!("event");
+            let _span = _span.enter();
             self.root.event(&mut ctx, &event, data, env);
+            std::mem::drop(_span);
```
JAicewizard, line 233: unstead of doing this, could you try using scoping to drop the value? I am not sure if thats how it works in rust but it would be cleaner and lesshacky since a span over a scope makes more sense logically (imo)

druid/src/window.rs
```diff
@@ -291,7 +296,12 @@ impl<T: Data> Window<T> {
             state: &mut state,
             widget_state: &mut widget_state,
         };
+
+        let _span = info_span!("lifecycle");
+        let _span = _span.enter();
         self.root.lifecycle(&mut ctx, event, data, env);
+        std::mem::drop(_span);
```
JAicewizard, line 303: same

druid/src/window.rs
```diff
@@ -308,7 +318,11 @@ impl<T: Data> Window<T> {
             env,
         };
 
+        let _span = info_span!("update");
+        let _span = _span.enter();
         self.root.update(&mut update_ctx, data, env);
+        std::mem::drop(_span);
```
JAicewizard, line 324: same

druid/src/window.rs
```diff
@@ -391,7 +405,12 @@ impl<T: Data> Window<T> {
             WindowSizePolicy::User => BoxConstraints::tight(self.size),
             WindowSizePolicy::Content => BoxConstraints::UNBOUNDED,
         };
+
+        let _span = info_span!("layout");
+        let _span = _span.enter();
         let content_size = self.root.layout(&mut layout_ctx, &bc, data, env);
+        std::mem::drop(_span);
```
JAicewizard, line 412: same

JAicewizard:
Could you also remove all default features besides "fmt" from tracing subscriber? It wont do anything now but will have add compile time improvement later.

druid/src/window.rs
Ciantic: I think the proper syntax is this:

```rust
let span = info_span!("my span");
span.in_scope(|| {
    // do work inside the span...
});
```

I have no idea about this info_span thingie, but that's from tracing docs.

druid/src/widget/clip_box.rs
PoignardAzur: It's more readable this way, and the function is small enough to let the inliner optimize it.

druid/src/app.rs
PoignardAzur: With no filters the output is way too verbose. This is a good compromise for a default. If the user wants to fine-tune filter they can implement their own subscriber.

druid/src/window.rs
PoignardAzur: I'll look into scoping, but the closure-based solution doesn't work because of the borrow checker.

PoignardAzur:
> Could you also remove all default features besides "fmt" from tracing subscriber? It wont do anything now but will have add compile time improvement later.

Already done in a later commit, which I'm not pushing to keep the review process simple.

But yeah, it's a good point.

druid/src/contexts.rs
PoignardAzur: Wow, nice catch! I wish clippy could detect those copy-paste errors.

druid/src/widget/button.rs
PoignardAzur: Nice catch. This was supposed to be a change to help me debugging, I'm removing it.

druid/src/window.rs
PoignardAzur: Done.

cmyr (commented):
First: thank you or your patience so far, this has definitely been a trickier set of patches than we thought, at first.

Second: I don't think we're *quite* there, yet.

Having to instrument each widget independently feels awkward. Ultimately, widgets are *generally* not where the important stuff happens (the exception being the container widgets, and complex things like scrolling). Since all of the logic around invalidation and event delivery is scoped to the `WidgetPod`, it feels to me like we should be able to get most of the benefits of this patch just be instrumenting `WidgetPod`, and then also instrumenting the various `Ctx` methods.

There's a method on the `Widget` trait, [`type_name`](https://docs.rs/druid/0.7.0/src/druid/widget/widget.rs.html#197).  This returns `&str` with a verbose type name, like: `druid::widget::radio::Radio<druid::widget::flex::MainAxisAlignment>`. I think that we could use this method, in combination with `WidgetPod`, to essentially instrument all widgets, in one place.

Did you consider this approach? Are there some limitations that I've overlooked?

druid/src/app.rs
```diff
@@ -183,14 +176,14 @@ impl<T: Data> AppLauncher<T> {
     /// # Panics
     ///
     /// Panics if the subscriber fails to initialize.
-    pub fn use_env_tracing(self) -> Self {
+    pub fn log_to_console(self) -> Self {
```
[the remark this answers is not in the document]
PoignardAzur: Oh, good catch. Yeah, I need to update the doc, it's no longer possible.

druid/src/widget/list.rs
```diff
@@ -217,6 +219,7 @@ impl<S: Data, T: Data> ListIter<(S, T)> for (S, Arc<Vec<T>>) {
 }
 
 impl<C: Data, T: ListIter<C>> Widget<T> for List<C> {
+    #[instrument(name = "Widget", level = "trace", skip(self, ctx, event, data, env))]
```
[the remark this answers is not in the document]
PoignardAzur: Oops

PoignardAzur:
> Did you consider this approach? Are there some limitations that I've overlooked?

Paraphrasing what I said on zulip for visibility:

Okay, so I've tried the `short_type_name()` approach, and it fundamentally doesn't work :neutral:

Basically, if I have `debug_span!("{}", self.short_type_name)` with `self.short_type_name = "Button"` or something. Instead having a span named `Button`, I get a span that says `{}{self.short_type_name="Button"}`.

This is because tracing only accepts string literals as span names; as I understand it, the goal is to cache is whether or not to use a given span at startup, to avoid doing string prefix comparison every time the program needs to compare a span against filter rules.

So I'll keep the `#[instrument]` approach unless you have objections.

cmyr (approved):
Okay so:

Ultimately I think there are clear arguments for having most of the tracing being on `WidgetPod`, but if that is too difficult then I understand wanting to prioritize getting *something* working now, that is reasonably effective.

Thank you for your patience @PoignardAzur!

PR #1627, opened 2021-02-27
Linkable text in RichText.

Maan2003:
I am not sure if adding event function to `TextStorage` is a good thing.

### DIFF
--- a/druid/examples/markdown_preview.rs
+++ b/druid/examples/markdown_preview.rs
@@ -30,11 +30,11 @@ const TEXT: &str = "*Hello* ***world***! This is a `TextBox` where you can \
 		    use limited markdown notation, which is reflected in the \
 		    **styling** of the `Label` on the left.
 
-A very good place to ask questions and discuss development work is our [Zulip chat instance],\
-in the #druid-help and #druid channels, respectively.
-
-[Zulip chat instance]: https://xi.zulipchat.com
-";
+		    If you're curious about Druid, a good place to ask questions \
+		    and discuss development work is our [Zulip chat instance], \
+		    in the #druid-help and #druid channels, respectively.\n\n\n\
+		    
+		    [Zulip chat instance]: https://xi.zulipchat.com";
 
 const SPACER_SIZE: f64 = 8.0;
 const BLOCKQUOTE_COLOR: Color = Color::grey8(0x88);
--- a/druid/examples/markdown_preview.rs
+++ b/druid/examples/markdown_preview.rs
@@ -23,7 +23,6 @@ use druid::{
     AppDelegate, AppLauncher, Color, Command, Data, DelegateCtx, FontFamily, FontStyle, FontWeight,
     Handled, Lens, LocalizedString, MenuDesc, Selector, Target, Widget, WidgetExt, WindowDesc,
 };
-use tracing::warn;
 
 const WINDOW_TITLE: LocalizedString<AppState> = LocalizedString::new("Minimal Markdown");
 
@@ -83,7 +82,7 @@ impl<T: Data> AppDelegate<T> for Delegate {
             #[cfg(not(target_arch = "wasm32"))]
             open::that_in_background(url);
             #[cfg(target_arch = "wasm32")]
-            warn!("opening link({}) not supported on web yet.", url);
+            tracing::warn!("opening link({}) not supported on web yet.", url);
             Handled::Yes
         } else {
             Handled::No
--- a/druid/examples/markdown_preview.rs
+++ b/druid/examples/markdown_preview.rs
@@ -23,6 +23,7 @@ use druid::{
     AppDelegate, AppLauncher, Color, Command, Data, DelegateCtx, FontFamily, FontStyle, FontWeight,
     Handled, Lens, LocalizedString, MenuDesc, Selector, Target, Widget, WidgetExt, WindowDesc,
 };
+use tracing::warn;
 
 const WINDOW_TITLE: LocalizedString<AppState> = LocalizedString::new("Minimal Markdown");
 
@@ -79,7 +80,10 @@ impl<T: Data> AppDelegate<T> for Delegate {
         _env: &Env,
     ) -> Handled {
         if let Some(url) = cmd.get(OPEN_LINK) {
+            #[cfg(not(target_arch = "wasm32"))]
             open::that_in_background(url);
+            #[cfg(target_arch = "wasm32")]
+            warn!("opening link({}) not supported on web yet.", url);
             Handled::Yes
         } else {
             Handled::No
--- a/druid/examples/markdown_preview.rs
+++ b/druid/examples/markdown_preview.rs
@@ -14,27 +14,32 @@
 
 //! An example of live markdown preview
 
-use std::ops::Range;
-
 use pulldown_cmark::{Event as ParseEvent, Parser, Tag};
 
-use druid::text::{Attribute, AttributeSpans, RichText};
+use druid::text::{AttributesAdder, RichText, RichTextBuilder};
 use druid::widget::prelude::*;
 use druid::widget::{Controller, LineBreaking, RawLabel, Scroll, Split, TextBox};
 use druid::{
-    AppLauncher, Color, Data, FontFamily, FontStyle, FontWeight, Lens, LocalizedString, MenuDesc,
-    Widget, WidgetExt, WindowDesc,
+    AppDelegate, AppLauncher, Color, Command, Data, DelegateCtx, FontFamily, FontStyle, FontWeight,
+    Handled, Lens, LocalizedString, MenuDesc, Selector, Target, Widget, WidgetExt, WindowDesc,
 };
 
 const WINDOW_TITLE: LocalizedString<AppState> = LocalizedString::new("Minimal Markdown");
 
 const TEXT: &str = "*Hello* ***world***! This is a `TextBox` where you can \
-                   use limited markdown notation, which is reflected in the \
-                   **styling** of the `Label` on the left.";
+		    use limited markdown notation, which is reflected in the \
+		    **styling** of the `Label` on the left.
+
+A very good place to ask questions and discuss development work is our [Zulip chat instance],\
+in the #druid-help and #druid channels, respectively.
+
+[Zulip chat instance]: https://xi.zulipchat.com
+";
 
 const SPACER_SIZE: f64 = 8.0;
 const BLOCKQUOTE_COLOR: Color = Color::grey8(0x88);
 const LINK_COLOR: Color = Color::rgb8(0, 0, 0xEE);
+const OPEN_LINK: Selector<String> = Selector::new("druid-example.open-link");
 
 #[derive(Clone, Data, Lens)]
 struct AppState {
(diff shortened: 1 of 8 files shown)


druid/src/text/rich_text.rs
```diff
@@ -19,16 +19,17 @@ use std::sync::Arc;
...
-use crate::{ArcStr, Data, Env, FontDescriptor, KeyOrValue};
+use crate::{ArcStr, Data, Env, Event, EventCtx, FontDescriptor, KeyOrValue, TextLayout};
 
 /// Text with optional style spans.
-#[derive(Debug, Clone, Data)]
+#[derive(Clone, Data)]
 pub struct RichText {
     buffer: ArcStr,
     attrs: Arc<AttributeSpans>,
+    links: Arc<[(Range<usize>, Box<dyn Fn(&mut EventCtx, &Env)>)]>,
```
cmyr, line 32: instead of storing the text ranges I would be inclined to store the actual hit-boxes? it just feels like a small simplification.

cmyr (commented):
Interesting! this looks quite promising, I think.

I don't mind passing events to the richtext, although the signature doesn't need to exactly match the signature in `Widget`. For instance, it might be easier to just have `mouse_moved` and `mouse_clicked` methods, or something?

I guess we need an API for launching other applications and opening links, etc? 😬 

Let me know where you'd like to go with this, I think it's on a good path.

druid/src/text/rich_text.rs
Maan2003: We don't have hit boxes when we create `RichText`. So we need to add widget_added or something, to compute the hit boxes. But still we will have to store text ranges.

cmyr: right, that make sense.

Maybe we would want to store the hit boxes in the `TextLayout` object, and compute them whenever we rebuild the layout?

Maan2003:
I am storing `TextStorage`'s `Data` inside `RawLabel` because storing it inside `TextLayout` added a lot of bounds.

cmyr (changes requested):
Okay I think I maybe misunderstood previously, and I see now how having the event handlers be on `RichText` confuses things. Having the `Data` associated type feels hacky, and it is a sign to me that we are not adequately "separating our concerns".

That said, I think this is pretty easy to deal with.

- The `RichText` stuff looks correct; in addition to the spans that piet knows about, we separately track our links.
- Instead of storing handler functions, I would make a link just store a `Command`. This should be enough flexibility that the user can do whatever they need.
- the `TextStorage` trait gets a method like `links(&self) -> &[Link]`.
- the `TextLayout` does all the work. When it lays itself out, it builds its own internal `Rc<[Rect]>` (question: I guess a single link can resolve to multiple rects? in which case maybe it stores `Rc<[(Rect, usize)]>` where the `usize` corresponds to the index of the link that is referred to.
- The `TextLayout` exposes these rects, and then the `RawLabel` can check them in its event methods. There are some options here. I think what might be easiest is a single method,
    ```rust
    fn link_for_mouse_pos(&self, pos: &Point) -> Option<&Link> { ..}
    ```
    this could be used both to check if we're over a link (on hover) or to retrieve the link (on click)

Does this make sense? It feels to me like a cleaner separation of concerns, and likely a simpler implementation. If I'm overlooking anything, please let me know!


Sorry about the runaround. 😒

Maan2003:
This looks better and simpler!

Is `TextLayout` expected to be cheaply clonable? If yes, then cairo backend shouldn't be using `line_metrics: Vec<LineMetric>`. If no, we should use `Vec<(Rect, usize)>`

cmyr:
> This looks better and simpler!
> 
> Is `TextLayout` expected to be cheaply clonable? If yes, then cairo backend shouldn't be using `line_metrics: Vec<LineMetric>`. If no, we should use `Vec<(Rect, usize)>`

Yes, we expect `TextLayout` to be cheaply clonable. The `Vec<LineMetric>` will be fixed when the new linux text work lands, in https://github.com/linebender/piet/pull/389.

cmyr:
@Maan2003 can you add a few links to one of the examples, so that I can play around with this a little bit for testing?

Maan2003:
1. added `open` as a dev dependency. We may want to add a way to open things to druid.
2. rewrote markdown example to use `RichTextBuilder` because we can't add links to `RichText`(because `Arc[Link]`).
3. We should add `len()` to `RichTextBuilder`. So that we don't have to manually keep track of `current_pos`

cmyr:
I think for now I would have the links be 'dummy' links, and not actually open anything? Maybe just log to the console? I'm more interested in seeing that that the hover works and the command is received than what we actually do with it.

Maan2003:
Given that this example doesn't work on web anyways, I have removed dev dependency on `open` for web target

druid/examples/markdown_preview.rs
```diff
@@ -14,27 +14,32 @@
...
-                   use limited markdown notation, which is reflected in the \
-                   **styling** of the `Label` on the left.";
+		    use limited markdown notation, which is reflected in the \
+		    **styling** of the `Label` on the left.
+
+A very good place to ask questions and discuss development work is our [Zulip chat instance],\
+in the #druid-help and #druid channels, respectively.
+
+[Zulip chat instance]: https://xi.zulipchat.com
+";
```
cmyr, line 37: formatting nit, I'm sorry, I can't help myself 😭 

```suggestion
		    **styling** of the `Label` on the left.

		    If you're curious about Druid, a good place to ask questions \
		    and discuss development work is our [Zulip chat instance], \
		    in the #druid-help and #druid channels, respectively.\n\n\n\
		    
		    [Zulip chat instance]: https://xi.zulipchat.com";
```

druid/examples/markdown_preview.rs
```diff
@@ -62,6 +67,28 @@ impl<W: Widget<AppState>> Controller<AppState, W> for RichTextRebuilder {
...
+impl<T: Data> AppDelegate<T> for Delegate {
+    fn command(
+        &mut self,
+        _ctx: &mut DelegateCtx,
+        _target: Target,
+        cmd: &Command,
+        _data: &mut T,
+        _env: &Env,
+    ) -> Handled {
+        if let Some(url) = cmd.get(OPEN_LINK) {
```
cmyr, line 81: This is so cool!

druid/src/text/attribute.rs
```diff
@@ -17,7 +17,14 @@
 use std::ops::Range;
 
 use crate::piet::{Color, FontFamily, FontStyle, FontWeight, TextAttribute as PietAttr};
-use crate::{Env, FontDescriptor, KeyOrValue};
+use crate::{Command, Env, FontDescriptor, KeyOrValue};
+
+/// A link represents a clickable part in the text.
```
cmyr, line 22:
```suggestion
/// A clickable range of text with an associated [`Command`].
```

druid/src/text/attribute.rs
```diff
@@ -93,6 +100,18 @@ pub enum Attribute {
     Descriptor(KeyOrValue<FontDescriptor>),
 }
 
+impl Link {
+    /// Create a new `Link`.
+    pub fn new(range: Range<usize>, command: Command) -> Self {
+        Self { range, command }
+    }
+
+    /// Get Link's range.
```
cmyr, line 109:
```suggestion
    /// Get this `Link`'s range.
```

druid/src/text/layout.rs
```diff
@@ -280,6 +284,17 @@ impl<T: TextStorage> TextLayout<T> {
             .unwrap_or_else(|| Line::new(Point::ZERO, Point::ZERO))
     }
 
+    /// Returns the link at the mouse position.
```
cmyr, line 287:
```suggestion
    /// Returns the [`Link`] at the provided point (relative to the layout's origin) if one exists.
    ///
    /// This can be used both for hit-testing (deciding whether to change the mouse cursor,
    /// or performing some other action when hovering) as well as for retrieving a [`Link`]
    /// on click.
```

druid/src/text/layout.rs
```diff
@@ -280,6 +284,17 @@ impl<T: TextStorage> TextLayout<T> {
...
+    pub fn link_for_mouse_pos(&self, pos: Point) -> Option<&Link> {
```
cmyr, line 288:
```suggestion
    pub fn link_for_pos(&self, pos: Point) -> Option<&Link> {
```

It is not *strictly* necessary that the position originate from a mouse?

druid/src/text/rich_text.rs
```diff
@@ -219,4 +233,12 @@ impl AttributesAdder<'_> {
         self.add_attr(Attribute::font_descriptor(font));
         self
     }
+
+    /// Make this text clickable.
```
cmyr, line 237:
```suggestion
    /// Add a [`Link`] attribute.
```

druid/src/text/storage.rs
```diff
@@ -27,6 +29,11 @@ pub trait TextStorage: PietTextStorage + Data {
     fn add_attributes(&self, builder: PietTextLayoutBuilder, env: &Env) -> PietTextLayoutBuilder {
         builder
     }
+
+    /// The links inside the text.
```
cmyr, line 33:
```suggestion
    /// Any additional [`Link`] attributes on this text.
    ///
    /// If this `TextStorage` object manages link attributes, it should implement this
    /// method and return any attached [`Link`]s.
    ///
    /// Unlike other attributes, links are managed in Druid, not in [`piet`]; as such they
    /// require a separate API.
```

druid/src/widget/label.rs
```diff
@@ -520,9 +522,33 @@ impl<T: TextStorage> Widget<T> for RawLabel<T> {
...
-        skip(self, _ctx, _event, _data, _env)
+        skip(self, ctx, event, _data, _env)
     )]
-    fn event(&mut self, _ctx: &mut EventCtx, _event: &Event, _data: &mut T, _env: &Env) {}
+    fn event(&mut self, ctx: &mut EventCtx, event: &Event, _data: &mut T, _env: &Env) {
+        match event {
+            Event::MouseUp(event) => {
+                // Account for the padding
+                let mut pos = event.pos;
+                pos.x -= LABEL_X_PADDING;
```
cmyr, line 532: just a thought, but I might just write this as,

```suggestion
                let pos = event.pos - Vec2::new(LABEL_X_PADDING, 0.0);
```

druid/src/widget/label.rs
```diff
@@ -520,9 +522,33 @@ impl<T: TextStorage> Widget<T> for RawLabel<T> {
...
+                if let Some(link) = self.layout.link_for_mouse_pos(pos) {
+                    ctx.submit_command(link.command.clone());
+                }
+            }
+            Event::MouseMove(event) => {
+                // Account for the padding
+                let mut pos = event.pos;
+                pos.x -= LABEL_X_PADDING;
```
cmyr, line 540:
```suggestion
                let pos = event.pos - Vec2::new(LABEL_X_PADDING, 0.0);
```

cmyr (approved):
Okay I have a bunch of doc fixups but overall this is great, thanks for taking it on and sorry for the run-around on the implementation. :)

Maan2003:
> sorry for the run-around on the implementation.

thanks for that, it lead to better code and thanks for the doc comments help :)

PR #1628, opened 2021-02-28
Touch support WIP

doppioandante:
Hello, as I need touch support in an application I'm developing with druid, I'm opening this draft with the changes I made to hack a very early  touch support.

The code is not ready for review or inclusion, but it can be a first base for discussion or experimentation. Please feel free to chime in if you are interested in this.

Current code limitations:
1) Only linux (gtk) support. I may try to write to windows backend in the future, but I don't own a mac.
2) Touch events (TouchBegin/TouchUpdate/TouchEnd) are propagated by WidgetPod's. Inside WidgetPod, an instance of `TouchProcessor` is created that tries to convert touch events to standard mouse events. A mechanism to permit widgets to by-pass the processor is needed, I'm trying to think of something.
3) Limitation: touch events conversion is only done inside WidgetPod's, so some lightweight widgets won't have it. This is way I slightly modified the scroll example.
4) Current logic for converting events is hacky to say the least. Mouse click and on finger scroll (pan) are supported.
5) I still don't know what the best way is to push synthesized events back into the event loop.

Every platform has gesture recognition utilities (e.g. gtk's GtkGesture), but I think having druid implement its own platform-independent logic would be a plus?

### DIFF
--- a/druid-shell/src/lib.rs
+++ b/druid-shell/src/lib.rs
@@ -69,7 +69,7 @@ pub use error::Error;
 pub use hotkey::{HotKey, RawMods, SysMods};
 pub use keyboard::{Code, IntoKey, KbKey, KeyEvent, KeyState, Location, Modifiers};
 pub use menu::Menu;
-pub use mouse::{Cursor, CursorDesc, MouseButton, MouseButtons, MouseEvent};
+pub use mouse::{Cursor, CursorDesc, MouseButton, MouseButtons, MouseEvent, PointerType};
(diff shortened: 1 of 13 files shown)


doppioandante:
I'll postpone working on this due to uni for a couple of weeks, but after some discussion on zulip we figured that a better strategy could be to  put the processor at the top of the event chain and mark synthetic events as such, so that widgets can discard them. Doing the processing at every WidgetPod node is not going to work.

PR #1629, opened 2021-03-01
Small refactor of ClipBox

cmyr:
This breaks up the impl blocks a bit so that fewer methods
require the `W: Widget<T>` bound.

It also includes a few small tweaks to the docs and the order
of method declarations.

This patch makes no functional changes.

### DIFF
--- a/druid/src/widget/clip_box.rs
+++ b/druid/src/widget/clip_box.rs
@@ -123,15 +123,32 @@ pub struct ClipBox<T, W> {
     constrain_vertical: bool,
 }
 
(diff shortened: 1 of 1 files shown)


jneem (approved):
Looks fine to me. I guess I'm not sure why it's better to have fewer bounds, since I can't imagine why you'd ever want a `W` without `W: Widget<T>`, but anyway :shrug:

cmyr:
@jneem it was being used in another widget where I wanted to be able to have setters that called `child_mut`, and where the trait bounds on the widget impl were more complex. Definitely not a requirement or anything, just a little help. :)

PR #1630, opened 2021-03-01
Add ClipBox::content_must_fill

cmyr:
This adds a flag that forces the child to be at least as large
as the parent, which is a common behaviour for scroll views.

I'm playing around with using this with the textbox, instead
of having our own manual scroll implementation.

----

- I don't love this name
- There's an argument that this should be per-axis, but I'm not really sure that would be very useful?


jneem:
I don't love it, but I don't have a better suggestion. And at least it's consistent with `Flex`...

jneem (approved):

cmyr:
yea I can't think of anything better either.  😒

PR #1634, opened 2021-03-05
Make Padding generic of child type and add child getters

cmyr:
This is another small thing I wanted for the textbox rework.

### DIFF
--- a/druid/src/widget/padding.rs
+++ b/druid/src/widget/padding.rs
@@ -14,20 +14,20 @@
 
 //! A widget that just adds padding during layout.
 
-use crate::widget::prelude::*;
+use crate::widget::{prelude::*, WidgetWrapper};
 use crate::{Data, Insets, Point, WidgetPod};
 
 /// A widget that just adds padding around its child.
-pub struct Padding<T> {
+pub struct Padding<T, W> {
     left: f64,
     right: f64,
     top: f64,
     bottom: f64,
 
-    child: WidgetPod<T, Box<dyn Widget<T>>>,
+    child: WidgetPod<T, W>,
 }
 
-impl<T> Padding<T> {
+impl<T, W: Widget<T>> Padding<T, W> {
     /// Create a new widget with the specified padding. This can either be an instance
     /// of [`kurbo::Insets`], a f64 for uniform padding, a 2-tuple for axis-uniform padding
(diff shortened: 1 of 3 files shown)


druid/src/widget/padding.rs
```diff
@@ -51,25 +51,37 @@ impl<T> Padding<T> {
...
+    /// A reference to this widget's child.
+    pub fn child(&self) -> &W {
+        self.child.widget()
+    }
+
+    /// A mutable reference to this widget's child.
+    pub fn child_mut(&mut self) -> &mut W {
+        self.child.widget_mut()
+    }
+}
```
Maan2003, line 82: I think you can use `widget_wrapper_body` here

cmyr: mm, I'd completely forgotten about that. It looks like it's only implemented for types that don't contain a `WidgetPod`? The intent when that was merged was just to facilitate robert's experiments with bindings (or maybe scope?) and I'm not sure if it makes sense as a general mechanism? (In particular it's hard to discover, and requires having the trait in scope).

Maybe @rjwittams can chime in.

Maan2003: > It looks like it's only implemented for types that don't contain a WidgetPod? 

there is also `widget_wrapper_pod_body`.

> I'm not sure if it makes sense as a general mechanism?

I am not sure either but implementing a trait feels better to me.

> In particular it's hard to discover, and requires having the trait in scope

I wish we had ability to mark trait as "important", so that rustdoc highlights it.

Maan2003 (approved):
Looks good! We can implement the `WidgetWrapper` trait later if needed

rjwittams:
Sorry missed this. WidgetWrapper was implemented for every 'widget type preserving' wrapper, ie those which don't dyn away the api of what they are wrapping. The reason I need it for bindings was to reach inside things like lenswrap and scope, because binding *before* the data change doesn't work, as the binding host needs to know about it. I doubt it is worth going into details of why here... I think the whole way we deal with data and leaf widgets etc is a bit broken at the moment (they should all be widgets of () and something like bindings should be used for their reactivity). But all that I expect is up for debate with arch changes etc.

On the trait vs not trait thing. Dealing with trait imports just seems like the cost of doing business in Rust. Throwing out polymorphism to avoid imports seems like a very odd thing to even consider to me.

PR #1635, opened 2021-03-05
Add a few new methods to Scroll

cmyr:
These make it possible to configure a few more aspects of the
scrollview's behaviour after it is created.

This will be used for upcoming changes to the textbox.

This also relaxes the type constraints required for some methods.

based off of #1630


cmyr:
@jneem @ForLoveOfCats either of you have any concerns with this?

ForLoveOfCats (approved):
I have no objections. Perhaps they should get a changelog entry where appropriate?

cmyr:
Oh good call I keep forgetting about the changelog 😬

PR #1636, opened 2021-03-05
IME integration (and a new TextBox widget)

cmyr:
This is a big rollup of work on integrating platform text input APIs into druid.

It is based on (and currently includes commits from) #1619. It also includes commits from a number of other smaller PRs, such as #1630, #1634, and #1635.

## Input methods

The highlight of this work is that the textbox is now designed to work with platform text input APIs. Currently these are only supported on macOS; there is a very simple shim in place for other platforms, so that they should continue to work as before.

https://user-images.githubusercontent.com/3330916/110175925-813f2200-7dd0-11eb-9e8d-23064bf91fb5.mov

https://user-images.githubusercontent.com/3330916/110177544-222edc80-7dd3-11eb-87b3-5dd8756ccbd0.mov

There are various little bugs to be worked out here (the size of the text seems to jump around, the position of the text isn't always reported correctly) but overall I think it's good enough to merge.

## Textbox changes

As this work involved ripping everything out anyway, I decided to overhaul the textbox. In particular the textbox now uses a normal `Scroll` widget internally to manage text-position, instead of having its own custom clip/hscroll code. A major benefit of *this* change is that the multi-line textbox is much better; it will grow vertically until it hits a constraint, and once it is constrained it will show sliders:

https://user-images.githubusercontent.com/3330916/110178355-51921900-7dd4-11eb-8c4b-5473b11aa8e9.mov

## Known issues
My current plan is to get the other PRs merged, rebase this, and get it merged, more or less as-is. This will introduce some bugs, but we can address them individually instead of just trying to cram everything into here.

That said, there are some problems I'm aware of:

- whitespace measurement on macOS is buggy
- line height doesn't appear to be correct with some scripts, which means we aren't drawing the composing region correctly
- composing region and selection region seem to get confused, sometimes
- empty lines that are included in a selection are not visibly selected
- we do not always flash the scrollbars when expected, during editing
- text input behaviour regresses somewhat on linux & windows, because we use a new shim to simulate IME on these platforms, and there's a bunch of things we don't currently handle.

I'm sure more things will crop up over time.


cmyr:
Okay I've opened #1652 to track follow-up work, let's see what breaks?

PR #1639, opened 2021-03-10
widget: list: impl ListIter for VecDeque

raymanfx:
Signed-off-by: Christopher N. Hesse <raymanfx@gmail.com>

### DIFF
--- a/druid/src/widget/list.rs
+++ b/druid/src/widget/list.rs
@@ -15,6 +15,7 @@
 //! Simple list view widget.
 
 use std::cmp::Ordering;
+use std::collections::VecDeque;
 use std::f64;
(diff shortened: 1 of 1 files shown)


raymanfx:
I've got to admit I'm not _too_ fond of duplicating the code for `Vec<T>` here.
Then again, I'm not sure how to handle this otherwise. A macro does not help, since the functions differ (e.g. `push` vs `push_back`).

FYI I'm using a `VecDeque<T>` to implement a log buffer as a ring-like structure. Push at the end, pop from the front, that is.

cmyr (approved):
This seems reasonable, I don't see an easy way around the duplication either. 🤷

PR #1640, opened 2021-03-10
Notification fixups

cmyr:
Two small fixes:

- notifications submitted while handling a `Notification` are now allowed
- notifications are no longer sent to the widget that originated them

cc @Maan2003 does this fix that issue for you?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ You can find its changes [documented below](#070---2021-01-01).
 - Shell: IME API and macOS IME implementation ([#1619] by [@lord])
 - Scroll::content_must_fill and a few other new Scroll methods ([#1635] by [@cmyr])
 - New `TextBox` widget with IME integration ([#1636] by [@cmyr])
+- `Notification`s can be submitted while handling other `Notification`s ([#1640] by [@cmyr])
 - Added ListIter implementations for OrdMap ([#1641] by [@Lejero])
 
 ### Changed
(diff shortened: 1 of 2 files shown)


Maan2003 (approved):
It fixes the issue. But I have doubt about order of notifications. 

Iets say a widget has notifications in order `A` `B` `C` . The widget doesn't handle `A` or `C`. When handling `B`, widget submits a notification `B'`.
The order for the parent will be `A` `C` `B'`.
But `A` `B'` `C` makes more sense to me.

cmyr:
hmm, thinking through the orderings: your idea makes sense if we knew that newly added notifications were always replacements for other notifications, but I'm not sure we do? If a widget submits two new notifications in response to one old one, should they both be inserted at the position of the old one?

my gut feeling is that we shouldn't be relying on notification order; if two notifications cause interfering effects they should really be combined into a single notification?

I think I've talked myself into the implementation that I did here; it feels just slightly simpler. 🤷

Maan2003:
> my gut feeling is that we shouldn't be relying on notification order; if two notifications cause interfering effects they should really be combined into a single notification?

Yes I feel the same :)

> I think I've talked myself into the implementation that I did here; it feels just slightly simpler. 

To me `A B' C` implementation feels simpler: just use `parent_notifications` inside `inner_ctx`. And it saves allocation :wink:

cmyr:
> > my gut feeling is that we shouldn't be relying on notification order; if two notifications cause interfering effects they should really be combined into a single notification?
> 
> Yes I feel the same :)
> 
> > I think I've talked myself into the implementation that I did here; it feels just slightly simpler.
> 
> To me `A B' C` implementation feels simpler: just use `parent_notifications` inside `inner_ctx`. And it saves allocation 😉

oh right, totally, that's much better. Thanks!

PR #1641, opened 2021-03-10
Listiter impl

Lejero:
Added ListIter implementations for HashMap and OrdMap. Still unsure if handling key modifications the best way, but values should be good. (This is my first PR here)


Lejero:
I'm not very familiar with web assembly but I can't get the test to run on my computer. Not sure if there's additional setup I need to do it or possibly its having trouble on an M1-mac which I'm using.

futurepaul:
I believe you're missing an entry in `druid/examples/web/src/lib.rs` for your example. Maybe that will solve the test failure?

(I abandoned a [similar PR](https://github.com/linebender/druid/pull/1033/files#diff-a7eea337067b8ad99f746f44d1e29e40975721dab754371f1403ad58bbd514aa) a while back glad to see this being picked up!)

Lejero:
Yes, that was exactly it. I just pushed the change adding to the web examples and all the tests ran successfully. Thanks.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ You can find its changes [documented below](#070---2021-01-01).
 - Shell: windows implementation from content_insets ([#1592] by [@HoNile])
 - Shell: IME API and macOS IME implementation ([#1619] by [@lord])
 - Scroll::content_must_fill and a few other new Scroll methods ([#1635] by [@cmyr])
+- Added ListIter implementations for HashMap and OrdMap ([#1641] by [@Lejero])
```
cmyr, line 25:
```suggestion
- Added ListIter implementations for OrdMap ([#1641] by [@Lejero])
```

druid/examples/list_sources.rs
```diff
@@ -0,0 +1,138 @@
+// Copyright 2019 The Druid Authors.
```
cmyr, line 1:
```suggestion
// Copyright 2021 The Druid Authors.
```

druid/src/widget/list.rs
```diff
@@ -117,6 +117,69 @@ impl<T: Data> ListIter<T> for Vector<T> {
...
+        }
+    }
+
+    fn data_len(&self) -> usize {
+        self.len()
+    }
+}
+
+#[cfg(feature = "im")]
+//TODO: This implementation just disregards any changes to the key. It makes it visible, but nothing else.
```
cmyr, line 150: trying to reconcile this with the code: the code looks like it respects changes to the key?

Edit: actually, I think there's a fundamental problem with exposing the key mutably.

This is obvious if you actually let the user edit the key, which you can do if you replace the `Label` in the `list_sources` example with the following (and import `druid::lens::Identity and `druid::LensExt`):


```rust
Flex::row()
    .with_flex_child(
        TextBox::new().lens(Identity.map(
            |kv: &(String, u64)| kv.0.clone(),
            |kv, new| kv.0 = new.into(),
        )),
        1.0,
    )
    .with_flex_child(
        Label::new(|item: &(String, u64), _env: &_| {
            format!("{0}: {1} square kilometres", item.0, item.1)
        }),
        1.0,
    )
```

(this is a hacky way of adding a textbox that can edit the key)


In any case: the problem is that the items are reordered during editing, but the underlying widgets don't understand this; as far as they can tell it's just that their data has changed, but it's a pretty weird experience for the user; when they edit the text, the items get rearranged, but the focus stays on the original widget?

https://user-images.githubusercontent.com/3330916/110858852-944a6a00-8288-11eb-9fcc-96e6dc8d1f0d.mov

I'm not sure what I think about this, except that it *probably* means that doing things that change the order of list items, *inside the list iter methods*, is probably not a good idea.

It's also possible that the `List` is just generally broken in this sort if situation, and that it isn't really a problem with this PR at all. But my gut instinct still says that we shouldn't mutate the list in the `iter_mut` method, probably?

druid/src/widget/list.rs
```diff
@@ -117,6 +117,69 @@ impl<T: Data> ListIter<T> for Vector<T> {
...
+    fn for_each(&self, mut cb: impl FnMut(&V, usize)) {
+        for (i, item) in self.iter().enumerate() {
+            let ret = (item.0.to_owned(), item.1.to_owned());
+            cb(&ret.1, i);
+        }
+    }
+
+    fn for_each_mut(&mut self, mut cb: impl FnMut(&mut V, usize)) {
+        for (i, item) in self.clone().iter().enumerate() {
+            let mut ret = (item.0.clone(), item.1.clone());
```
cmyr, line 135: I don't think we need to clone the key here? I think we can just use the key from `item`.

cmyr (commented):
A couple little things and then one genuine concern I ran into; take a look and let me know what you think?

druid/src/widget/list.rs
Lejero: As to the 'disregards changes to the key' comment, I apparently forgot to get rid of it after an earlier implementation.

It does seem like the inability to handle reordering is a problem with List, but we could remove the tuple implementation and just keep the value only one for now. The alternative would be going back to the way that comment works and disregarding key changes, but that could be confusing in its own right.

druid/src/widget/list.rs
Lejero: Yes, the clone looks unnecessary. I'm guessing it snuck in because I wrote the key and value implementation and then copy-pasted.

druid/src/widget/list.rs
Lejero: Removing the `ListIter<(K, V)>` implementation might make the list_sources.rs example a bit unnecessary if its only for one Implementation. Should I remove list_sources.rs?

cmyr: I always love having less code! Yes, I agree with your thinking. Just using the values makes sense and is uncontroversial, and I agree that `list_sources` isn't going to be adding much in that case.

Lejero: Alright, this should be good to go. I removed the excess stuff. Left a little comment about why we only implement for values right now. I think that was the last thing to address.

cmyr (approved):
Thanks!

PR #1643, opened 2021-03-11
[win] Fix double-sending of KeyDown event

### LINKED ISSUE #1642: Receiving two KeyDown events for every key press (Windows 10)
Hi team,

First of all, thanks for all the lovely work on this amazing Druid ecosystem!

After the very recent round of changes, I'm now getting two `KeyDown` events coming in for every key press, even very short ones, so all text input and keyboard shortcuts are being doubled. There is still only one `KeyUp` event generated. This is on Windows 10.

No other events (mouse clicks, etc.) seem to be affected.

Any thoughts would be greatly appreciated.

cmyr:
We send this as part of `simulate_input`, so we don't need to
also send it directly from the runloop.

- fixes #1642


cmyr:
see #1642 for confirmation this is fixing the reported issue.

PR #1645, opened 2021-03-12
Aspect ratio box

arthmis:
So there are some things to fix. 
- First, I don't understand what the `instrument` macro is about so would like advice on how to add that to the widget impl.
- I left some comments in the code because I wasn't too sure what I should do for certain things. 
- The last test I wrote I have a comment on an edge case I'm not too sure what the preferred default is.
- I still have to add the change to the changelog.
- Also I think it would nice to have a `ConstrainedBox` to pass in arbitrary constraints to children for testing.
- This has only been tested on Windows 10.

I think I will need to make some changes after reading about Flutter's constraints model.


druid/src/widget/aspect_ratio_box.rs
```diff
@@ -0,0 +1,279 @@
...
+                } else if box_height >= box_width && box_height * self.ratio > box_width {
+                    box_height = box_width / self.ratio;
+                } else {
+                    // I'm not sure if these sections are necessary or correct
+                    // dbg!("ratio can't be preserved {}", self.ratio);
+                    warn!("The aspect ratio cannot be preserved because one of dimensions is tight and the other dimension is too small: bc.max() = {}, bc.min() = {}", bc.max(), bc.min());
+                }
+
+                BoxConstraints::tight(Size::new(box_width, box_height))
+            } else if self.ratio > 1.0 {
```
cmyr, line 134: It's it an invariant that the ratio is <= 1.0?

druid/src/widget/aspect_ratio_box.rs
```diff
@@ -0,0 +1,279 @@
...
+                if box_width > box_height && box_height * self.ratio <= box_width {
+                    box_width = box_height * self.ratio;
+                }
+                // this condition might be wrong if box_width and height are equal to each other
+                // and the aspect ratio is something like 1.00000000000000001, in this case
+                // the  box_height * self.ratio could be equal to box_width
+                // however 1.00000000000000001 does equal 1.0
```
cmyr, line 141: this confuses me a bit. If you are not sure about the conditions, you can add `assert`s that state your assumptions, so we will at least find out when they are not being met.

druid/src/widget/aspect_ratio_box.rs
```diff
@@ -0,0 +1,279 @@
...
+    pub fn empty(ratio: f64) -> Self {
+        Self {
+            inner: None,
+            ratio: AspectRatioBox::<T>::clamp_ratio(ratio),
+        }
+    }
+
+    /// Set the ratio of the box.
+    ///
+    /// The ratio has to be a value between 0 and 1, excluding 0. It will be clamped
```
cmyr, line 40: this does not sound correct to me? An aspect ratio of 2.0 should mean that for your result, height is 2x width.

cmyr (commented):
Thanks! 

Some preliminary notes, but before I dig in much more, have you looked at the `BoxConstraints::constrain_aspect_ratio` method? I'm not sure if it will do exactly what you need, but if not it is a very good example of how to write this sort of "16 different possible combinations of inputs" code in a reasonably clear way. :)

druid/src/widget/aspect_ratio_box.rs
arthmis: I am assuming a value of 2.0 means width is 2x height. If it is 0.5 then height is 2x width. So i'm going aspect ratio is width / height. I think that's how it is assumed everywhere else.

druid/src/widget/aspect_ratio_box.rs
```diff
@@ -0,0 +1,279 @@
...
+use druid::Data;
+use tracing::{instrument, warn};
+
+/// A widget that preserves the aspect ratio given to it.
+///
+/// If given a child, this widget forces the child to have a width and height that preserves
+/// the aspect ratio.
+///
+/// If not given a child, The box will try to size itself  as large or small as possible
+/// to preserve the aspect ratio.
```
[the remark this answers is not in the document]
arthmis: Not particularly. I think I was mostly thinking about small mini games. Maybe there is a grid of different color blocks and they might want each block to maintain their aspect ratio? I don't know if this is a compelling case. It's hard for me to imagine. I guess one can always put a blank box within this and achieve what I'm talking about. So I'll change this.

druid/src/widget/aspect_ratio_box.rs
arthmis: What is being asked here? Are you asking if it is an invariant that the ratio is <= 1.0?

arthmis:
> Some preliminary notes, but before I dig in much more, have you looked at the `BoxConstraints::constrain_aspect_ratio` method? I'm not sure if it will do exactly what you need, but if not it is a very good example of how to write this sort of "16 different possible combinations of inputs" code in a reasonably clear way. :)

Ok so I think I understand the issue. I assumed the aspect ratio `constrain_aspect_ratio` was `width / height`. So when I used it I got wildly incorrect results. Is there a reason it is using height / width instead of width / height?

druid/src/widget/aspect_ratio_box.rs
arthmis: I think I want to stick to aspect ratio being width / height because it is used that way in most other imaging mediums as far as I'm aware.

druid/src/widget/aspect_ratio_box.rs
arthmis: I mostly kept this here to see if you had any opinions on it. I don't think it will be an actual issue. I did test out those values to see if they would be a problem. I'm just paranoid with floating point numbers.

druid/src/widget/aspect_ratio_box.rs
cmyr: I'm just going by what we do with the [`BoxConstraints`](https://docs.rs/druid/0.7.0/druid/struct.BoxConstraints.html#method.constrain_aspect_ratio).

I don't have strong feelings here, but what's most important is that we are consistent and explicit; that means we should be talking very clearly in the docs here about what the argument represents.

cc @derekdreery, who did the `BoxConstraints` impl, and maybe has some thoughts.

cmyr:
I don't know much about the other impl, but let's see if dodj has any suggestions?

druid/src/widget/aspect_ratio_box.rs
cmyr: oops yes, sorry.

This comment was slightly out of date; I was confused because elsewhere in some comments it mentioned that `ratio` should be in the range 0..=1.0, but that wasn't being upheld in the implementation (and didn't make sense, so I assume it's just that the comments were out of date)

druid/src/widget/aspect_ratio_box.rs
arthmis: Ok yes definitely, I'll specifically say what I'm expecting as far as aspect ratio.

druid/src/widget/aspect_ratio_box.rs
arthmis: Oh oops, I don't know what I was thinking when I wrote that comment.

druid/src/widget/aspect_ratio_box.rs
arthmis: After changing the implementation I think the function in `BoxConstraints` is trying to achieve something else.

arthmis:
Ok so, I have two questions now. Should I create a new example that demonstrates the different box type widgets? The other is am I handling correctly infinite constraints and tight constraints?

arthmis:
So I asked @derekdreery about his implementation. They should achieve the same things but his also allows "finding the closest possible aspect ratio that is possible". I think for this widget it isn't necessary to have that behavior. If it can't choose the appropriate size to fit the aspect ratio then it will choose the max constraints as its size.

derekdreery:
> 
> 
> So I asked @derekdreery about his implementation. They should achieve the same things but his also allows "finding the closest possible aspect ratio that is possible". I think for this widget it isn't necessary to have that behavior. If it can't choose the appropriate size to fit the aspect ratio then it will choose the max constraints as its size.

Yes there are a lot of subtly different choices the user could make. Its good to give multiple options I think.

cmyr:
Sorry, I've been distracted the past little while. What's the status of this? What can I do to help get it finished?

arthmis:
Well I think you can review the layout function again to make sure I'm doing what's expected for box constraints. I think you said that child's box constraints should be respected and the parent would clip if necessary? Otherwise this works correctly with respect to the Flutter version. The implementation is basically a copy. 

Also I had posed the question of whether there should be an example that shows all the layouts and put the `AspectRatioBox` in there. Otherwise I will add it to the layout example.

druid/examples/aspect_ratio_box.rs
```diff
@@ -0,0 +1,53 @@
+use druid::{
+    widget::{AspectRatioBox, Flex, Label, LineBreaking, MainAxisAlignment, SizedBox},
+    AppLauncher, Color, Data, Env, Lens, Widget, WidgetExt, WindowDesc,
+};
+
+fn main() {
+    let window = WindowDesc::new(ui());
+    let fixed_message =
+        "Hello there, this is a fixed size box and it will not change no matter what.";
```
SecondFlight, line 9: Maybe just a nitpick, but this isn't true no matter what. `SizedBox` is respecting `BoxConstraints::max`:
image

druid/src/widget/aspect_ratio_box.rs
```diff
@@ -0,0 +1,164 @@
...
+    #[instrument(
+        name = "AspectRatioBox",
+        level = "trace",
+        skip(self, ctx, bc, data, env)
+    )]
+    fn layout(&mut self, ctx: &mut LayoutCtx, bc: &BoxConstraints, data: &T, env: &Env) -> Size {
+        bc.debug_check("AspectRatioBox");
+
+        if bc.max() == bc.min() {
+            warn!("Box constraints are tight. Aspect ratio box will not be able to preserve aspect ratio.");
```
SecondFlight, line 140: The conditions for this warning are not complete. I think the layout algorithm will fail if `bc.min().height * ratio < bc.max().width`, or if `bc.min().height / ratio < bc.max().width`. Could have my math wrong there but it's something similar.

That said, I'm not convinced this always represents invalid state. Aside from precedent (`SizedBox` doesn't warn if it can't meet its constraints), I think there's an argument here for allowing "messy" box constraints. What if I sometimes want to control the width and height and sometimes want to let the constraints work? I'd at least want a way to turn off the warning.

druid/src/widget/aspect_ratio_box.rs
```diff
@@ -0,0 +1,164 @@
...
+
+            return self.inner.layout(ctx, &bc, data, env);
+        }
+
+        if bc.max().width == f64::INFINITY && bc.max().height == f64::INFINITY {
+            warn!("Box constraints are INFINITE. Aspect ratio box won't be able to choose a size because the constraints given by the parent widget are INFINITE.");
```
SecondFlight, line 146: This is another case where I'm not sure the warning is always indicating "incorrect" behavior, because infinite box constraints are sometimes desirable. That said, I think this warning is far less likely to be a false positive than the one above.

SecondFlight:
I looked through the layout function. Looks correct as far as I can tell :+1:

druid/examples/aspect_ratio_box.rs
```diff
@@ -0,0 +1,53 @@
+use druid::{
```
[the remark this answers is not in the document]
SecondFlight: After looking at both examples, I'm not sure this fits well in `--example layout` without reworking that example. I think it does make sense to have a comprehensive layout example, but I'm worried that the current example will become convoluted without a thoughtful rework, which I think is outside the scope of this PR.

druid/src/widget/aspect_ratio_box.rs
Zarenor: So, generally, we've been implementing warnings like this in widgets where the infinite constraint creates an impossible situation. This means Flex (sometimes, if there is a flexible child), Split, and probably others I'm forgetting. So I think it makes sense here. The warning is not that an infinite constraint is *always wrong*, but that it causes a situation the widget having to handle it cannot solve.

druid/src/widget/aspect_ratio_box.rs
Zarenor: I agree with you here, that there are other situations where the widget can't meet the desired aspect ratio. I wouldn't be opposed to warning in those cases, but I think it would be kinda noisy.
Conversely, if the BoxConstraints are tight, and don't match the desired ratio already, this widget is not permitted to solve that problem... And there are only a few ways to get tight bcs - usually, it's because of a SizedBox setting the size, or because you're the root widget of the window (or any direct descendant before splitting or padding has been added).
I don't have a strong opinion here about whether the warning stays, but I do see the logic in warning about this situation.

SecondFlight: Hmm, I see what you're saying. I can still imagine a warning in a "valid" case where the box constraints are not constant, but it's probably enough of an edge case to not matter too much? If a warning or two can be considered okay in a valid but far-out edge case then I'm not worried about it.

Zarenor: Right, I think this could show up spuriously in some edge cases, for sure. And I think a broader "I can't hit the desired ratio" warning would come up a lot more in spurious or transient situations... but I think either or both warnings could be good.
I'm not sure what the right signal-noise ratio is here, but we've usually tended toward only high-snr warnings.

druid/examples/layout.rs
```diff
@@ -60,6 +60,17 @@ fn build_app() -> impl Widget<u32> {
             weight,
         );
     }
+
+    // aspect ratio box
+    let aspect_ratio_label = Label::new("Hello there, this box it's aspect-ratio. Notice text will overflow if box becomes too small.")
```
SecondFlight, line 65:
```suggestion
    let aspect_ratio_label = Label::new("Hello there, this is an aspect-ratio box. Notice how the text will overflow if the box becomes too small.")
```

arthmis: Haha sorry, I did not proof read that at all.

SecondFlight: No worries 😄 Feel free to just replace my suggestion with your own wording if you want

SecondFlight (approved):

SecondFlight:
Looks good as long as CI passes. I didn't look at the unit tests but I trust your judgement there.

druid/src/widget/aspect_ratio_box.rs
arthmis: I think maybe what I can do is instead of warning there, I can run the generate constraints, then check if the final size maintains the aspect ratio. If it doesn't then I can warn and print the box constraints it was given and the aspect ratio?

arthmis: That way I'm not checking for all the conditions where it could fail, though that may be less helpful to the user?

arthmis:
For the warnings, I don't think I will add any more for now. I guess if it proves to be a problem then more checks can be added in the future.

PR #1647, opened 2021-03-13
Correctly capture and use stroke properties when rendering SVG paths

SecondFlight:
This patch updates the SVG widget to fetch and use properties about stroke style, such as [line cap](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-linecap) and [dash array](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-dasharray).

This has not been thoroughly tested since the new properties have a one-to-one mapping with `StrokeStyle` in piet. It has passed a smoke test in a project I'm working on, which has 12 unique SVG icons. Some of these icons also now render correctly where they didn't before.

I'm a little iffy about is the `Vec` clone. I don't think it's avoidable here, but I want to point it out since Druid goes to great lengths to avoid it elsewhere.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ You can find its changes [documented below](#070---2021-01-01).
 
 - Fixed docs of derived Lens ([#1523] by [@Maan2003])
 - Use correct fill rule when rendering SVG paths ([#1606] by [@SecondFlight])
+- Correctly capture and use stroke properties when rendering SVG paths ([#1647] by [@SecondFlight])
 
 ### Visual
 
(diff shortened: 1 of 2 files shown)


Maan2003 (approved):
Thanks! 

I don't think there is a way to avoid clone here. :)

(Edit: there is a way but not practical: Make `p` `&mut usvg::Path`, then `mem::take` the `dasharray` and reassigning it again after `stroke_style()`)

cmyr:
I think the best answer here is to not have `dash_array` be a vec; `StrokeStyle` should be cheap to clone.

A simple answer is `Rc<[]>`; and maybe const generics will let us use arrays in more places like this?

PR #1649, opened 2021-03-13
adding constrained box

arthmis:
- I hope the instrumentation stuff is correct, I copied from `SizedBox`.
- I want to add methods to do things like `set_min_width`, `set_min_height`, `set_max_height`, `set_max_width`, but I don't think they are absolutely necessary right now. It would be nice to have methods on `BoxConstraints` to change individual min and max dimensions.

I did not realize there was already an attempt at this widget. I saw that the other attempt used a closure. I guess that would be nice to set arbitrary box constraints based on the ctx, data, and env. I would like to keep it simple and only allow `BoxConstraints`. If using the closure is preferred then I will change it to use that. That is more powerful and flexible.

Now I'm questioning whether there should be an option to have no child. It now doesn't seem to make sense to have this widget without a child.


cmyr:
Cool, will try to get to these two tomorrow. :)

Maan2003 (commented):
Looks good to me. 

Do we need to support optional child?

druid/src/widget/constrained_box.rs
```diff
@@ -0,0 +1,297 @@
...
+    fn update(&mut self, ctx: &mut UpdateCtx, old_data: &T, data: &T, env: &Env) {
+        if let Some(ref mut inner) = self.inner {
+            inner.update(ctx, old_data, data, env);
+        }
+    }
+
+    #[instrument(
+        name = "ConstrainedBox",
+        level = "trace",
+        skip(self, ctx, bc, data, env)
```
Maan2003, line 109: We shouldn't skip bc here, given this is all about constraints

PoignardAzur:
Looks good to me as well. I like that the PR comes with unit tests.

@arthmis Can you rebase the PR and fix the CI errors? If you do, I'll quickly test it on my side and then merge it.

xStrom:
What's the state of this @arthmis? Are you still interested in moving this forward?

If so, then a good place to start is to rebase this on `master` and getting the CI to pass without errors.

PR #1651, opened 2021-03-16
Relax constraints on who can change focus

cmyr:
Instead of just the focus widget this is now available to any of
that widget's direct ancestors (such as a controller).

This is needed for the new TextBox, since the widget with
the logic (TextBox) is not the widget that necessarily
has focus (the SharedComponent).

This is broken out of #1636.


Maan2003 (approved):
:rocket: Sounds reasonable

Issue #1652, opened 2021-03-16
New TextBox & IME integration tracking issue (report new TextBox issues here!)


cmyr:
When #1636 lands, how druid handles text will change fundamentally, and there are almost certainly some issues that we will encounter as we start using the new implementation.

This issue is intended as a tracking/discussion issue for various problems with the new text work.

## polish
There are a number of outstanding fixups I would like to land:
- [ ] better implementation of 'select word' and 'select line': these don't currently behave well when clicks occur on boundaries.
- [ ] preserve double/triple-click state through to drag gestures: drag after double-click should select at "word granularity", after triple click at "line granularity"
- [ ] handle more keys when 'simulating input' on windows/linux
- [ ] unify selection & movement types between druid-shell and druid
- [ ] TextBox without focus should use an 'inactive' color for selections.
- [ ] restructure `druid::text` module, centralizing all text-related types here

## features
Larger related projects it might be nice to do while we're in here
- [ ] select-all, copy, paste (and undo/redo, eventually) should work without the menus present (#1030) 
- [ ] 'submit on return': there should be a mechanism for hooking the return key in order to perform some custom action.
- [ ] allow toggling selectable & editable flags: you may want a text field that can be selected but not edited, or a text field that cannot be edited *or* selected
- [ ] TextBox should have built-in support for undo/redo

## bugs
- [ ] selections that include lines with emergency breaks (breaks that are not at whitespace) are not drawn correctly: https://github.com/linebender/piet/pull/412
- [ ] IME composing region is drawn incorrectly sometimes
- [ ] IME composing window can draw at incorrect position
- [ ] line height calculations are incorrect for some scripts (japanese)

jpochyla:
Thank you, it's very impressive!

- I think a lot of people were previously having custom controllers on TextBoxes, dispatching EditActions on shortcuts. This would be now done via `textbox.text().input_handler().acquire().handle_action()`, correct?
- `TEXTBOX_INSETS` is a constant again! :(

madbonkey:
> 'submit on return': there should be a mechanism for hooking the return key in order to perform some custom action.

To be honest, I feel this is out of scope for text handling. What "submitting" or "commiting" means in regard to changes to state highly depends on the application and the control/inout type. You might think search boxes and go 'yeah that seems nice to have', but in reality the search might automatically "commit" after a debounce, or it's also committed by an icon next to it. It might require entirely different ways of "submitting" the input depending on devices and peripherals. 

[...]

PR #1653, opened 2021-03-16
Unify druid & druid-shell selection types, rm src/text/selection.rs

cmyr:
This removes the druid selection type, and adds any API on that type that was missing to `druid-shell::Selection`

- progress on #1652


lord:
A few thoughts/questions:

* When I've implemented text editing before, I've modeled selections as having two `h_pos`s, one for the active end and one for the anchor. But I'm struggling to actually find a case where this makes a difference, so it could be that these two representations are equivalent? If they are in fact equivalent, this way with just one seems maybe better? Not sure if you've thought too much about this or not.
* How would a user create a selection with an `h_pos` of `None`?
* I think the way this interacts with `InputHandler::set_selection` is worth documenting? (Should the input method be able to update the `h_pos` directly? I feel like not, right?) Similarly, maybe worth writing down if the input method can observe the `h_pos`, and the `h_pos` changes, does that result in `Event::SelectionChanged`, or (my vote) `Event:: LayoutChanged`
* If these types are unified, you'll eventually want to add affinity as well, right? If `Selection` continues to grow, may want to do something to limit what the input method can access, or at least what it can implicitly set through `set_selection`

cmyr:
> A few thoughts/questions:
> 
> * When I've implemented text editing before, I've modeled selections as having two `h_pos`s, one for the active end and one for the anchor. But I'm struggling to actually find a case where this makes a difference, so it could be that these two representations are equivalent? If they are in fact equivalent, this way with just one seems maybe better? Not sure if you've thought too much about this or not.

This is interesting, it would not have occurred to me to track the h-pos of the anchor edge, but now that you mention it you *probably* want to track the hpos of the secondary cursor, if you're doing BiDi.

> * How would a user create a selection with an `h_pos` of `None`?

`h_pos` is none by default; it is only set when there is a vertical movement.

> * I think the way this interacts with `InputHandler::set_selection` is worth documenting? (Should the input method be able to update the `h_pos` directly? I feel like not, right?) Similarly, maybe worth writing down if the input method can observe the `h_pos`, and the `h_pos` changes, does that result in `Event::SelectionChanged`, or (my vote) `Event:: LayoutChanged`

All explicit setting of the selection (that are not the product of vertical motion) clears `h_pos`, and `h_pos` is not part of visible selection state. Also it should not be possible to change `h_pos` without also changing some other aspect of the selection? In any case I don't think `h_pos` interacts with IME.


> * If these types are unified, you'll eventually want to add affinity as well, right? If `Selection` continues to grow, may want to do something to limit what the input method can access, or at least what it can implicitly set through `set_selection`

Yes, I was thinking of making this be our single selection type. I can't really think of a very good way to limit what is available to IME, but I do think it would be worth documenting the expectations a bit better.

Thanks for the notes, very happy to have another pair of eyes on some of this stuff!

lord:
sounds good! :)

PR #1654, opened 2021-03-17
Implement Lens directly on a pair

Maan2003:
I see no reason not to implement it directly :shrug: and it feels more intuitive


Maan2003:
Maybe we should implement `Lens` for other tuples as well.

cmyr:
@derekdreery does this make sense to you? I think so, but curious if you had a more specific need that required `Tuple2` to be a named type?

I've run into this before; I think it *does* make sense to have lenses on the tuples, maybe from rank 1-8? I think an 8-tuple is a horrible idea but I think it's okay as an upper-bound..

Maan2003:
@cmyr @derekdreery can we get this done?

cmyr (approved):
Yes, sorry, happy to see this merged.

SecondFlight:
@Maan2003 Just noticed this doesn't have a matching changelog entry. Just want to make sure you have the chance if you forgot.

cmyr (approved):
Looks good!

PR #1656, opened 2021-03-17
Export Link type

cmyr:
This type is part of public API and is linked in some doc comments
so it makes sense that the type be accessible.


Maan2003 (approved):
Thanks! I totally forgot about this. I thought module `attribute` was public

cmyr:
so did I, I just noticed I was getting a warning when testing the docs. 🤷

cmyr:
note to self: we should have `--document-private-items` set in the docs step of CI, that would have caught this

Issue #1386, opened 2020-11-09
Rewrite EnvScope


cmyr:
The current implementation of `EnvScope` has fundamental performance problems and is not fit for purpose.

I would like to propose a two-part solution: first, we more eficiently accomodate the common case, where the items in the `Env` do not change, and secondly that we do more caching in order to decrease the frequency with which we need to construct a new `Env`.

### First part:

Most of the time, `EnvScope` is used to override some static set of attributes; there is no special logic. We should accomodate this directly, by allowing the user to specify just some set of keys/values; and we will actually construct a new `Env` only if the incoming `Env` differs from the previous incoming `Env`.

This would look something like this; anytime you needed to get the env for your child, you would call `child_env`, and it would rebuild only if necessary.

```rust
struct EnvScope2<T> {
    prev_super_env: Option<Env>,
    current_child_env: Option<Env>,
    overrides: Env,
    // also a child widget etc here
}

impl EnvScope2 {
    /// The `overrides` is an env containing only keys/values that should be overwritten at this scope.
    pub fn new(overrides: Env, child: impl Widget<T>) -> Self {
        // ...
    }

    fn child_env(&mut self, super_env: &Env) -> &Env {
        let super_same = self.pre_super_env.as_ref().map(|old| old.same(super_env)).unwrap_or(false);
        if !super_same {
            self.current_child_env = Some(super_env.with_overrides(&self.overrides));
            self.prev_super_env = Some(super_env.to_owned());
        }
        self.current_child_env.as_ref().unwrap()
    }
}
```

### Second part:

the second part is similar, and equally simple: If you *really* need to dynamically change the `Env`, you can still do that with a closure, but you need a second closure that is used to determine whether or not the change needs to happen **right now**. This would be called in `update`, and would return a bool, and we would only actually rebuild the env if this was `true`, using logic similar to the logic above.

futurepaul:
My main use case for `EnvScope` right now is to set a `DISABLED` key based on some inspection of my app state, so it looks like I would be using the dual closure version you propose in part ii. Not a huge deal for me but possibly this helps settle the question in #746 about putting disable functionality in Env.

PR #1658, opened 2021-03-17
updating EnvScope to cache Env

arthmis:
I added an `Env::new` function but this might be redundant since there is a default implementation for `Env`. I think I closely followed what you outlined in #1386. 
- I still have to fix errors in the examples


druid/src/widget/env_scope.rs
```diff
@@ -21,8 +21,13 @@ use tracing::instrument;
...
 /// A widget that accepts a closure to update the environment for its child.
 pub struct EnvScope<T, W> {
-    pub(crate) f: Box<dyn Fn(&mut Env, &T)>,
     pub(crate) child: WidgetPod<T, W>,
+    pub(crate) current_child_env: Option<Env>,
+    pub(crate) prev_super_env: Option<Env>,
+    pub(crate) overrides: Env,
+    pub(crate) modify_env: Option<Box<dyn Fn(&T, &mut Env)>>,
+    #[allow(clippy::type_complexity)]
+    pub(crate) should_modify_env_now: Option<Box<dyn Fn(&T, &T, &Env) -> bool>>,
```
cmyr, line 30: I would listen to clippy here and add a `type EnvInvalidationCheck = dyn Fn(&T, &T, &Env) -> bool` declaration at the top of the file.

cmyr (commented):
This looks pretty interesting.

I'd be curious if you've actually profiled this? If you have somewhere that's using `EnvScope` heavily it should show up pretty consistently in a profiler, and it would be nice to verify that this has a clear impact.

druid/src/widget/env_scope.rs
arthmis: Ok, I wasn't even sure what to name it.

arthmis:
I have never profiled anything before. Do you have any recommendations for that on Windows? There is a paid software but I don't want to spend money on that right now. https://nnethercote.github.io/perf-book/profiling.html I found this so I'll try the AMD one. Hopefully that will be fine.

cmyr:
I unfortunately don't have any good suggestions for profiling on windows, but I'm sure there are some profilers out there; you might ask on zulip and see if anyone has any suggestions?

druid/src/widget/env_scope.rs
cmyr: heh me neither, just brainstorming. :)

arthmis:
Ok, so from my shallow attempts at profiling. So take some of these things as, maybe there is something to it. I'm going to call the original way "original env" and the new way "cached env". 

In the original env calling the closure within `event` dominates the time spent in the function. It spends about 75% of the function time calling that and about 20% calling the child's event function. In my simple example, `EnvScope`'s `event` function used about 48 milliseconds of total execution time for the entire time the app was profiled for. So calling the closure took 36 milliseconds of that total execution time. 

The cached env spends about 2-3% time on getting the child env. 98% of time spent for cached env is calling it's child's event function. Cached env's event function used a total of 12 ms for the entire execution time.

All the profiles were executed for aprox. 10 seconds. So it doesn't seem like massive savings but there are some. I'll profile some more. I didn't test dynamically updating part of cached env.

cmyr:
Cool, I'll give this another close look shortly but your results sound encouraging.

arthmis:
Ah I forgot to mention. I used an Env where I added about 85 new entries and set the value to be `Color`. Yeah there is still more to look at, it's a lot of information when profiling.

arthmis:
I used some of the instrumentation api available for the profiler I'm using so I was able to see the approximate average execution times. So the cached env has on average ~6-7 microsecond execution time and the old env took about ~22-24 microseconds on average.

druid/src/widget/env_scope.rs
```diff
@@ -21,10 +21,16 @@ use tracing::instrument;
 
 /// A widget that accepts a closure to update the environment for its child.
 pub struct EnvScope<T, W> {
-    pub(crate) f: Box<dyn Fn(&mut Env, &T)>,
     pub(crate) child: WidgetPod<T, W>,
+    pub(crate) current_child_env: Option<Env>,
+    pub(crate) prev_super_env: Option<Env>,
+    pub(crate) overrides: Env,
+    pub(crate) modify_env: Option<Box<dyn Fn(&T, &mut Env)>>,
+    pub(crate) should_modify_env_now: EnvInvalidationCheck<T>,
```
cmyr, line 29: My understanding is that you can provide *either* an `Env` or a `Fn(&T, &mut Env)`; these are the two supported ways to update the env. Is this correct? what does it mean if the user supplies both of these? Is that fine? I don't see this documented anywhere, which will be important.

druid/src/widget/env_scope.rs
```diff
@@ -48,61 +54,109 @@ impl<T, W: Widget<T>> EnvScope<T, W> {
...
+
+    /// This function takes a closure that allows you to modify the Env
+    /// based on the provided data.
+    ///
+    /// The first closure gives you access to app data and a mutable `Env`
+    /// which will be passed to `EnvScope`'s children.
+    ///
+    /// The second closure will determine if the first closure is called
+    /// when there are updates to the provided data.
+    pub fn modify_env(
```
cmyr, line 89: Maybe this should be an alternate constructor method? Like: Maybe we have,

```rust
fn new(overrides: Env) -> Self {}
fn dynamic(
    modify_env: impl Fn(&T, &mut Env) + 'static,
    should_modify_env: impl Fn(&T, &T, &Env) -> bool + 'static,
) -> Self {}
```

?

druid/src/widget/env_scope.rs
arthmis: I was thinking you can set the base environment, but having the closure allows you to dynamically modify only part of the environment you've set or maybe even a different set of variables from the original static set. If it were either dynamic or static then, the dynamic part would have some parts that are static and some parts dynamic, though this wouldn't actually matter practically. 

You're thinking of something like what `Label` does where it has an Enum that holds a static Env or a closure that will be called when data is updated?

Would there ever be a reason to want to change the closure? Say you wanted to update your app some kind of way, but the user changed some setting, so the environment variables you want to change are different. Maybe there should be a way to set the closure like `set_dynamic`?

arthmis: I can't think of good use cases for them being two separate fields versus being in an enum that has `Static` and `Dynamic`.

druid/src/widget/env_scope.rs
arthmis: Yes this sounds good. It depends on the above conversation. If this becomes dynamic or static then I'll create this constructor.

druid/src/widget/env_scope.rs
cmyr: > I was thinking you can set the base environment, but having the closure allows you to dynamically modify only part of the environment you've set or maybe even a different set of variables from the original static set. If it were either dynamic or static then, the dynamic part would have some parts that are static and some parts dynamic, though this wouldn't actually matter practically.
> 
> You're thinking of something like what `Label` does where it has an Enum that holds a static Env or a closure that will be called when data is updated?
> 

This is *exactly* what I was thinking.


> Would there ever be a reason to want to change the closure? Say you wanted to update your app some kind of way, but the user changed some setting, so the environment variables you want to change are different. Maybe there should be a way to set the closure like `set_dynamic`?

I get a sense of YAGNI about this; I think if someone ever actually encounters this problem we can consider it? A closure has a lot of flexibility, you could have various branches in the closure that respond to existing env keys, for instance.

arthmis: Wait, so which part was what you were thinking? Is it fine that the `Env` be set statically and have the option to dynamically update? Or should it be similar in design to how `Label` is constructed?

cmyr: I think something like `Label` would make a lot of sense and be pretty easy to understand.

arthmis:
I have not forgotten this, I will add the changes and see how it works.

arthmis:
Ok So I've implemented this for now. I've run into a problem and I think it makes more sense for their to be both a static and dynamic method instead of one or the other. I'm going to upload the problem as I have it. 

Before the closure, that modified the env, would be called immediately and make the `Key`,`Value` pair available. With this new dynamic version, the closure to modify env is only called if data changes. So it ends up that the program tries to access a key that hasn't been set yet.

```rust
        .env_scope_dynamic(
            |data: &AppData, env: &mut druid::Env| {
                let new_font = if data.mono {
                    FontDescriptor::new(FontFamily::MONOSPACE)
                } else {
                    FontDescriptor::new(FontFamily::SYSTEM_UI)
                }
                .with_size(data.size);
                env.set(MY_CUSTOM_FONT, new_font);
            },
            |old_data, data, _env| {
                !data.mono.same(&old_data.mono)
            },
        );
```
This will crash the styled_text example as it is. I added this to the example
```rust
        .configure_env(|env, data| {
             const MY_CUSTOM_FONT: Key<FontDescriptor> =
                 Key::new("org.linebender.example.my-custom-font");
             let font = FontDescriptor::new(FontFamily::SYSTEM_UI).with_size(data.size);
             env.set(MY_CUSTOM_FONT, font);
        })
```
This prevents the crash from happening. I tried using the `WidgetAdded` lifecycle event to modify the env before it's accessed, maybe I'm not doing it right, but it still crashed. So back to what i was saying at the top. I think it makes more sense to be able to set the environment statically, and also dynamically change it if you want to. I'll look into fixing the issue for now though.

arthmis:
Oh, the crash probably happens because of this
```rust
    let styled_label = Label::new(|data: &AppData, _env: &_| format!("{}", data))
        .with_text_color(theme::PRIMARY_LIGHT)
        .with_font(MY_CUSTOM_FONT)
```
This part the env_scope wraps this label after `with_font` is called. This only works if `MY_CUSTOM_FONT` is already set by the env_scope. With the new way of dynamic env, it doesn't happen. So there needs to be a way to set the environment before dynamic things can happen.

arthmis:
Ok, I fixed a bug that was in the code 😅 . So the only weakness of this, is not being able to set environment variable prior to using dynamic. I'm open to adding one more argument in the `dynamic` method. This would take either an `Env` or closure. Though, thinking about it, it might be a bit tricky to get that working well.

arthmis:
Well I've made changes to the dynamic method. This will take a `static_overrides` argument which is basically a closure that allows you to set the necessary environment variables. This is important because, if your dynamic overrides assume some env keys are there, then this `static_overrides` closure will set those keys first. I figured it would be better to use a closure instead of a static `Env` because it gives the user the option to have access to the data when setting the Env.

arthmis:
Well I've made changes to the dynamic method. This will take a `static_overrides` argument which is basically a closure that allows you to set the necessary environment variables. This is important because, if your dynamic overrides assume some env keys are there, then this `static_overrides` closure will set those keys first. I figured it would be better to use a closure instead of a static `Env` because it gives the user access to data.

druid/examples/sub_window.rs
```diff
@@ -326,8 +326,14 @@ impl<W: Widget<bool>> Controller<bool, W> for CancelClose {
 }
 
 fn build_root_widget() -> impl Widget<HelloState> {
+    let env = {
+        let mut env = Env::new();
+        env.set(theme::TEXT_COLOR, env.get(theme::PRIMARY_LIGHT));
+        env
```
cmyr, line 332: the easier way to do this is,

```rust
let env = Env::new() .adding(theme::TEXT_COLOR, Color::SOME_COLOR);
```

I'm not sure why we're trying to get a value out of the new env, here, it should be empty?

druid/src/env.rs
```diff
@@ -240,6 +240,19 @@ impl Env {
     /// [`WidgetExt::debug_widget`]: trait.WidgetExt.html#method.debug_widget
     pub const DEBUG_WIDGET: Key<bool> = Key::new("org.linebender.druid.built-in.debug-widget");
 
+    /// Creates a new empty Env with localization information.
+    pub fn new() -> Self {
+        let l10n = L10nManager::new(vec!["builtin.ftl".into()], "./resources/i18n/");
```
cmyr, line 245: huh, I hadn't really considered this. We really don't want to be re-creating the `L10Manager` everytime we create an env, if we're creating them regularly to just use as a key/value store; we'll need a different approach to that.

druid/src/env.rs
```diff
@@ -371,6 +384,21 @@ impl Env {
         let color_num = id as usize % self.0.debug_colors.len();
         self.0.debug_colors[color_num].clone()
     }
+
+    /// Replaces and adds to current `Env`'s values with values form the provided `Env`.
+    pub(crate) fn with_overrides(&self, changes: &Env) -> Env {
+        let mut new_env = self.to_owned();
+        let mut internal_env = Arc::make_mut(&mut new_env.0).to_owned();
```
cmyr, line 391: we shouldn't need to do the `to_owned`, we can just use the mutable reference directly?

cmyr (commented):
A couple observations while looking through your zulip question

druid/src/env.rs
arthmis: Ok, I wasn't sure how to go about that. I added this so someone can create an Env when they use it statically.

druid/examples/sub_window.rs
arthmis: Oh ok, I was mostly trying to keep the example the same.

arthmis: Oh I see, I had just copied it without thinking about what it was actually doing.

druid/src/env.rs
arthmis: which mutable reference? I can't access any of them because self isn't `mut`.

druid/src/env.rs
arthmis: is there a better way to come up with empty Env? I think instead of the static variant using a Env, it should take a closure like this
| data: &T, env: &mut Env | {}
This way the user doesn't concern themselves with creating an env and instead should modify the one provided. We won't have to worry about creating an env without localization resources or with it. The env provided to the closure would be a clone of the env provided by the widget method.

druid/src/env.rs
cmyr: I guess there are maybe two problems here:

- `Env` has a weird internal layout. It wasn't really designed with this particular pattern (overrides from another env) in mind. The rationale for `EnvImpl` is just to make sure that cloning an `Env` requires a single pointer clone. There's the additional problem that the `Env` is sort of conceptually vague; to the user the `Env` is just a key/value store, but it also happens to contain localization info etc.
- This is not helped by the fact that the API we've designed here represents the overrides as a new `Env`. The issue here is that each new `Env` is going to contain a duplicate version of the localization manager, and this is silly.

Some possible solutions:

- make the localization stuff something that has to be set up explicitly, and `Env::default()` just has no localization info.
- have a separate `EnvMap` type that is used for specifying the overrides

I don't *love* having a new type, since it's another import, another piece of public api, etc; I think it might make more sense to move to having the localization setup be something that you have to do manually, and which druid can do when setting up the initial env...


```rust
let mut new_env = self.clone();
let internal = Arc::make_mut(&mut new_env);

cmyr (commented):
Okay sorry for letting this sit so long, I think it's worth merging. I've left a comment inline; I'm going to make some tweaks to the `Env` API that should make this patch cleaner.

arthmis:
Ok cool, give me a few days to get back into this. Thanks for looking it over again.

PoignardAzur:
@arthmis Any progress?

druid/src/env.rs
arthmis: Hmm I tried doing what you suggested but the borrow checker doesn't allow that. Maybe I'm approaching it the wrong way. I'm quite rusty on rust right now lol.

arthmis:
Yes I think everything is fine. I have to run the tests and whatnot which I will do later today.

xStrom:
What's the state of this @arthmis? Are you still interested in moving this forward?

If so, then a good place to start is to rebase this on `master` to resolve the merge conflicts.

PR #1659, opened 2021-03-17
Add inactive selection color, tweak selection colors

cmyr:
This patch changes the selection color to *not* red, and adds
a new SELECTED_TEXT_INACTIVE_BACKGROUND_COLOR theme item; it
also now draws the selection with that color when the text is
not focused.

This requires a bit of a hack because of how focus is working
in the textbox; the TextBox has focus, not the InputComponent,
but the InputComponent is responsible for drawing the selection
rects, so we need to explicitly pass down changes in focus state.

C'est la vie.

This also renames SELECTION_COLOR to SELECTED_TEXT_BACKGROUND_COLOR,
for clarity & symmetry.

https://user-images.githubusercontent.com/3330916/111514971-a6188b00-8728-11eb-982b-a87778a16118.mov

- more progress on #1652


Maan2003 (approved):
:rocket: Awesome!

PR #1660, opened 2021-03-17
Handle standard text keyboard shortcuts without menu

### LINKED ISSUE #1030: Copying and pasting in textbox does not work (at least on windows)
When pressing "ctrl+v" on windows inside the textbox, the text "\u{16}" is inserted. I am not sure where this exactly comes from but the event "Event::Paste" is never raised and therefore clipboard code never ran. Instead the code execution goes into BasicTextInput's handle_event() and the is_printable() if-check. 

Pressing "ctrl+c" and "ctrl+x" both deletes the text inside the textbox when only "ctrl+x" should delete it. The text is never copied to the clipboard in both cases.

cmyr:
The textbox will now receive copy/cut/paste/undo/redo/select-all
without a menu being present.

Several of these (select-all, undo/redo) do not currently have
implementations, but we will at least get the commands.

- progress on #1652
- closes #1030

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ You can find its changes [documented below](#070---2021-01-01).
 
 ### Fixed
 - `Notification`s will not be delivered to the widget that sends them ([#1640] by [@cmyr])
+- `TextBox` can handle standard keyboard shortcuts without needing menus ([#1660] by [@cmyr])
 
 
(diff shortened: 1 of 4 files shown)


kud1ing:
> The textbox will now receive copy/cut/paste/undo/redo/select-all

Observation on macOS with the `widget_gallery`:
* Cmd+C/V work for copying/pasting.
* Cmd+A does not select the text

Should i open an issue?

PR #1662, opened 2021-03-18
Padding takes KeyOrValue, TEXTBOX_INSETS moves back to Env

cmyr:
two related commits; the first enabling the second.

This is more progress on #1652.

### DIFF
--- a/druid/src/widget/padding.rs
+++ b/druid/src/widget/padding.rs
@@ -15,24 +15,23 @@
 //! A widget that just adds padding during layout.
 
 use crate::widget::{prelude::*, WidgetWrapper};
-use crate::{Data, Insets, Point, WidgetPod};
+use crate::{Data, Insets, KeyOrValue, Point, WidgetPod};
 
 use tracing::{instrument, trace};
 
 /// A widget that just adds padding around its child.
 pub struct Padding<T, W> {
-    left: f64,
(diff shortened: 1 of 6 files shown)


druid/src/widget/padding.rs
```diff
@@ -15,24 +15,26 @@
...
-    left: f64,
-    right: f64,
-    top: f64,
-    bottom: f64,
-
+    insets: KeyOrValue<Insets>,
     child: WidgetPod<T, W>,
 }
 
+/// `Insets`, types that impl `Into<Insts>`, or KeyOrValue<Insets>.
```
futurepaul, line 28: typo? `Into<Insets>`

Also, is there a reason this had to be a trait instead of just `impl Into`? I haven't been following closely so pls disregard if this is obvious.

cmyr: You're right, I'd been over-thinking this. The problem is that we don't just accept `Insets`, we also accept things like `f64` and `(f64, f64)`, and I was thinking that to make those work we would need to have `impl <T: Into<Insets>> Into<KeyOrValue<Insets>> for T`, which wouldn't work because of coherence, but we can instead just manually impl `Into<KeyOrValue<Insets>>` for the various other arguments we want to accept, and that works fine. Thanks!

PR #1665, opened 2021-03-18
Fix word selection logic

cmyr:
This is particularly relevant for when the user double-clicks a word.
Previously if the click fell on a word boundary we would not
recognize that; now if the click falls on a word boundary we will
treat that as the start of the new selection range.

In addition, word select now selects *anything* between two word
boundaries; we do not care if it is actually a word. As an example:
if the user double clicks in whitespcae, we will select any
contiguous whitespace.

progress on #1652

This will conflict with #1653, so I should maybe start merging some of these.


cmyr:
I'm going to merge this because I have another patch that will heavily conflict. If there are any problems in the logic here I'm sure we'll hear about them. 😅

Issue #1609, opened 2021-02-20
Implement HasRawWindowHandle for x11, maybe GTK?


cmyr:
This was added for the mac and windows in #1586.

> curious to hear if this was a decision or an omission.

This was discussed. I have no access to non-Windows platforms and little knowledge about them, so rather than doing a blind implementation likely to be buggy, implementations on those missing platforms is left for a future change by whoever can contribute it. I did Windows and @rjwittams did macOS.

_Originally posted by @djeedai in https://github.com/linebender/druid/issues/1586#issuecomment-782599567_

djeedai:
Some notes: I looked at GTK and X11, and X11 is probably trivial (but I'd rather someone do it who can also test it). GTK however is an odd one because it's not a platform but already a platform abstraction itself, so this would require a bit of work to go fetch the right raw platform handle for the current platform config, if at all possible (I think it is, but couldn't find the appropriate wrapper in the Rust binding).

emi2k01:
GTK is also easy (for x11 atleast) but I don't think the bindings have an easy way to get the xdisplay.

There's (gdk_x11_window_get_xid)[https://developer.gnome.org/gdk3/stable/gdk3-X-Window-System-Interaction.html#gdk-x11-window-get-xid] to get the xlib window. The corresponding function in the bindings is (here)[https://docs.rs/gdkx11/0.9.0/gdkx11/struct.X11Window.html#method.get_xid].

To get the xdisplay we have (gdk_x11_display_get_xdisplay)[https://developer.gnome.org/gdk3/stable/gdk3-X-Window-System-Interaction.html#gdk-x11-display-get-xdisplay] but in the (bindings)[https://docs.rs/gdkx11/0.9.0/gdkx11/struct.X11Display.html] I don't see how I can get the raw pointer.

I opened an (issue)[https://github.com/gtk-rs/gtk-rs/issues/346] in the gtk-rs repo to see if I can get some help.

PR #1667, opened 2021-03-19
Implement HasRawWindowHandle for X11

i509VCB:
This covers the X11 part of #1609

I did test that this works and returns the handle, implementation pending approval of course.


jneem (approved):
Looks good, thanks!

PR #1677, opened 2021-03-28
Update for piet 0.4.0 and kurbo 0.8.0

### LINKED ISSUE #1622: Unable to load a font and use it in a label (Windows)
Tested on Windows 10 with the latest master (36610300).

Repo: https://github.com/SecondFlight/druid-font-repro

image

I include a font (Noto Sans Mono in this case) with

```rust
pub const FONT: &[u8] = include_bytes!("NotoSansMono-Regular.ttf");
pub const FONT_KEY: Key<FontDescriptor> = Key::new("fonts.NotoSansMono-Regular");
```

I then load it into the environment in my parent widget, like so:

```rust
fn lifecycle(&mut self, ctx: &mut LifeCycleCtx, event: &LifeCycle, data: &(), env: &Env) {
    if let LifeCycle::WidgetAdded = event {
        let mono_regular_family = ctx.text().load_font(FONT).unwrap();

        self.content = WidgetPod::new(Box::new(Padding::new(3., Child::new()).env_scope(
            move |env, _data| {
                let mono_regular = FontDescriptor::new(mono_regular_family.clone())
                    .with_weight(FontWeight::REGULAR);

                env.set(FONT_KEY, mono_regular.clone());
            },
        )));
    }
    
    // ...
}
```

Then, in a child widget, I try to use the font:
```rust
fn lifecycle(&mut self, ctx: &mut LifeCycleCtx, event: &LifeCycle, data: &(), env: &Env) {
    if let LifeCycle::WidgetAdded = event {
        self.content = WidgetPod::new(Box::new(
            Flex::column()
                .with_child(Label::new("This does have a font").with_font(env.get(FONT_KEY).with_size(15.)))
                .with_spacer(10.)
                .with_child(Label::new("This does not have a font")),
        ));
        ctx.children_changed();
    }

    // ...
}
```

I expect the first label to have a monospace font, but it has the same font as the second label.

cmyr:
Not a lot of breakage here, although there might be a bit more
upstream.

edit: this now includes an extra commit, which should fix #1622.

@SecondFlight would you like to verify that font loading works with this patch?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
       - name: install libx11-dev
         run: |
           sudo apt update
-          sudo apt install libx11-dev
+          sudo apt install libx11-dev libpango1.0-dev
         if: contains(matrix.os, 'ubuntu')
 
       - name: install stable toolchain
@@ -263,7 +263,7 @@ jobs:
       - name: install libx11-dev
         run: |
           sudo apt update
-          sudo apt install libx11-dev
+          sudo apt install libx11-dev libpango1.0-dev
         if: contains(matrix.os, 'ubuntu')
 
       - name: install nightly toolchain
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ You can find its changes [documented below](#070---2021-01-01).
 
 ### Highlights
 - International text input support (IME) on macOS.
+- Rich text and complex script support on Linux.
 
 ### Added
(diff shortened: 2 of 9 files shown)


SecondFlight:
I will verify this later today.

rschulman:
I started writing an app in druid earlier today, ran into https://github.com/linebender/piet/issues/427 a couple hours ago and came here to find a PR already in final review before I could even write up an issue. Now that's service!

SecondFlight:
:tada:
image

PR #1679, opened 2021-03-30
Ime position

cmyr:
Fixes an issue where the ime composition window could initially show up in the wrong position.

There are still problems with the window being positioned *slightly* badly (it can be a few pixels higher than expected) but I've experimented quite a bit and haven't had much more luck, and would prefer to move on.

progress on #1652


SecondFlight:
Closing per [this conversation on Zulip](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/druid.231679).

PR #1688, opened 2021-03-31
Changed Radio button color

arifd:
`theme::LABEL_COLOR` conflicts when `LABEL_COLOR` is changed. Seems more appropriate that it behaves similar to a cursor than to text. (which if I have dark text, the Radio button does not appear to change when selected)


arifd:
Oh okay, that works too. I'm really not an authority on how it should be implemented, but how it is currently implemented is giving me a problem :)

cmyr (approved):
Okay I don't think this patch is a great long-term solution but I don't think using `LABEL_COLOR` here is correct anyway so I'm happy to go with this as a "short-term" (😬) fix.

PR #1691, opened 2021-04-02
Adding protections for invalid flex parameters

arthmis:
Ok, not sure if there are better ways to clamp a value above a certain threshold. Also, is it fine to do the tests like that? I had to do it that way otherwise the test fails in release and passes debug.

### DIFF
--- a/druid/src/widget/flex.rs
+++ b/druid/src/widget/flex.rs
@@ -338,7 +338,6 @@ impl FlexParams {
     /// can pass an `f64` to any of the functions that take `FlexParams`.
     ///
     /// By default, the widget uses the alignment of its parent [`Flex`] container.
-    /// If the provided flex value is 0.0 or less, the flex value will default to 1.0.
     ///
     ///
     /// [`Flex`]: Flex
--- a/druid/src/widget/flex.rs
+++ b/druid/src/widget/flex.rs
@@ -1081,22 +1081,8 @@ mod tests {
         assert_eq!(vec(a, 38., 5), vec![4., 7., 8., 8., 7., 4.]);
         assert_eq!(vec(a, 39., 5), vec![4., 8., 7., 8., 8., 4.]);
     }
-    #[test]
-    #[cfg(not(debug_assertions))]
-    fn test_invalid_flex_params() {
-        use float_cmp::approx_eq;
-        let params = FlexParams::new(0.0, None);
-        approx_eq!(f64, params.flex, 1.0, ulps = 2);
-
-        let params = FlexParams::new(-0.0, None);
-        approx_eq!(f64, params.flex, 1.0, ulps = 2);
-
-        let params = FlexParams::new(-1.0, None);
-        approx_eq!(f64, params.flex, 1.0, ulps = 2);
-    }
 
     #[test]
-    #[cfg(debug_assertions)]
     #[should_panic]
     fn test_invalid_flex_params() {
         use float_cmp::approx_eq;
--- a/druid/src/widget/flex.rs
+++ b/druid/src/widget/flex.rs
@@ -341,20 +341,14 @@ impl FlexParams {
     /// If the provided flex value is 0.0 or less, the flex value will default to 1.0.
     ///
     ///
-    /// [`Flex`]: crate::widget::Flex
-    /// [`CrossAxisAlignment`]: crate::widget::CrossAxisAlignment
+    /// [`Flex`]: Flex
+    /// [`CrossAxisAlignment`]: CrossAxisAlignment
     pub fn new(flex: f64, alignment: impl Into<Option<CrossAxisAlignment>>) -> Self {
-        let flex = if flex > 0.0 {
-            flex
-        } else {
-            debug_assert!(
-                flex > 0.0,
-                "flex value should not be less than equal to 0.0. Flex given was: {}",
-                flex
-            );
-            tracing::warn!("Provided flex value was less than equal to 0.0: {}", flex);
-            1.0
-        };
+        if flex <= 0.0 {
+            debug_panic!("Flex value should be > 0.0. Flex given was: {}", flex);
+        }
+
+        let flex = flex.max(0.0);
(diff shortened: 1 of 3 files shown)


druid/src/widget/flex.rs
```diff
@@ -652,6 +680,15 @@ impl<T: Data> Widget<T> for Flex<T> {
...
                     *calculated_siz = kv.resolve(env);
+                    *calculated_siz = if *calculated_siz >= 0.0 {
+                        *calculated_siz
+                    } else {
+                        tracing::warn!(
+                            "Length provided to fixed spacer was les than 0: {}",
+                            *calculated_siz
+                        );
+                        0.0
+                    };
```
cmyr, line 691: I would probably just write this as,

```suggestion
                    if *calculated_siz < 0.0 {
                        tracing::warn!("Length provided to fixed spacer was less than 0");
                    }
                    *calculated_size = calculated_size.max(0.0);
```

cmyr (commented):
Thanks! A couple notes inline :)

druid/src/widget/flex.rs
```diff
@@ -416,7 +429,7 @@ impl<T: Data> Flex<T> {
     ///
     /// Convenient for assembling a group of widgets in a single expression.
     pub fn with_child(mut self, child: impl Widget<T> + 'static) -> Self {
-        self.add_flex_child(child, 0.0);
+        self.add_child(child);
```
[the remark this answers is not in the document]
arthmis: Ah so, if it's negative or 0.0 I should make it a non flex child? Would there still be a warning about invalid flex value?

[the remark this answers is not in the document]
cmyr: I'm just trying to think this through.

My basic thinking is that flex values are relative, and we have no real way of knowing that `1.0` is a useful or desirable value, and so we could go non-flex instead. I *think* that most people who would pass `0.0` *probably* want the non-flex behaviour? 

I could really go either way. One thing I would probably do is note in the log message that maybe they want `add_child` instead of `add_flex_child`, or whatever, and also a pointer to the Flex docs.

okay, I say: if flex is `0.0` add a non-flex child, and print  the following:

"Flex value should be > 0.0. To add a non-flex child use the add_child or with_child methods. See the Flex docs for more information."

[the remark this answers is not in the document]
arthmis: Ok yeah, I see your point. It's hard to know what the appropriate default flex value would be for that Flex widget. I'll do that. You said to point to the flex docs for info, but I feel like it's kind of barely mentioned in the reading. 
>Unlike a non-flex child, a child with a non-zero flex factor has a maximum allowed size on the main axis;

Wish there was a way to directly link that part. If this is somehow in the future being ignored by people after the warning then it's only a doc comment fix away to make it more transparent.

Anyways, I'll add these changes.

arthmis:
I opted to keep `add_child` within `with_child` instead of relying on the `add_flex_child` function checking for flex values that are 0.0 or less.

druid/src/widget/flex.rs
```diff
@@ -340,9 +340,15 @@ impl FlexParams {
     /// By default, the widget uses the alignment of its parent [`Flex`] container.
     ///
     ///
-    /// [`Flex`]: struct.Flex.html
-    /// [`CrossAxisAlignment`]: enum.CrossAxisAlignment.html
+    /// [`Flex`]: Flex
+    /// [`CrossAxisAlignment`]: CrossAxisAlignment
```
cmyr, line 344:
```suggestion
```
We can just get rid of this completely and it should work.

druid/src/widget/flex.rs
```diff
@@ -533,24 +543,25 @@ impl<T: Data> Flex<T> {
     /// my_row.add_flex_child(Slider::new(), FlexParams::new(1.0, CrossAxisAlignment::End));
     /// ```
     ///
-    /// [`FlexParams`]: struct.FlexParams.html
-    /// [`with_flex_child`]: #method.with_flex_child
+    /// [`FlexParams`]: FlexParams
+    /// [`with_flex_child`]: Flex::with_flex_child
```
cmyr, line 547:
```suggestion
    /// [`with_flex_child`]: Flex::with_flex_child
```

druid/src/widget/flex.rs
```diff
@@ -533,24 +543,25 @@ impl<T: Data> Flex<T> {
...
-            Child::Fixed {
+        let child = if params.flex > 0.0 {
+            Child::Flex {
                 widget: WidgetPod::new(Box::new(child)),
                 alignment: params.alignment,
+                flex: params.flex,
             }
         } else {
-            Child::Flex {
+            tracing::warn!("Flex value should be > 0.0. To add a non-flex child use the add_child or with_child methods. See the Flex docs for more information: https://docs.rs/druid/0.7.0/druid/widget/struct.Flex.html");
```
cmyr, line 561:
```suggestion
            tracing::warn!("Flex value should be > 0.0. To add a non-flex child use the add_child or with_child methods.\n\
                            See the docs for more information: https://docs.rs/druid/0.7.0/druid/widget/struct.Flex.html");
```

cmyr: my formatting might not be good; I just want to break this across two lines.

cmyr (approved):
Looks good! A few little formatting nits, but happy to see this merged. :)

druid/src/widget/flex.rs
arthmis: oooohhhh ok. I did not know this.

cmyr:
What's the conclusion here?

arthmis:
In regards to what? Are there any other issues? The only one that came up was the tracing events need a subscriber created before the app is launched but that is a separate issue. 

oh oops, I'm going to take out the tracing instrument for the add flex child function. I forgot. Other than that I think everything is fine.

cmyr (approved):
Thanks!

PR #1692, opened 2021-04-02
fixes #1690 (WindowState::Maximized on macOS doesn't lay out widgets …

### LINKED ISSUE #1690: Starting an app with `WindowState::Maximized` doesn't work quite right on macOS
I'm on macOS 10.15.5 and using the git version of Druid (post-0.7.0).

After posting a topic and repro on Zulip, we discovered that this code properly creates a maximized window on Windows, but not on macOS. The issue is that the window *is* maximized, but the widget *layout* is not. It still thinks the window is like 400x300 pixels or something. If I then resize the window, the layout corrects itself.

```rust
use druid::*;
use druid::widget::*;

#[derive(Clone, Data, Lens)]
struct AppState {}

impl Default for AppState {
    fn default() -> Self {
        Self {}
    }
}

fn build_ui() -> impl Widget<AppState> {
    Split::columns(Label::new("Split A"), Label::new("Split B"))
    .split_point(0.5)
    .draggable(true)
    .solid_bar(true)
    .min_size(250.0, 600.0)
}

    AppLauncher::with_window(
        WindowDesc::new(build_ui())
        .set_window_state(WindowState::Maximized)
    )
    .launch(AppState::default())
}
```

Screenshot of what it looks like:

image

JarrettBillingsley:
…properly at startup).

### DIFF
--- a/druid-shell/src/platform/mac/window.rs
+++ b/druid-shell/src/platform/mac/window.rs
@@ -294,6 +294,9 @@ impl WindowBuilder {
                 handle.set_level(level)
             }
 
+            // set_window_state above could have invalidated the frame size
+            let frame = NSView::frame(content_view);
+
             (*view_state).handler.connect(&handle.clone().into());
             (*view_state).handler.scale(Scale::default());
             (*view_state)


cmyr (approved):
Thanks!

PR #1695, opened 2021-04-04
GTK Shell: Manually get/set textual data in clipboard

### LINKED ISSUE #981: Strange clipboard carriage return behavior
To be honest I'm not even sure what to title this one. I am on master (875ad4055174072f167c8add1571cdaa51c9dfc5) and the GTK backend. The strange behaviour is that any carriage return which is placed into the clipboard by a druid app is either stripped or converted to a line feed but not really. Placing "\r\n" into the clipboard will result in druid only being able to retrieve "\n" from the clipboard afterwards but other applications can retrieve the expected "\r\n".  Placing "\r" in the clipboard results in druid retrieving a "\n" and other applications retrieving "\r\n". Similarly placing "\rtext" into the clipboard results in druid retrieving "\ntext" but other applications retrieving "\r\ntext". Finally copying "\r\n" from a separate application (VSCode in this case) will allow druid to retrieve the expected "\r\n".

I've looked at the GTK shell implementation of `put_string` and `get_string` and they seem dead simple which causes me to think that it might be an upstream issue with either gtk-rs or GTK itself. I'm not sure yet but I wanted to open this so other backends could be checked as a sanity check and anyone who might know something pertinent might be able to share.

Here is an example program which reproduces the issue: https://gist.github.com/ForLoveOfCats/c037b8a901cdbfbfb0f7700a99887cfe

ForLoveOfCats:
Fixes #981

~~It seems that newlines were not being mangled on the outbound direction, only on the inbound.~~ By manually retrieving the text and converting to a string we avoid GTK from performing that behavior when getting clipboard contents. This also appears to work on Wayland in both Plasma 5 KWin and Weston.

Verification that this behaves correctly on X11 and Wayland would be appreciated :)

### DIFF
--- a/druid-shell/src/platform/gtk/clipboard.rs
+++ b/druid-shell/src/platform/gtk/clipboard.rs
@@ -19,16 +19,36 @@ use gtk::{TargetEntry, TargetFlags};
 
 use crate::clipboard::{ClipboardFormat, FormatId};
 
+const CLIPBOARD_TARGETS: [&str; 5] = [
+    "UTF8_STRING",
+    "TEXT",
+    "STRING",
+    "text/plain;charset=utf-8",
+    "text/plain",
+];
+
 /// The system clipboard.
 #[derive(Debug, Clone)]
 pub struct Clipboard;
 
 impl Clipboard {
     /// Put a string onto the system clipboard.
-    pub fn put_string(&mut self, s: impl AsRef<str>) {
+    pub fn put_string(&mut self, string: impl AsRef<str>) {
+        let string = string.as_ref().to_string();
+
         let display = gdk::Display::get_default().unwrap();
         let clipboard = gtk::Clipboard::get_default(&display).unwrap();
-        clipboard.set_text(s.as_ref())
+
+        let targets: Vec<TargetEntry> = CLIPBOARD_TARGETS
+            .iter()
+            .enumerate()
+            .map(|(i, target)| TargetEntry::new(target, TargetFlags::all(), i as u32))
+            .collect();
+
+        clipboard.set_with_data(&targets, move |_, selection, _| {
+            const STRIDE_BITS: i32 = 8;
+            selection.set(&selection.get_target(), STRIDE_BITS, string.as_bytes());
+        });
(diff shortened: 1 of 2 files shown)


druid-shell/src/platform/gtk/clipboard.rs
```diff
@@ -62,7 +62,25 @@ impl Clipboard {
...
+            "text/plain;charset=utf-8",
+            "text/plain",
+        ];
+
+        for target in &targets {
+            let atom = Atom::intern(target);
+            if let Some(selection) = clipboard.wait_for_contents(&atom) {
+                return std::str::from_utf8(&selection.get_data())
+                    .ok()
+                    .map(|text| text.to_string());
```
cmyr, line 79: Probably easier to just use the method on `String` directly, like:

```suggestion
                return String::from_utf8(&selection.get_data()).ok();
```

cmyr (approved):
Looks good!

druid-shell/src/platform/gtk/clipboard.rs
ForLoveOfCats: I believe that I've mentioned before that I love the Rust standard library

ForLoveOfCats: Ah ha, and because `String::from_utf8` can take ownership of the `Vec<u8>` returned by `.get_data()` we save a clone :)

cmyr: oh, right, so my suggestion isn't quite right, we don't want that borrow in there.

ForLoveOfCats:
Okay so after playing with this some more I realized that the newline chars *were* being mangled by GTK but only on application shutdown due to some interaction with the clipboard manager. This PR now also manually provides the textual data to the requesting application instead of using GTK's `clipboard.set_text`. I wish I had caught this earlier but the X11 clipboard has so many strange edge cases and interactions so I simply didn't think about how it looked post-application exit :shrug:

cmyr (approved):
Looks good!

PR #1701, opened 2021-04-07
Add support for transparent titlebar on Mac

jpochyla:
Adds two `WindowBuilder` methods, `show_title` and `set_transparent_titlebar`.

When used together, only the window controls are left:

image

### DIFF
--- a/druid-shell/src/platform/gtk/window.rs
+++ b/druid-shell/src/platform/gtk/window.rs
@@ -33,7 +33,7 @@ use gtk::{AccelGroup, ApplicationWindow, DrawingArea, SettingsExt};
 use tracing::{error, warn};
 
 #[cfg(feature = "raw-win-handle")]
-use raw_window_handle::{HasRawWindowHandle, RawWindowHandle};
+use raw_window_handle::{unix::XcbHandle, HasRawWindowHandle, RawWindowHandle};
 
 use crate::kurbo::{Insets, Point, Rect, Size, Vec2};
 use crate::piet::{Piet, PietText, RenderContext};
--- a/druid-shell/src/platform/gtk/window.rs
+++ b/druid-shell/src/platform/gtk/window.rs
@@ -220,8 +220,16 @@ impl WindowBuilder {
         self.show_titlebar = show_titlebar;
     }
 
+    pub fn show_title(&mut self, _show_title: bool) {
+        // Ignored
+    }
+
     pub fn set_transparent(&mut self, transparent: bool) {
         self.transparent = transparent;
+    }
+
+    pub fn set_transparent_titlebar(&mut self, _transparent_titlebar: bool) {
+        // Ignored
     }
 
     pub fn set_position(&mut self, position: Point) {
--- a/druid-shell/src/platform/mac/window.rs
+++ b/druid-shell/src/platform/mac/window.rs
@@ -25,6 +25,7 @@ use block::ConcreteBlock;
 use cocoa::appkit::{
     CGFloat, NSApp, NSApplication, NSAutoresizingMaskOptions, NSBackingStoreBuffered, NSColor,
     NSEvent, NSView, NSViewHeightSizable, NSViewWidthSizable, NSWindow, NSWindowStyleMask,
+    NSWindowTitleVisibility,
 };
 use cocoa::base::{id, nil, BOOL, NO, YES};
 use cocoa::foundation::{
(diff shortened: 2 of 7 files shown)


cmyr:
I have some concerns with these; do they have reasonable cross-platform implementations? If not, it might be time to introduce a platform-specific `WindowBuilderExt` trait, and only expose these methods on the mac?

jpochyla:
Yes I understand the concern. I'm not familiar with Windows customs, it seems that at least [with WPF it's possible](https://docs.microsoft.com/en-us/windows/uwp/design/shell/title-bar#full-customization), but i.e.`winit` seems to [do this with an ext trait](https://github.com/rust-windowing/winit/pull/423). I might get some inspiration there and try a similar approach?

cmyr:
This is a good point. I think we would need separate `Ext` traits in shell and druid proper.

A good place to look for prior art is in `std`, where there are a bunch of examples of this pattern; see [`DirBuilderExt`](http://doc.rust-lang.org/1.51.0/std/os/unix/fs/trait.DirBuilderExt.html) and [`ExitStatusExt`](http://doc.rust-lang.org/1.51.0/std/os/unix/process/trait.ExitStatusExt.html) for example; there are probably ten or so of these traits.

m-lima:
@jpochyla I'm happy to help with this PR. Just ping me if I can be of **any** help.

This looks like a great addition, in my opinion :)
Allows the apps to look more modern and a better UX when built for web + desktop with closer rendering between the two targets.

jpochyla:
Hi @m-lima, I'm AFK until this weekend, I'll have time to focus on this then.

ikeofkc:
I think every major platform has moved mainly to non traditional, customizable titlebars, sometimes called "headerbars".  I think this is actually the default design in major platforms except in edge cases when it's not viable.  

The Mac and Windows implementations have been mentioned above.  On Linux this is the case also with GNOME and Elementary, which uses GTK as well.  GNOME or at least GTK apps are the default on most distros [ubuntu, fedora] and GTK3 & 4 encourages headbar (custom titlebar) apps whenever possible. KDE doesn't use them, but their convergent [Mauikit apps](https://mauikit.org/) do.

- [GNOME Human Interface Guidelines](https://developer.gnome.org/hig/patterns/containers/header-bars.html)
- [Gtk.Headerbar docs](https://docs.gtk.org/gtk4/class.HeaderBar.html)
- [Gnome's CSD Initiative](https://blogs.gnome.org/tbernard/2018/01/26/csd-initiative/)

So not only do I think this should be thought of as a platform agnostic goal, I think it should really be the default way to style apps in druid, as they are more likely to look modern, and it's easier to control the look (including titlebar) on each platform, even if they may look less native. However, I feel like the modern desktop experience is accepting inconsistent app looks (varying toolkits) in favor of apps consistently looking as devs intended.

longmathemagician:
> Yes I understand the concern. I'm not familiar with Windows customs, it seems that at least [with WPF it's possible](https://docs.microsoft.com/en-us/windows/uwp/design/shell/title-bar#full-customization), but i.e.`winit` seems to [do this with an ext trait on Mac](https://github.com/rust-windowing/winit/pull/423).

You don't actually need WPF, I got it to work with druid. 

On Windows 11 builds 22000 through 22494 a window can request to be drawn with mica through an undocumented attribute (u32: 1029) sent to DwmSetWindowAttribute in combination with manually drawing the window title in the non client area. In newer builds you can request mica or acrylic just by passing a DWM_SYSTEMBACKDROP_TYPE.

Any interest in reviving this?

xStrom:
What's the status here? Can we move this forward?

I don't think we need to worry about the extension traits right now in this PR. That can be a follow-up PR later.

I could review this, but I need this rebased on `master` for the merge conflicts to be resolved.

@jpochyla do you still have any interest in helping land this?

longmathemagician:
If we do want to move this forward I can contribute support for Windows (11, 22H2+) and GNOME/KDE (to the extent that such things are supported, which is minimal). Since it's a new feature and requires a rebase anyway, it might be worth focusing on pushing it to Glazier instead?

xStrom:
We would port the changes to glazier for sure. Whether to do the work only there or not, I don't really care. I think @jpochyla's opinion matters, because he has a big app using Druid, so this feature might be of direct value for him.

jpochyla:
Hi @xStrom, sadly I don't have bandwidth to land this ATM :( Thanks a lot for thinking of me!

longmathemagician:
@xStrom I wasn't aware that @jpochyla had an actively developed use case for this. I can rebase and test but it could be a week or more (my ``mac'' didn't survive the upgrade to Ventura and I'll have to deal with that first). Is this a blocker for 0.8.0?

xStrom:
No it is not a blocker for 0.8.0. It is an interesting feature though and so we can still look into it, but no rush.

Issue #1709, opened 2021-04-09
TextBox may paint cursor outside of its bounds


andrewhickman:
There are a couple of ways to reproduce this:

- Scroll within the textbox such that the cursor position is no longer visible
- Resize the textbox so the cursor position is hidden

In both cases the cursor is still painted

![image](https://user-images.githubusercontent.com/19718247/114250568-0a301700-9996-11eb-96e6-c2d5a600b350.png)

andrewhickman:
This also occurs with the scrollbar

![image](https://user-images.githubusercontent.com/19718247/114250903-21bbcf80-9997-11eb-838d-d472a84d4501.png)

PR #1712, opened 2021-04-10
Fix clipping of textbox cursor

andrewhickman:
This fixes the cursor clipping issues described in #1709. (I think the scrollbar clipping issue is purely an issue with the `Scroll` component, I'll fix that bit next.

As mentioned [on Zulip](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/druid.231709.20.28textbox.20clipping.29), this changes the appearance of the textbox when scrolling:

**Before:**

before

**After:**

image

Note the padding around the scroll bar is now smaller. We can change that if desired by overriding `theme::SCROLLBAR_PAD` but for now I've left it.

### DIFF
--- a/druid/src/widget/textbox.rs
+++ b/druid/src/widget/textbox.rs
@@ -284,7 +284,7 @@ impl<T> TextBox<T> {
         self.inner.child().wrapped()
     }
 
-    /// An mutable reference to the inner [`TextComponent`].
+    /// A mutable reference to the inner [`TextComponent`].
     ///
     /// Using this correctly is difficult; please see the [`TextComponent`]
     /// docs for more information.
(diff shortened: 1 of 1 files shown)


druid/src/widget/textbox.rs
```diff
@@ -282,15 +281,15 @@ impl<T> TextBox<T> {
     /// Using this correctly is difficult; please see the [`TextComponent`]
     /// docs for more information.
     pub fn text(&self) -> &TextComponent<T> {
-        self.inner.wrapped().child()
+        self.inner.child().wrapped()
     }
 
-    /// An immutable reference to the inner [`TextComponent`].
+    /// An mutable reference to the inner [`TextComponent`].
```
jneem, line 287: *A* mutable

jneem (approved):
Looks good to me, just one tiny typo.

druid/src/widget/textbox.rs
andrewhickman: [Muphry's law](https://en.wikipedia.org/wiki/Muphry%27s_law) strikes again 🤦

cmyr (approved):
Thanks!

PR #1715, opened 2021-04-10
Improve layout of the scrollbar

andrewhickman:
This PR fixes the scrollbar clipping observed here https://github.com/linebender/druid/issues/1709#issuecomment-817032228. There are a few small changes, best viewed as individual commits

- Merged some code which was duplicated for horizontal/vertical cases
- I did a bunch of testing with super small scroll components and fixed some cases where the scrollbar was drawn with negative size or outside the boundaries
- The scroll component reserves some vertical padding space for the horizontal scroll bar (and vice versa). I've changed it to only do this if both scroll bars are enabled, otherwise it looks a bit odd
- `TextBox` wasn't actually benefiting from the last point because it sets `Scroll::set_horizontal_scroll_enabled` but that didn't change whether the scrollbar is actually enabled - that is now fixed.


SecondFlight:
Hey, looks like this is missing a [changelog](https://github.com/linebender/druid/blob/master/CHANGELOG.md) update. Just want to make sure that it's intentional or otherwise that you get a chance to add this there.

andrewhickman:
> Hey, looks like this is missing a [changelog](https://github.com/linebender/druid/blob/master/CHANGELOG.md) update. Just want to make sure that it's intentional or otherwise that you get a chance to add this there.

Completely forgot, thanks 👍 . I've updated it.

xarvic (approved):
Looks good, thank you!

Issue #143, opened 2019-09-07
Add "enabled" state for interactive widgets


futurepaul:
As mentioned in #134,  we need some sort of enabled / disabled state for the button widget. It also makes sense for other widgets: slider, checkbox (I'll have a PR for that soon), textbox, scrollbar, and I'm sure other widgets to come.

To kick things off, here's @raphlinus's comment:
> ...what I was thinking (and this is borrowed to some extent from SwiftUI) is that the enabled state becomes part of env (probably with specific API surface for it; doesn't have to go through the typed key/value mechanism) and widgets respond to it. There are definitely questions though.

This sounds pretty straightforward to implement. I'm not sure what the questions are, though!

SecondFlight:
Closing as resolved via #1717.

PR #1702, opened 2021-04-07
DisabledIf widget

xarvic:
Adds a widget wrapper that disables the inner widget if the provided closure returns true.
For the moment the only effect is that the text boxes annotated with "(disabled)" can't be focused by pressing Tab. The visual changes and changes in behavior are implemented on the widgets themselves in a following PR.

PR #1717, opened 2021-04-13
Implement disabled for druid controls

xarvic:


xarvic:
This is weird. Github believes that i made the change from #1702 here because i merged the branches before creating the PR :/ I hope this will be fine.

cmyr:
@xarvic if you like you could just squash locally? Otherwise we can squash afterwards, I'm not too concerned.

druid/src/widget/button.rs
```diff
@@ -175,7 +177,16 @@ impl<T: Data> Widget<T> for Button<T> {
             .inset(-stroke_width / 2.0)
             .to_rounded_rect(env.get(theme::BUTTON_BORDER_RADIUS));
 
-        let bg_gradient = if is_active {
+        let bg_gradient = if ctx.is_disabled() {
```
cmyr, line 180: I don't find this particularly communicates "disabledness"; what's most important here for me is dimming the text color.

(I acknowledge that the current widget styling wasn't really designed with a concept of "disabledness",  so it's hard to make it fit)

druid/src/widget/checkbox.rs
```diff
@@ -145,7 +147,13 @@ impl Widget<bool> for Checkbox {
...
                 .line_join(LineJoin::Round);
 
-            ctx.stroke_styled(path, &env.get(theme::LABEL_COLOR), 2., &style);
+            let brush = if ctx.is_disabled() {
+                env.get(theme::DISABLED_FOREGROUND_LIGHT)
+            } else {
+                env.get(theme::LABEL_COLOR)
+            };
+
+            ctx.stroke_styled(path, &brush, 2., &style);
```
cmyr, line 156: Unfortunately this makes the checkmark not be visible at all, for me?

also: repeating myself, but I think the most important thing is to draw the text dimmed.

druid/src/widget/slider.rs
```diff
@@ -184,31 +190,37 @@ impl Widget<f64> for Slider {
...
+                (
+                    env.get(theme::FOREGROUND_LIGHT),
+                    env.get(theme::FOREGROUND_DARK),
+                ),
+            )
         };
 
         //Paint the border
-        let border_color = if is_hovered || is_active {
+        let border_color = if (is_hovered || is_active) && !ctx.is_disabled() {
```
cmyr, line 223: this looks pretty jarring, I might use a darker color when disabled?

druid/src/widget/switch.rs
```diff
@@ -256,27 +275,33 @@ impl Widget<bool> for Switch {
...
-            UnitPoint::BOTTOM,
-            (
-                env.get(theme::FOREGROUND_DARK),
-                env.get(theme::FOREGROUND_LIGHT),
-            ),
-        );
-
-        let knob_gradient = if is_active {
-            flipped_knob_gradient
+        let knob_gradient = if ctx.is_disabled() {
```
cmyr, line 278: I think the most important thing here is that we not show the blue accent color, even if the switch is disabled in the 'on' state.

cmyr (commented):
A few notes inline, but basically I think this is good; I just have two real notes about the styling.

- Anything that displays text, when disabled, should dim that text. This means the textbox, but also labels anywhere they appear, like the checkbox or the the radio button, or the text in the switch.

SecondFlight:
Will this resolve #143 once it merges?

xarvic:
Yes and it should also close #746

druid/src/widget/slider.rs
xarvic: i made the knob brighter again, it looks a bit better now. What do you think?
image

druid/src/widget/checkbox.rs
xarvic: Im using `TEXT_COLOR` and `DISABLED_TEXT_COLOR` now, does this change anything?

druid/src/theme.rs
```diff
@@ -34,6 +34,10 @@ pub const BACKGROUND_LIGHT: Key<Color> = Key::new("org.linebender.druid.theme.ba
 pub const BACKGROUND_DARK: Key<Color> = Key::new("org.linebender.druid.theme.background_dark");
 pub const FOREGROUND_LIGHT: Key<Color> = Key::new("org.linebender.druid.theme.foreground_light");
 pub const FOREGROUND_DARK: Key<Color> = Key::new("org.linebender.druid.theme.foreground_dark");
+pub const DISABLED_FOREGROUND_LIGHT: Key<Color> =
+    Key::new("org.linebender.druid.theme.disabled_foreground_light");
+pub const DISABLED_FOREGROUND_DARK: Key<Color> =
+    Key::new("org.linebender.druid.theme.disabled_foreground_dark");
```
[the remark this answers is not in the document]
xarvic: Ok

xarvic:
I changed the look:
image
image

druid/src/widget/slider.rs
cmyr: This looks good!

druid/src/theme.rs
```diff
@@ -23,7 +23,9 @@ use crate::{Env, FontDescriptor, FontFamily, FontStyle, FontWeight, Insets, Key}
 pub const WINDOW_BACKGROUND_COLOR: Key<Color> =
     Key::new("org.linebender.druid.theme.window_background_color");
 
-pub const LABEL_COLOR: Key<Color> = Key::new("org.linebender.druid.theme.label_color");
```
cmyr, line 26: Because this is heavily used elsewhere (including external projects) I would keep it around with a deprecation notice, like:

```rust
#[deprecated(since = "0.8.0", note = "renamed to TEXT_COLOR")]
pub const LABEL_COLOR: Key<Color> = TEXT_COLOR;
```

druid/src/widget/label.rs
```diff
@@ -96,6 +96,9 @@ pub struct Label<T> {
 pub struct RawLabel<T> {
     layout: TextLayout<T>,
     line_break_mode: LineBreaking,
+
+    control_text: bool,
```
cmyr, line 100: I would be more explicit about this; it's really only for widget implementors, so I would give it a longer, more explicit name. "is_control_component"?

druid/src/widget/label.rs
```diff
@@ -548,14 +585,23 @@ impl<T: TextStorage> Widget<T> for RawLabel<T> {
...
-        if matches!(event, LifeCycle::WidgetAdded) {
-            self.layout.set_text(data.to_owned());
+    #[instrument(name = "RawLabel", level = "trace", skip(self, ctx, event, data, _env))]
+    fn lifecycle(&mut self, ctx: &mut LifeCycleCtx, event: &LifeCycle, data: &T, _env: &Env) {
+        match event {
+            LifeCycle::WidgetAdded => {
+                self.layout.set_text(data.to_owned());
+            }
+            LifeCycle::DisabledChanged(disabled) if self.control_text => {
+                dbg!("Control!");
```
cmyr, line 595:
```suggestion
```

druid/src/widget/label.rs
```diff
@@ -175,6 +180,14 @@ impl<T: TextStorage> RawLabel<T> {
         self
     }
 
+    /// Builder-style method for making this label into a control-label.
+    ///
+    /// If this label is a control-label it will change its color when disabled.
+    pub fn control_text(mut self) -> Self {
```
cmyr, line 186: ditto, I'd go with something more explicit maybe? like `control_component` or even `dims_when_disabled`?


Did you consider having labels *always* draw themselves dimmed when they're disabled? I could imagine there maybe being some downsides to this but they aren't totally clear to me..

cmyr (commented):
This looks much better, visually. a few little comments and questions inline, but I think we're very close!

druid/src/widget/label.rs
xarvic: > Did you consider having labels always draw themselves dimmed when they're disabled? I could imagine there maybe being some downsides to this but they aren't totally clear to me..

I don't have a strong opinion how to do this, if you want i change it. If we do this we could at least have the option to disable the color change, but enabling it by default.

xarvic:
Here is the next version:
image

cmyr:
Sorry to let this sit, will try to give it one more thorough look-through tomorrow. :)

cmyr (approved):
Sorry for the delay, I think this looks great. If there are any remaining issues we can sort them out afterwards.

Thanks again for taking this on!

xarvic:
No problem, thanks for the review :)

Issue #1716, opened 2021-04-12
Clarify the relation of focus and `propagate_to_hidden`


xarvic:
As i understand it, hidden in `propagate_to_hidden()` means the widget, is inside the widget tree, but can't be seen. Not because it is overlaid by another component, but because it doesn't have a position inside the current window (like a widget inside a hidden Tab). On the other hand, a component in a scroll would not be "hidden" according to my understanding, even if you can't see it right now.
Therefore a component can participate in focus if and only if it is not hidden. 
Which means for one thing that my implementation for the `Lifecycle::BuildFocusChain` in #1632 is wrong since it includes all hidden tabs. ~~But also, to fix this we would probably need a context method `children_state_changed()` which would request a rebuild of the focus_chain (for an example when selecting another Tab).~~ Calling children changed should be enough.

Another thing we should think about is what happens to focused widgets which become hidden.

cmyr:
Your understanding is correct; a 'hidden' widget is a widget that is part of the widget tree but is not currently a functional part of the UI, generally because it is part of a switch/either/tab that is not visible.

I also agree that a widget that is in a scroll view and is not onscreen is not "hidden".

xarvic:
Ok, then i will create a PR with the changes. I would like test in `BuildFocusChain` if the focused widget is still a part of the non hidden tree, send ResignFocus otherwise. Is that ok? Also to  unfocus the widget in the hidden part, i would like to change `Lifecycle::should_propagate_to_hidden() -> bool` to `Lifecycle::propagate_to_hidden() -> Option<Self>`. Then we could change the focus event into a focus event with only the resign part.

xarvic:
Ok, then i will create a PR with the changes. I would like to test in `BuildFocusChain` if the focused widget is still a part of the non hidden tree, send ResignFocus otherwise. Is that ok? I also would like to change `Lifecycle::should_propagate_to_hidden() -> bool` to `Lifecycle::propagate_to_hidden() -> Option<Self>`. Then we could change the focus event into a focus event with only the resign part. Therefore even if someone tries to focus a hidden widget he won't succeed, but were still able send the unfocus to the specific widget.

cmyr:
[...]

PR #1724, opened 2021-04-17
Focus-chain update

xarvic:
This is a fix of the part of #1716 @cmyr and i agreed on:

- `should_propagate_to_hidden()` now returns `false` for `Lifecycle::BuildFocusChain`
- `Lifecycle::BuildFocusChain` requests a a resign of focus if `state.has_focus` is set but neither one of its hidden children nor it self has focus.

The other point: What should happen if try to focus a WidgetID which is not allowed to recieve focus like `ctx.set_focus(WidgetID::new())` or focusing a hidden widget is not addressed here.


druid/src/core.rs
```diff
@@ -1082,6 +1088,17 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
             // Update focus-chain of our parent
             LifeCycle::BuildFocusChain => {
                 self.state.update_focus_chain = false;
+
+                // had_focus is the old focus value. state.has_focus was repaced with ctx.is_focused().
+                // Therefore is had_focus is true but state.has_focus is false the widgets, which is
+                // currently focused is not part of the functional tree anymore
+                // (Lifecycle::BuildFocusChain.should_propagate_to_hidden() is false!) and should
+                // resign the focus.
```
cmyr, line 1096:
```suggestion
                // had_focus is the old focus value. state.has_focus was repaced with ctx.is_focused().
                // Therefore if had_focus is true but state.has_focus is false then the widget which is
                // currently focused is not part of the functional tree anymore
                // (Lifecycle::BuildFocusChain.should_propagate_to_hidden() is false!) and should
                // resign the focus.
```

druid/src/core.rs
```diff
@@ -1082,6 +1088,17 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
...
+                if had_focus && !self.state.has_focus {
```
cmyr, line 1097: I'm confused; we haven't done `merge_up` yet, so how would `has_focus` have changed?

cmyr (commented):
I'm a bit confused here, and will need to take a second look (or maybe we should have a quick chat about it) I'm having a slow time remembering the context for this. In any case, one question inline.

druid/src/core.rs
xarvic: The `merge_up` of our children happened already in line 1058 / 1062. If someone hides a focused widget and then calls `children_changed()`, we clear `has_focus` in `BuildFocusChain` if has_clear is not set after the calls in 1058 and 1062 we know, that the focused widget is not reachable

cmyr (approved):
Okay, that explanation works for me, thanks!

Issue #1725, opened 2021-04-17
Split cursor flickering due to layout lag w.r.t. cursor fast moves


djeedai:
The `list` widget does some hit-test of the mouse over the split bar during each mouse move. When moving fast (at least on Windows 10), the hardware cursor behind much faster than the Druid layout, it can happen that the cursor briefly moves out of the bar area, and back again the next frame. Because the cursor is set for each mouse move event, this produces some flickering.

https://github.com/linebender/druid/blob/5958963a05c18c7943256fb8724f7637bd524e8e/druid/src/widget/split.rs#L304-L313

The cursor should instead be locked when the bar is active (mouse down) and reset when it becomes inactive (mouse up). I've implemented a fix and it looks nicer, closer to the behavior of other controls like a scroll bar knob being dragged even if the mouse is not perfectly hovering it.

PoignardAzur:
I think this issue can be closed?

PoignardAzur:
I tested the split example and I think the PR fixed the problem. Closing this issue.

PR #1726, opened 2021-04-17
Lock split cursor state when active

djeedai:
Ensure the cursor state of the `split` widget is locked when the split
bar is active, to avoid flickering when the mouse moves faster than the
widget can re-layout.

Bug: #1725

### DIFF
--- a/druid/src/widget/split.rs
+++ b/druid/src/widget/split.rs
@@ -35,6 +35,10 @@ pub struct Split<T> {
     /// if the mouse moves faster than the layout and temporarily gets outside of the
     /// bar area while still being dragged.
     is_bar_hover: bool,
+    /// Offset from the split point (bar center) to the actual mouse position when the
+    /// bar was clicked. This is used to ensure a click without mouse move is a no-op,
+    /// instead of re-centering the bar on the mouse.
+    click_offset: f64,
     child1: WidgetPod<T, Box<dyn Widget<T>>>,
     child2: WidgetPod<T, Box<dyn Widget<T>>>,
 }
@@ -59,6 +63,7 @@ impl<T> Split<T> {
             solid: false,
             draggable: false,
             is_bar_hover: false,
+            click_offset: 0.0,
             child1: WidgetPod::new(child1).boxed(),
             child2: WidgetPod::new(child2).boxed(),
         }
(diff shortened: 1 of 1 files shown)


druid/src/widget/split.rs
```diff
@@ -30,6 +30,11 @@ pub struct Split<T> {
     min_bar_area: f64,    // Integers only
     solid: bool,
     draggable: bool,
+    /// The split bar is hovered by the mouse. This state is locked to `true` if the
+    /// widget is active (the bar is being dragged) to avoid cursor and painting jitter
+    /// if the mouse moves faster than the layout and temporarily gets outside of the
+    /// bar area while still being dragged.
+    is_bar_hover: bool,
```
cmyr, line 37: Instead of adding a new variable, can we just check `ctx.is_active`?

djeedai: Makes sense I guess. I was porting that from a custom control, didn't realize it's probably not needed. I'll update that PR over the week-end.

djeedai: @cmyr : Actually yes and no (but mostly no from a performance standpoint).

`ctx.is_active()` has already changed when the `MouseMove` event is raised, so we don't know its old value, and therefore cannot know if there is a change between the last and current state. Now, a better way is to move the `set_cursor()` to the `MouseDown` and `MouseUp` in the active case, which doesn't require any variable. And keep the hover check in the non-active case in `MouseMove`. But, in that latter case, if we don't know the previous hover state then we are still forced to be conservative and set the cursor / request a repaint for every single `MouseMove` event. For the price of a single `bool` variable we can avoid that, and I think that's a reasonable price to pay.

druid/src/widget/split.rs
```diff
@@ -30,6 +30,15 @@ pub struct Split<T> {
...
+    /// Offset from the split point (bar center) to the actual mouse position when the
+    /// bar was clicked. This is used to ensure a click without mouse move is a no-op,
+    /// instead of re-centering the bar on the mouse.
+    click_offset: f64,
```
djeedai, line 41: @cmyr I actually found and fixed some other behavior while refactoring as explained in the other comment : when clicking and releasing without moving the mouse, because the split point is conceptually 1 unit but the bar is generally larger, we were snapping the bar to the actual mouse cursor position. I checked in other implementations like VSCode and they don't do that; instead they consider that no mouse move is a no-op. To do that, I added a delta between the click position and the actual split point. That ensures that the bar is always moved such that the mouse cursor appears to be static (in the split direction) relative to the bar.

cmyr (approved):
Okay, thank you for the clear explanation!

PR #1730, opened 2021-04-21
Fix link in documentation of widget::Image

RichardPoole42:
struct ImageBuf now lives in piet not in druid-shell

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,7 @@ You can find its changes [documented below](#070---2021-01-01).
 - Fixed docs describing `ViewSwitcher` widget functionality ([#1693] by [@arthmis])
 - Added missing documentation on derived lens items ([#1696] by [@lidin])
 - Fix example code in `Get started with Druid` chapter of book ([#1698] by [@ccqpein])
+- Fix link in documentation of widget::Image ([#1730] by [@RichardPoole42])
 
 ### Examples
 
@@ -457,6 +458,7 @@ Last release without a changelog :(
 [@xarvic]: https://github.com/xarvic
 [@arthmis]: https://github.com/arthmis
 [@ccqpein]: https://github.com/ccqpein
+[@RichardPoole42]: https://github.com/RichardPoole42
 
 [#599]: https://github.com/linebender/druid/pull/599
 [#611]: https://github.com/linebender/druid/pull/611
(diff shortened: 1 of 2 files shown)


druid/src/widget/image.rs
```diff
@@ -69,7 +69,7 @@ use tracing::{instrument, trace};
 ///
 /// [scaling a bitmap image]: ../struct.Scale.html#pixels-and-display-points
 /// [SVG files]: https://en.wikipedia.org/wiki/Scalable_Vector_Graphics
-/// [`ImageBuf`]: ../druid_shell/struct.ImageBuf.html
+/// [`ImageBuf`]: ../piet/struct.ImageBuf.html
 /// [`FillStrat`]: ../widget/enum.FillStrat.html
 /// [`InterpolationMode`]: ../piet/enum.InterpolationMode.html
```
jneem, line 74:
```suggestion
/// [`ImageBuf`]: piet::ImageBuf
/// [`FillStrat`]: crate::widget::FillStrat
/// [`InterpolationMode`]: piet::InterpolationMode
```

jneem (commented):
Good catch! I think we can do even better, though, now that intra-doc links are stable. This way, the compiler will catch it next time it breaks.

druid/src/widget/image.rs
cmyr: we can do even better: if these types are in scope, we don't need to include explicit links to them at all, rustdoc will automatically resolve links with the format ```[`Type`]``` if `Type` is in scope.

RichardPoole42: I've updated with those three lines completely removed. All three links do work in the built documentation.

CHANGELOG.md
```diff
@@ -69,6 +69,7 @@ You can find its changes [documented below](#070---2021-01-01).
 - Fixed docs describing `ViewSwitcher` widget functionality ([#1693] by [@arthmis])
 - Added missing documentation on derived lens items ([#1696] by [@lidin])
 - Fix example code in `Get started with Druid` chapter of book ([#1698] by [@ccqpein])
+- Fix link in documentation of widget::Image ([#1730] by [@RichardPoole42])
```
SecondFlight, line 72: [#1730] needs a matching definition below as well

SecondFlight (changes requested):

CHANGELOG.md
RichardPoole42: Done.

cmyr (approved):
Thanks!

PR #1698, opened 2021-04-05
Change simple code in book

ccqpein:
In `Get started with Druid` of book, the simple code:

```rust
fn main() -> Result<(), PlatformError> {
    AppLauncher::with_window(WindowDesc::new(build_ui)).launch(())?;
    Ok(())
}
```

doesn't pass the compiler.  Change it to:

```rust
fn main() -> Result<(), PlatformError> {
    AppLauncher::with_window(WindowDesc::new(build_ui())).launch(())?;
    Ok(())
}
```

will work

PR #1732, opened 2021-04-22
Book: Fix the intro tutorial

SecondFlight:
The examples are written against `Druid@latest`, and this patch clarifies that and fixes the incorrect import code.

See also #1698.


cmyr (approved):
Thanks!

PR #1733, opened 2021-04-24
Test out Swatinem's cache.

jneem:
Based on the discussion [here](https://www.reddit.com/r/rust/comments/muw61h/github_actions_best_practices_for_rust_projects/)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,6 +48,9 @@ jobs:
           profile: minimal
           override: true
 
+      - name: restore cache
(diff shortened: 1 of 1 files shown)


cmyr (approved):
This looks like a clear win to me, do you have any concerns with merging?

jneem:
No, I don't have any concerns. I was planning to check the performance when there are non-trivial changes to dependencies, but given how fast the cache save/restore seems to be, I guess there isn't much risk in having a slowdown there.

PR #1737, opened 2021-04-25
Implement tooltip level on Windows platform

djeedai:
Implement `WindowLevel::Tooltip` on Windows platform.

**Before change:**
- Tooltip window appear in taskbar
- Window has strange min-size (probably due to OS min constraints)
- Window steals focus from main app window (see grayed out title bar of main app)

image

**After change:**
- Tooltip window doesn't appear in taskbar
- Window fits any size (tested with nursery's `tooltip.rs`)
- Window doesn't steal focus from main app window

image

**Details**

The change adds `SWP_NOACTIVATE` in many calls to `ShowWindow()` to prevent unwanted Win32 activation stealing the focus. `SWP_NOOWNERZORDER` is also added, inspired by Qt5 (see [here](https://github.com/qt/qtbase/blob/2ce94889a192ea67ba343bf0f6b16ede2648f374/src/plugins/platforms/windows/qwindowswindow.cpp#L870)).

Some `is_focusable` boolean is added to the state to discriminate between a regular "restore" and a non-activating show when `ShowWindow()` is called, because the former implicitly activates the window, stealing focus from the main app window and giving it to the tooltip.

Tentative `WS_`  and `WS_EX_` flags are set for `WindowLevel::Modal` and `WindowLevel::DropDown`. They have not been tested, but feel like a good starting point for whoever's coming next to implement those. In particular note that `DropDown` almost surely wants to be `WS_CHILD` to allow keyboard focus while not graying out the main app window title bar. However, because they have not been tested, `WindowBuilder::set_level()` continues to emit a warning for those levels, and ignore them for now for consistency with the current behavior.

### DIFF
--- a/druid-shell/src/platform/windows/window.rs
+++ b/druid-shell/src/platform/windows/window.rs
@@ -588,18 +588,20 @@ impl MyWndProc {
                     self.with_window_state(|s| s.is_resizable.set(resizable));
                     set_style(hwnd, resizable, self.has_titlebar());
                 }
-                DeferredOp::SetWindowState(val) => unsafe {
-                    if self.handle.borrow().is_focusable() {
-                        let s = match val {
+                DeferredOp::SetWindowState(val) => {
+                    let show = if self.handle.borrow().is_focusable() {
+                        match val {
                             window::WindowState::Maximized => SW_MAXIMIZE,
                             window::WindowState::Minimized => SW_MINIMIZE,
                             window::WindowState::Restored => SW_RESTORE,
-                        };
-                        ShowWindow(hwnd, s);
+                        }
                     } else {
-                        ShowWindow(hwnd, SW_SHOWNOACTIVATE);
+                        SW_SHOWNOACTIVATE
+                    };
+                    unsafe {
+                        ShowWindow(hwnd, show);
                     }
-                },
+                }
                 DeferredOp::SaveAs(options, token) => {
                     let info = unsafe {
                         get_file_dialog_path(hwnd, FileDialogType::Save, options)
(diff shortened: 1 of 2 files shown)


arthmis:
Ok where did you find how to make the tooltip not steal focus? I was searching on how to do that but I came up empty. Where did you find about the calls to `ShowWindow()` stealing focus? And thank you for doing this!

druid-shell/src/platform/windows/window.rs
```diff
@@ -1682,12 +1730,16 @@ impl WindowHandle {
...
+                if w.is_focusable {
+                    let show = match self.get_window_state() {
+                        window::WindowState::Maximized => SW_MAXIMIZE,
+                        window::WindowState::Minimized => SW_MINIMIZE,
+                        _ => SW_SHOWNORMAL,
+                    };
+                    ShowWindow(hwnd, show);
+                } else {
+                    ShowWindow(hwnd, SW_SHOWNOACTIVATE);
+                }
```
jneem, line 1742: If you say
```rust
let show = if w.is_focusable { ...
} else {
  SW_SHOWNOACTIVATE
};
```
Then you can have the `unsafe` just around the `ShowWindow`...

jneem (approved):
I'd be happier if someone who knows windows took a look, but I think it's been sitting long enough and it looks pretty much fine to me.

druid-shell/src/platform/windows/window.rs
djeedai: Fixed as suggested, and other occurrence on L592-L600.

Also added entry in CHANGELOG and fixed 2-3 lint errors in there, I hope that's OK @jneem?

jneem: Sounds good to me!

djeedai:
Thanks! 👍

PR #1744, opened 2021-04-28
Provide access to lens in LensWrap

SecondFlight:
I'm writing a widget for the nursery where I needed to modify a field in a lens owned by a `LensWrap`, and the easiest way was to make this patch.


cmyr (approved):
I think this makes sense, thanks!

PR #1746, opened 2021-04-28
druid-shell: Added Delete key handling

bjorn:
As I was trying the examples on Linux I noticed that whereas Backspace was handled, the Delete key was not. I hope I found the right way to add this functionality (it works at least).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ You can find its changes [documented below](#070---2021-01-01).
 - `DisabledIf` widget wrapper to disable based on the state of Data and Env ([#1702] by [@xarvic])
 - GTK: added support for `content_insets` ([#1722] by [@jneem])
 - `chrono` feature with `Data` support for [chrono](https://docs.rs/chrono/) types ([#1743] by [@r-ml])
+- Text input handles Delete key ([#1746] by [@bjorn])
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -684,6 +687,7 @@ Last release without a changelog :(
 [#1713]: https://github.com/linebender/druid/pull/1713
 [#1715]: https://github.com/linebender/druid/pull/1715
 [#1724]: https://github.com/linebender/druid/pull/1724
+[#1726]: https://github.com/linebender/druid/pull/1726
```
jneem, line 690: 1746, no?

jneem (changes requested):
Thanks, this makes sense to me, at least. Just one typo in the changelog...

jneem (approved):

cmyr:
Yep, this is definitely right; there's a bunch of stuff missing from this "simulate_input" method, and fixes are welcome!

bjorn:
@cmyr I had just run into the missing Delete handling myself. What other stuff is known to be missing?

cmyr:
I suspect most modifier combos like ctrl+arrow are missing. Also it might be nice, on linux, to support all the basic readline combos, like ctrl+e, ctrl+a, etcetera?

PR #1747, opened 2021-04-28
Perform widget layout directly after events

ForLoveOfCats:
In the case of a newly added child it may receive events between getting WidgetAdded and having a layout which is warned against. I'm not entirely confident in these changes. ~~I figured it would be safer (more obvious if breakage occurs) to always do the layout if required, instead of only when children had changed. Additionally with that choice the layout call on line 432 at the beginning of `do_paint` may be dead code now. (I would have these both be inline comments but GH does not allow commenting on un-modified lines)~~

### DIFF
--- a/druid/src/window.rs
+++ b/druid/src/window.rs
@@ -222,6 +222,32 @@ impl<T: Data> Window<T> {
         }
     }
 
+    fn layout_if_needed(
+        &mut self,
+        widget_state: &mut WidgetState,
+        queue: &mut CommandQueue,
+        data: &T,
+        env: &Env,
+    ) {
+        /*
+         * If a Widget is newly created it would not have an initial layout until
+         * paint time whichs bad because we might send more events down the tree
+         * before then. This is bad because a widget is not supposed to recieve a
+         * WidgetAdded event and then have layout before recieving any other events.
+         * Avoid this by calling this function after `post_event_processing` has
+         * sent the WidgetAdded event but before any other events could be sent.
+         *
+         * NOTE: Not called by lifecycle because lifecycle and layout can end up
+         * calling each other recursively. This is fine because Druid makes no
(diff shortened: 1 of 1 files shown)


druid/src/window.rs
```diff
@@ -180,6 +180,19 @@ impl<T: Data> Window<T> {
...
+         * before then. This is bad because a widget is not supposed to recieve a
+         * WidgetAdded event and then have layout before recieving any other events.
+         * Avoid this by performing a layout now that it has recieved WidgetAdded
+         * but before any other events could be sent.
+         */
+        if widget_state.needs_layout {
+            self.invalidate_and_finalize();
+            self.layout(queue, data, env);
+        }
+
```
ForLoveOfCats, line 195: I figured it would be simpler to always perform this layout as opposed to placing it in the `if widget_state.children_changed` block directly above. I am not confident about this

cmyr (commented):
The problem we have is that layout is expensive, and we want to let layout requests stack up, only performing *actual* layout once per frame.

I'd be curious to know more about the problem which motivated this patch? Maybe we could attempt a more narrowly targeted solution.

SecondFlight:
I ran into this while working on [this nursery widget](https://github.com/linebender/druid-widget-nursery/pull/49). The example has widgets created on `MouseDown`, and I sometimes get warnings (I assume on `MouseUp`?). I've ignored the warnings so far as they don't persist and I don't think they indicate an issue.

cmyr:
yes, I think you're right, and your example is a good one. I can see how this becomes a matter of correctness.

I think the best course of action right now will be to create an issue that describes this problem, and we will address it as part of the next architecture pass.

ForLoveOfCats:
That sounds like a reasonable path forward :+1:

Issue #1748, opened 2021-04-29
the readme.md example is not building


hammsvietro:
after copying the README.md example in a new cargo project, using druid 0.7.0, I get  the following error:

```
error[E0277]: expected a `FnOnce<()>` closure, found `impl druid::Widget<u32>`
   --> src/main.rs:5:39
    |
5   |     let main_window = WindowDesc::new(ui_builder());
    |                                       ^^^^^^^^^^^^ expected an `FnOnce<()>` closure, found `impl druid::Widget<u32>`
    | 
   ::: /home/hvm/.cargo/registry/src/github.com-1ecc6299db9ec823/druid-0.7.0/src/app.rs:347:28
    |
347 |         F: FnOnce() -> W + 'static,
    |                            ------- required by this bound in `WindowDesc::<T>::new`
    |
    = help: the trait `FnOnce<()>` is not implemented for `impl druid::Widget<u32>`
    = note: wrap the `impl druid::Widget<u32>` in a closure with no arguments: `|| { /* code */ }`
```

chanble:
you should run this example  [https://github.com/linebender/druid/tree/v0.7.0](url)

xarvic:
druid recently changed the signature of `WindowDesc::new()`. In druid 0.7 it needs a function to build the widget later. In the latest commit it just needs the widget. Since you are using druid 0.7 you need to remove the () brackets after `ui_builder`.

hammsvietro:
thanks! but shouldn't the readme example be in sync with the version it's refering to?

SecondFlight:
I may read it differently from you, but I don't believe the readme states or implies that the example uses `0.7.0`. The closest I see is this:

    ...Druid is available on [crates.io] and should work as
    a lone dependency (it re-exports all the parts of `druid-shell`, `piet`, and `kurbo`
    that you'll need):

    ```toml
    druid = "0.7.0"
    ```

    Since Druid is currently in fast-evolving state, you might prefer to drink from
    the firehose:

    ```toml
    druid = { git = "https://github.com/linebender/druid.git" }
    ```

This doesn't read to me as preferring one or the other, and it occurs below the example. It is the only mention of 0.7.0 I see.

Still, if you think this could be clearer, feel free to PR!

PR #1750, opened 2021-04-30
Update README.md example to match version 0.7.0

hammsvietro:
As discussed in #1748, although not implied, the README.md example isn't building if using the version mentioned below, 0.7.0. maybe this could avoid troubleshooting for a newcomer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,10 +39,10 @@ use druid::widget::{Button, Flex, Label};
 use druid::{AppLauncher, LocalizedString, PlatformError, Widget, WidgetExt, WindowDesc};
 
 fn main() -> Result<(), PlatformError> {
-    let main_window = WindowDesc::new(ui_builder());
+    let main_window = WindowDesc::new(ui_builder);
     let data = 0_u32;
     AppLauncher::with_window(main_window)
-        .log_to_console()
+        .use_simple_logger()
         .launch(data)
 }
 
--- a/README.md
+++ b/README.md
@@ -32,17 +32,17 @@ We gladly accept contributions via GitHub pull requests. Please see
 
 ## Example
 
-Here's a simple counter example app.
+Here's a simple counter example app:
 
 ```rust
 use druid::widget::{Button, Flex, Label};
 use druid::{AppLauncher, LocalizedString, PlatformError, Widget, WidgetExt, WindowDesc};
 
 fn main() -> Result<(), PlatformError> {
-    let main_window = WindowDesc::new(ui_builder);
+    let main_window = WindowDesc::new(ui_builder());
     let data = 0_u32;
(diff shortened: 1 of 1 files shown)


SecondFlight:
The issue to me is not that the example implies 0.7.0; it's that the version in the example is ambiguous. Both `0.7.0` and `master` are mentioned in the readme, and it's not clear which one the example is targeting. My gut feeling is that, unless something is clarified, changing the example to target `0.7.0` will help some people but confuse others.

I think there is a case to be made for having the example on GitHub target the version on GitHub; the [docs](https://docs.rs/druid/0.7.0/druid/) provide an example targeting 0.7.0, so that's covered. The potential issue with the readme is that it's not clear that it targets master.

Any other thoughts? I'd like to avoid completely driving this, so I'll wait for someone else to chime in with thoughts/suggestions.

jneem:
I think it makes sense to have the readme follow master, with a link to the 0.7.0 docs for those who want a slower-moving target.

cmyr:
Agreed, it's going to be too much of a headache to not keep the readme in line with master.

SecondFlight:
@hammsvietro I don't think this PR will merge as-is, but I think a possible solution here would be to
1. Revert the example change
2. Make a note saying the example uses master
3. Link to the docs for a 0.7.0 example

I think this is a good change and I'll do it soon if there's no new activity here, but I want to give you the chance to if you'd like, or to suggest something else.

hammsvietro:
that makes sense, I did as you suggested

README.md
```diff
@@ -58,6 +58,9 @@ fn ui_builder() -> impl Widget<u32> {
     Flex::column().with_child(label).with_child(button)
 }
 ```
+##### This example is compatible with master, [here's an example using 0.7.0](https://crates.io/crates/druid/0.7.0#example)
+<br>
+
```
SecondFlight, line 63:
```suggestion
This example works with master. Here's an example for 0.7.0: https://docs.rs/druid/latest/druid/#examples.
```

SecondFlight (changes requested):
Sorry for the late review. I haven't had time to dedicate to this until today.

These are just some minor nitpicks. I've replaced the crates.io link with docs.rs - this is more of a preference thing. docs.rs has a lot more documentation than just examples, so I like the idea of pointing there.

I'll apply these changes and merge tomorrow.

SecondFlight (approved):

SecondFlight:
Thanks!

cmyr:
Thaanks @hammsvietro and @SecondFlight, I think this is a good compromise!

hammsvietro:
really like the project, glad that I could help in some way 😃

PR #1751, opened 2021-05-01
X11 Scaling

### LINKED ISSUE #939: X11 shell does not handle DPI scaling.

Maan2003:
~~I am using DRUID_X11_SCALE env variable to detect scale, because I dont know to properly detect scaling.~~
Now using `Xft.dpi`
Fixes #939

### DIFF
--- a/druid-shell/src/platform/x11/window.rs
+++ b/druid-shell/src/platform/x11/window.rs
@@ -800,7 +800,12 @@ impl Window {
     }
 
     fn add_invalid_rect(&self, rect: Rect) -> Result<(), Error> {
-        borrow_mut!(self.state)?.invalid.add_rect(rect.expand());
+        let mut state = borrow_mut!(self.state)?;
+        let scale = state.scale;
+        // We prefer to invalidate an integer number of pixels.
+        state
+            .invalid
+            .add_rect(rect.to_px(scale).expand().to_dp(scale));
         Ok(())
     }
 
--- a/druid-shell/src/platform/x11/window.rs
+++ b/druid-shell/src/platform/x11/window.rs
@@ -24,6 +24,7 @@ use std::rc::{Rc, Weak};
 use std::sync::{Arc, Mutex};
 use std::time::Instant;
 
+use crate::scale::Scalable;
 use anyhow::{anyhow, Context, Error};
 use cairo::{XCBConnection as CairoXCBConnection, XCBDrawable, XCBSurface, XCBVisualType};
 use tracing::{error, info, warn};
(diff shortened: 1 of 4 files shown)


druid-shell/src/platform/x11/window.rs
```diff
@@ -1212,15 +1260,19 @@ impl PresentData {
...
-            .map(|r| Rectangle {
-                x: r.x0 as i16,
-                y: r.y0 as i16,
-                width: r.width() as u16,
-                height: r.height() as u16,
+            .map(|r| {
+                let r = r.to_px(scale);
+                Rectangle {
+                    x: r.x0 as i16,
+                    y: r.y0 as i16,
```
jneem, line 1272: I think you'll want to round these. `r.px(scale)` should essentially have integer coordinates, but I think it's possible for the coordinates to be just slightly smaller than an integer. In that case, we should round up, but this will round down.

druid-shell/src/platform/x11/window.rs
```diff
@@ -720,13 +743,15 @@ impl Window {
...
         if let Some(present) = borrow_mut!(self.present_data)?.as_mut() {
-            present.present(self.app.connection(), pixmap, self.id, &invalid)?;
+            present.present(self.app.connection(), pixmap, self.id, &invalid, scale)?;
             buffers.idle_pixmaps.pop();
         } else {
-            for r in invalid.rects() {
-                let (x, y) = (r.x0 as i16, r.y0 as i16);
-                let (w, h) = (r.width() as u16, r.height() as u16);
+            for rect in invalid.rects() {
+                let rect = rect.to_px(scale);
```
jneem, line 752: I think you need to round this rectangle (for the same reason as explained below)

druid-shell/src/platform/x11/window.rs
```diff
@@ -220,6 +221,24 @@ impl WindowBuilder {
         let screen_num = self.app.screen_num();
         let id = conn.generate_id()?;
         let setup = conn.setup();
+
+        let scale = match self.app.rdb.get_value::<f64>("Xft.dpi", "") {
```
jneem, line 225: I agree that this is the right way to do it, but could we have an environment variable override also? They're handy for testing...

druid-shell/src/platform/x11/window.rs
```diff
@@ -856,17 +888,15 @@ impl Window {
 
     #[allow(clippy::unnecessary_wraps)]
```
jneem, line 889: I think this is unnecessary now.

jneem (changes requested):
It looks good, thanks! Just a few minor things

druid-shell/src/platform/x11/window.rs
```diff
@@ -813,9 +843,11 @@ impl Window {
 
     fn invalidate(&self) {
         match self.size() {
-            Ok(size) => self.invalidate_rect(size.to_rect()),
+            Ok(size) => self.add_invalid_rect(size.size_dp().to_rect()).unwrap(),
```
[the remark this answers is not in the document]
Maan2003: Sorry, completely missed this one

jneem (approved):
Thanks!

Issue #1723, opened 2021-04-17
Book doesn't explain pattern to register env key from library


djeedai:
The force in my opinion of Druid is the simplicity of getting started to build new customized widgets. This is also (for now) a weakness as the list of available built-in widgets is thin.

Assuming a scenario where a user builds a library of widgets for use in third-party apps. [The Env section of the Druid book](https://linebender.org/druid/env.html) explains about theming and how to declare keys. It then proceed to explain keys must be registered by the app. Unfortunately in this scenario the library author is not the one who launches the app, so has no access to registration. There doesn't seem to be any alternative to register the keys automatically.

The pattern I see to work around this is, instead of assuming keys are set like built-in Druid widget do (calling `Env::get()`), the library author should probably use `Env::try_get()` and provide the same default as the key if it is not set. I think this workaround should at least be documented, or better a solution provided for this use case, possibly by adding a default value to the key itself.

PR #1752, opened 2021-05-02
Add default values to Env

xarvic:
This is an possible fix for #1723, but its not ready yet: 
- The crate used to make this possible [`doesn't work in debug mode`](https://github.com/dtolnay/inventory/issues/24).
- Inventory doesn't work with webassembly
- This does`nt work work with shared libraries, but this is probably not that important.


cmyr (commented):
I think this is interesting as an experiment, but overall I'm hesitant. I would prefer some mechanism that is more explicit; for instance adopting an idiom where custom variables are explicitly added via `configure_env`.

From a higher-level, I think the main problem

xarvic:
Ok, then i will just keep it as a draft.

PR #1755, opened 2021-05-03
X11 cursors

Maan2003:
Custom cursors are still TODO.

### DIFF
--- a/druid-shell/src/platform/x11/application.rs
+++ b/druid-shell/src/platform/x11/application.rs
@@ -25,7 +25,7 @@ use anyhow::{anyhow, Context, Error};
 use x11rb::connection::Connection;
 use x11rb::protocol::present::ConnectionExt as _;
 use x11rb::protocol::xfixes::ConnectionExt as _;
-use x11rb::protocol::xproto::{ConnectionExt, CreateWindowAux, EventMask, WindowClass};
+use x11rb::protocol::xproto::{self, ConnectionExt, CreateWindowAux, EventMask, WindowClass};
 use x11rb::protocol::Event;
 use x11rb::resource_manager::Database as ResourceDb;
 use x11rb::xcb_ffi::XCBConnection;
@@ -53,6 +53,7 @@ pub(crate) struct Application {
 
     /// The X11 resource database used to query dpi.
     pub(crate) rdb: ResourceDb,
+    pub(crate) cursors: Cursors,
     /// The default screen of the connected display.
     ///
     /// The connected display may also have additional screens.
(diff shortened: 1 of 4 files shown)


druid-shell/src/platform/x11/application.rs
```diff
@@ -126,13 +138,27 @@ impl Application {
...
+        let load_cursor = |name| handle.load_cursor(connection.as_ref(), name);
+
+        let cursors = Cursors {
+            default: load_cursor("default")?,
+            text: load_cursor("text")?,
+            pointer: load_cursor("pointer")?,
+            crosshair: load_cursor("crosshair")?,
+            not_allowed: load_cursor("not-allowed")?,
+            row_resize: load_cursor("row-resize")?,
+            col_resize: load_cursor("col-resize")?,
```
jneem, line 151: Is there a chance that these loads could fail? My understanding is that because these belong to the default cursor font, the only way they could fail is through some catastrophic failure (e.g. losing the connection). If this is the case, it's probably worth noting it in a comment. If I'm mistaken and there are other failure modes, I think we just want to fail to load that single cursor instead of aborting the whole function.

druid-shell/src/platform/x11/window.rs
```diff
@@ -884,6 +884,31 @@ impl Window {
...
+            Cursor::NotAllowed => cursors.not_allowed,
+            Cursor::ResizeLeftRight => cursors.col_resize,
+            Cursor::ResizeUpDown => cursors.row_resize,
+            // TODO: (x11/custom cursor)
+            Cursor::Custom(_) => cursors.default,
+        };
+        let conn = self.app.connection();
+        let changes = ChangeWindowAttributesAux::new().cursor(cursor);
+        if let Err(e) = conn.change_window_attributes(self.id, &changes) {
+            error!("Changing window attributes failed {}", e);
```
jneem, line 908: Maybe a cursor-specific error message would help in tracking things down? Especially if we use `change_window_attributes` for other attributes in the future...

jneem (approved):
Thanks for taking this on! I made two minor nitpicks; you can decide for yourself what to do about them.

jneem:
btw, have you seen the [list of default cursors](https://tronche.com/gui/x/xlib/appendix/b/)? It's pretty hilarious...

Maan2003:
> btw, have you seen the [list of default cursors](https://tronche.com/gui/x/xlib/appendix/b/)? It's pretty hilarious...

Oh, I missed this :rofl: . I just followed the [freedesktop cursor spec](https://www.freedesktop.org/wiki/Specifications/cursor-spec/)

jneem:
I can't imagine why freedesktop decided to exclude XC_gumby...

druid-shell/src/platform/x11/application.rs
Maan2003: after taking a look at x11rb cursor loading code, even though these are in default fonts, but are named differently in the default font.

It makes sense to make all these `Option<xproto::Cursor>` and then not setting the cursor if it is not loaded and `warn`ing.

PR #1756, opened 2021-05-03
X11 shell locale detection

Maan2003:
Based on [gettext manual](https://www.gnu.org/software/gettext/manual/html_node/Locale-Environment-Variables.html#Locale-Environment-Variables).

Also on Linux, there is a priority list of locales rather than a single locale. Not sure if `Application::get_locale` should return a priority list.

Fixes: #938

### DIFF
--- a/druid-shell/src/platform/x11/application.rs
+++ b/druid-shell/src/platform/x11/application.rs
@@ -504,12 +504,12 @@ impl Application {
         // from gettext manual
         // https://www.gnu.org/software/gettext/manual/html_node/Locale-Environment-Variables.html#Locale-Environment-Variables
         var_non_empty("LANGUAGE")
+            // the LANGUAGE value is priority list seperated by :
+            // See: https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html#The-LANGUAGE-variable
+            .and_then(|locale| locale.split(':').next().map(String::from))
             .or_else(|| var_non_empty("LC_ALL"))
             .or_else(|| var_non_empty("LC_MESSAGES"))
             .or_else(|| var_non_empty("LANG"))
-            // the value is priority list seperated by :
-            // See: https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html#The-LANGUAGE-variable
-            .and_then(|locale| locale.split(':').next().map(String::from))
             .unwrap_or_else(|| "en-US".to_string())
     }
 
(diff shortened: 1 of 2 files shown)


druid-shell/src/platform/x11/application.rs
```diff
@@ -495,9 +495,22 @@ impl Application {
...
+        // from gettext manual
+        // https://www.gnu.org/software/gettext/manual/html_node/Locale-Environment-Variables.html#Locale-Environment-Variables
+        var_non_empty("LANGUAGE")
+            .or_else(|| var_non_empty("LC_ALL"))
+            .or_else(|| var_non_empty("LC_MESSAGES"))
+            .or_else(|| var_non_empty("LANG"))
+            // the value is priority list seperated by :
+            // See: https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html#The-LANGUAGE-variable
+            .and_then(|locale| locale.split(':').next().map(String::from))
+            .unwrap_or_else(|| "en-US".to_string())
```
jneem, line 513: Is there a difference between "en-US" and "en_US"? My system uses the latter...

druid-shell/src/platform/x11/application.rs
```diff
@@ -495,9 +495,22 @@ impl Application {
...
+        };
+
+        // from gettext manual
+        // https://www.gnu.org/software/gettext/manual/html_node/Locale-Environment-Variables.html#Locale-Environment-Variables
+        var_non_empty("LANGUAGE")
+            .or_else(|| var_non_empty("LC_ALL"))
+            .or_else(|| var_non_empty("LC_MESSAGES"))
+            .or_else(|| var_non_empty("LANG"))
+            // the value is priority list seperated by :
+            // See: https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html#The-LANGUAGE-variable
```
jneem, line 511: My understanding of the pages you linked is that only `LANGUAGE` is a priority list; the other variables just contain a single value.

Maan2003: Oh just noticed that. But I don't think this should make a difference. The locale should not have `:`. I have also checked in my `/etc/locale.gen`. But the comment needs to be updated.

druid-shell/src/platform/x11/application.rs
Maan2003: I think it is better idea to `en_US`. I used `en-US` because it was used in the stub earlier.

cmyr: > The identifiers can vary in case and in the separator characters. The "-" and "_" separators are treated as equivalent, although "-" is preferred.



(from [Unicode TR #35](https://unicode.org/reports/tr35/#Identifiers))

SecondFlight (approved):

PR #1761, opened 2021-05-05
updating docs to use Selector.with instead of Command::new

### LINKED ISSUE #1731: Explanation of Target::Auto
Is there meant to be an explanation of how the target is determined when using `Target::Auto` for command submissions? It seems it sends to the host window of the widget?

arthmis:
- adding some more explanation for Target::Auto
- updating doc links to use new doc links

closes #1731

### DIFF
--- a/druid/src/command.rs
+++ b/druid/src/command.rs
@@ -77,7 +77,7 @@ pub struct Selector<T = ()>(SelectorSymbol, PhantomData<T>);
 ///
 /// let selector = Selector::new("process_rows");
 /// let rows = vec![1, 3, 10, 12];
-/// let command = Command::new(selector, rows, Target::Auto);
+/// let command = selector.with(rows);
 ///
 /// assert_eq!(command.get(selector), Some(&vec![1, 3, 10, 12]));
 /// ```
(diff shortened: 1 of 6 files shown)


druid/src/command.rs
```diff
@@ -382,11 +387,11 @@ impl<T: Any> Selector<T> {
...
     /// By default, the command will have [`Target::Auto`].
-    /// The [`Command::to`] method can be used to override this.
+    /// The [`Selector::to`] method can be used to override this.
     ///
-    /// [`Command::new`]: struct.Command.html#method.new
-    /// [`Command::to`]: struct.Command.html#method.to
-    /// [`Target::Auto`]: enum.Target.html#variant.Auto
+    /// [`Command::new`]: Command::new
+    /// [`Selector::to`]: Selector::to
+    /// [`Target::Auto`]: Target::Auto
```
jneem, line 394: I haven't quite got the hang on intra-doc links, but I think these are redundant?

jneem (approved):
Looks good!

druid/src/command.rs
arthmis: oh yeah I think that's the case.

PR #1764, opened 2021-05-07
Supports accessing nested fields with lens macro

Maan2003:
Now this is valid

``` rust
struct Foo { x: Bar }
struct Bar { y: [i32; 10] }
lens!(Foo, x.y[5]);
```


jneem (approved):
Cool!

PR #1769, opened 2021-05-09
Fix alignment offset for selection in TextBox

### LINKED ISSUE #1767: Selection of centered text is wrongly positioned
(Also it seems like text that is too long is wrongly clipped (by like 1 pixel))

This is on 8e83610d19f1ccdbd3ebaacde98c189f2bfeffca (recent master)

CryZe:
The alignment offset was accidentally added instead of subtracted from the cursor position.

Fixes #1767

### DIFF
--- a/druid/src/text/input_component.rs
+++ b/druid/src/text/input_component.rs
@@ -719,7 +719,7 @@ impl<T: TextStorage + EditableText> EditSession<T> {
     }
 
     fn do_mouse_down(&mut self, point: Point, mods: Modifiers, count: u8) {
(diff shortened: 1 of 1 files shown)


CryZe:
It might be wrong in more places such as `cursor_line_for_text_position` as well.

cmyr (approved):
Cool, I was quite confused by this but it looks like I broke it somewhere during the IME work.

`cursor_line_for_text_position` is fine; it's supposed to be the inverse operation (figuring out where to visually draw the cursor, for a given text position, as opposed to figuring out the text position for a cursor position).

Thanks!

Issue #936, opened 2020-05-15
X11 shell does not support file dialogs.


luleyleo:
See the [Zulip conversation](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/FileDialogOptions.20no.20effect.20on.20linux/near/196377266).

Finnerale:
There is a new Rust library which implements file dialogs using `zenity` and `kdialog`:
https://github.com/balthild/native-dialog-rs

xStrom:
I noticed that library too and checked its alert code, as I've been working in that area recently. The implementation is super naive and simple. More power to them and great for little toy projects, but I don't think there's anything there for us.

raphlinus:
This is of course a known limitation of using a low level library. My personal feeling is that the best way going forward is to write a file dialog in druid, but that's a nontrivial undertaking of course.

xStrom:
The `zenity`/`kdialog` vs custom druid implmentation question comes up with X11 alerts as well. I've been going back and forth in my mind, but lately leaning towards custom alerts. Alerts are easier to implement than file dialogs, so this could serve as a test case of sorts to see how the work/result compares to just using the binaries.

In the alert case the custom alerts can be useful on Windows/macOS too for more complex cases, because it would be easy to embed custom widgets into the alert dialog for some advanced functionality.

saucesaft:
https://github.com/x42/sofd bindings of this c library could be useful, so the wheel is not rewritten.

jplatte:
On Linux, it would be useful to use the desktop portal DBus protocol for this. See https://belmoussaoui.com/article/13-oxidizing-portals :)

In addition to not requiring a druid file dialog implementation (though I would love to eventually see one), this would allow flatpak-sandboxed applications to open files that aren't otherwise accessible to them AFAIK.

Finnerale:
That post is a nice overview, not sure how I missed it…

I haven't brought up portals so far because they used to have no support for opening directories for whatever reason, but looks like they finally fixed that with the last edition. Definitely worth a try.

jneem:
Most likely you "missed" it because it was only just written :)

Maan2003:
Fixed by #2153

PR #1774, opened 2021-05-11
Support for open/save dialogs on x11.

jneem:
This is a first stab at doing open/save in the x11 backend (addressing #936). Very much in a WIP state, but I'm opening this because I have a few unresolved questions.

This is using the freedesktop.org DBus specification to launch a native file dialog, as suggested in #936. Of course it would be nice to have a druid-native dialog also, but this has the advantages of (a) being less work and (b) having support in the flatpak sandbox.

## Dependencies

This introduces a number of new dependencies:
- `zbus`/`zvariant`/`serde`: these are pretty much required for using DBus, I think.
- `zvariant_derive`: could be avoided; it's currently saving about 50 lines of serde boilerplate.

I'm not currently depending on the [`ashpd`](https://github.com/bilelmoussaoui/ashpd), although I'd kind of like to. It would save probably around 100 lines right now, but has some advantages and disadvantages:
- it's new, and doesn't seem to be used much yet
- on the other hand, the maintainer is super responsive (and has been to me in the past)
- it has a bunch of stuff we don't need
- but we might want some of them in the future (e.g. notifications, trash, printing)

### DIFF
--- a/druid-shell/src/platform/x11/dbus.rs
+++ b/druid-shell/src/platform/x11/dbus.rs
@@ -34,17 +34,12 @@ fn response_thread(
                     let response = msg.body::<ResponseResult<FileDialogResponse>>()?;
                     let file_info = if let ResponseType::Success = response.0 {
                         // FIXME: what are we supposed to do if there are multiple files?
-                        response.1.uris.first().and_then(|s| {
-                            // FIXME: we get a URI, but we want a path. There must be a less hacky
-                            // way to do this...
-                            if s.starts_with("file://") {
-                                Some(FileInfo {
-                                    path: s[7..].into(),
-                                })
-                            } else {
-                                None
-                            }
-                        })
+                        response
+                            .1
+                            .uris
+                            .first()
+                            .and_then(|s| s.strip_prefix("file://"))
+                            .map(|s| FileInfo { path: s.into() })
                     } else {
                         None
                     };
@@ -79,7 +74,10 @@ impl DbusHandle {
         let connection = Connection::new_session()?;
         let proxy = FileChooserProxy::new(&connection)?;
 
-        let title = options.title.clone().unwrap_or("Open file".to_owned());
+        let title = options
+            .title
+            .clone()
+            .unwrap_or_else(|| "Open file".to_owned());
         let opts = OpenOptions::from(options);
         let reply = proxy.open_file(&format!("x11:{}", window), &title, opts)?;
         let tok = FileDialogToken::next();
@@ -98,7 +96,10 @@ impl DbusHandle {
         let connection = Connection::new_session()?;
         let proxy = FileChooserProxy::new(&connection)?;
 
-        let title = options.title.clone().unwrap_or("Save as".to_owned());
+        let title = options
+            .title
+            .clone()
+            .unwrap_or_else(|| "Save as".to_owned());
         let opts = SaveOptions::from(options);
         let reply = proxy.save_file(&format!("x11:{}", window), &title, opts)?;
         let tok = FileDialogToken::next();
@@ -166,7 +167,7 @@ impl From<crate::FileDialogOptions> for OpenOptions {
             directory: opts.select_directories,
             filters: opts
                 .allowed_types
-                .unwrap_or(Vec::new())
+                .unwrap_or_else(Vec::new)
                 .into_iter()
                 .map(Into::into)
                 .collect(),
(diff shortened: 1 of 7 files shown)


druid-shell/src/platform/x11/dbus.rs
```diff
@@ -0,0 +1,264 @@
...
+        {
+            if let Some(reply_path) = msg.header()?.path()? {
+                if &path == reply_path {
+                    let response = msg.body::<ResponseResult<FileDialogResponse>>()?;
+                    let file_info = if let ResponseType::Success = response.0 {
+                        // FIXME: what are we supposed to do if there are multiple files?
+                        response.1.uris.first().and_then(|s| {
+                            // FIXME: we get a URI, but we want a path. There must be a less hacky
+                            // way to do this...
+                            if s.starts_with("file://") {
```
ForLoveOfCats, line 40: [strip_prefix](https://doc.rust-lang.org/std/primitive.str.html#method.strip_prefix)?

That's what I use in my editor for handling URIs from LSP servers but my understanding is that there are other complications that can arise

jneem:
Yeah, unfortunately `zbus` isn't small (it adds 500k to a release binary), but I definitely don't think we should roll our own DBus implementation. The alternative seems to be [`dbus`](https://github.com/diwic/dbus-rs), which is a wrapper around the C library. I'm more inclined to go for the pure-rust version, though (plus, it's part of the official upstream DBus organization).

The other dependencies are less critical, but also much smaller :shrug:

runiq:
Nothing but love for this PR <3

psychon:
If you want, I can actually try this out, but just from looking at the code: What happens if `org.freedesktop.portal.FileChooser` is unowned? I'd expect the DBus daemon to reply with some kind of error, but the code seems to ignore that error...?

(I might also just be misunderstanding how DBus works and this might all be fine)

Maan2003:
works! 
But if I have 2 X servers running at tty1 and tty2. application is running on tty2. the file selector opens on tty1 :confused:

jneem:
>  I'd expect the DBus daemon to reply with some kind of error, but the code seems to ignore that error...?

Good point! I think long term we'll want to fall back to our own druid-implemented file dialog. But that's a lot of work, so for now I'll just log an error.

jneem:
> works!
> But if I have 2 X servers running at tty1 and tty2. application is running on tty2. the file selector opens on tty1 confused

hm, that's worrying. My understanding (from [here](https://www.freedesktop.org/wiki/Software/dbus/) was that dbus runs a daemon for each user login session, so I was expecting that to help somehow. I'll see if I can find some time to play around with this later in the week.

jneem:
So despite what the dbus page says, arch linux (and maybe other distros, I didn't check) has been doing per-user (instead of per-login-session) dbusses [since 2015](https://archlinux.org/news/d-bus-now-launches-user-buses/). I see some discussion [here](https://mail.gnome.org/archives/desktop-devel-list/2014-January/msg00079.html) about involving the `$DISPLAY` variable somehow, but I don't know about more recent status...

cmyr:
What's the status of this? I don't have a great set up for testing on linux right now, but I'd like to help get this merged if there's anything I can do?

jneem:
It's currently blocked because xdg-desktop-portal [doesn't support](https://github.com/flatpak/xdg-desktop-portal/issues/586) having more than one X session open, and I'm not sure whether they have any plans to...

Maan2003:
I am in favor of leaving multiple X servers as a todo and merging this because running multiple X servers isn't very common anyways :shrug:

jneem:
I agree that it isn't common, but the main thing that bothers me is that fixing this isn't actually something that we *can* do: we're using someone else's protocol and so we rely on them accepting a fix. If they aren't willing to do that, then this isn't a long-term solution.

JAicewizard:
Is there any reason we rely on flatpak?

jneem:
`xdg-desktop-portal` defines the dbus protocols that we use. It was originally created for flatpak because applications that use those protocols are easy to sandbox.

bilelmoussaoui:
note that the portals are also used by snap

jneem:
Superseded by #2153

goddaneel:
Same problem, at first I thought it was a steam issue. I just found this repository, forgive me if I may have filed an issue in the wrong place.

https://github.com/flatpak/flatpak/issues/6209

I'd actually be more interested in determining if this issue is limited to x11, if not this could be a big security issue, obviously the sandbox is accessing sockets it shouldn't be.

PR #1775, opened 2021-05-12
Correctly clip the TextBox

CryZe:
By adjusting only the x component of the size before turning it into a rectangle, the right side of the rectangle gets moved to the left, which allowed the text to peak out on the right side. Inset already seems to do what subtracting the border width there was possibly intended to do, so just using the unadjusted size and then later insetting, resolves the problem entirely.

Here for reference the bug that this resolves:


Maan2003 (approved):
Looks Good. Thanks!

cmyr:
cool, I think this was a relic of an earlier version, where the padding was calculated and applied manually.

PR #1776, opened 2021-05-12
Use point/size in ClipBox instead of rect.

jneem:
This fixes a layout bug I was having, but I think it's also an
indication of a bigger issue.

The issue is that in various parts of the layout code, we expand
dimensions to the nearest display point. Besides the fact that we should
be using pixels intead of display points (IMO), this is numerically
unstable: if because of some numerical stuff you have a dimension that's
a rounding error larger than an integer number of pixels, the layout
"jumps".

What does this have to do with ClipBox? Well, ClipBox was storing its
layout rect as a `Rect`, which is stored as the coordinates of opposite
corners. When scrolling the ClipBox, the size should stay fixed but the
origin should move. But moving the origin can *change the size* very
slightly because of numerical errors converting to/from the Rect
representation. These slight errors got magnified by the pixel expansion
in other parts of the layout code.

This PR fixes the numerical imprecision in the ClipBox code, but I
suspect that the real fix is to eliminate the numerical instability.

### DIFF
--- a/druid/src/scroll_component.rs
+++ b/druid/src/scroll_component.rs
@@ -218,9 +218,9 @@ impl ScrollComponent {
     }
 
     fn calc_bar_bounds(&self, axis: Axis, port: &Viewport, env: &Env) -> Option<Rect> {
-        let viewport_size = port.rect.size();
+        let viewport_size = port.view_size;
         let content_size = port.content_size;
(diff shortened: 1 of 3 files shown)


Maan2003 (approved):
Looks good! The `.expand()` is tricky to manage :frowning_face:

PR #1777, opened 2021-05-12
Don't redo the whole layout when a ClipBox scrolls.

### LINKED ISSUE #1762: Scrolling anything causes a full re-layout
If I have a `ClipBox` anywhere in my widget tree, scrolling it (e.g. using `pan_by`) triggers a full re-layout. This was introduced in de5f88ad9, and the comment in `WidgetPod::set_viewport_offset` says:
```
// We need the parent_window_origin recalculated.
// It should be possible to just trigger the InternalLifeCycle::ParentWindowOrigin here,
// instead of full layout. Would need more management in WidgetState.
```

jneem:
Fixes #1762

### DIFF
--- a/druid/src/core.rs
+++ b/druid/src/core.rs
@@ -127,6 +127,9 @@ pub(crate) struct WidgetState {
 
     pub(crate) needs_layout: bool,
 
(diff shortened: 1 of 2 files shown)


cmyr (approved):
I think this is sensible.

PR #1779, opened 2021-05-13
X11 shell keyboard mapping

### LINKED ISSUE #1633: Remapped keys in X11 have their original value in druid
I remap my Caps Lock key to Escape using the following file in `/etc/X11/xorg.conf.d/00-keyboard.conf`:
```
Section "InputClass"
        Identifier "system-keyboard"
        MatchIsKeyboard "on"
        Option "XkbLayout" "us"
        Option "XkbOptions" "caps:escape"
EndSection
```
When I press the escape key in my druid application, using the X11 backend, I receive a `KeyEvent` where both `key` and `code` are mapped to `CapsLock`.

The remapping works in every other application installed in my system, so this was pretty surprising behaviour.

Maan2003:
Resolves #1633

~~I am using a fork of the `xcbcommon-sys` crate because it doesn't build with x11 feature enabled, waiting for meh/rust-xkbcommon-sys#5 to merged.~~ Now just running bindgen

The xkb.rs is from #1498


jneem:
Nice to see so much progress on x11! My main concern at this point is the dependence on `xkbcommon-sys`, which seems "lightly" maintained and also has a strange license...

I wonder how much work it is just to run `bindgen` ourselves?

Maan2003:
I feel same. Maintaining a `xkbcommon-sys` crate ourselves feels the best solution to me.

Maan2003:
Ready for review :)

druid-shell/src/platform/x11/xkb.rs
```diff
@@ -0,0 +1,456 @@
...
+use std::{convert::TryFrom, ptr};
+use x11rb::xcb_ffi::XCBConnection;
+
+pub struct DeviceId(c_int);
+
+/// A global xkb context object.
+///
+/// Reference counted under the hood.
+// Assume this isn't threadsafe unless proved otherwise. (e.g. don't implement Send/Sync)
+// TODO do we need UnsafeCell?
```
jneem, line 35: What would we need to check to find out? Is it a question of whether the FFI calls modify anything through the pointer?

druid-shell/src/platform/x11/xkb.rs
```diff
@@ -0,0 +1,456 @@
...
+        };
+        if key_map.is_null() {
+            return None;
+        }
+        Some(Keymap(key_map))
+    }
+
+    /// Set the log level using `log` levels.
+    ///
+    /// Becuase `xkb` has a `critical` error, each rust error maps to 1 above (e.g. error ->
```
jneem, line 76: Because

druid-shell/src/platform/x11/xkb.rs
```diff
@@ -0,0 +1,456 @@
...
+                shift: mod_idx(XKB_MOD_NAME_SHIFT),
+                alt: mod_idx(XKB_MOD_NAME_ALT),
+                super_: mod_idx(XKB_MOD_NAME_LOGO),
+                caps_lock: mod_idx(XKB_MOD_NAME_CAPS),
+                num_lock: mod_idx(XKB_MOD_NAME_NUM),
+            },
+        }
+    }
+
+    pub fn key_event(&self, scancode: u32, state: KeyState) -> KeyEvent {
```
jneem, line 164: This feels like it should be `&mut self`

druid-shell/src/platform/x11/xkb.rs
```diff
@@ -0,0 +1,456 @@
...
+            let len = xkb_state_key_get_utf8(self.state, scancode, ptr::null_mut(), 0);
+            if len == 0 {
+                return None;
+            }
+            // add 1 because we will get a null-terminated string.
+            let len = usize::try_from(len).unwrap() + 1;
+            let mut buf: Vec<u8> = Vec::new();
+            buf.resize(len, 0);
+            xkb_state_key_get_utf8(self.state, scancode, buf.as_mut_ptr() as *mut c_char, len);
+            debug_assert!(buf[buf.len() - 1] == 0);
```
jneem, line 245: Maybe `assert!` instead of `debug_assert!` (because it's cheap and a possible source of memory errors)

jneem (approved):
Nice work! I just had a few minor questions/suggestions.

druid-shell/src/platform/gtk/keycodes.rs
```diff
@@ -46,51 +90,125 @@ pub fn raw_key_to_key(raw: RawKey) -> Option<Key> {
...
+        // F19 => Key::F19,
+        // F20 => Key::F20,
+        // F21 => Key::F21,
+        // F22 => Key::F22,
+        // F23 => Key::F23,
+        // F24 => Key::F24,
+        // Calculator => Key::LaunchCalculator,
+        // MyComputer | Explorer => Key::LaunchMyComputer,
+        // ISO_Level3_Latch => Key::AltGraphLatch,
+        // ISO_Level5_Shift => Key::ShiftLevel5,
```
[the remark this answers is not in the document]
Maan2003: because these are **not** available in keyboard-types. not sure, how I missed commenting the reason here(it is in the x11 backend one), will do

druid-shell/src/platform/x11/xkb.rs
Maan2003: uh took code from wayland backend and missed this part :grimacing: . I will look into it

Maan2003: I am not sure, why we need it; we only interact with `*mut xkb_context`

cc @derekdreery

druid-shell/src/platform/x11/xkb.rs
Maan2003: again a bit unsure, given we only work with `*mut`s and this is not `Sync/Send`. this can be thought of as interior mutability

BUT, the `&mut` doesn't hurt us in this case, lets do it anyways

druid-shell/src/platform/x11/xkb.rs
derekdreery: I've honestly forgot about how I did this with wayland, sorry!

Maan2003: I don't think we need UnsafeCell, UnsafeCell is used to get `*mut T` from `&T`(well `&UnsafeCell<T>`)  but already have `*mut T`

cmyr:
Anything I can do to push this forward?

Maan2003:
still no consensus on if `UnsafeCell` is needed. everything else is settled I think

druid-shell/src/platform/x11/xkb.rs
Maan2003: thought about this again, I have no doubts. 
I will merge in 2 days if no one objects.

PR #1780, opened 2021-05-14
Fix typos in Clipboard docs

Maan2003:


druid-shell/src/clipboard.rs
```diff
@@ -36,7 +36,7 @@ pub use crate::platform::clipboard as platform;
 /// data available as an SVG, for other editors, and a bitmap image for applications
 /// that can accept general image data.
 ///
-/// ## `FormatId`entifiers
+/// ## `FormatId` entifiers
```
jneem, line 39: Maybe this one was intentional? because, like, "FormatIdentifiers". It isn't great, though

jneem (approved):

druid-shell/src/clipboard.rs
Maan2003: oh :sweat_smile: missed it. 
Any suggestions? I would make it just `FormatId` :man_shrugging:

cmyr: This was intentional, but maybe not helpful. I'd either leave it as it was, or I would replace it with plain `FormatId`.

Maan2003: leaving it

PR #1781, opened 2021-05-15
Make cursor for EditableText generic

Juan65Lira:


cmyr (commented):
In the medium term I would like to get rid of the `EditableText` trait, but I haven't figured out exactly what it should be replaced with, yet, so I'm not opposed to some minor improvements for the time being.

That said, given that `EditableText` is already implemented for `Arc<String>`, I'm not sure if this PR is really making anything possible that wasn't possible already, unless I'm missing something? And I think it's important that the cursor not clone the underlying string.

Was there any background or motivation for this patch that I might have overlooked?

PoignardAzur:
@Juan65Lira Any update?

Going from Colin's response, I'm probably going to close this issue in a few days unless somebody objects.

cmyr:
Closing for now, @Juan65Lira feel free to reopen if you're still interested in discussing

PR #1784, opened 2021-05-16
Convert some state from RefCell to Cell in x11 backend 

Maan2003:
This makes some getting some state infallible.


jneem (approved):
Thanks for the clean-up!

PR #1785, opened 2021-05-18
Implement some window properties for x11 shell

Maan2003:
Implements:
* set_size
* get_size
* set_min_size
* resizable
* get_position
* set_position
* set_level
* set_window_state (only for WindowBuilder)

### DIFF
--- a/druid-shell/src/platform/x11/window.rs
+++ b/druid-shell/src/platform/x11/window.rs
@@ -366,6 +366,7 @@ impl WindowBuilder {
             .set_normal_hints(conn.as_ref(), id)
             .context("set wm normal hints"));
 
+        // TODO: set _NET_WM_STATE
         let mut hints = WmHints::new();
         if let Some(state) = self.state {
             hints.initial_state = Some(match state {
@@ -375,6 +376,34 @@ impl WindowBuilder {
             });
         }
         log_x11!(hints.set(conn.as_ref(), id).context("set wm hints"));
+
+        // set level
+        {
+            let window_type = match self.level {
+                WindowLevel::AppWindow => atoms._NET_WM_WINDOW_TYPE_NORMAL,
+                WindowLevel::Tooltip => atoms._NET_WM_WINDOW_TYPE_TOOLTIP,
+                WindowLevel::Modal => atoms._NET_WM_WINDOW_TYPE_DIALOG,
+                WindowLevel::DropDown => atoms._NET_WM_WINDOW_TYPE_DROPDOWN_MENU,
+            };
+
+            let conn = self.app.connection();
+            log_x11!(conn.change_property32(
+                xproto::PropMode::REPLACE,
+                id,
+                atoms._NET_WM_WINDOW_TYPE,
+                AtomEnum::ATOM,
+                &[window_type],
+            ));
+            if matches!(
+                self.level,
+                WindowLevel::DropDown | WindowLevel::Modal | WindowLevel::Tooltip
+            ) {
+                log_x11!(conn.change_window_attributes(
+                    id,
+                    &ChangeWindowAttributesAux::new().override_redirect(1),
+                ));
+            }
+        }
 
         let window = Rc::new(Window {
             id,
@@ -396,7 +425,6 @@ impl WindowBuilder {
             active_text_field: Cell::new(None),
         });
 
-        window.set_level(self.level);
         window.set_title(&self.title);
         if let Some(pos) = self.position {
             window.set_position(pos);
@@ -562,7 +590,7 @@ atom_manager! {
         _NET_WM_WINDOW_TYPE_NORMAL,
         _NET_WM_WINDOW_TYPE_DROPDOWN_MENU,
         _NET_WM_WINDOW_TYPE_TOOLTIP,
-        _NET_WM_WINDOW_TYPE_DAILOG,
+        _NET_WM_WINDOW_TYPE_DIALOG,
     }
 }
 
@@ -863,30 +891,6 @@ impl Window {
                 .width(size.width as u32)
                 .height(size.height as u32),
         ));
-    }
-
-    fn set_level(&self, level: WindowLevel) {
-        let window_type = match level {
-            WindowLevel::AppWindow => self.atoms._NET_WM_WINDOW_TYPE_NORMAL,
-            WindowLevel::Tooltip => self.atoms._NET_WM_WINDOW_TYPE_TOOLTIP,
-            WindowLevel::Modal => self.atoms._NET_WM_WINDOW_TYPE_DAILOG,
-            WindowLevel::DropDown => self.atoms._NET_WM_WINDOW_TYPE_DROPDOWN_MENU,
-        };
-
-        let conn = self.app.connection();
-        log_x11!(conn.change_property32(
-            xproto::PropMode::REPLACE,
-            self.id,
-            self.atoms._NET_WM_WINDOW_TYPE,
-            AtomEnum::ATOM,
-            &[window_type],
-        ));
-        if let WindowLevel::DropDown = level {
-            log_x11!(conn.change_window_attributes(
-                self.id,
-                &ChangeWindowAttributesAux::new().override_redirect(1),
-            ));
-        }
     }
 
     /// Bring this window to the front of the window stack and give it focus.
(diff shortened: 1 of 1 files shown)


druid-shell/src/platform/x11/window.rs
```diff
@@ -336,6 +350,29 @@ impl WindowBuilder {
...
+        let min_size = self.min_size.to_px(scale);
+        if self.resizeable {
+            size_hints.min_size = Some((min_size.width as i32, min_size.height as i32));
+        } else {
+            size_hints.min_size = Some(size);
+            size_hints.max_size = Some(size);
+        }
+        log_x11!(size_hints
+            .set_normal_hints(conn.as_ref(), id)
+            .context("set wm normal hints"));
```
Maan2003, line 364: This is similar to `Window::resizable` not sure how to de duplicate it

druid-shell/src/platform/x11/window.rs
```diff
@@ -763,15 +810,86 @@ impl Window {
...
+        let conn = self.app.connection();
+        let scale = self.scale.get();
+        let pos = pos.to_px(scale).expand();
+        log_x11!(conn.configure_window(
+            self.id,
+            &ConfigureWindowAux::new().x(pos.x as i32).y(pos.y as i32),
+        ));
+    }
+
+    fn _set_size(&self, size: Size) {
```
Maan2003, line 857: looking for better name

druid-shell/src/platform/x11/window.rs
jneem: Maybe have a `fn(size: Size, resizable: bool) -> WmSizeHints` somewhere?

Maan2003: oh good idea

Maan2003:
Also, dropdown example from nursery is now functional on x11 backend.

druid-shell/src/platform/x11/window.rs
```diff
@@ -510,6 +558,11 @@ atom_manager! {
         _NET_WM_PID,
         _NET_WM_NAME,
         UTF8_STRING,
+        _NET_WM_WINDOW_TYPE,
+        _NET_WM_WINDOW_TYPE_NORMAL,
+        _NET_WM_WINDOW_TYPE_DROPDOWN_MENU,
+        _NET_WM_WINDOW_TYPE_TOOLTIP,
+        _NET_WM_WINDOW_TYPE_DAILOG,
```
jneem, line 565:
```suggestion
        _NET_WM_WINDOW_TYPE_DAILOG,
```

druid-shell/src/platform/x11/window.rs
```diff
@@ -763,15 +816,79 @@ impl Window {
...
+
+        let conn = self.app.connection();
+        log_x11!(conn.change_property32(
+            xproto::PropMode::REPLACE,
+            self.id,
+            self.atoms._NET_WM_WINDOW_TYPE,
+            AtomEnum::ATOM,
+            &[window_type],
+        ));
+        if let WindowLevel::DropDown = level {
```
jneem, line 884: We probably want this for tooltip and modal also? At least, that's what I did in the gtk backend.

psychon (commented):
I just received a notification due to jneem's comments and apparently I have an old, unsubmitted review laying around... huh?

druid-shell/src/platform/x11/window.rs
```diff
@@ -763,15 +816,79 @@ impl Window {
...
+            log_x11!(conn.change_window_attributes(
+                self.id,
+                &ChangeWindowAttributesAux::new().override_redirect(1),
```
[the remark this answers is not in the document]
jneem: Could we remove the window-level-changing methods from `WindowHandle` and only allow them on `WindowBuilder`? Is there a use case on any platform for changing a `WindowLevel` after the window was created?

druid-shell/src/platform/x11/window.rs
Maan2003: Done

druid-shell/src/platform/x11/window.rs
```diff
@@ -124,36 +142,41 @@ impl WindowBuilder {
     }
 
     pub fn set_size(&mut self, size: Size) {
-        self.size = size;
+        // zero sized window results in server error
```
psychon, line 145: Actually... zero width *or* zero height is not allowed. You only check for "*and*".

psychon (approved):

jneem (approved):

jneem:
Thanks! I was actually wondering if we need `WindowHandle::set_level` on *any* platform, but I'll open another issue for that.

Maan2003:
either `set_position` or `get_position` is not correct. When using `dropdown` example from nursery on a window manager with a top bar. the drop down is not positioned correctly. [Zulip discussion](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/Triage/near/240336948)

image

druid-shell/src/platform/x11/window.rs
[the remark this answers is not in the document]
Maan2003: Looks like [gtk backend](https://github.com/linebender/druid/blob/0b6e1c3868328296cc713e2c2406683db0a0dd01/druid-shell/src/platform/gtk/window.rs#L901-L931) is also doing this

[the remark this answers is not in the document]
psychon: Hm, you are right. Gtk really does this: https://github.com/GNOME/gtk/blob/77f32a69c0a35ffd3a1a542a6057c1c6d13c793d/gdk/x11/gdkwindow-x11.c#L3534-L3553

Still seems like a really bad idea to me. If the WM already reparented the window, the resulting state will not work properly. You can move and resize your window insides of the WM's container window, but the container itself is not affected / does not follow...

psychon:
Well, since you are not assigning a window gravity (and IMO you should not), the functions might not really do exactly what you expect. The ConfigureRequest generated by `set_position` will be interpreted by the WM to be the position for the top left corner of the frame window. The implementation for `get_position` however calculates the position of the top left corner of your child window.

Both positions are correct, they just do not "do the same thing". Which one is correct depends on the exact definition of these method (and I guess is not really defined).

druid-shell/src/platform/x11/window.rs
```diff
@@ -763,15 +844,55 @@ impl Window {
...
     }
 
+    fn get_position(&self) -> Point {
+        fn _get_position(window: &Window) -> Result<Point, Error> {
+            let conn = window.app.connection();
+            let scale = window.scale.get();
+            let geom = conn.get_geometry(window.id)?.reply()?;
+            let cord = conn
+                .translate_coordinates(window.id, geom.root, geom.x, geom.y)?
+                .reply()?;
```
psychon, line 866: Shouldn't the above be `translate_coordinates(window.id, geom.root, 0, 0)`? You want to know which coordinates the point (0,0) in your window has in the root window. As written, this function figures out the position of the window inside of its parent window (which will be the size of the top/left window decorations in a reparenting WM) and then translates this to the root window.

Maan2003:
`translate_coordinates(window.id, geom.root, 0, 0)` fixes it from gnome and xmonad( with top bar)

druid-shell/src/platform/x11/window.rs
Maan2003: Thanks :)

Maan2003:
<details>
<summary>Different desktop environments</summary>
## Gnome
Screenshot from 2021-06-13 10-02-25

## Xmonad
xmonad+topbar

</details>

psychon:
`_NET_FRAME_EXTENTS` does not appear in https://github.com/xmonad/xmonad, so I guess it is unsupported. And the open issues seem to suggest that xmonad does not implement ICCCM's window gravities anyway, but I guess a tiling WM can get away with that easily...

No idea if it is possible with xmonad for a window to request to be floating, but I would expect that a floating window requesting position (0,0) will end up with its window decorations off-screen. :-(

Maan2003:
<details>
<summary>working with gnome, kde, lxde, xmonad</summary>

## gnome 

Screenshot from 2021-06-13 10-02-25

## kde

kde

## lxde

lxde

## xmonad

xmonad+topbar

</details>

PR #1786, opened 2021-05-18
Improve `simulate_input`

CryZe:
Various common operations were not implemented. I'm not 100% why these weren't implemented, but I'm assuming this is because the IME work isn't done yet.

### DIFF
--- a/druid-shell/src/text.rs
+++ b/druid-shell/src/text.rs
@@ -477,7 +477,11 @@ pub fn simulate_input<H: WinHandler + ?Sized>(
             input_handler.set_selection(Selection::caret(new_caret_index));
         }
         KbKey::ArrowLeft => {
-            let movement = Movement::Grapheme(Direction::Left);
+            let movement = if event.mods.ctrl() {
+                Movement::Word(Direction::Left)
+            } else {
+                Movement::Grapheme(Direction::Left)
+            };
             if event.mods.shift() {
                 input_handler.handle_action(Action::MoveSelecting(movement));
             } else {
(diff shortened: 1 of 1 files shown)


druid-shell/src/text.rs
```diff
@@ -531,6 +549,38 @@ pub fn simulate_input<H: WinHandler + ?Sized>(
...
             input_handler.handle_action(action);
         }
+        KbKey::Home => {
+            let movement = Movement::ParagraphStart;
+            if event.mods.shift() {
+                input_handler.handle_action(Action::MoveSelecting(movement));
+            } else {
+                input_handler.handle_action(Action::Move(movement));
+            }
+        }
```
Maan2003, line 559: this doesn't looks correct to me, home moves to the start of line, and ctrl-home to the start of document. 
(similarly for end)

druid-shell/src/text.rs
```diff
@@ -531,6 +549,38 @@ pub fn simulate_input<H: WinHandler + ?Sized>(
...
+        KbKey::End => {
+            let movement = Movement::ParagraphEnd;
+            if event.mods.shift() {
+                input_handler.handle_action(Action::MoveSelecting(movement));
+            } else {
+                input_handler.handle_action(Action::Move(movement));
+            }
+        }
+        KbKey::PageUp => {
+            let movement = Movement::Vertical(VerticalMovement::DocumentStart);
```
Maan2003, line 569: `VerticalMovement::PageUp` looks like a better candidate

CryZe:
Alright I made the changes. However it seems like page up and down movement simply isn't implemented in the textbox for now, but I guess that's fine for now.

CryZe:
Seems like the CI is trolling me.

Maan2003 (approved):
Thanks!

cmyr:
Yea pageup/pagedown aren't implemented because they're surprisingly tricky; they don't modify the selection or cursor at all, and are instead handled by the scrollview.

PR #1787, opened 2021-05-19
Give a WinHandler to the idle callback.

jneem:
I asked about this on zulip and no one replied, so I'm guessing that the reasons for the existing signature of `add_idle_callback` were lost in the mists of time. It used to be that an idle callback got a `&dyn Any`, but that `Any` is always just the `WinHandler` (so, `DruidHandler` in all `druid` apps). Since `&DruidHandler` isn't much use (all the interesting things are `&mut`), I'm guessing that no one was really using it.


raphlinus (approved):
This seems right to me. I'm the one who did the original dyn Any, but I don't think the justification for that was ever very strong (I wanted to loose coupling so druid-shell wouldn't overly specify the higher levels, but I don't see this change really taking away from that).

PR #1792, opened 2021-05-19
[web backend] Handle keydown events correctly

Maan2003:
We were sending KeyDown event before calling `simulate_input` which
also sends KeyDown event internally. Also now we call
`event.prevent_default()` if the event was handled. So we can now
remove the Backspace special case. Tab navigation is also working
correctly now


cmyr (approved):
Thanks, this seems reasonable, we had a similar issue on windows iirc. :)

PR #1793, opened 2021-05-19
[web backend] Limit max level log to debug and update tracing-wasm

Maan2003:
We were logging everything to console, even the trace events. This was a huge performance issue in Firefox(keyboard latency of ~1sec in disabled example!)


SecondFlight:
I haven't tested or anything, but this change looks trivial and seems correct. ¯\_(ツ)_/¯

SecondFlight (approved):

PR #1794, opened 2021-05-19
Snap the TextBox cursor to the pixel grid

CryZe:
This snaps the cursor of the TextBox to the grid to ensure it's always rendered as sharp as possible.

## Screenshot:

Top is with grid snapping, bottom without.

### DIFF
--- a/druid/src/widget/textbox.rs
+++ b/druid/src/widget/textbox.rs
@@ -606,11 +606,15 @@ impl<T: TextStorage + EditableText> Widget<T> for TextBox<T> {
 
             let padding_offset = Vec2::new(textbox_insets.x0, textbox_insets.y0);
 
-            let cursor = if data.is_empty() {
+            let mut cursor = if data.is_empty() {
                 cursor_line + padding_offset
             } else {
                 cursor_line + padding_offset - self.inner.offset()
             };
+
+            // Snap the cursor to the pixel grid so it stays sharp.
+            cursor.p0.x = cursor.p0.x.trunc() + 0.5;
+            cursor.p1.x = cursor.p0.x;
 
             ctx.with_save(|ctx| {
                 ctx.clip(clip_rect);


xarvic (approved):
Thank you!

cmyr:
I'd like to take a look at this before it gets merged please. (I can't figure out how to self-request a review from my phone 😕)

druid/src/widget/textbox.rs
```diff
@@ -606,12 +606,16 @@ impl<T: TextStorage + EditableText> Widget<T> for TextBox<T> {
...
-            let cursor = if data.is_empty() {
+            let mut cursor = if data.is_empty() {
                 cursor_line + padding_offset
             } else {
                 cursor_line + padding_offset - self.inner.offset()
             };
 
+            // Snap the cursor to the pixel grid so it stays sharp.
+            cursor.p0.x = cursor.p0.x.trunc() + 0.5;
+            cursor.p1.x = cursor.p0.x;
```
cmyr, line 617: - this should be correct for coregraphics and skia; i'm not sure if directwrite behaves the same way?
- might `round` make more sense than `trunc`?
- I'm actually also not sure if this will work at all scale factors. it should work at 1x and 2x? I'm less sure about other cases.


All that said: I'm not opposed to this at all, it's just exposing some of my general confusion about pixels and scale factor. I would really love to see a sort of 'unified' theory of this stuff; until we have that it's going to feel like our drawing code is an inconsistently applied series of heuristics. :/

cmyr (commented):
Thanks for this! I do think it's important to get this stuff right, and my main concern is totally unrelated to this patch, and is more about how need some mechanisms for being more consistent about this, across the board. There's a discussion in #1088, and there's been various efforts in the past (https://github.com/linebender/kurbo/pull/93, https://github.com/linebender/kurbo/pull/107, https://github.com/linebender/druid/pull/904).

Anyway, main *actual* question is whether we should `round` to `trunc`?

cmyr:
I'm going to merge this, it's an improvement on the current code and we can always fix other problems later.

PR #1795, opened 2021-05-19
Don't collect invalid rects into vector

CryZe:
It's a slice that we can just easily iterate over. Collecting into an owned Vec might make sense to solve some borrow issue, but there is none here.


jneem (approved):
Thanks!

PR #1797, opened 2021-05-21
Add request_local_layout

dzhou121:
When called from a widget,
it only triggers a re-laoyout on this Widget
and its immediate parents all the way up to the root Widget.
No paint invlidation is done, so the caller needs to take care of it.

As the sibling Widgets' sizes wouldn't be recalculated,
it could cause a bad layout if this Widget's size is changed.

One good example to use this is in a scroll. You can call this
when the scroll content's size is changed to update the state of
the scroll widget, without needs to re-layout and redraw the whole window,
which can be expensive.

Another use case would be if you've got some kind of "floating" Widget.
Because of the fact it's "floating", the origin and size of it doens't
affect any other Widgets.

### DIFF
--- a/druid/src/contexts.rs
+++ b/druid/src/contexts.rs
@@ -112,6 +112,7 @@ pub struct LayoutCtx<'a, 'b> {
     pub(crate) state: &'a mut ContextState<'b>,
     pub(crate) widget_state: &'a mut WidgetState,
     pub(crate) mouse_pos: Option<Point>,
+    pub(crate) needs_layout: bool,
 }
 
 /// Z-order paint operations with transformations.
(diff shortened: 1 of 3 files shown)


jneem:
Thanks for the PR! I worry, though, about whether this approach is composable. For example, a `Label` needs to cause a full re-layout when it changes its text, because it doesn't know whether it belongs to a `Scroll`. You could only use this local re-layout in situations where you have lots of custom widgets and control the whole widget tree.

There was a discussion of a similar feature [on zulip](https://xi.zulipchat.com/#narrow/stream/255910-druid-help/topic/New.20panic/near/217160817), pasted below for convenience, that outlines an approach that seems (to me, at least) more comprehensive.

```text
Zarenor: I'm a little curious about if we should try to pare that down. As an example, any flexible child of a flex won't have new BCs unless the flex gets re-laid out (as a result of a changed BC or inflexible sibling)

Zarenor: Similarly, a SizedBox with fixed width and height is a two-way barrier - it's children won't need to re-layout other than for the Pods to update their origin, and it's children laying out can't affect anything uptree. Are these optimizations worth exploring, or are you saying you don't think they're yet worth the complexity?

Zarenor: Totally just curious, FWIW. If we don't start missing frames because of layout time, I don't know that it's a big deal yet either way (though it could eventually be a battery savings, by decreasing CPU time)

Colin Rofls: my feeling is that yea, it just hasn't creeped up the priority list.

Colin Rofls: there's tons of room for short-circuiting; basically you just need to track the actual widgets that have requested layout, and then in any container where the children that needed layout have returned the same size they previously had, we can skip laying out the other children.
```

JAicewizard:
Could we not have some code in `Scroll` that caches the old layouts and only calculated the ones that need to be refreshed. Then it can lay itself out based on the old (and still correct) layout data from the other widgets. I think if this PR adds support for this + some benchmarks it could be useful for minimizing layout cost.

PR #1798, opened 2021-05-21
Add `Scroll` widget snap feature

becky112358:
For the `Scroll` widget, add the ability such that whenever the scrolled data changes, the view snaps to view the most recent data.

Add an example demonstrating use of this feature.

### DIFF
--- a/druid/src/widget/scroll.rs
+++ b/druid/src/widget/scroll.rs
@@ -34,6 +34,8 @@ use tracing::{instrument, trace};
 pub struct Scroll<T, W> {
     clip: ClipBox<T, W>,
     scroll_component: ScrollComponent,
+    scroll_snap_vertical: Box<dyn Fn(&T, &Env) -> bool>,
+    scroll_snap_horizontal: Box<dyn Fn(&T, &Env) -> bool>,
 }
 
 impl<T, W: Widget<T>> Scroll<T, W> {
@@ -46,6 +48,8 @@ impl<T, W: Widget<T>> Scroll<T, W> {
         Scroll {
             clip: ClipBox::new(child),
             scroll_component: ScrollComponent::new(),
+            scroll_snap_vertical: Box::new(|_, _| false),
+            scroll_snap_horizontal: Box::new(|_, _| false),
         }
     }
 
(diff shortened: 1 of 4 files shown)


druid/src/widget/scroll.rs
```diff
@@ -203,19 +227,37 @@ impl<T: Data, W: Widget<T>> Widget<T> for Scroll<T, W> {
...
+            let viewport = self.clip.viewport();
+            let mut distance = Vec2::ZERO;
+            if (self.scroll_snap_vertical)(data, env) {
+                distance.y = viewport.content_size.height
+                    - (viewport.view_origin.y + viewport.view_size.height);
+            }
+            if (self.scroll_snap_horizontal)(data, env) {
+                distance.x = viewport.content_size.width
+                    - (viewport.view_origin.x + viewport.view_size.width);
+            }
```
becky112358, line 257: Alternatively, we could scroll by the viewport content size and allow the clip box to handle panning us back into the visible area. For example:
```
if (self.scroll_snap_vertical)(data, env) {
    distance.y = viewport.content_size.height;
}
if (self.scroll_snap_horizontal)(data, env) {
    distance.x = viewport.content_size.width;
}
```

Is it sloppy to give an excessive scroll distance and to rely on the clip box to pan us back into the visible area?
Or is it better to give an excessive scroll distance, because the clip box will check our view anyway so then we only do the computation once (and the computation location is centralised)?

cmyr (commented):
I think this is a really nice idea, but I have practical concerns. The data can change in lots of ways, and in different places at the same time, and I think it will be difficult to design this as an automated mechanism in such a way that it actually fulfils the expectations of the user in all cases.

My gut feeling is that doing this correctly requires the app author to handle the scrolling manually. This is possible with the `scroll_to` method on `Scroll`.

I had intended to expose a mechanism that lets a particular child widget request that it's nearest parent scroll view scroll to make it visible; this would involve sending some `Notification` up the tree during an event, which the `Scroll` widget would know how to handle. Would that cover your use case?

becky112358:
@cmyr Yes, I think what you describe would cover my case.

The parent `Scroll` widget would need to be able to handle that `Notification` after the child widget has performed its `layout()` call. Otherwise the user will always be displayed the second-to-last data that is being scrolled, and not the last. But that is how the ordering is already handled, right?

cmyr:
Good point, that's a hiccup. I recall bumping into something like this with the `TextBox`, and there's special handling there for "needs_scroll_after_layout". It might be worth taking a look at that, and seeing if something can be generalized from it?

becky112358:
Thanks for the suggestion. I had a go at implementing a `Scroll` snap feature according to my understanding of your suggestion, but my code came out basically the same as my first implementation (maybe just without line 246 in `scroll.rs`
```
&& self.clip.viewport().content_size != old_content_size
```
)

So I will not be continuing on this. Should I close the PR?

cmyr:
I'm sorry not to be able to offer more useful guidance here but my focus is definitely elsewhere at the moment, and you've stumbled (as I'm sure you've realized) into a subtly difficult problem.

I'm not sure what the best longer-term design is, for this? The fact that you sometimes need to perform some action *after* layout (particularly for scroll views) is definitely real, but I want to make sure that whatever mechanism exists for this works in a flexible and controllable way, and I think that just tying it to data changes leaves to much to chance.

I'm not sure what the best actual design is, though. maybe some mechanism by which child views can specify an anchor from some position in the child to some position in the scroll space? I'm just thinking out loud.

PR #1799, opened 2021-05-21
Transform box

xarvic:
This RP implements a widget wrapper which allows arbitrary transforms of the inner widget using the `TransformPolicy` trait and some implementations of the `TransformPolicy` trait.

TODO:
- [ ] fix `Widget::parent_window_origin`. At the moment this value wrong inside of a `TransformBox`
- [ ] calculate damaged-region correctly


jneem:
I can't actually find the conversation now, but around the time that partial invalidation was first introduced, there was some discussion about allowing arbitrary transformations. My recollection was that we decided to wait and see if there was a compelling use case first, and then figure out if we can support it. Do you have some specific situations where you want, e.g., an arbitrary rotation?

The main issue is that having arbitrary transformations interacts in complicated ways with other features (such as detecting window-relative mouse position for launching pop-ups, and the fact that the current layout code that tries to align things to pixel boundaries will break if a scaling is applied).

xarvic:
I don't really need it. It would make some things easier, but its not that important to me. I heard other people talk about it (i also can't find the comments) and thought maybe someone else needs this, but i understand your concerns.

jneem:
Yeah, it's definitely come up a couple of times. I think the closest we had to a version with a real use-case was for the Tabs widget (#1160), which was only for 90,180,270-degree rotations. But even in that case, we decided to punt...

xarvic:
`AaTransform` does exactly this. One option would be to export only `AaTransform` until someone needs the rest.

cmyr:
Thank you for the patch, but I remain reluctant here. In general, UI elements are designed with an orientation, and should not be arbitrarily transformed. The only case I can personally really think of is something like an animated transition, but that remains entirely theoretical.

If someone can demonstrate a compelling case where this is helpful then I am open to that, but in general it feels like bringing in a bunch of code that will be very rarely used, and which will have to be maintained.

xarvic:
Ok

PR #1801, opened 2021-05-22
x11: Add support for custom cursors

psychon:
This adds custom cursor support to the X11 backend, fixing an open todo from #1755.

Half of the added code is "RENDER initialisation" (is the necessary RENDER version supported? What is the id of the standard ARGB32 format?). The other half of the code "upload this image to the X11 server". I guess the later could be simplified somehow, but I was lazy and did it "manually". This could also be made faster, but I don't think creating cursors is performance critical...

Oh and: Nothing ever deletes custom cursors, so they stay around forever, eating memory in the X11 server. I wonder if one can make the X11 server go out-of-memory with cursors...

This code is completely untested since I can't build druid and thus cannot run the cursor example due to a too old Rust compiler.

### DIFF
--- a/druid-shell/src/platform/x11/window.rs
+++ b/druid-shell/src/platform/x11/window.rs
@@ -14,6 +14,7 @@
 
 //! X11 window creation and window management.
 
+use std::borrow::Cow;
 use std::cell::{Cell, RefCell};
 use std::collections::BinaryHeap;
 use std::convert::{TryFrom, TryInto};
@@ -29,7 +30,10 @@ use cairo::{XCBConnection as CairoXCBConnection, XCBDrawable, XCBSurface, XCBVis
 use tracing::{error, info, warn};
 use x11rb::atom_manager;
 use x11rb::connection::Connection;
+use x11rb::errors::ReplyOrIdError;
+use x11rb::image::{BitsPerPixel, Image, ImageOrder, ScanlinePad};
 use x11rb::protocol::present::{CompleteNotifyEvent, ConnectionExt as _, IdleNotifyEvent};
+use x11rb::protocol::render::{ConnectionExt as _, Pictformat};
 use x11rb::protocol::xfixes::{ConnectionExt as _, Region as XRegion};
 use x11rb::protocol::xproto::{
     self, AtomEnum, ChangeWindowAttributesAux, ConfigureNotifyEvent, ConnectionExt, CreateGCAux,
@@ -883,8 +887,7 @@ impl Window {
             Cursor::NotAllowed => cursors.not_allowed,
             Cursor::ResizeLeftRight => cursors.col_resize,
             Cursor::ResizeUpDown => cursors.row_resize,
-            // TODO: (x11/custom cursor)
-            Cursor::Custom(_) => None,
+            Cursor::Custom(custom) => Some(custom.0),
         };
         if cursor.is_none() {
             warn!("Unable to load cursor {:?}", cursor);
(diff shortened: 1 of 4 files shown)


psychon:
Hm, I think the x11rb-image dependency is not actually needed here. That code doesn't actually do anything useful here. Future-me should double-check that and possibly remove the dependency. (Note to self)

Maan2003:
looks like channels are wrong. Also it is less clear
cursor example x11:
x11-cursor
cursor example gtk:
cursor-gtk

druid-shell/src/platform/x11/window.rs
```diff
@@ -1669,3 +1694,59 @@ unsafe impl HasRawWindowHandle for WindowHandle {
...
+
+fn make_cursor(
+    conn: &XCBConnection,
+    root_window: u32,
+    argb32_format: Pictformat,
+    desc: &CursorDesc,
+) -> Result<Cursor, ReplyOrIdError> {
+    // BEGIN: Lots of code just to get the image into a RENDER Picture
+
+    // No idea how to sanely get the pixel values, so I'll go with 'insane':
```
Maan2003, line 1706: [`ImageBuf::raw_pixels`](https://docs.rs/piet/0.4.0/piet/struct.ImageBuf.html#method.raw_pixels) should work

psychon: Yeah, but then I need to handle all the different possible `ImageFormat` values. The GTK backend does not do that (I guess it only handles two of the four possible values properly), so this seems complicated/untested.

And with `raw_pixels`, `RgbaSeparate` still needs to be converted to premultiplied "by hand". The simplest way for that could be the already existing code, I guess.

psychon:
Updated. Here is a screenshot. On the left is the actual cursor and on the right is a screenshot of gimp showing the same image. I added some more colors to the image for testing:

Unbenannt

Seems like I got the component order completely wrong. And treating unpremultiplied pixels as premultiplied looks worse than I expected, so I also implemented that todo. The big-endian code is completely untested, but might work. Or not...

I have no idea which component order are required. `xdpyinfo -ext RENDER` here has nine picture formats with `depth: 32`. Some of them have no alpha component, some have depth 10 per color component. There are only two with 8 bits for each of r/g/b/a: argb and bgra. No idea if both of them are required, the spec only requires:

> The server must support a Direct PictFormat with 8 bits each of red, green,
> blue and alpha as well as a Direct PictFormat with 8 bits of red, green and
> blue and 0 bits of alpha.  The server must also support Direct PictFormats
> with 1, 4 and 8 bits of alpha and 0 bits of r, g and b.


Also: Thanks a lot for the hint on how to get druid to work with my ancient Rust compiler.

Maan2003 (approved):
Looks good! the cursor example is working correctly

psychon:
It just occurred to me that one could let piet-cairo and cairo handle the "complicated" part of uploading stuff to the X11 server in the right format. I briefly followed down this road, but when I figured out I'd need to call `XCBSurface::create_with_xrender_format`, I gave up. This requires more fun like the already existing `xcb_visualtype_t` hack that exists in the code: Basically providing a Rust translation of two more xcb C structs. Perhaps the existing code is good enough.

Maan2003:
I am in favor of "existing code is good enough"

Maan2003:
:pensive: conflict resolution failed

psychon:
Here is my attempt at merging with master. I also added the use for `TryInto` in the merge commit (a8867666cdd94f4eaee9 removed it).

I didn't test anything besides "it builds".

PR #1803, opened 2021-05-23
X11: Add support for transparent windows

psychon:
First, some history lesson:

X11 originally did not support transparency.  Visuals describe how pixel
values are interpreted, but they can only describe red/green/blue
components. Then, the RENDER extension came and added transparency. Now
we have picture formats. That's basically the same as a visual, but it
also describes an alpha component. There is a mapping between picture
formats and visual. Also, there is a standard ARGB32 format that is
always supported.

Then came the composite extension. Classically, the X11 server "drew
stuff" directly to the output, but there is no alpha compositing
involved.  With composite, a composite manager can request drawing to be
redirected to an off-screen pixmap. The composite manager will then
produce the "real" visible screen contents. And in doing so, it can take
alpha values into account.

This commit adds support for transparent windows to the X11 backend. For
that, it has to find the standard ARGB32 render picture format and the
corresponding visual. When a transparent window is created, it checks if
a composite manager is running. If not, nothing changes. If everything
worked out (RENDER is supported, we found the ARGB32 picture format, a
composite manager is running), transparent windows are created with this
visual instead of the root window's visual.

When a window has a different bit depth than its parent window, it has
to "fully define" its rendering. Since it has a different bit depth than
its parent, the X11 server cannot just copy background and window border
from the parent. Thus, when we create an ARGB32 window (which has
depth=32), we are creating a window with a different depth than the root
window (most likely depth=24) and have to provide some additional
settings. Even though no border will be used, we still have to define
its color.

------

Since I can only build druid-shell and not druid, this PR only received light testing: I let it open some window and asked `xwininfo` for the window's depth. I only tested that this produces the expected result with a hardcoded `transparent: true` and with/without a composite manager. I did not actually test anything involving transparent window background, but I expect  that to work since cairo should handle all that is necessary. Hopefully...

### DIFF
--- a/druid-shell/src/platform/x11/util.rs
+++ b/druid-shell/src/platform/x11/util.rs
@@ -19,11 +19,11 @@ use std::rc::Rc;
 use std::time::Instant;
 
 use anyhow::{anyhow, Error};
-use x11rb::errors::ReplyError;
 use x11rb::connection::RequestConnection;
+use x11rb::errors::ReplyError;
 use x11rb::protocol::randr::{ConnectionExt, ModeFlag};
-use x11rb::protocol::xproto::{Screen, Visualtype, Window};
 use x11rb::protocol::render::{self, ConnectionExt as _};
+use x11rb::protocol::xproto::{Screen, Visualtype, Window};
 use x11rb::xcb_ffi::XCBConnection;
 
 use crate::window::TimerToken;
@@ -87,17 +87,24 @@ pub fn get_visual_from_screen(screen: &Screen) -> Option<Visualtype> {
     find_visual_from_screen(screen, screen.root_visual)
 }
 
-pub fn get_argb_visual_type(conn: &XCBConnection, screen: &Screen) -> Result<Option<Visualtype>, ReplyError> {
+pub fn get_argb_visual_type(
+    conn: &XCBConnection,
+    screen: &Screen,
+) -> Result<Option<Visualtype>, ReplyError> {
     // Getting a visual is already funny, but finding the ARGB32 visual is even more fun.
     // RENDER has picture formats. Each format corresponds to a visual. Thus, we first find the
     // right picture format, then find the corresponding visual id, then the Visualtype.
-    if conn.extension_information(render::X11_EXTENSION_NAME)?.is_none() {
+    if conn
+        .extension_information(render::X11_EXTENSION_NAME)?
+        .is_none()
+    {
         // RENDER not supported
         Ok(None)
     } else {
         let pict_formats = conn.render_query_pict_formats()?.reply()?;
         // Find the ARGB32 standard format
-        let res = pict_formats.formats
+        let res = pict_formats
+            .formats
             .iter()
             .find(|format| {
                 format.type_ == render::PictType::DIRECT
@@ -111,21 +118,26 @@ pub fn get_argb_visual_type(conn: &XCBConnection, screen: &Screen) -> Result<Opt
                     && format.direct.alpha_shift == 24
                     && format.direct.alpha_mask == 0xff
             })
-        // Now find the corresponding visual ID
-        .and_then(|format|
-            pict_formats.screens
-                .iter()
-                .filter_map(|screen| screen.depths
+            // Now find the corresponding visual ID
+            .and_then(|format| {
+                pict_formats
+                    .screens
+                    .iter()
+                    .filter_map(|screen| {
+                        screen
+                            .depths
                             .iter()
-                            .filter_map(|depth| depth.visuals
-                                        .iter()
-                                        .find(|visual| visual.format == format.id)
-                                        .map(|visual| visual.visual)
-                                        )
+                            .filter_map(|depth| {
+                                depth
+                                    .visuals
+                                    .iter()
+                                    .find(|visual| visual.format == format.id)
+                                    .map(|visual| visual.visual)
+                            })
                             .next()
-                            )
-                .next()
-                )
+                    })
+                    .next()
+            })
             // And finally, we can find the visual
             .and_then(|visual_id| find_visual_from_screen(screen, visual_id));
         Ok(res)
--- a/druid-shell/src/platform/x11/util.rs
+++ b/druid-shell/src/platform/x11/util.rs
@@ -91,20 +91,20 @@ pub fn get_argb_visual_type(
     conn: &XCBConnection,
     screen: &Screen,
 ) -> Result<Option<Visualtype>, ReplyError> {
-    fn find_visual_for_format(reply: &render::QueryPictFormatsReply, id: render::Pictformat) -> Option<Visualid> {
-        let find_in_depth = |depth: &render::Pictdepth| depth
-            .visuals
-            .iter()
-            .find(|visual| visual.format == id)
-            .map(|visual| visual.visual);
-        let find_in_screen = |screen: &render::Pictscreen| screen
-            .depths
-            .iter()
-            .find_map(find_in_depth);
-        reply
-            .screens
-            .iter()
-            .find_map(find_in_screen)
+    fn find_visual_for_format(
+        reply: &render::QueryPictFormatsReply,
+        id: render::Pictformat,
+    ) -> Option<Visualid> {
+        let find_in_depth = |depth: &render::Pictdepth| {
+            depth
+                .visuals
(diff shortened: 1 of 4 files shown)


Maan2003:
Just tested the transparency example, does **not** work on kde, xfce or picom (didn't check on others).
<details>
<summary>Errors</summary>
<pre>
 INFO druid_shell::platform::x11::application: X server supports Present version 1.0
 INFO druid_shell::platform::x11::application: X server supports XFIXES version 5.0
DEBUG druid::localization: available locales [en-US, de-DE, fr-CA], current en-US
DEBUG druid::localization: resolved: [en-US]
 WARN druid_shell::platform::x11::window: WindowBuilder::resizable is currently unimplemented for X11 platforms.
 WARN druid_shell::platform::x11::window: WindowBuilder::show_titlebar is currently unimplemented for X11 platforms.
DEBUG druid::window: TextFieldToken(1) added
ERROR druid_shell::platform::x11::application: Error handling event: X11 error X11Error { error_kind: Match, error_code: 8, sequence: 107, bad_value: 38, minor_opcode: 4, major_opcode: 139 }
ERROR druid_shell::platform::x11::application: Error handling event: X11 error X11Error { error_kind: RenderPicture, error_code: 143, sequence: 108, bad_value: 58720289, minor_opcode: 26, major_opcode: 139 }
ERROR druid_shell::platform::x11::application: Error handling event: X11 error X11Error { error_kind: RenderPicture, error_code: 143, sequence: 109, bad_value: 58720289, minor_opcode: 26, major_opcode: 139 }
ERROR druid_shell::platform::x11::application: Error handling event: X11 error X11Error { error_kind: RenderPicture, error_code: 143, sequence: 110, bad_value: 58720289, minor_opcode: 26, major_opcode: 139 }
ERROR druid_shell::platform::x11::application: Error handling event: X11 error X11Error { error_kind: RenderPicture, error_code: 143, sequence: 112, bad_value: 58720289, minor_opcode: 10, major_opcode: 139 }
ERROR druid_shell::platform::x11::application: Error handling event: X11 error X11Error { error_kind: RenderPicture, error_code: 143, sequence: 114, bad_value: 58720289, minor_opcode: 10, major_opcode: 139 }
ERROR druid_shell::platform::x11::application: Error handling event: X11 error X11Error { error_kind: RenderPicture, error_code: 143, sequence: 116, bad_value: 58720289, minor_opcode: 10, major_opcode: 139 }
ERROR druid_shell::platform::x11::application: Error handling event: X11 error X11Error { error_kind: RenderPicture, error_code: 143, sequence: 118, bad_value: 58720289, minor_opcode: 10, major_opcode: 139 }
ERROR druid_shell::platform::x11::application: Error handling event: X11 error X11Error { error_kind: RenderPicture, error_code: 143, sequence: 143, bad_value: 58720289, minor_opcode: 8, major_opcode: 139 }
ERROR druid_shell::platform::x11::application: Error handling event: X11 error X11Error { error_kind: RenderPicture, error_code: 143, sequence: 146, bad_value: 58720289, minor_opcode: 10, major_opcode: 139 }
ERROR druid_shell::platform::x11::application: Error handling event: X11 error X11Error { error_kind: RenderPicture, error_code: 143, sequence: 154, bad_value: 58720289, minor_opcode: 24, major_opcode: 139 }
ERROR druid_shell::platform::x11::application: Error handling event: X11 error X11Error { error_kind: Match, error_code: 8, sequence: 156, bad_value: 58720293, minor_opcode: 1, major_opcode: 148 }
</pre>
</details>

I think your build error can solved by commenting out the [const generic data impl for array](https://github.com/linebender/druid/blob/e862b42aa1c17bb014cfc754c3427623b5fcf0b6/druid/src/data.rs#L614)

psychon:
Thanks for testing. Since the following errors are Picture errors, I guess that Match error is a RENDER CreatePicture request. `minor_opcode: 4` matches that theory.

According to https://gitlab.freedesktop.org/xorg/proto/renderproto/-/blob/master/renderproto.txt

>  It is a Match error
>	to specify a format with a different depth than the drawable.  If
>	the drawable is a Window then the Red, Green and Blue masks must
>	match those in the visual for the window else a Match error is
>	generated.

Since this PR doesn't add any CreatePicture requests, I guess that this comes from cairo. Weird...

> I think your build error can solved by commenting out the const generic data impl for array

Thanks. I didn't even look at the code to see how simple it might be to get this to work.

I won't have time to investigate for the next seven days. Afterwards, I'll take a look.

psychon:
As I already quoted from the spec above:

> It is a Match error
> to specify a format with a different depth than the drawable. If
> the drawable is a Window then the Red, Green and Blue masks must
> match those in the visual for the window else a Match error is
> generated.

The window in question uses visual 0x72. The picture is supposed to use format 0x26. (Both values taken from the output of ` xtrace -o /tmp/t cargo run --no-default-features --features=x11 --example=transparency`.) According to `xdpyinfo -ext RENDER`:
```
  visual:
    visual id:    0x72
    class:    TrueColor
    depth:    32 planes
    available colormap entries:    256 per subfield
    red, green, blue masks:    0xff0000, 0xff00, 0xff
    significant bits in color specification:    8 bits
```
and
```
  pict format:
        format id:    0x26
        type:         Direct
        depth:        32
        alpha:        24 mask 0xff
        red:          16 mask 0xff
        green:         8 mask 0xff
        blue:          0 mask 0xff
```
That looks pretty much the same to me....?
The same output later even says that the two go together (which doesn't really surprise me since the code is using the same table to map from visual to picture):
```
      visual format:
        visual id:      0x72
        pict format id: 0x26
```
I really don't know what is going on here and will take a closer look later this week.

psychon:
Urgh. Turns out I was mis-reading "stuff" previously. The code was really using a wrong depth in one place (`screen.root_depth`). I replaced that with the correct value and now the `transparency` example actually works fine here with `xcompmgr`.

druid-shell/src/platform/x11/util.rs
```diff
@@ -69,17 +72,78 @@ pub fn refresh_rate(conn: &Rc<XCBConnection>, window_id: Window) -> Option<f64>
...
+                    && format.direct.red_mask == 0xff
+                    && format.direct.green_shift == 8
+                    && format.direct.green_mask == 0xff
+                    && format.direct.blue_shift == 0
+                    && format.direct.blue_mask == 0xff
+                    && format.direct.alpha_shift == 24
+                    && format.direct.alpha_mask == 0xff
+            })
+            // Now find the corresponding visual ID
+            .and_then(|format| {
```
Maan2003, line 122: These chain are getting a bit large. is helpful to extract these into a functions that returns an Option?

druid-shell/src/platform/x11/util.rs
```diff
@@ -69,17 +72,78 @@ pub fn refresh_rate(conn: &Rc<XCBConnection>, window_id: Window) -> Option<f64>
...
+                pict_formats
+                    .screens
+                    .iter()
+                    .filter_map(|screen| {
```
Maan2003, line 126: not sure why clippy is not warning about `filter_map` and `next`. You can use `find_map`

Maan2003 (approved):
Working :tada: . Sorry for being late.

> Do you want me to merge master into this PR whenever that occurs? Should I rewrite some git history?

You can do either. We squash and merge so it doesn't matter

druid-shell/src/platform/x11/util.rs
psychon: Done. You may decide on the "is it helpful". ;-)

psychon:
CI fails due to `cargo fmt` in code that is not touched in this PR. If you want, I can fix that in the PR, but I don't really like touching "random code" outside of the PR just to please CI. I'd rather wait for this to be fixed in master (since this must make all PRs fail...?)

Maan2003:
> CI fails due to cargo fmt in code that is not touched in this PR.

You touched it :sweat_smile: (`screen.root_depth` to `depth`)

`> cargo fmt -- --check`
```diff
         let buffers = RefCell::new(Buffers::new(
-            conn,
-            id,
-            buf_count,
-            width_px,
-            height_px,
-            depth,
+            conn, id, buf_count, width_px, height_px, depth,
         )?);
```

psychon:
Argh, sorry. Apparently I am not qualified to read git histories with merges. :-(

Issue #937, opened 2020-05-15
X11 shell does not support clipboards.


Finnerale:

psychon:
The relevant standard here is ICCCM. I recently-ish had the joy of implementing support for this. Since this X11, of course you need to create windows. Read the sections of ICCCM about *selections*.

The general process (for getting the selection) is that you send a `ConvertSelection` request and some time later receive a `SelectionNotify` event. Large data transfers then additionally involve some dance with window properties. This is called `INCR` and described in ICCCM.

For setting a selection, you create a window and make it the selection owner with `SetSelectionOwner`. Then you have to react to `SelectionRequest` events.  `SelectionClear` tells you when something else became the selection owner.

The relevant selections are `CLIPBOARD`. Perhaps you also want to handle the `PRIMARY` selection. This is the currently selected text. Most (?) programs support pasting this via the middle mouse button.

psychon:
I took another look at the API and... I really would prefer a callback-based clipboard API in druid. That would be easier to implement.

The GTK backend uses `gtk_clipboard_wait_for_contents()` which is documented as: https://developer.gnome.org/gtk3/stable/gtk3-Clipboards.html#gtk-clipboard-wait-for-contents

> This function waits for the data to be received using the main loop, so events, timeouts, etc, may be dispatched during the wait.

Something similar would be needed for the x11 backend for a synchronous API. And no idea what then should happen when "the other end" does not reply to a clipboard event. Perhaps one could just "collect" all X11 events and stuff them into a list for later processing, waiting only for clipboard events.

On the other hand, a synchronous API is likely way nicer for the API user, so... yeah, I'm not really sure. At least I'll remove this from my list of "things to try to implement in druid-x11 during vacation".

psychon:
#1805 handles the "get clipboard contents"-part. I also implemented setting clipboard contents: https://github.com/linebender/druid/compare/master...psychon:x11-put-clipboard?expand=1

I want to wait for #1805 to "go through" before tackling setting the clipboard contents. So far I only tested this with something like the following (I also tested this with a shorter string):
```rs
fn main() {
    tracing_subscriber::fmt().init();
    let app = druid_shell::Application::new().unwrap();
    let mut contents = String::from("1234567890");
    while contents.len() < 17 * 1024 * 1024 {
        let copy = contents.clone();
        contents = contents + &copy;
    }
    app.clipboard().put_string(contents);
    app.run(None);
}
```
[...]

PR #1805, opened 2021-05-23
Implement getting X11 clipboard contents

psychon:
This PR adds support for getting the clipboard contents. I only tested this with strings, but would expect that other kinds of data should work as well. There is still lots of TODO left here: Only `get_string` and `get_format` are implemented. Of course all the `put_*` are missing, but also `available_type_names` and `preferred_format` are missing since I ran out of time for now.

Edit: Updated to also implement `available_type_names` and `preferred_format`. `put_*` is best left for another PR, I think, because that requires keeping some state and my best idea for that so far is to change `Clipboard` into `Rc<ClipboardState>` so that the clipboard contents can be kept in `ClipboardState`.

### DIFF
--- a/druid-shell/src/platform/x11/clipboard.rs
+++ b/druid-shell/src/platform/x11/clipboard.rs
@@ -75,11 +75,10 @@ impl Clipboard {
     }
 
     pub fn get_string(&self) -> Option<String> {
-        STRING_TARGETS
-            .iter()
-            .find_map(|target| self
-                 .get_format(target)
-                 .and_then(|data| String::from_utf8(data).ok()))
+        STRING_TARGETS.iter().find_map(|target| {
+            self.get_format(target)
+                .and_then(|data| String::from_utf8(data).ok())
+        })
     }
 
     pub fn preferred_format(&self, formats: &[FormatId]) -> Option<FormatId> {
--- a/druid-shell/src/platform/x11/clipboard.rs
+++ b/druid-shell/src/platform/x11/clipboard.rs
@@ -77,8 +77,9 @@ impl Clipboard {
     pub fn get_string(&self) -> Option<String> {
         STRING_TARGETS
             .iter()
-            .find_map(|target| self.get_format(target))
-            .and_then(|data| String::from_utf8(data).ok())
+            .find_map(|target| self
+                 .get_format(target)
+                 .and_then(|data| String::from_utf8(data).ok()))
     }
 
     pub fn preferred_format(&self, formats: &[FormatId]) -> Option<FormatId> {
--- a/druid-shell/src/platform/x11/clipboard.rs
+++ b/druid-shell/src/platform/x11/clipboard.rs
@@ -75,12 +75,10 @@ impl Clipboard {
     }
 
     pub fn get_string(&self) -> Option<String> {
-        for target in STRING_TARGETS.iter() {
-            if let Some(data) = self.get_format(target) {
-                return String::from_utf8(data).ok();
-            }
-        }
-        None
+        STRING_TARGETS
+            .iter()
+            .find_map(|target| self.get_format(target))
+            .and_then(|data| String::from_utf8(data).ok())
     }
 
     pub fn preferred_format(&self, formats: &[FormatId]) -> Option<FormatId> {
--- a/druid-shell/src/platform/x11/clipboard.rs
+++ b/druid-shell/src/platform/x11/clipboard.rs
@@ -14,13 +14,56 @@
 
 //! Interactions with the system pasteboard on X11.
 
-use crate::clipboard::{ClipboardFormat, FormatId};
-use tracing::warn;
+use std::cell::{Cell, RefCell};
+use std::collections::VecDeque;
+use std::rc::Rc;
 
-#[derive(Debug, Clone, Default)]
-pub struct Clipboard;
+use x11rb::connection::Connection;
+use x11rb::errors::ReplyOrIdError;
+use x11rb::protocol::xproto::{
+    AtomEnum, ChangeWindowAttributesAux, ConnectionExt, EventMask, GetPropertyReply,
+    GetPropertyType, Property, Timestamp, WindowClass,
+};
+use x11rb::protocol::Event;
+use x11rb::xcb_ffi::XCBConnection;
+
+use crate::clipboard::{ClipboardFormat, FormatId};
+use tracing::{debug, warn};
+
+// We can pick an arbitrary atom that is used for the transfer. This is our pick.
+const TRANSFER_ATOM: AtomEnum = AtomEnum::CUT_BUFFE_R4;
+
+const STRING_TARGETS: [&str; 5] = [
+    "UTF8_STRING",
+    "TEXT",
+    "STRING",
+    "text/plain;charset=utf-8",
+    "text/plain",
+];
+
+#[derive(Debug, Clone)]
+pub struct Clipboard {
+    connection: Rc<XCBConnection>,
+    screen_num: usize,
+    event_queue: Rc<RefCell<VecDeque<Event>>>,
+    timestamp: Rc<Cell<Timestamp>>,
+}
 
 impl Clipboard {
+    pub(crate) fn new(
+        connection: Rc<XCBConnection>,
+        screen_num: usize,
+        event_queue: Rc<RefCell<VecDeque<Event>>>,
+        timestamp: Rc<Cell<Timestamp>>,
+    ) -> Self {
+        Clipboard {
+            connection,
+            screen_num,
+            event_queue,
+            timestamp,
+        }
+    }
+
     pub fn put_string(&mut self, _s: impl AsRef<str>) {
         // TODO(x11/clipboard): implement Clipboard::put_string
         warn!("Clipboard::put_string is currently unimplemented for X11 platforms.");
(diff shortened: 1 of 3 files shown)


psychon:
I forgot to mention in any of the commits: This is a first step towards #937. (Now that I left this here, GH will surely link the issue to this PR)

druid-shell/src/platform/x11/clipboard.rs
```diff
@@ -43,10 +89,107 @@ impl Clipboard {
...
+                Event::SelectionNotify(notify) if notify.requestor == window.window => {
+                    break notify
+                }
+                event => self.event_queue.borrow_mut().push_back(event),
+            }
+        };
+
+        if notify.property == x11rb::NONE {
+            // Selection is empty
+            Ok(None)
```
Maan2003, line 139: can early return

druid-shell/src/platform/x11/clipboard.rs
```diff
@@ -43,10 +89,107 @@ impl Clipboard {
...
+                    window.window,
+                    TRANSFER_ATOM,
+                    GetPropertyType::ANY,
+                    0,
+                    u32::MAX,
+                )?
+                .reply()?;
+
+            if property.type_ != incr_atom {
+                Ok(Some(property.value))
```
Maan2003, line 158: can early return

Maan2003 (commented):
Thanks! I have tested this for strings.

druid-shell/src/platform/x11/clipboard.rs
psychon: Uhm, what do you mean? This `if` is the last expression in the function. This is the return value.

Do you want me to add an explicit `return` for clarity?

psychon:
Oh and: Somewhen, this might need timeouts. If the transfer did not finish after [time], abort and return None.

However, now is not the time for that. (As in, I don't have the time right now.)

druid-shell/src/platform/x11/clipboard.rs
Maan2003: Yes, this is the last expression. but using return here the code in the else block can be moved out. Also yes, explicit return looks clearer

psychon: Okay, done.

psychon:
I updated this to also implemented `available_type_names()` and `preferred_format()`.

`available_type_names()` is just a transfer of type `TARGETS`. This produces a list of `Atom`s (i.e.: `Vec<u32>`). This list then just has to go through `self.connection.get_atom_name` (element-wise).

`preferred_format()` finds the first entry in the result of `available_type_names()`, which seems to be what the gtk backend is doing.

I tested this with:
```rs
fn main() {
    let app = druid_shell::Application::new().unwrap();
    println!("{:?}", app.clipboard().available_type_names());
    println!("{:?}", app.clipboard().preferred_format(&["FOOBAR", "UTF8_STRING"]));
}
```
Output is:
```
["TIMESTAMP", "TARGETS", "MULTIPLE", "_VIMENC_TEXT", "_VIM_TEXT", "UTF8_STRING", "COMPOUND_TEXT", "TEXT", "STRING"]
Some("UTF8_STRING")
```
With the gtk backend I get:
```
["TIMESTAMP (133)", "TARGETS (129)", "MULTIPLE (132)", "_VIMENC_TEXT (136)", "_VIM_TEXT (137)", "UTF8_STRING (70)", "COMPOUND_TEXT (138)", "TEXT (139)", "STRING (31)"]
Some("UTF8_STRING")
```
Dunno why the gtk backend also includes the number of the atom. That does not seem useful to me and would break my implementation of `preferred_format()`, so I did not do that.

cmyr:
Taking a look at this, and just wanted to make sure that the changes related to the screen number and timestamps is intended to be part of this PR?

Maan2003:
timestamp is needed for the clipboard, screen_num looks like a unrelated refactor

psychon:
To clean up this PR, I rebased it on current `master` and squashed all the commits (since you use squash merges anyway). I also wrote a hopefully good commit message for this new commit.

cmyr:
@psychon if that's the only unrelated change then no worries, but if there's other stuff in here that's unrelated breaking them out into a separate PR will definitely make review easier.

druid-shell/src/platform/x11/clipboard.rs
```diff
@@ -32,26 +75,193 @@ impl Clipboard {
...
 
     pub fn get_string(&self) -> Option<String> {
-        // TODO(x11/clipboard): implement Clipboard::get_string
-        warn!("Clipboard::set_string is currently unimplemented for X11 platforms.");
+        for target in STRING_TARGETS.iter() {
+            if let Some(data) = self.get_format(target) {
+                return String::from_utf8(data).ok();
+            }
+        }
         None
```
Maan2003, line 83: I couldn't resist :sweat_smile: . But feel free to ignore this
```suggestion
        STRING_TARGETS
            .iter()
            .find_map(|target| self.get_format(target))
            .and_then(|data| String::from_utf8(data).ok())
```

druid-shell/src/platform/x11/clipboard.rs
```diff
@@ -32,26 +75,193 @@ impl Clipboard {
     }
 
     pub fn get_string(&self) -> Option<String> {
-        // TODO(x11/clipboard): implement Clipboard::get_string
-        warn!("Clipboard::set_string is currently unimplemented for X11 platforms.");
+        for target in STRING_TARGETS.iter() {
+            if let Some(data) = self.get_format(target) {
+                return String::from_utf8(data).ok();
```
Maan2003, line 80: Should we ignore the invalid UTF-8 items instead?

druid-shell/src/platform/x11/clipboard.rs
```diff
@@ -32,26 +75,193 @@ impl Clipboard {
...
+
+        let conn = &*self.connection;
+        let (format_atom, clipboard_atom, incr_atom) = {
+            let format = conn.intern_atom(false, format.as_bytes())?;
+            let clipboard = conn.intern_atom(false, b"CLIPBOARD")?;
+            let incr = conn.intern_atom(false, b"INCR")?;
+            (
+                format.reply()?.atom,
+                clipboard.reply()?.atom,
+                incr.reply()?.atom,
```
Maan2003, line 152: if we delaying this reply after window creation, is it worth the less readability?

Maan2003 (approved):
Looks good! Just some minor comments

druid-shell/src/platform/x11/clipboard.rs
psychon: Uhm. What exactly do you mean with "is it worth"? What's the benefit?

I guess you want to understand a round-trip to the X11 server (and I applaud that goal - I sometimes feel like I am the only one caring about that). However, I don't see the reduced number of round trips. There is one round trip getting the result for the `InternAtom` requests and another round trip waiting for the result of `ConvertSelection`. You suggestions just moves a `CreateWindow` request from the second to the first round-trip.

Why does that make a difference? Or am I misunderstanding you?

(I think I applied fixes for the other comments. Thanks for those comments!)

Maan2003: Sorry I don't quite understand x11rb, I thought as a rule of thumb: the later `.reply()`, the better. Thanks for explaining :)

psychon: > I thought as a rule of thumb: the later .reply(), the better.

Well, yeah. That's the short version. Let me try to write down a long version:

When you send a request (e.g. `conn.intern_atom(...)`), the request is serialised and its bytes are put into an output buffer. The next request is also put into the output buffer etc. This buffer is actually sent to the X11 server in three cases:
- `conn.flush()` is called
- A request is sent but the buffer is full (for example because of lots of previous `intern_atom()` calls, but this can also happen with a big image in a `put_image()` call). In this case, all or parts of the new request can again end up in the buffer.
- You call `.reply()` for a request that is still in the output buffer.

Next part is `.reply()`: This function, well, waits for the reply to a request! There is again a buffer (actually a `VecDeque`, I think) of replies that were already read from the socket but not yet returned from API functions. When you call `.reply()`, it first checks this buffer. If the right reply is not found, the code waits for a new packet from the X11 server and then checks the queue again.

-------

With the above in mind, we can look at code like this:
```
let foo = conn.intern_atom("foo")?.reply()?;
let bar = conn.intern_atom("bar")?.reply()?;
```
The first line puts a request into the output buffer (`intern_atom()`), and then immediately flushes the output buffer and waits for a reply from the X11 server. This "I just sent a request and now wait for its reply" is what I call a round trip since when you measure the time it takes to execute the code, these round-trips are what take most of the time.

The second line does the same again and again takes some time.

With code like this instead:
```
let foo_cookie = conn.intern_atom("foo")?;
let bar_cookie = conn.intern_atom("bar")?;
let foo = foo_cookie.reply()?;
let bar = bar_cookie.reply()?;
```
This code will send both requests in `foo_cookie.reply()`, because both are already in the output buffer. It now has to wait for the reply, but the X11 server basically gets both requests "together".

In `bar_cookie.reply()`, the X11 connection will not yet actually have `read()` the reply from the X11 connection, since the reply for "foo" comes in first and the code stops reading from the connection once it got the right reply. Still, this will take less time.

----

As an extreme example, there is this code: https://github.com/psychon/x11rb/blob/5b9aafa515d2bc3b2c7fd8f1bfbd01cc5dff4431/examples/tutorial.rs#L203-L239

This code will first do lots of `conn.intern_atom().reply()` and print how long this took. Then it does the same thing again, but with first doing all the `intern_atom()` calls so that they all end up in the output buffer, and afterwards doing all the `reply()` calls. Thus, the second version only needs one round trip while the first one does 500 round trips.

The output of this code with `--release`:
```
bad use time:  5.812688ms
good use time: 970.326µs
ratio:         5.990448570892669
```
The output without `--release`:
```
bad use time:  16.929757ms
good use time: 3.958736ms
ratio:         4.276556203798384
```
So, yeah... avoiding round trips is a bit faster, but this example needs 500 round trips to make a measurable difference. However, I bet that if I used ssh X11 forwarding (´ssh -X`) over the internet, the round-trip-time to the X11 server would go up and the difference would be higher. Dunno if anyone still uses this forwarding and if one should care a lot about it when coding...

----

Now for the code in this PR: Actually... no, I think my original comment says enough about this. It's still the same number of round-trips ("send a request and wait for its reply"). I hope that this comment explains everything a bit better and makes it clearer what is going on.

PR #1806, opened 2021-05-24
Make book example code compile

JeftavanderHorst:
This example code doesn't work since it calls `build_ui` rather than passing it directly. Removing the parentheses fixes the problem. Note that this is my first interaction with druid, so my fix may not be idiomatic? Would love to hear of a better solution if that's the case.

### DIFF
--- a/docs/src/get_started.md
+++ b/docs/src/get_started.md
@@ -26,7 +26,7 @@ fn build_ui() -> impl Widget<()> {
 }
 
 fn main() -> Result<(), PlatformError> {
-    AppLauncher::with_window(WindowDesc::new(build_ui())).launch(())?;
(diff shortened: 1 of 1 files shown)


CryZe:
I believe both the book and the README are intended to talk about 0.7 rather than the master branch.

JeftavanderHorst:
I'm using druid 0.7.0, rustc 1.52.1, windows 10. Is this api changing in druid 0.8?

Maan2003:
Yes the [book is targeting the git version](https://linebender.org/druid/get_started.html#set-up-a-druid-project)

> I'm using druid 0.7.0, rustc 1.52.1, windows 10. Is this api changing in druid 0.8?

Yes See #1559

JeftavanderHorst:
Cool, my bad then. I'll close the pr

PR #1807, opened 2021-05-25
Window Drag Handle

blemelin:
Hello!

This PR proposes to add a `WindowDragHandle` widget that, when dragged around, moves the window with it. It wraps one child widget, so the handle can be anything (anything not clickable, that is).

# Proposed changes
- Add a `WindowDragHandle` widget to drag the window around (to move it). Works on *Windows* and *gtk* only.
- Add a `begin_move_drag` and a `end_move_drag` method to the `WindowHandle` struct on all platforms. Implemented only for *gtk* currently.
- Add an example named `custom_titlebar` that demonstrate how to use the `WindowDragHandle` widget.

# Questions to answer
1. This new widget was not first submitted through the [nursery](https://github.com/linebender/druid-widget-nursery). I thought it wasn't possible, since it requires new things in `druid-shell` (i.e `begin_move_drag` and a `end_move_drag`). Should I submit it through that first and cancel this PR ?
2. To make this work on *gtk*, I had to modify the `WindowHandle` struct, adding it a field (`last_mouse_button_press_event`). It stores the last mouse press event received, because I required other values from that "raw" event (like the timestamp or the root position of the cursor). That solution is not elegant, but it does not pollute the `MouseEvent` struct with *gtk only* related values. Is that an acceptable solution ?

# Other comments

This is my first contribution to an open-source project, so I hope I've not made any huge mistake. Please feel free to ask for modifications.

### DIFF
--- a/druid-shell/src/platform/gtk/window.rs
+++ b/druid-shell/src/platform/gtk/window.rs
@@ -996,7 +996,8 @@ impl WindowHandle {
     /// Tell the window manager to start a window drag using current mouse position.
     pub fn begin_move_drag(&self) {
         if let Some(state) = self.state.upgrade() {
-            match state.last_mouse_button_press_event.take() { //Consume the mouse button event.
+            match state.last_mouse_button_press_event.take() {
+                //Consume the mouse button event.
                 Some(event) if event.get_event_type() == gdk::EventType::ButtonPress => {
                     state.window.begin_move_drag(
                         event.get_button() as i32,
@@ -1005,7 +1006,7 @@ impl WindowHandle {
                         event.get_time(),
                     )
                 }
-                _ => warn!("Cannot begin window drag if no mouse button is currently down.")
+                _ => warn!("Cannot begin window drag if no mouse button is currently down."),
             }
         }
     }
--- a/druid-shell/src/platform/gtk/window.rs
+++ b/druid-shell/src/platform/gtk/window.rs
@@ -990,20 +990,10 @@ impl WindowHandle {
     }
 
     pub fn handle_titlebar(&self, _val: bool) {
-        warn!("WindowHandle::handle_titlebar is currently unimplemented for gtk.");
-    }
-
-    /// Close the window.
-    pub fn close(&self) {
-        if let Some(state) = self.state.upgrade() {
-            state.closing.set(true);
-            state.window.close();
-        }
+        warn!("WindowHandle::handle_titlebar is currently unimplemented for gtk. Use WindowHandle::begin_move_drag instead.");
     }
 
     /// Tell the window manager to start a window drag using current mouse position.
-    ///
-    /// On Gtk, this must be called immediately after a MouseEvent with a button press is received.
     pub fn begin_move_drag(&self) {
         if let Some(state) = self.state.upgrade() {
             match state.last_mouse_button_press_event.take() { //Consume the mouse button event.
@@ -1015,14 +1005,22 @@ impl WindowHandle {
                         event.get_time(),
                     )
                 }
-                _ => warn!("Cannot drag window if no mouse button is currently down.")
+                _ => warn!("Cannot begin window drag if no mouse button is currently down.")
             }
         }
     }
 
     /// Tell the window manager to end the current window drag, if any.
     pub fn end_move_drag(&self) {
-        //Nothing to do on gtk.
+        //Nothing to do on gtk. Window drag immediately ends when mouse button is released.
+    }
+
+    /// Close the window.
+    pub fn close(&self) {
+        if let Some(state) = self.state.upgrade() {
+            state.closing.set(true);
+            state.window.close();
+        }
     }
 
     /// Bring this window to the front of the window stack and give it focus.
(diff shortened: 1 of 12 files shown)


blemelin:
Seems like CI failed on the new example...because of an unused `main` function ?  Well.....adding `#[allow(unused)]` on `main` will fix it, but that's not how it should be done. Am I missing something ?

jneem:
There's some extra magic that's needed for wrapping examples in the `wasm` backend, so you'll need to add your example to `druid/examples/web/src/lib.rs`

blemelin:
CI should work now. I've tested the `cargo clippy` command on my side and it passed.

druid/src/widget/window_drag_handle.rs
```diff
@@ -0,0 +1,80 @@
...
+/// let title_label = WindowDragHandle::new(Label::new("My Application"));
+/// let exit_button = Button::new("X").on_click(|ctx, _data : &mut (), _env| ctx.submit_command(CLOSE_WINDOW));
+/// let title_bar = Flex::row().with_child(title_label).with_child(exit_button);
+/// ```
+///
+/// # Platform notes
+///
+/// ## Windows
+///
+/// Click events aren't passed through to the children.
```
jneem, line 47: They aren't? I don't see what's stopping them...

druid/src/widget/window_drag_handle.rs
```diff
@@ -0,0 +1,80 @@
...
+
+use tracing::instrument;
+
+use crate::widget::prelude::*;
+use crate::widget::{Controller, ControllerHost};
+
+/// A [`Controller`] widget that makes the target acts as a window grab handle.
+///
+/// On most platforms, click events aren't passed through to the children. As such, it is
+/// recommended to only put text or small icons in a window drag handle.
```
jneem, line 27: Regarding the cross-platform behavior, I think we want as many things as possible/reasonable to behave the same between platforms. It's fine to leave things stubbed out if it's too much effort to do it *for now* on other platforms, but I'd argue against most other plaform-specific stuff. (This is mainly a complaint about the docs, though, because AFAICT the implementation doesn't actually swallow the click events)

druid-shell/src/platform/gtk/window.rs
```diff
@@ -982,7 +990,30 @@ impl WindowHandle {
...
+                    )
+                }
+                _ => warn!("Cannot begin window drag if no mouse button is currently down."),
+            }
+        }
+    }
+
+    /// Tell the window manager to end the current window drag, if any.
+    pub fn end_move_drag(&self) {
+        //Nothing to do on gtk. Window drag immediately ends when mouse button is released.
```
jneem, line 1016: Is it ever valid to end the drag earlier? If not, why not just delete this? It does nothing on other platforms anyway...

jneem (changes requested):
Thanks for this, it looks good!

My main concern isn't really the fault of this PR (and I won't block on it), but I want to bring it up here anyway: I think `druid-shell` has the wrong API for custom title bars. It should be abstracting over the platform-specific behavior, but (and correct me if I'm wrong) I don't think there's any intention to ever implement `handle_titlebar` on GTK or `begin_move_drag` on windows. And I don't know anything about macOS, but [it seems like](https://stackoverflow.com/questions/54353576/colorizing-the-titlebar-in-macos-with-multiple-colors) they have a different mechanism altogether for customizing the titlebar.

Do you have any idea what an API could look that for allowing custom title bars on most platforms? Would it make sense to pass `druid-shell` a `Rect` and say "treat this as a title bar as far as dragging the window around is concerned"?

jneem:
Oh, I forgot to say: as far as this PR is concerned, could you add a warning to the API docs for `WindowHandle::begin_move_drag` and `WIndowHandle::handle_titlebar` saying essentially that these will never be cross-platform and will probably go away?

blemelin:
> It should be abstracting over the platform-specific behavior

I agree.

> Would it make sense to pass `druid-shell` a `Rect` and say "treat this as a title bar as far as dragging the window around is concerned"?

This might be the best solution indeed. I pretty sure I can implement this for *Windows* and *gtk*, but I won't be able to do anything for *MacOs* and *x11*. Is that an issue ?

*Side note here : I'm lacking time to tackle this right now (since I only need this for a personal project), so this might take some time. *

druid-shell/src/platform/gtk/window.rs
blemelin: I wanted to make sure it had a reverse operation (like `open` and `close`, `start` and `stop`, `begin` and `end`), but in retrospect, that was a bad idea.

druid/src/widget/window_drag_handle.rs
blemelin: I think the [WM_NCHITTEST](https://docs.microsoft.com/en-us/windows/win32/inputdev/wm-nchittest) is eating them, but don't quote me on that.

jneem:
> This might be the best solution indeed (as I hinted above). I pretty sure I can implement this for Windows and gtk, but I won't be able to do anything for MacOs and x11. Is that an issue ?

That's definitely fine. It would be nice if a macOS user could give input on whether it's *feasible* to do for macOS in the future, but I won't insist on that either...

jpochyla:
#1701 adds support for the Mac-specific part of this, and the feedback was that we should do this with extention trait.

PR #1809, opened 2021-05-25
Allow gtk window to resize when titlebar is hidden 

blemelin:
Hello again!

This PR proposes to add a [GtkHeaderBar](https://developer.gnome.org/gtk3/stable/GtkHeaderBar.html) *gtk* widget to the *gtk* window on the *gtk* platform, acting as the window titlebar. Doing so allows the window to be resized even if the titlebar is hidden. Functionality and appearance of the titlebar seems unaffected (i.e it looks and acts the same as before).

# Proposed changes
- Add a [GtkHeaderBar](https://developer.gnome.org/gtk3/stable/GtkHeaderBar.html) *gtk* widget to the *gtk* window on the *gtk* platform.
- On the *gtk* platform, windows will always be decorated. Adding that `GtkHeaderBar` to the window actually replace the default one. Hiding that *gtk* widget hide the titlebar.

# Questions to answer
1. Like said before, on the *gtk* platform, windows will **always** be decorated. If one day someone want a true non-decorated window, he wont be able to. I do think this is a non-issue, since the non-decorated behaviour can be replicated by hiding the titlebar and making the window not resizable, but that is something to consider.

# Other comments

Sorry for that double PR. I'm currently working on a project that require these features, and I thought that just forking and not contributing back would be a waste. That would be the last one though.

By the way, an other solution to this problem was to put a dummy titlebar when the user requested no titlebar with the `show_titlebar` method. Removing that dummy restored the default titlebar. Unfortunately, this is [highly discouraged by the developers](https://developer.gnome.org/gtk3/stable/GtkWindow.html#gtk-window-set-titlebar), and even triggered a warning. I think this is the best solution, but I could be wrong.


jneem:
Thanks for the PR! So we definitely want to support undecorated windows, for example for things like tooltips and dropdown lists. Maybe that means we need to have an extra property to set? The docs currently say that `show_titlebar` does both the titlebar and window decorations, but we could have a separate `show_titlebar` and `show_decorations`...

blemelin:
Oh you're right. I didn't see that the docs on `WindowConfig` mention decorations (both `WindowHandle` and `WindowDesc` don't).

I should be able to implement what you propose on both *gtk* and *Windows* platforms. When undecorated, a window won't have a titlebar or resize controls, even if `show_titlebar` or `resizable` is `true`. I'll make sure to document that behaviour on each method, so there won't be any surprise for developers. While I'm at it, I'll remove the mention about decorations from the docs of `show_titlebar` on `WindowConfig`, because we'll be adding a `show_decoration` method.

The `show_titlebar` method is not implemented for *x11*, *mac* (only partially), and *web*, so I won't do anything for theses platforms (except a warning). As a hint for the *mac* implementation, I'll add a comment with useful information I've looked up while researching this. I would have been happy to do it, but I don't have a mac to test it.

druid-shell/src/platform/gtk/window.rs
```diff
@@ -854,7 +863,13 @@ impl WindowHandle {
...
     pub fn show_titlebar(&self, show_titlebar: bool) {
         if let Some(state) = self.state.upgrade() {
-            state.window.set_decorated(show_titlebar)
+            state.titlebar.set_visible(show_titlebar);
+        }
+    }
+
+    pub fn set_title(&self, title: impl Into<String>) {
+        if let Some(state) = self.state.upgrade() {
+            state.titlebar.set_title(Some(&(title.into())));
```
blemelin, line 872: Set the title on the window as well.
Maybe setting it on the window sets it on the HeaderBar widget as well ? I'll have to test this.

blemelin: Just to be sure, I decided to set it on both.

blemelin:
Okay, so I've added a `show_decorations` method that enable/disable decorations on *Windows* and *gtk*. From my own testing, this seems to work perfectly fine on both platforms. I've added a `window_decorations` example with a basic application for you to try it out.

Honestly, this is a lot more changes that I had previously anticipated. Since this is a very critical part of this library, I'll review the code one more time myself and then submit it for a final review.

Thanks for your comments.

druid-shell/src/platform/mac/window.rs
```diff
@@ -194,6 +194,15 @@ impl WindowBuilder {
         self.min_size = Some(size);
     }
 
+    pub fn show_decorations(&mut self, _show_decorations: bool) {
+        //Note for future implementor.
+        //
+        //There is a "NSBorderlessWindowMask" flag on the "NSWindowStyleMask" struct that
+        //you could use to implement this feature. Unfortunately, this has been deprecated
+        //by Apple (see https://developer.apple.com/documentation/appkit/nswindowstylemask).
+        warn!("WindowHandle::show_decorations is currently unimplemented for Mac.");
```
blemelin: Wrong warning message : `WindowBuilder`, not `WindowHandle`.

blemelin: Done.

blemelin:
Ready for review. Thanks you and have a nice day.

blemelin:
I'm closing this because it need more work.

PR #1816, opened 2021-06-01
Make test code available in release builds

PoignardAzur:
Having instrumentation code (eg DebugRequestState) be `#[test]`
makes it impossible to use from parent crates.
This is a problem if you want to do unit tests in something like
panoramix or crochet.

### DIFF
--- a/druid/src/core.rs
+++ b/druid/src/core.rs
@@ -73,7 +73,7 @@ pub struct WidgetPod<T, W> {
 /// [`paint`]: trait.Widget.html#tymethod.paint
 /// [`WidgetPod`]: struct.WidgetPod.html
 #[derive(Clone)]
-pub(crate) struct WidgetState {
+pub struct WidgetState {
     pub(crate) id: WidgetId,
     /// The size of the child; this is the value returned by the child's layout
     /// method.
@@ -975,7 +975,6 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
                     self.state.needs_window_origin = false;
                     true
                 }
-                #[cfg(test)]
                 InternalLifeCycle::DebugRequestState { widget, state_cell } => {
                     if *widget == self.id() {
                         state_cell.set(self.state.clone());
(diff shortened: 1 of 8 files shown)


druid/src/tests/helpers.rs
```diff
@@ -303,42 +312,12 @@ impl<T: Data, W: Widget<T>> Widget<T> for Recorder<W> {
...
-
-pub fn widget_id4() -> (WidgetId, WidgetId, WidgetId, WidgetId) {
-    (
-        WidgetId::next(),
-        WidgetId::next(),
-        WidgetId::next(),
-        WidgetId::next(),
-    )
-}
+pub fn widget_ids<const N: usize>() -> [WidgetId; N] {
```
cmyr, line 315: 🎉

druid/src/core.rs
```diff
@@ -975,7 +975,6 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
                     self.state.needs_window_origin = false;
                     true
                 }
-                #[cfg(test)]
```
cmyr, line 978: did you consider putting this behind a feature? e.g. this would be `#[cfg(any(test, feature="test-harness"))]`? This might cause minor problems downstream if anyone is trying to exhaustively handle events anywhere.

cmyr (commented):
Looks good, only question is about whether a feature makes sense?

druid/src/core.rs
PoignardAzur: Yeah, I did consider it, and yes, the added complexity and risk for breakage is a concern. I've ranted about cfg not being great before.

I guess we could simply mark the `InternalLifeCycle` enum as `#[non_exhaustive]`.

I'm very on the fence on whether adding a feature gate is worth the hassle (linkers will remove most unused code anyway), but I'm willing to add it as long as I don't have to update the CI and stuff myself.

cmyr (approved):
Okay, I think given that these are part of `InternalLifeCycle` I won't feel particularly guilty about any breakage. thanks!

Issue #1814, opened 2021-05-31
gtk: Unsound call to std::str::from_utf8_unchecked


psychon:
This line calls `std::str::from_utf8_unchecked`: https://github.com/linebender/druid/blob/057f9f9bbf529c6fd60f00afaa8494fa3a7a4300/druid-shell/src/platform/gtk/clipboard.rs#L68

The given data is directly provided by the library user, thus making it trivial to violate the [safety contract](https://doc.rust-lang.org/std/str/fn.from_utf8_unchecked.html#safety) of this function.

jneem:
Fixed by #1820

PR #1820, opened 2021-06-06
Improve unsafe code in GTK clipboard code

psychon:
gtk: Replace unsound std::str::from_utf8_unchecked
    
    When putting text data into the clipboard, the platform code was calling
    std::str::from_utf8_unchecked() on the data. That's not really valid,
    because the caller can provide arbitrary bytes. Thus, this commit
    replaces that with a call to std::str::from_utf8().
    
    In case invalid utf8 was provided, the selection is instead set as raw
    bytes. I am not sure what will happen "down the road" with this, but it
    seems to be the best that can be done.
    
    Fixes: https://github.com/linebender/druid/issues/1814

I also reported https://github.com/gtk-rs/gtk3-rs/issues/564 to make `gdk::Atom::value()` safe (or at least to figure out why it is unsafe).

### DIFF
--- a/druid-shell/src/platform/gtk/clipboard.rs
+++ b/druid-shell/src/platform/gtk/clipboard.rs
@@ -130,7 +130,8 @@ impl Clipboard {
         let targets = clipboard.wait_for_targets().unwrap_or_default();
         targets
             .iter()
-            .map(|atom| unsafe { format!("{} ({})", atom.name(), atom.value()) })
+            // SAFETY: Atom::value() is 'self.0 as usize'. No idea why that is unsafe.
+            .map(|atom| format!("{} ({})", atom.name(), unsafe { atom.value() }))
             .collect()
     }
 }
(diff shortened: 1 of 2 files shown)


druid-shell/src/platform/gtk/clipboard.rs
```diff
@@ -64,9 +64,17 @@ impl Clipboard {
             let idx = idx as usize;
             if idx < formats.len() {
                 let item = &formats[idx];
+                let mut not_set = true;
                 if item.identifier == ClipboardFormat::TEXT {
```
jneem, line 68: I think you can do
```rust
if let (ClipboardFormat::TEXT, Ok(data)) = (item.identifier, std::str::from_utf(&item.data)) {
    sel.set_text(data);
} else {
    //...
}
```
and then you don't need the `not_set` logic.

psychon: Good idea, done.

Note however that this might always call `from_utf8`, even with stuff other than `ClipboardFormat::TEXT`.

jneem (approved):

PR #1822, opened 2021-06-07
Add example to ViewSwitcher

arifd:
Let me know if you don't like it, I should modify anything? I'm hoping to add examples to many more areas of Druid that could benefit from them.


jneem:
Did you reach a conclusion about this? FWIW, I don't think it's necessary to have a `main`, because if you repeat that for every example, it's a lot of boilerplate...

arifd:
There has been no feedback to date. And so it's kind of stuck.

jneem:
Well in that case I suggest we just merge it. Having an example is clearly better than not having one, so there's no point in being stuck on it. Any objections?

jneem (approved):

Maan2003:
the example should be in top level docs of `ViewSwitcher` not `ViewSwitcher::new`?

arifd:
Good question. I wanted it in `new()` because that's when I need to be
reminded of the order of the parameters in the closures.

Maan2003:
Merging, as no objections

PR #1823, opened 2021-06-12
New clippy lints

Maan2003:


jneem (approved):
That seems like a pretty useful new lint!

PR #1825, opened 2021-06-13
Add public `has_focus` to `WidgetPod`

ForLoveOfCats:
Very tiny change to allow a widget's parent to check if it has focus (as in it or one of its descendants is focused). I ran up against the lack of this and it makes sense given the active/hot methods.


Maan2003 (approved):
:+1:

PR #1826, opened 2021-06-16
Fix error in README.md sample code

nilsmartel:
Fix faulty code.


Maan2003:
This code works with master. From readme:
> This example works with master. Here's an example for 0.7.0: https://docs.rs/druid/latest/druid/#examples

also see #1748 , #1750

nilsmartel:
I See, thank you for double checking and have a nice day!

PR #1829, opened 2021-06-21
Clippy

Maan2003:


djeedai (approved):

druid-shell/src/platform/gtk/window.rs
```diff
@@ -1209,23 +1209,19 @@ impl IdleHandle {
     {
         let mut queue = self.idle_queue.lock().unwrap();
         if let Some(state) = self.state.upgrade() {
+            queue.push(IdleKind::Callback(Box::new(callback)));
```
jneem, line 1212: This won't work, because after you push, the queue won't be empty...

druid-shell/src/platform/gtk/window.rs
```diff
@@ -1209,23 +1209,19 @@ impl IdleHandle {
...
-            } else {
-                queue.push(IdleKind::Callback(Box::new(callback)));
             }
         }
     }
 
     pub fn add_idle_token(&self, token: IdleToken) {
         let mut queue = self.idle_queue.lock().unwrap();
         if let Some(state) = self.state.upgrade() {
+            queue.push(IdleKind::Token(token));
```
jneem, line 1222: Here too

jneem (changes requested):

druid-shell/src/platform/gtk/window.rs
Maan2003: :man_facepalming: `clippy --fix` failed me. Good catch  :)

cmyr (approved):
Thanks!

PR #1832, opened 2021-06-21
Avoid undefined behaviour in GTK animation frames

JAicewizard:
Requestion an animation frame is and always has been undefined behaviour. 
This is not documented ANYWHERE besides the brains of whoever coded this.
See https://gitlab.gnome.org/GNOME/gtk/-/issues/3703.

### DIFF
--- a/druid-shell/src/platform/gtk/window.rs
+++ b/druid-shell/src/platform/gtk/window.rs
@@ -382,7 +382,7 @@ impl WindowBuilder {
                                 state.request_animation.set(false);
                                 let vbox = state.window.get_children().first().unwrap().clone()
                                     .downcast::<gtk::Box>()
-                                    .unwrap().clone();
+                                    .unwrap();
                                 let first_child = vbox.get_children().last().unwrap().clone();
                                 if first_child.is::<gtk::DrawingArea>() {
                                     first_child.queue_draw();
--- a/druid-shell/src/platform/gtk/window.rs
+++ b/druid-shell/src/platform/gtk/window.rs
@@ -382,8 +382,8 @@ impl WindowBuilder {
                                 state.request_animation.set(false);
                                 let vbox = state.window.get_children().first().unwrap()
                                     .downcast::<gtk::Box>()
-                                    .unwrap();
-                                let first_child = &vbox.get_children().last().unwrap();
+                                    .unwrap().clone();
+                                let first_child = vbox.get_children().last().unwrap().clone();
                                 if first_child.is::<gtk::DrawingArea>() {
                                     first_child.queue_draw();
                                 }
@@ -1212,7 +1212,7 @@ impl WindowHandle {
                 .downcast::<gtk::Box>()
                 .unwrap();
 
-            let first_child = vbox.get_children().first().unwrap();
+            let first_child = vbox.get_children().first().unwrap().clone();
             if let Some(old_menubar) = first_child.downcast_ref::<gtk::MenuBar>() {
                 old_menubar.deactivate();
                 vbox.remove(old_menubar);
--- a/druid-shell/src/platform/gtk/window.rs
+++ b/druid-shell/src/platform/gtk/window.rs
@@ -380,7 +380,7 @@ impl WindowBuilder {
                             state.in_draw.set(false);
                             if state.request_animation.get() {
                                 state.request_animation.set(false);
-                                let vbox = state.window.get_children().first().unwrap()
+                                let vbox = state.window.get_children().first().unwrap().clone()
                                     .downcast::<gtk::Box>()
                                     .unwrap().clone();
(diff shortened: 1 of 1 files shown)


JAicewizard:
Before merging, someone else please confirm that this code works. I currently cannot compile druid (AGAIN!!!! FFS rust how is this even possible. How can it never just WORK).

This is copied straight out of my gtk4 code which did work.
To test you can try running the anim example, and also see if the text widget updates properly (both without moving your mouse, as that causes a new frame anyways).

Maan2003:
Both examples are working fine.

> Requesting an animation frame is and always has been undefined behaviour.
> This is not documented ANYWHERE

sounds horrible :grimacing: 

I believe this only affects druid-shell because druid doesn't let you request animation frame when painting (at type level!)

CI is red due to new clippy lints (in 1.53)

> I currently cannot compile druid (AGAIN!!!! FFS rust how is this even possible. How can it never just WORK).

I would love to know why. if you are bellow rustc 1.51 then it can solved by commenting out the [const generic data impl for array](https://github.com/linebender/druid/blob/e862b42aa1c17bb014cfc754c3427623b5fcf0b6/druid/src/data.rs#L614)

JAicewizard:
> I believe this only affects druid-shell because druid doesn't let you request animation frame when painting (at type level!)

No!!! because we base the whole cycle druid lifecycle on GTK draw events, including update. This is the only real option, since there isnt many other events this could be put into.

> I would love to know why. if you are bellow rustc 1.51 then it can solved by commenting out the const generic data impl for array

no I tried all channels. I didnt have much time today, I will try to reinstall rust all together. That didnt fix things last time, but it might now :shrug:

Maan2003:
rebasing should solve the CI issues now

JAicewizard:
CI seems fixed :+1:

druid-shell/src/platform/gtk/window.rs
```diff
@@ -1165,10 +1212,10 @@ impl WindowHandle {
                 .downcast::<gtk::Box>()
                 .unwrap();
 
-            let first_child = &vbox.get_children()[0];
+            let first_child = vbox.get_children().first().unwrap().clone();
```
Maan2003, line 1215: I don't understand this change. can you please explain more

JAicewizard: uhh, oh it was unrelated. I can remove it, it just seemed more explicit about the panic.

This is part of a path to move some things from my GTK4 fork into GTK3. In GTK4 you done have a general `get_children` only `first/last_child`. So when copy-pasting to the GTK3 branch this became `vbox.get_children().first()` etc. 

Its just an artifact from copy pasting, nothing meaningful. I can undo all of these if you want.

Maan2003: Given this pr is related *Undefined behavior*, I would keep this separate

druid-shell/src/platform/gtk/window.rs
```diff
@@ -780,7 +812,22 @@ impl WindowState {
...
+        } else {
+            let vbox = &self
+                .window
+                .get_children()
+                .first()
+                .unwrap()
+                .clone()
+                .downcast::<gtk::Box>()
+                .unwrap();
+            let first_child = vbox.get_children().last().unwrap().clone();
```
Maan2003, line 826: Is this also due to GTK4 -> GTK3, or required?

JAicewizard: well this is all new code. I do now see that this is way to complicated, we store the drawingarea in the state so we can just use that. So I assume that this comment is no longer relevant

JAicewizard:
welp, as github seems to have broken itself, CI fails. I tested the new code and works as before.

druid-shell/src/platform/gtk/window.rs
```diff
@@ -361,6 +366,33 @@ impl WindowBuilder {
...
+                        if let Some(state) = handle.state.upgrade() {
+                            state.in_draw.set(false);
+                            if state.request_animation.get() {
+                                state.request_animation.set(false);
+                                let vbox = state.window.get_children().first().unwrap().clone()
+                                    .downcast::<gtk::Box>()
+                                    .unwrap();
+                                let first_child = vbox.get_children().last().unwrap().clone();
+                                if first_child.is::<gtk::DrawingArea>() {
+                                    first_child.queue_draw();
```
Maan2003, line 388: I think we can also use state.drawing_area here? Sorry for missing this earlier

JAicewizard: I also looked over everything to see if it could be simplified more, and also missed it. NP just pushed this too.

Maan2003 (approved):
Thanks! Both animation and textbox examples are working

PR #1836, opened 2021-06-28
Keep debug colors as a static slice

cmyr:
I don't think there's any advantage to storing these in the env, and it
potentially makes working with the Env more expensive.

Motivated by revisiting #1658 while catching up on reviews


Maan2003 (approved):
Looks good!

PR #1837, opened 2021-06-28
Remove Env::default

cmyr:
This was a bit of a footgun; it would always configure localization
resources, which doesn't make sense if we want to use an `Env` as just a
bag of key/values in some contexts.

Specifically, we may want to do this for #1658.

The fix is a two-parter:

- the `LocalizationManager` owned by the env is now optional, although
  it is expected to exist for the root env. The framework makes sure this
  is the case.
- There is a new `Env::empty` method that creates a new empty
  environment, with no localization resources. This is suitable for the
  'specifying overrides' case.

@arthmis if you're still interested in finishing up #1658, then I would merge this as is, but if that is in doubt I would probably remove the `empty()` method since it isn't really useful anywhere else I can think of?

### DIFF
--- a/druid/src/theme.rs
+++ b/druid/src/theme.rs
@@ -183,8 +183,3 @@ pub(crate) fn add_to_env(env: Env) -> Env {
                 .with_size(15.0),
         )
 }
-
-#[deprecated(since = "0.7.0", note = "use Env::default() instead")]
-pub fn init() -> Env {
-    Env::default()
-}
--- a/druid/src/app.rs
+++ b/druid/src/app.rs
@@ -242,7 +242,7 @@ impl<T: Data> AppLauncher<T> {
         let mut env = self
             .l10n_resources
             .map(|it| Env::with_i10n(it.0, &it.1))
-            .unwrap_or_default();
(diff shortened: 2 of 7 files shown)


arthmis:
Yes this sounds good. Give me a few days to get back into this.

druid/src/theme.rs
```diff
@@ -186,5 +186,5 @@ pub(crate) fn add_to_env(env: Env) -> Env {
 
 #[deprecated(since = "0.7.0", note = "use Env::default() instead")]
```
pleshevskiy, line 187: I think we should also change the note to following
```
#[deprecated(since = "0.7.0", note = "use Env::with_default_i10n() instead")]
```

or `Env::empty()`. You know best :)

Maan2003: Can we just remove this? I don't think this is a popular function :thinking:

PoignardAzur (approved):
LGTM. Merging in a few days unless someone objects.

druid/src/theme.rs
cmyr: agree, there's no good way to use this right now.

PR #1839, opened 2021-06-29
Hacky support for protected text fields

cmyr:
This adds TextField::protected, which creates a TextField that obscures
its content, suitable for things like passwords.

This is an expedient (rather than principled) implementation. In
particular, cursor positions will get weird if you enter glyphs that are
not present in the system monospace font.

This has come up enough and involves tricky enough parts of the codebase that I thought it would be easiest to just dig in and see what I could do. This implementation isn't great, but I think it might be *good enough*.

Doing this 'properly' will be a more invasive change; I haven't thought it through too much, but there are a number of problems: for instance, in order to correctly display selections we will need to have a mapping from the selection state of the actual text to the selection state of the replacement text (because we show one replacement glyph per grapheme cluster, the selections in the underlying text are not necessarily valid in the replacement text.)

I'll only merge this if it seems genuinely useful to someone, so try it out and let me know?

### DIFF
--- a/druid/examples/textbox.rs
+++ b/druid/examples/textbox.rs
@@ -21,8 +21,8 @@ use std::sync::Arc;
 
 use druid::widget::{Flex, Label, TextBox};
 use druid::{
-    AppLauncher, Color, Data, Env, Lens, LocalizedString, Menu, Widget, WidgetExt, WindowDesc,
-    WindowId,
+    AppLauncher, Color, Data, Env, FontDescriptor, FontFamily, Lens, LocalizedString, Menu, Widget,
+    WidgetExt, WindowDesc, WindowId,
 };
 
 const WINDOW_TITLE: LocalizedString<AppState> = LocalizedString::new("Text Options");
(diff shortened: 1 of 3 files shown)


jplatte:
Not sure whether anybody cares, but here's a version rebased on the latest `master`: https://github.com/jplatte/druid/tree/hacky-password-support

cmyr:
I'm going to close this since it didn't seem particularly interesting to anybody.

jplatte:
Well I care.. I wouldn't have rebased otherwise 😅

ngugcx:
I care a lot. When I realized that the Druid did not have a protected text box, I was shocked.

cloudybyte:
Same thing here.
I know druid is discontinued, but having a GUI framework, which does not support protected input is kinda bad.
I guess we're gonna have to wait for [Xilem](https://github.com/linebender/xilem) to support this, as putting the time into a discontinued codebase is frankly not worth it.

PR #1842, opened 2021-06-30
Fix parse error in formatter example

### LINKED ISSUE #1838: value formatter panic when entering `24,-` as euro amount.
This is customary in dutch, and although it might not be needed for an example to handle all the local variations, it might be good to not panic on unreachable. Rather mark it as invalid or something along those lines.

cmyr:
- closes #1838

### DIFF
--- a/druid/examples/value_formatting/src/formatters.rs
+++ b/druid/examples/value_formatting/src/formatters.rs
@@ -165,6 +165,7 @@ impl Formatter<f64> for NaiveCurrencyFormatter {
                 Validation::failure(CurrencyValidationError::TooManyCharsAfterDecimal)
             }
             (Some(c), None, _) if c.is_ascii_digit() => Validation::success(),
+            (Some(c), None, _) => Validation::failure(CurrencyValidationError::InvalidChar(c)),
             (None, None, _) => Validation::success(),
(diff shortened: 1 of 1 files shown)


druid/examples/value_formatting/src/formatters.rs
```diff
@@ -173,7 +174,7 @@ impl Formatter<f64> for NaiveCurrencyFormatter {
                 let bad_char = if c1.is_ascii_digit() { other } else { c1 };
                 Validation::failure(CurrencyValidationError::InvalidChar(bad_char))
             }
-            _ => unreachable!(),
+            other => panic!("unexpected: {:?}", other),
```
JAicewizard, line 177: Why `other` vs `_`? Just curious.

cmyr: Just so that I had a simple binding I could print to see what the failure was. Probably don't need to commit, but seems more useful anyway..

JAicewizard: ohh haha I see now, I didnt notice you use it too.

JAicewizard (approved):
Looks good.
Not sure if I'm allowed to approve, but I'm doing it anyways!

Maan2003 (approved):
Looks good to me!

PR #1843, opened 2021-06-30
LinuxApplicationExt: experiment with platforms specific extensions

Maan2003:


druid-shell/src/application.rs
```diff
@@ -209,3 +209,20 @@ impl Application {
         platform::Application::get_locale()
     }
 }
+
+#[cfg(any(doc, target_os = "linux"))]
+pub trait LinuxApplicationExt {
```
cmyr, line 214: Let's document this.

druid-shell/src/application.rs
```diff
@@ -209,3 +209,20 @@ impl Application {
...
+    /// Returns a handle to the primary system clipboard.
+    /// Useful for middle mouse paste.
```
cmyr, line 216:
```suggestion
    /// Returns a handle to the primary system clipboard.
    ///
    /// This is useful for middle mouse paste.
```

druid-shell/src/lib.rs
```diff
@@ -81,3 +81,6 @@ pub use window::{
 };
 
 pub use keyboard_types;
+
+#[cfg(target_os = "linux")]
+pub use application::LinuxApplicationExt;
```
cmyr, line 86: how we should expose these is a good question.

I like the pattern in `std` of having the `std::os` module, which contains `std::os::unix`, `std::os:windows`, etcetera; I could imagine having something like that?

hmhm:

If we made the `platform` module public we could put these in the particular root modules of the specific platforms, but that might get a bit messy.

Maybe what I would prefer instead would be that we have an `ext` module, and then `ext::linux`? Another option would be to rename `platform` to `backend` or something, and then we could use `platform` for this, which feels like a better name...

cmyr (commented):
Cool, I think this makes sense. I've left some comments inline, but to me it's really just a matter of sorting out the details.

druid-shell/src/lib.rs
Maan2003: I like `platform::LinuxApplicationExt` or `platform::linux::LinuxApplicationExt`. renaming existing `platform` to `backend` sound reasonable.

I would prefer shorter use statement:
`use druid::shell::platform::LinuxApplicationExt;`
over
`use druid::shell::platform::linux::LinuxApplicationExt;`

cmyr: we could also have `druid::platform::linux` that reexports things from shell?

Now should we generally distinguish between `platform::gtk` and `platform::x11`?

Maan2003: I guess `platform::linux::{GtkApplicationExt, X11ApplicationExt}`. X11 and Wayland should be different though

Maan2003:
these commits are best understandable separately

* Move druid-shell/src/platform to druid-shell/src/backend c46c0e3
  renames platform to backend

* platform -> backend in code 765016f
  fixes all the build errors and changes internal comments

* public docs and error messages 976bc93
  changes the docs, error messages and README

Maan2003:
Removed x11 impl for now because clipboard in changing #1851

cmyr:
@Maan2003 can we split this up, and do the 'platform -> backend' change as a single separate PR?

Maan2003:
Sure :+1:

Maan2003:
rebased

druid-shell/src/platform/linux.rs
```diff
@@ -0,0 +1,38 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Linux specific extensions.
+use crate::Clipboard;
+
+/// Linux specific extensions to [`Application`]
+///
+/// [`Application`]: crate::Application
+pub trait LinuxApplicationExt {
```
cmyr, line 21: Is this  'linux' thing, or only  a gtk thing?

cmyr (approved):
Cool, I've been wanting to see something like this for ages, and this feels like a very nice little test case to try it out with.

One question inline, otherwise I'm happy to get this merged!

druid-shell/src/platform/linux.rs
Maan2003: this is a Linux thing

psychon: Is this a Linux thing? The primary selection (I wouldn't really call it "clipboard"...) is AFAIK an X11 thing and not implemented on Wayland. But my knowledge here might be out of date and I am too lazy to dig into Gtk's Wayland backend right now, so I am just leaving this as a comment here...

Maan2003: :man_facepalming: I really forgot wayland's existence.

Maan2003: there is a [unstable wayland protocol](https://github.com/wayland-project/wayland-protocols/blob/36cee4bdbcaf3a75f567315809953b5e21d703da/unstable/primary-selection/primary-selection-unstable-v1.xml) and is supported by gtk, qt and by kde plasma, gnome and sway, so :man_shrugging:

cmyr:
@Maan2003: I just realized something about this, and naming: basically I think this should just be called `ApplicationExt`, and the fact that it is for linux is determined by it's presence in the `platform::linux` module. This mimics what `std` does, where there are, say, four different `MetadataExt` traits, in four different submodules of `std::os`.

Maan2003:
makes sense, will fix this soon

Issue #1835, opened 2021-06-28
NSMenu missing init call


cmyr:
We currently only call `NSMenu::alloc()` but i think we also need to call `init(title: String)`, which is [NSMenu's only constructor](https://developer.apple.com/documentation/appkit/nsmenu).
Cacao calls [msg_send![alloc, initWithTitle:&*title]](https://github.com/ryanmcgrath/cacao/blob/trunk/src/macos/menu/menu.rs).
Not sure how to solve this since `Menu::new()`does not have a title.

_Originally posted by @kud1ing in https://github.com/linebender/druid/issues/1307#issuecomment-832937084_

Regardless of anything else discussed, we *should* be calling init, and not just alloc.

kud1ing:
Can be closed.

cmyr:
Thanks!

kud1ing:
Nah. Thank @terhechte and you!

PR #1846, opened 2021-07-01
Add nsmenu title

terhechte:
This is an initial fix for #1835.

It uses `::alloc().initWithTitle(...)` to initialize `NSMenu` items with a correct title.
However, it is not clear to me yet what the behaviour would be for the `Menu` that is exposed from the `druid` crate.
Right now I'm calling `Menu("")` but I suppose that's wrong. My guess would be that the public druid API would also expose a title property? But then it would affect all platforms. Open to suggestions here :)


cmyr:
I think that initing with an empty string and then modifying it afterwards to the correct title should be okay?

terhechte:
Updated it accordingly to use an empty string and added `set_title` which only works on macOS and is a NOP on the other platforms

druid-shell/examples/shello.rs
```diff
@@ -123,7 +123,8 @@ impl WinHandler for HelloState {
 
 fn main() {
     tracing_subscriber::fmt().init();
-    let mut file_menu = Menu::new("File Menu");
+    let mut file_menu = Menu::new();
+    file_menu.set_title("File Menu");
```
kud1ing, line 127: maybe just "File"?

cmyr:
Okay so I'm looking at this again, and I want to double-check my understanding:

- druid-shell's `Menu::new` method is only called for root menus; that means the root application menu, as well as root popup menus.
- all other items (such as the file menu) are created with the `add_item` and `add_dropdown`, which do correctly set titles

And in both the case of. the root menu and of a context menu, the title is (I believe) unused?

Ultimately I'm wondering if we need to bother with the whole `set_title` addition. My main concern with #1835 isn't with the title not being set, it's with a missing `init` call that may be performing other important work. If we can just init with an empty string and leave it at that, I'd be happy for the simplification..

terhechte:
@cmyr I agree. I've removed the `set_title` implementations. Just using an empty string should be enough.

cmyr (approved):
Thanks!

PR #1847, opened 2021-07-02
Add file format popup mac

### LINKED ISSUE #998: macOS: Support choosing file formats in the file dialog.
When a user is saving a file they should be able to choose from among the supported file formats.
The Windows and GTK implementations already support this and other apps on macOS support this.

file-formats

According to Apple's [Human Interface Guidelines](https://developer.apple.com/design/human-interface-guidelines/macos/user-interaction/file-handling/) this should be possible even without a custom extension.

> **Let the user choose an output format if your app can save in multiple file formats.** Output formats, when present, are displayed via a pop-up button beneath the output location.

It's unclear how these *output formats* can be specified. [`NSSavePanel`](https://developer.apple.com/documentation/appkit/nssavepanel) only seems to have [`allowedFileTypes`](https://developer.apple.com/documentation/appkit/nssavepanel/1534419-allowedfiletypes) which we are using already. However the way we are using them is just an array of strings, so we couldn't even specify the format name like *Rich Text Document*.

The `allowedFileTypes` also seems to support [UTIs](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/understanding_utis/understand_utis_intro/understand_utis_intro.html). Would we get the format selection if we defined UTIs? Are there other ways to get the format selection?

terhechte:
This fixes #998 by adding a file format popup. A couple of caveats (as also explained in my comment [here](https://github.com/linebender/druid/issues/998#issuecomment-872871430):

- MacOS usually shows these for `NSDocument` based apps. In order to have a similar experience, we have to build our own view hierarchy. This might not be pixel perfect similar to, say, how TextEdit does it.
- The implementation does not return the selected save path and the selected file format. Instead, the save path will be mutated based on the selected file format. So if the user enters "example.pdf" as the path and set set file format to "Jpg" the result will be that the path returned from calling the save panel is "example.jpg". 
- The string `File Format:` is currently hard coded into the implementation. This is fine but will break with localization. There's no way to retrieve the operating system resource that is used in apps like TextEdit. 

This is what it looks like:
Screen Shot 2021-07-02 at 11 41 21

### DIFF
--- a/druid-shell/src/dialog.rs
+++ b/druid-shell/src/dialog.rs
@@ -21,7 +21,16 @@ use std::path::{Path, PathBuf};
 /// This path might point to a file or a directory.
 #[derive(Debug, Clone)]
 pub struct FileInfo {
+    /// The path to the selected file.
+    /// On macOS, this is already rewritten to use the extension that the user selected
+    /// with the `file format` property.
     pub(crate) path: PathBuf,
+    /// The selected file format. If there're multiple different formats available
+    /// this allows understanding the kind of format that the user expects the file
+    /// to be written in. Examples could be Blender 2.4 vs Blender 2.6 vs Blender 2.8.
+    /// The `path` above will already contain the appropriate extension chosen in the
+    /// `format` property, so it is not necessary to mutate `path` any further.
+    pub(crate) format: Option<FileSpec>,
 }
 
 /// Type of file dialog.
--- a/druid-shell/src/dialog.rs
+++ b/druid-shell/src/dialog.rs
@@ -25,15 +25,15 @@ pub struct FileInfo {
     ///
     /// On macOS, this is already rewritten to use the extension that the user selected
     /// with the `file format` property.
-    pub(crate) path: PathBuf,
-    /// The selected file format. 
+    pub path: PathBuf,
+    /// The selected file format.
     ///
     /// If there're multiple different formats available
     /// this allows understanding the kind of format that the user expects the file
     /// to be written in. Examples could be Blender 2.4 vs Blender 2.6 vs Blender 2.8.
     /// The `path` above will already contain the appropriate extension chosen in the
     /// `format` property, so it is not necessary to mutate `path` any further.
-    pub(crate) format: Option<FileSpec>,
+    pub format: Option<FileSpec>,
 }
 
 /// Type of file dialog.
--- a/druid-shell/src/dialog.rs
+++ b/druid-shell/src/dialog.rs
@@ -22,6 +22,7 @@ use std::path::{Path, PathBuf};
 #[derive(Debug, Clone)]
 pub struct FileInfo {
     /// The path to the selected file.
+    ///
     /// On macOS, this is already rewritten to use the extension that the user selected
     /// with the `file format` property.
     pub(crate) path: PathBuf,
--- a/druid-shell/src/dialog.rs
+++ b/druid-shell/src/dialog.rs
@@ -26,7 +26,9 @@ pub struct FileInfo {
     /// On macOS, this is already rewritten to use the extension that the user selected
     /// with the `file format` property.
     pub(crate) path: PathBuf,
-    /// The selected file format. If there're multiple different formats available
+    /// The selected file format. 
+    ///
+    /// If there're multiple different formats available
     /// this allows understanding the kind of format that the user expects the file
     /// to be written in. Examples could be Blender 2.4 vs Blender 2.6 vs Blender 2.8.
     /// The `path` above will already contain the appropriate extension chosen in the
--- a/druid-shell/examples/shello.rs
+++ b/druid-shell/examples/shello.rs
@@ -58,12 +58,24 @@ impl WinHandler for HelloState {
                 ]);
                 self.handle.open_file(options);
             }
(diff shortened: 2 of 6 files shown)


druid-shell/src/dialog.rs
```diff
@@ -21,7 +21,16 @@ use std::path::{Path, PathBuf};
 /// This path might point to a file or a directory.
 #[derive(Debug, Clone)]
 pub struct FileInfo {
+    /// The path to the selected file.
+    /// On macOS, this is already rewritten to use the extension that the user selected
+    /// with the `file format` property.
```
cmyr, line 26:
```suggestion
    /// The path to the selected file.
    ///
    /// On macOS, this is already rewritten to use the extension that the user selected
    /// with the `file format` property.
```

Kind of like git commit messages, rust doc comments treat the first line as a 'summary', and then want a newline between it and the rest of the docs. This is displayed inline beside an item when the docs are rendered. 👌

druid-shell/src/dialog.rs
```diff
@@ -21,7 +21,16 @@ use std::path::{Path, PathBuf};
...
     pub(crate) path: PathBuf,
+    /// The selected file format. If there're multiple different formats available
+    /// this allows understanding the kind of format that the user expects the file
+    /// to be written in. Examples could be Blender 2.4 vs Blender 2.6 vs Blender 2.8.
+    /// The `path` above will already contain the appropriate extension chosen in the
+    /// `format` property, so it is not necessary to mutate `path` any further.
```
cmyr, line 32:
```suggestion
    /// The selected file format. 
    ///
    /// If there're multiple different formats available
    /// this allows understanding the kind of format that the user expects the file
    /// to be written in. Examples could be Blender 2.4 vs Blender 2.6 vs Blender 2.8.
    /// The `path` above will already contain the appropriate extension chosen in the
    /// `format` property, so it is not necessary to mutate `path` any further.
```

druid-shell/src/platform/mac/dialog.rs
```diff
@@ -142,3 +164,116 @@ pub(crate) unsafe fn build_panel(ty: FileDialogType, mut options: FileDialogOpti
     }
     panel
 }
+
+// AppKit has a build-in file format accessory view. However, this is only
```
cmyr, line 168:
```suggestion
// AppKit has a built-in file format accessory view. However, this is only
```

druid-shell/src/platform/mac/dialog.rs
```diff
@@ -142,3 +164,116 @@ pub(crate) unsafe fn build_panel(ty: FileDialogType, mut options: FileDialogOpti
...
+    popup_button.autorelease()
+}
+
+unsafe fn file_format_label() -> (id, NSSize) {
+    let label: id = msg_send![class!(NSTextField), new];
+    let _: () = msg_send![label, setBezeled:false];
+    let _: () = msg_send![label, setDrawsBackground:false];
+    // FIXME: As we have to roll our own view hierachy, we're not getting a translated
+    // title here. So we ought to find a way to translate this.
+    let title = NSString::alloc(nil).init_str("File Format:").autorelease();
```
cmyr, line 240: This is interesting. our architecture does not expect druid-shell to ever need to be locale aware; the idea is that localization will happen higher up in the stack, with correctly localized strings passed in to druid-shell where needed. I'm not sure how best to do that here, though. For the time being I'm happy to punt on this, but we will want to revisit it eventually..

druid-shell/src/dialog.rs
```diff
@@ -21,7 +21,16 @@ use std::path::{Path, PathBuf};
...
+    pub(crate) format: Option<FileSpec>,
```
cmyr, line 33: maybe we need this to be `pub`? Do we expect the user to potentially inspect it?

cmyr (approved):
Cool, this looks good! A few comments inline, and CI is complaining in a few places, but all trivial things. Thanks!

druid-shell/src/dialog.rs
terhechte: But wouldn't `path` then also need to be `pub`🤔? I've changed both to be `pub` for now

cmyr (approved):
Great, thanks!

It looks like we used to expose the `path` field through a getter, but I think having both of these be `pub` is reasonable.

cmyr:
Okay, CI is failing because `FileInfo` is created in the other backends, and they aren't setting the new field. Other than that this looks ready to merge. #1846 should merge without any problem 🤞, so I think once that's merged you'll be a "contributor" and CI will run automatically on this when you make changes; if it doesn't just ping me or someone else in order to click the "run" button. 😅

terhechte:
Thanks! @cmyr I'll have a look now. Sorry for the long delay, the last couple of days were rather busy.

cmyr:
I think this needs to be rebased to fix the clippy lints. :(

terhechte:
@cmyr I did a rebase, `cargo clippy` runs fine on my system. I hope this can be merged soon :-) 

I guess there is no easy way to run the Windows test suite locally except by booting up a Windows machine or running a VM, right?

cmyr (approved):
Great, thank you very much @terhechte!

PR #1849, opened 2021-07-03
[druid-derive] Donot use gtk backend in dev dependencies

Maan2003:
The cause of wasm/Ubuntu CI [building the gtk backend](https://github.com/linebender/druid/runs/2979898351).

### DIFF
--- a/druid-derive/Cargo.toml
+++ b/druid-derive/Cargo.toml
@@ -20,6 +20,6 @@ quote = "1.0.7"
 proc-macro2 = "1.0.19"
 
 [dev-dependencies]
-druid = { version = "0.7.0", path = "../druid" }
+druid = { version = "0.7.0", path = "../druid", default-features = false, features = ["x11"] }
 
 float-cmp = { version = "0.8.0", features = ["std"], default-features = false }


cmyr (approved):
Thanks!

PoignardAzur:
This commit gives me the opposite problem: now when I build the project from the root folder (say, to run an example), the x11 backend is always used, even when I want the gtk backend.

Not sure how to fix this. The new cargo resolver might help.

Maan2003:
oh I see what is happening. x11 feature gets activated because examples are built with dev dependencies. I think if at workspace level building an example enables dev dependencies for all members.

PoignardAzur:
Man, we are *not* seeing the end of our "cargo features for backend selection" problems.

Maan2003:
should we revert this? x11 backend is not very complete. some examples don't work(file dialog, menus) which is not a good first impression for the users of the library.

PoignardAzur:
I mean, if you had to commit this change to stop the CI from breaking, then we still need to fix the underlying problem. I think we should move this to Zulip.

Issue #1827, opened 2021-06-16
Svg Widget doesn't render fonts


DrGabble:
I've got an `.svg` I'm trying to render which has text on it. The image comes up fine but none of the text is rendered. 

I turned on console debug logging and it looks like `fontdb` (via `usvg`) is failing to find any of the fonts on my system (Ubuntu 20.04). Tried it with a bunch of fonts I know there are files for, no luck.

Digging around in the Druid source, looks like it might be the default `usvg` constructor not initializing its `fontdb` properly (`load_system_fonts()` etc). If I haven't got the wrong end of the stick, would you like me to open a merge request to add that?

### How It Looks In Druid (screenshot)

![Screenshot from 2021-06-16 22-48-33](https://user-images.githubusercontent.com/15156457/122298864-2cde2f00-cef5-11eb-9d86-32c8089879c1.png)

### How It Looks in Firefox

![Screenshot from 2021-06-16 22-49-06](https://user-images.githubusercontent.com/15156457/122298893-35cf0080-cef5-11eb-9362-8e142bc6306f.png)

### SVG I was using

```xml
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN"
 "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<!-- Generated by graphviz version 2.43.0 (0)
 -->
<!-- Title: graph_0 Pages: 1 -->
<svg width="191pt" height="301pt"
 viewBox="0.00 0.00 191.00 301.00" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
<g id="graph0" class="graph" transform="scale(1 1) rotate(0) translate(4 297)">
<title>graph_0</title>
<polygon fill="white" stroke="transparent" points="-4,4 -4,-297 187,-297 187,4 -4,4"/>
<text text-anchor="middle" x="91.5" y="-7.8" font-family="Times,serif" font-size="14.00">Case: 1 / 1, Step: 0 / 10</text>
<g id="clust1" class="cluster">
<title>cluster_1</title>
<polygon fill="none" stroke="black" points="16.5,-31 16.5,-285 166.5,-285 166.5,-31 16.5,-31"/>
<text text-anchor="middle" x="91.5" y="-269.8" font-family="Times,serif" font-size="14.00">And</text>
</g>
<!-- node_0 -->
[...]

PR #1850, opened 2021-07-03
Added font support for SVG text

DrGabble:
Fixes https://github.com/linebender/druid/issues/1827


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ You can find its changes [documented below](#070---2021-01-01).
 
 ### Added
 
+- System fonts loaded so that SVG images render text ([#1850] by [@DrGabble])
```
cmyr, line 15: you'll also need to add the link to the PR below. :)

cmyr (approved):
This seems reasonable. It would be nice if we could somehow defer font loading until we knew that fonts were needed, since it likely has some cost, but I'm fine with worrying about that later if it seems like a problem. :)

DrGabble:
> This seems reasonable. It would be nice if we could somehow defer font loading until we knew that fonts were needed, since it likely has some cost, but I'm fine with worrying about that later if it seems like a problem. :)

Yeah doing that automatically is probably hard. Could add a function for "new_with_text" but seems like people might miss it. Agreed it will make things more expensive but we can fix that later if it becomes an issue :)

CHANGELOG.md
cmyr: @DrGabble this is ready to merge except for this missing link!

DrGabble:
Yes sorry, VM packed in, it's on my to-do list don't worry

Get Outlook for Android<https://aka.ms/AAb9ysg>

________________________________

CHANGELOG.md
DrGabble: @cmyr done!

PR #1851, opened 2021-07-04
X11: Implement putting things into the clipboard

psychon:
This PR fixes https://github.com/linebender/druid/issues/937: The X11 platform now fully supports the clipboard.

When putting things into the clipboard, a new (invisible (= not mapped)) window is created and this window becomes the selection owner. Nothing else "x11-y" happens, so the clipboard contents are just kept in memory. Later, we might get `SelectionRequest` events, which are replied to by setting a window property and sending an event. INCR requests (for "incremental" (for "lots of data")) are again more complicated, just as when getting the clipboard contents.

Another complication are self-transfer: If we own the CLIPBOARD selection and request the clipboard contents, we would hang: The X11 server sends us an event that something wants a selection transfer, but the code would just put the event into a queue and not answer it. I briefly tried to implement "we still have to handle these requests", but ran into borrowing problems (the code for getting the selection and handling selection requests had conflicting borrows). Since it is way easier this way anyway, I implement a short-circuit: When we are the selection owner, "nothing X11" is done and we just directly provide the requested data.

Testing done with the following program and pasting into `gvim` (I also tested without the `while` loop for non-INCR transfers):
```rust
fn main() {
    tracing_subscriber::fmt().init();
    let app = druid_shell::Application::new().unwrap();
    let mut contents = String::from("1234567890");
    while contents.len() < 17 * 1024 * 1024 {
        let copy = contents.clone();
        contents = contents + &copy;
    }
    app.clipboard().put_string(contents);
    println!("{:?}", app.clipboard().available_type_names());
    println!("{:?}", app.clipboard().preferred_format(&["FOOBAR", "UTF8_STRING"]));
    println!("{:?}", app.clipboard().get_string().map(|s| s.len()));
    println!("Running!");
    app.run(None);
}
```


druid-shell/src/platform/x11/clipboard.rs
```diff
@@ -231,6 +351,245 @@ impl Clipboard {
...
+}
+
+#[derive(Debug)]
+struct IncrementalTransfer {
+    requestor: Window,
+    selection: Atom,
+    target: Atom,
+    property: Atom,
+    time: Timestamp,
+    data: Rc<Vec<u8>>,
```
Maan2003, line 543: here as well

druid-shell/src/platform/x11/clipboard.rs
```diff
@@ -231,6 +351,245 @@ impl Clipboard {
...
+            }
+        }
+        Ok(())
+    }
+}
+
+#[derive(Debug)]
+struct ClipboardContents {
+    owner_window: Window,
+    data: Vec<(Atom, String, Rc<Vec<u8>>)>,
```
Maan2003, line 482: Can we use `Rc<[u8]>`?

Maan2003 (commented):
Thanks! Really happy to see x11 backend moving forward. Just some minor nits:

druid-shell/src/platform/x11/clipboard.rs
psychon: I needed a moment to figure out how to construct that, but yeah, it seems to work.

CHANGELOG.md
```diff
@@ -46,8 +46,12 @@ You can find its changes [documented below](#070---2021-01-01).
 - X11: Added support for `get_monitors` ([#1804] by [@psychon])
 - x11: Remove some unnecessary casts ([#1851] by [@psychon])
 - `has_focus` method on `WidgetPod` ([#1825] by [@ForLoveOfCats])
+<<<<<<< HEAD
+- x11: Add support for getting and setting clipboard contents ([#1805] and [#1851] by [@psychon])
+=======
 - x11: Add support for getting clipboard contents ([#1805] by [@psychon])
 - Linux extension: primary_clipboard ([#1843] by [@Maan2003])
+>>>>>>> master
```
Maan2003, line 54: looks like some unresolved conflicts

psychon: When a file has two conflicts and you only fix one of them... :facepalm: 

Sorry!

Maan2003 (approved):
Thanks!

PR #1852, opened 2021-07-04
X11: Always save screen number as usize

psychon:
The screen number is only ever used for indexing into the list of
screens. Thus, usize is the perfect type for this. x11rb even provides
the screen number as usize.

This commit thus can get rid of a couple of casts by just always using
usize.

(History guess: xcb-rs uses i32 for the screen number (because C) and
that's why the code was like this in the first place)


Maan2003 (approved):
Thanks!

PR #1853, opened 2021-07-04
Use `Cursor::Pointer` when hovering over `Button`

CryZe:
Usually buttons show that they are clickable by changing the cursor to a pointer.

### DIFF
--- a/druid/src/widget/button.rs
+++ b/druid/src/widget/button.rs
@@ -14,6 +14,7 @@
 
 //! A button widget.
 
+use crate::shell::Cursor;
 use crate::widget::prelude::*;
 use crate::widget::{Click, ControllerHost, Label, LabelText};
 use crate::{theme, Affine, Data, Insets, LinearGradient, UnitPoint};
@@ -127,6 +128,13 @@ impl<T: Data> Widget<T> for Button<T> {
                     trace!("Button {:?} released", ctx.widget_id());
                 }
                 ctx.set_active(false);
+            }
+            Event::MouseMove(_) => {
+                if !ctx.is_disabled() {
+                    ctx.set_cursor(&Cursor::Pointer);
+                } else {
+                    ctx.clear_cursor();
+                }
             }
             _ => (),
         }


druid/src/widget/button.rs
```diff
@@ -128,6 +129,13 @@ impl<T: Data> Widget<T> for Button<T> {
                 }
                 ctx.set_active(false);
             }
+            Event::MouseMove(_) => {
+                if !ctx.is_disabled() {
+                    ctx.set_cursor(&Cursor::Pointer);
+                } else {
+                    ctx.clear_cursor();
+                }
+            }
```
xarvic, line 138: I think this belongs in `lifecycle` instead of `event`. `Lifecycle::WidgetAdded` and `Lifecycle::DisabledChanged` are the events after which we need to update the cursor

CryZe: I'm not sure I understand. Modifying the cursor isn't a thing you can do from a lifecycle event and it doesn't make that much sense either as it needs to change based on whether you hover or not. No other widget seems to do it there either.

xarvic: You are right. I forgot, that `set_cursor` is only availible on `EventCtx`.
The problem is just that: If you click a button which then becomes disabled, the cursor only changes after the mouse moved again, you can see this in the disabled example.
But with the current implementation of cursor in `WidgetPod` and `Window` we probably can't change it :/

cmyr:
This is a bit tricky; on macOS, for instance, the cursor does not switch to the pointer when a button is hovered.

To me this feels mostly like a web convention? Is it the default behaviour on windows or in gtk?

CryZe:
TIL, this does indeed only seem to be a web convention. I guess everything is web nowadays to the point where it felt weird for druid not to do it. But native Windows apps do indeed not seem to behave that way. I don't know how it is on Linux.

CryZe:
Maybe this can be a `theme::BUTTON_CURSOR` setting?

Maan2003:
Also web doesn't use pointer cursor by default for `<button>`, see [mdn button docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button) for demo

PR #1861, opened 2021-07-07
Focus chain fix

xarvic:
This PR  updates the documentation of `children_changed()`, `should_propagate_to_hidden()`, and `register_for_focus()`.


druid/src/event.rs
```diff
@@ -397,8 +397,22 @@ impl Event {
...
+    /// this method: determine which widgets are allowed to receive focus. As a rule
+    /// of thumb a widget count as `hidden` if it makes no sense for it to receive focus
+    /// via pressing TAB.
+    ///
+    /// If a widget changes which children are hidden it must call [`children_changed`].
+    ///
+    /// See also [`LifeCycle::should_propagate_to_hidden`]
+    ///
+    /// [`children_changed`]: crate::EventCtx::children_changed
+    /// [`LifeCycle::should_propagate_to_hidden`]: LifeCycle::should_propagate_to_hidden
```
cmyr, line 415:
```suggestion
    /// Whether this event should be sent to widgets which are currently not visible and not
    /// accessible.
    ///
    /// For example: the hidden tabs in a tabs widget are `hidden` whereas the non-visible
    /// widgets in a scroll are not, since you can bring them into view by scrolling.
    ///
    /// This distinction between scroll and tabs is due to one of the main purposes of
    /// this method: determining which widgets are allowed to receive focus. As a rule
    /// of thumb a widget counts as `hidden` if it makes no sense for it to receive focus
    /// when the user presses thee 'tab' key.
    ///
    /// If a widget changes which children are hidden it must call [`children_changed`].
    ///
    /// See also [`LifeCycle::should_propagate_to_hidden`].
    ///
    /// [`children_changed`]: crate::EventCtx::children_changed
    /// [`LifeCycle::should_propagate_to_hidden`]: LifeCycle::should_propagate_to_hidden
```

druid/src/event.rs
```diff
@@ -424,8 +438,22 @@ impl Event {
...
+    /// this method: determine which widgets are allowed to receive focus. As a rule
+    /// of thumb a widget count as `hidden` if it makes no sense for it to receive focus
+    /// via pressing TAB.
+    ///
+    /// If a widget changes which children are hidden it must call [`children_changed`].
+    ///
+    /// See also [`Event::should_propagate_to_hidden`]
+    ///
+    /// [`children_changed`]: crate::EventCtx::children_changed
+    /// [`Event::should_propagate_to_hidden`]: Event::should_propagate_to_hidden
```
cmyr, line 456: Maybe it makes more sense to keep this shorter, and just link to the docs for the `Event` version? It will make it easier to ensure we keep these up to date.

```suggestion
    /// Whether this event should be sent to widgets which are currently not visible and not
    /// accessible.
    ///
    /// For example, the hidden tabs in a tabs widget are `hidden` whereas the non-visible
    /// widgets in a scroll are not, since you can bring them into view by scrolling.
    ///
    /// This distinction between scroll and tabs is due to one of the main purposes of
    /// this method: determining which widgets are allowed to receive focus. As a rule
    /// of thumb a widget count as `hidden` if it makes no sense for it to receive focus
    /// in response to the user pressing the 'tab' key.
    ///
    /// If a widget changes which children are hidden it must call [`children_changed`].
    ///
    /// See also [`Event::should_propagate_to_hidden`].
    ///
    /// [`children_changed`]: crate::EventCtx::children_changed
    /// [`Event::should_propagate_to_hidden`]: Event::should_propagate_to_hidden
```

druid/src/event.rs
```diff
@@ -439,8 +467,19 @@ impl LifeCycle {
 }
 
 impl InternalLifeCycle {
-    /// Whether this event should be sent to widgets which are currently not visible
-    /// (for example the hidden tabs in a tabs widget).
+    /// Whether this event should be sent to widgets which are currently not visible and not
```
cmyr, line 470: ditto, I don't think we need to duplicate this everywhere.

cmyr (approved):
Awesome, thanks for fixing this up, I think your version is much clearer.

A couple little tweaks & suggestions, but overall this looks great.

druid/src/event.rs
xarvic: Ok, done :)

druid/src/event.rs
xarvic: I agree

PR #1865, opened 2021-07-09
x11: Use only a single atom manager

psychon:
Instead of having one atom manager instance per window and another for the clipboard, this merges everything into a single instance. The idea is that this is a tiny, tiny bit more performant (but likely unmeasurable so). Also, constructing the clipboard becomes infallible, which is always nice.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,7 @@ You can find its changes [documented below](#070---2021-01-01).
 - Update look and feel of controls when disabled ([#1717] by [@xarvic])
 - Change the signature of `add_idle_callback` ([#1787] by [@jneem])
 - Move macOS only function to Mac extension trait ([#1863] by [@Maan2003])
+- x11: Only query atoms once instead of per window ([#1865] by [@psychon])
 
 ### Deprecated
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -65,6 +65,7 @@ You can find its changes [documented below](#070---2021-01-01).
 - Update look and feel of controls when disabled ([#1717] by [@xarvic])
 - Change the signature of `add_idle_callback` ([#1787] by [@jneem])
 - Move macOS only function to Mac extension trait ([#1863] by [@Maan2003])
+- x11: Only query atoms once instead of per window ([#1865] by [@psychon])
```
Maan2003, line 68: I am not sure if document internal changes in changelog. @cmyr ?

Maan2003 (approved):
Looks good!

CHANGELOG.md
psychon: I'm not convinced by this changelog entry either, but: https://github.com/linebender/druid/blob/master/CONTRIBUTING.md#changelog

> Every pull request should document all changes made in the changelog.

PR #1866, opened 2021-07-09
X11: Add timeout for getting clipboard contents

psychon:
When the selection does not answer, the code would just hang while
waiting for a reply. This could e.g. happen when the "other end"
crashes.

Handle this situation by applying a timeout of five seconds for getting
the clipboard contents. After this timeout, the transfer is aborted with
an error.

This was tested by batching handle_request() in druid-shell's clipboard
code to just return Ok(()). Then, I started the following program:

    fn main() { druid_shell::Application::new().unwrap().clipboard().put_string("whatever"); }

Afterwards, trying to get the clipboard contents with druid_shell would
just hang.

With this commit, the following happens instead:

     WARN druid_shell::backend::x11::clipboard: Error in Clipboard::do_transfer: ConnectionError(IOError(Custom { kind: TimedOut, error: "Timeout while waiting for selection owner to reply" }))


Maan2003 (approved):
Thanks!

PR #1868, opened 2021-07-10
x11: Set WM_CLASS on windows

psychon:
I compared "xprop" output between a random window with the gtk and x11
backends and noticed that WM_CLASS is missing, even though it is
required by ICCCM.

This commit implements setting the WM_CLASS property on x11.

This was tested via "xprop": gtk and x11 backends now produce the same
value for the WM_CLASS property.


Maan2003 (approved):
Thanks!

PR #1870, opened 2021-07-11
Fix localisation stipping in GTK, X11 and macos

### LINKED ISSUE #1864: Localization fails with .utf8 suffixed locales
On my system, by default, the locale is:
```
LANG=fr_FR.utf8
```

If I run a druid UI with it, it fails to localize using fr-FR locales:
```
$ echo $LANG
fr_FR.utf8
$ ./target/debug/rust-ui-druid
DEBUG [druid::localization] available locales [en-US, fr-FR], current en-US
DEBUG [druid::localization] resolved: [en-US]
```

It works fine with `LANG=fr_FR`:
```
$ LANG=fr_FR ./target/debug/rust-ui-druid
DEBUG [druid::localization] available locales [en-US, fr-FR], current fr-FR
DEBUG [druid::localization] resolved: [fr-FR, en-US]
```

It would be nice to:
 - Have more logs so that understanding what happens is easier (e.g which value is really found before any defaulting)
 - Ignore the .utf8 suffix

JAicewizard:
on macos we already strip the modifiers, this extends this to all backends and also stripping the charset.

fixes #1864

### DIFF
--- a/druid-shell/src/backend/gtk/application.rs
+++ b/druid-shell/src/backend/gtk/application.rs
@@ -84,7 +84,11 @@ impl Application {
     }
 
     pub fn get_locale() -> String {
-        glib::get_language_names()[0].as_str().into()
+        let mut locale: String = glib::get_language_names()[0].as_str().into();
+        if let Some(idx) = locale.chars().position(|c| c == '.') {
+            locale.truncate(idx);
+        }
+        locale
     }
 }
 
--- a/druid-shell/src/backend/gtk/application.rs
+++ b/druid-shell/src/backend/gtk/application.rs
@@ -85,7 +85,7 @@ impl Application {
 
     pub fn get_locale() -> String {
         let mut locale: String = glib::get_language_names()[0].as_str().into();
-        if let Some(idx) = locale.chars().position(|c| c == '.') {
+        if let Some(idx) = locale.chars().position(|c| c == '.' || c == '@') {
             locale.truncate(idx);
         }
         locale
--- a/druid-shell/src/backend/gtk/application.rs
+++ b/druid-shell/src/backend/gtk/application.rs
@@ -85,6 +85,7 @@ impl Application {
 
     pub fn get_locale() -> String {
         let mut locale: String = glib::get_language_names()[0].as_str().into();
+        // This is done because the locale parsing library we use expects an unicode locale, but these vars have an ISO locale
         if let Some(idx) = locale.chars().position(|c| c == '.' || c == '@') {
             locale.truncate(idx);
         }
--- a/druid-shell/src/backend/mac/application.rs
+++ b/druid-shell/src/backend/mac/application.rs
@@ -111,7 +111,7 @@ impl Application {
             let locale: id = msg_send![nslocale_class, currentLocale];
             let ident: id = msg_send![locale, localeIdentifier];
             let mut locale = util::from_nsstring(ident);
-            if let Some(idx) = locale.chars().position(|c| c == '@') {
+            if let Some(idx) = locale.chars().position(|c| c == '.' || c == '@') {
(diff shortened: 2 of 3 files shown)


druid-shell/src/backend/gtk/application.rs
```diff
@@ -84,7 +84,11 @@ impl Application {
     }
 
     pub fn get_locale() -> String {
-        glib::get_language_names()[0].as_str().into()
+        let mut locale: String = glib::get_language_names()[0].as_str().into();
```
ColinPitrat, line 87: I think it's worth a comment explaining why this is done (locale having one format but type used expecting a different format).

druid-shell/src/backend/x11/application.rs
```diff
@@ -778,14 +778,18 @@ impl Application {
 
         // from gettext manual
         // https://www.gnu.org/software/gettext/manual/html_node/Locale-Environment-Variables.html#Locale-Environment-Variables
-        var_non_empty("LANGUAGE")
+        let mut locale = var_non_empty("LANGUAGE")
```
ColinPitrat, line 781: Would it be possible to make var_non_empty log some info (ideally at some verbose level if such a thing is available). It can be useful to the user to know which var is used and which value it has. Currently if the variable cannot be parse you just see "en_US".

JAicewizard: I will leave this open for whoever is doing the final review. It does hinder readability but I can see its usefulness.

druid-shell/src/backend/gtk/application.rs
```diff
@@ -84,7 +84,12 @@ impl Application {
...
```
cmyr, line 87: this index would panic if there `get_language_names()` is empty. Is that possible?

druid-shell/src/backend/mac/application.rs
```diff
@@ -111,7 +111,8 @@ impl Application {
             let locale: id = msg_send![nslocale_class, currentLocale];
             let ident: id = msg_send![locale, localeIdentifier];
             let mut locale = util::from_nsstring(ident);
-            if let Some(idx) = locale.chars().position(|c| c == '@') {
+            // This is done because the locale parsing library we use expects an unicode locale, but these vars have an ISO locale
+            if let Some(idx) = locale.chars().position(|c| c == '.' || c == '@') {
```
cmyr, line 115: I believe this is okay as-is, because macOS doesn't include encoding in the locale?

druid-shell/src/backend/x11/application.rs
cmyr: I think this would be welcome, but I'd make it a separate PR.

cmyr (commented):
@ColinPitrat can you confirm that this fixes the original issue? It looks like it should, but would like to be sure.

druid-shell/src/backend/gtk/application.rs
JAicewizard: `Computes a list of applicable locale names, which can be used to e.g. construct locale-dependent filenames or search paths. The returned list is sorted from most desirable to least desirable and always contains the default locale "C".`

If druid fails to parse the locale it will be "en_US" so this should work in all cases.

druid-shell/src/backend/mac/application.rs
JAicewizard: I couldnt find any documentation on this. Having the code be the same across the backends seems nice to me. It hasnt caused any problems so far so it should indeed be fine without.

JAicewizard: I left the comment in since I think it still applies and is helpful

cmyr (approved):
Looks good, thanks!

PR #1871, opened 2021-07-12
Add serde feature to druid-shell

Kethku:
I noticed that kurbo wasn't using the serde feature for it's types (Point and such) even though serde was defined as a feature for druid. Adding serde to druid-shell's features fixed the issue for me.


cmyr (approved):
This is reasonable, thanks!

PR #1875, opened 2021-07-15
widget: slider: Add stepping functionality

raymanfx:
A new builder method `with_step(f64)` is added to set the stepping value.

Signed-off-by: Christopher N. Hesse <raymanfx@gmail.com>

### DIFF
--- a/druid/src/widget/slider.rs
+++ b/druid/src/widget/slider.rs
@@ -31,6 +31,7 @@ const KNOB_STROKE_WIDTH: f64 = 2.0;
 pub struct Slider {
     min: f64,
     max: f64,
+    step: Option<f64>,
     knob_pos: Point,
     knob_hovered: bool,
     x_offset: f64,
@@ -42,6 +43,7 @@ impl Slider {
         Slider {
             min: 0.,
             max: 1.,
+            step: None,
             knob_pos: Default::default(),
             knob_hovered: Default::default(),
             x_offset: Default::default(),
@@ -56,6 +58,24 @@ impl Slider {
         self.max = max;
         self
     }
+
+    /// Builder-style method to set the stepping.
+    ///
+    /// The default step size is `0.0` (smooth).
+    pub fn with_step(mut self, step: f64) -> Self {
+        if step < 0.0 {
+            tracing::warn!("bad stepping (must be positive): {}", step);
+            return self;
+        }
+        self.step = if step > 0.0 {
+            Some(step)
+        } else {
+            // A stepping value of 0.0 would yield an infinite amount of steps.
+            // Enforce no stepping instead.
+            None
+        };
+        self
+    }
 }
 
 impl Slider {
@@ -68,7 +88,14 @@ impl Slider {
         let scalar = ((mouse_x + self.x_offset - knob_width / 2.) / (slider_width - knob_width))
             .max(0.0)
             .min(1.0);
-        self.min + scalar * (self.max - self.min)
+        let mut value = self.min + scalar * (self.max - self.min);
+        if let Some(stepping) = self.step {
+            // Determine the number of steps and fit the slider position to a discrete step
+            let steps = (self.max / stepping).round();
+            let step = (value * steps / self.max).round();
+            value = step / steps * self.max;
+        }
+        value
     }
 
     fn normalize(&self, data: f64) -> f64 {
--- a/druid/src/widget/slider.rs
+++ b/druid/src/widget/slider.rs
@@ -90,10 +90,7 @@ impl Slider {
             .min(1.0);
         let mut value = self.min + scalar * (self.max - self.min);
         if let Some(step) = self.step {
-            let make_discrete = |value: f64, step: f64, min: f64, max: f64| {
-                (((value - min) / step).round() * step + min).min(max)
-            };
-            let max_step_value = make_discrete(self.max, step, self.min, self.max);
+            let max_step_value = ((self.max - self.min) / step).floor() * step + self.min;
             if value > max_step_value {
                 // edge case: make sure max is reachable
                 let left_dist = value - max_step_value;
@@ -105,7 +102,7 @@ impl Slider {
                 };
             } else {
                 // snap to discrete intervals
-                value = make_discrete(value, step, self.min, self.max);
+                value = (((value - self.min) / step).round() * step + self.min).min(self.max);
             }
         }
         value
--- a/druid/src/widget/slider.rs
+++ b/druid/src/widget/slider.rs
@@ -76,18 +76,6 @@ impl Slider {
         };
         self
     }
-
-    fn check_step(&self) {
-        if let Some(step) = self.step {
-            let max_step_value = (100.0 / step).round() * step;
-            if max_step_value < self.max {
-                warn!(
-                    "max ({}) - min ({}) should be a multiple of step ({})",
-                    self.max, self.min, step
-                );
-            }
-        }
-    }
 }
 
 impl Slider {
@@ -178,9 +166,6 @@ impl Widget<f64> for Slider {
     #[instrument(name = "Slider", level = "trace", skip(self, ctx, event, _data, _env))]
     fn lifecycle(&mut self, ctx: &mut LifeCycleCtx, event: &LifeCycle, _data: &f64, _env: &Env) {
         match event {
-            LifeCycle::WidgetAdded => {
-                self.check_step();
-            }
             LifeCycle::DisabledChanged(_) => {
                 ctx.request_paint();
             }
--- a/druid/src/widget/slider.rs
+++ b/druid/src/widget/slider.rs
@@ -162,11 +162,8 @@ impl Widget<f64> for Slider {
 
     #[instrument(name = "Slider", level = "trace", skip(self, ctx, event, _data, _env))]
     fn lifecycle(&mut self, ctx: &mut LifeCycleCtx, event: &LifeCycle, _data: &f64, _env: &Env) {
-        match event {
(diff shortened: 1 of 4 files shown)


druid/src/widget/slider.rs
```diff
@@ -56,6 +58,14 @@ impl Slider {
         self.max = max;
         self
     }
+
+    /// Builder-style method to set the stepping.
+    ///
+    /// The default step size is `None` (smooth).
+    pub fn with_step(mut self, step: f64) -> Self {
+        self.step = Some(step);
```
JAicewizard, line 66: You probably want a check to avoid negative numbers and log a warning

druid/src/widget/slider.rs
```diff
@@ -68,7 +78,18 @@ impl Slider {
         let scalar = ((mouse_x + self.x_offset - knob_width / 2.) / (slider_width - knob_width))
             .max(0.0)
             .min(1.0);
-        self.min + scalar * (self.max - self.min)
+        let mut value = self.min + scalar * (self.max - self.min);
+        if let Some(stepping) = self.step {
+            let steps = if stepping > 0.0 {
```
JAicewizard, line 83: If you move this check into the `with_step` function this could be simplified a lot. If stepping is `0.0` it is probably a bug, and a warning should be logged or something like that.

raymanfx:
What should a `0.0` step value represent? Smooth (no step at all) or a single step aka binary min/max range?

JAicewizard:
No step at all is reasonable. It makes logical sense to me, you have infinite steps.

Maybe its more rust-like use an option and disallow `0.0`, but I am the wrong person to ask that.

raymanfx:
This is good to go from my point of view now.
Paging @cmyr since he responded to my Zulip topic a few days ago.

druid/src/widget/slider.rs
```diff
@@ -56,6 +58,24 @@ impl Slider {
...
+            tracing::warn!("bad stepping (must be positive): {}", step);
+            return self;
+        }
+        self.step = if step > 0.0 {
+            Some(step)
+        } else {
+            // A stepping value of 0.0 would yield an infinite amount of steps.
+            // Enforce no stepping instead.
+            None
+        };
```
Maan2003, line 76: Should we also warn for 0.0 ?

druid/src/widget/slider.rs
```diff
@@ -56,6 +58,24 @@ impl Slider {
         self.max = max;
         self
     }
+
+    /// Builder-style method to set the stepping.
+    ///
+    /// The default step size is `0.0` (smooth).
+    pub fn with_step(mut self, step: f64) -> Self {
```
Maan2003, line 65: can we also get a setter method

druid/src/widget/slider.rs
```diff
@@ -68,7 +88,14 @@ impl Slider {
         let scalar = ((mouse_x + self.x_offset - knob_width / 2.) / (slider_width - knob_width))
             .max(0.0)
             .min(1.0);
-        self.min + scalar * (self.max - self.min)
+        let mut value = self.min + scalar * (self.max - self.min);
+        if let Some(stepping) = self.step {
+            // Determine the number of steps and fit the slider position to a discrete step
+            let steps = (self.max / stepping).round();
+            let step = (value * steps / self.max).round();
+            value = step / steps * self.max;
```
Maan2003, line 96: I think, this is wrong

say `self.min = 5`, `self.max = 25`, `stepping = 10` and `value = 13`

so 
`steps = 25/10 = (2.5).round() = 3`
`step = (13 * 3 / 25).round()` = `1.56` = `2`
`value = 2 / 3 * 25 = 16.66...`

but the value should be `15`

(sorry if maths is wrong :grimacing: )

Maan2003 (commented):
- What will/should happen if initial value is not aligned with the step?
- Should we show the available points(as little dots) on the slider?

druid/src/widget/slider.rs
raymanfx: I don't think so - see the earlier discussion between Jaap and me.
`0.0` is a valid value, indicating smooth stepping (as opposed to discrete stepping).

druid/src/widget/slider.rs
raymanfx: Do we need one? We don't have a setter for the range either. I'd like to keep the API consistent here I think.

druid/src/widget/slider.rs
raymanfx: Why should the value be `15`?
`25` % `10` != `0`, so it seems logical to me that the stepper values can be fractional. Let's list the possible steps:

`step = 0` => `value = 0 / 3 * 25 = 0.0`
`step = 1` => `value = 1 / 3 * 25 = 8.3..`
`step = 2` => `value = 2 / 3 * 25 = 16.66`
`step = 3` => `value = 3 / 3 * 25 = 25`

As you can see, this yields a uniform step length of `8.33..`, which is what I would expect.

JAicewizard:
Huh, the stepping is 10. I can see there being debate on of that should be in the form of `min+n*stepping`, but I'm confused how a stepping of 10 should create fractional results. (when min isn't a fraction)

With this rounding formula a stepping if 11 would create the same result as with 10.

druid/src/widget/slider.rs
Maan2003: normally in druid we have a setter method for every builder style. I am not sure why Slider was skipped

druid/src/widget/slider.rs
Maan2003: the value should start from `5`(self.min) and step by `10` and end at `25`
so, 5, 15, 25 should be the possible values :shrug:

druid/src/widget/slider.rs
Maan2003: so `.with_step(0.0)` is a way to clear the previous step if any

Maan2003:
> I'm confused how a stepping of 10 should create fractional results. (when min isn't a fraction)

because self.max was used in calculation. and self.max / step is fractional

druid/src/widget/slider.rs
```diff
@@ -56,6 +58,49 @@ impl Slider {
...
+        if let Some(stepping) = self.step {
+            // round down max such that max - min is a multiple of step
+            // say max - min = stepping * n + k
+
+            let n = (self.max - self.min) / stepping;
+
+            // check if n is choose enough to an integer
+
+            // f64::EPSILON doesn't work :/
+            if (n.round() - n).abs() > 10e-10 {
```
Maan2003, line 90: I don't like this. Suggestions welcome

Maan2003:
[different implementation](https://github.com/linebender/druid/compare/master...Maan2003:slider-steps-2) treats max as a separate step if it is not "well aligned"(min + n * step)

druid/src/widget/slider.rs
```diff
@@ -68,7 +88,25 @@ impl Slider {
         let scalar = ((mouse_x + self.x_offset - knob_width / 2.) / (slider_width - knob_width))
             .max(0.0)
             .min(1.0);
-        self.min + scalar * (self.max - self.min)
+        let mut value = self.min + scalar * (self.max - self.min);
+        if let Some(step) = self.step {
+            let max_step_value =
+                (((value - self.min) / step).floor() * step + self.min).min(self.max);
```
Maan2003, line 94: looks like a typo
```suggestion
                (((self.max - self.min) / step).floor() * step + self.min).min(self.max);
```
also I don't think .min(self.max) is needed now

druid/src/widget/slider.rs
```diff
@@ -68,7 +88,25 @@ impl Slider {
...
+                let left_dist = value - max_step_value;
+                let right_dist = self.max - value;
+                value = if left_dist < right_dist {
+                    max_step_value
+                } else {
+                    self.max
+                };
+            } else {
+                // snap to discrete intervals
+                value = (((value - self.min) / step).round() * step + self.min).min(self.max);
```
Maan2003, line 106: same i don't think `.min(self.max)` is needed because `value < max_step_value` and at most round to `max_step_value` which is less than `self.max`

raymanfx: > value < max_step_value

That is actually not true, since `value` could very well be `self.max` here.
I updated the code to remove the call for in the other location though.

Maan2003 (approved):
Thanks!

druid/src/widget/slider.rs
```diff
@@ -57,6 +59,24 @@ impl Slider {
...
+            warn!("bad stepping (must be positive): {}", step);
+            return self;
+        }
+        self.step = if step > 0.0 {
+            Some(step)
+        } else {
+            // A stepping value of 0.0 would yield an infinite amount of steps.
+            // Enforce no stepping instead.
+            None
+        };
```
cmyr, line 76: No need to change it, but just for posterity I'd have just written:
```suggestion
        if step < 0.0 {
            warn!("bad stepping (must be positive): {}", step);
        } else {
            self.step = Some(step);
        }
```

cmyr (approved):
Works for me! As always this stuff is trickier than expected, but I think it's also fair for us to expect the user to provide sane values.

druid/src/widget/slider.rs
raymanfx: That has a drawback IMO: you'd actually need to check for `step <= 0` since otherwise we would divide by zero elsewhere iirc.

We decided to allow `0` as a valid step value though, indicating smooth stepping (no intervals) - but we have to explicitly model this as `None`.

Maan2003: lets change that `step > 0.0` to `step != 0.0` to make it more explicit

PR #1876, opened 2021-07-16
x11: logging in Application::get_locale

Maan2003:
thanks @ColinPitrat

https://github.com/linebender/druid/pull/1870#discussion_r667471453

Also changed the name and made it a function


cmyr (approved):
Sure, this seems like it should be helpful. Thanks!

PR #1883, opened 2021-07-25
Implements WindowLevel::Modal for windows

GeorgOhneH:
I tried to implement the modal functionality for windows.  It accomplishes this by disabling all other windows.

### DIFF
--- a/druid-shell/src/backend/windows/window.rs
+++ b/druid-shell/src/backend/windows/window.rs
@@ -236,7 +236,7 @@ trait WndProc {
     fn cleanup(&self, hwnd: HWND);
 
     fn window_proc(&self, hwnd: HWND, msg: UINT, wparam: WPARAM, lparam: LPARAM)
-                   -> Option<LRESULT>;
+        -> Option<LRESULT>;
 }
 
 // State and logic for the winapi window procedure entry point. Note that this level
@@ -461,8 +461,8 @@ impl WndState {
 
 impl MyWndProc {
     fn with_window_state<F, R>(&self, f: F) -> R
-        where
-            F: FnOnce(Rc<WindowState>) -> R,
+    where
+        F: FnOnce(Rc<WindowState>) -> R,
     {
         f(self
             .handle
@@ -476,8 +476,8 @@ impl MyWndProc {
 
     #[track_caller]
     fn with_wnd_state<F, R>(&self, f: F) -> Option<R>
-        where
-            F: FnOnce(&mut WndState) -> R,
+    where
+        F: FnOnce(&mut WndState) -> R,
     {
         let ret = if let Ok(mut s) = self.state.try_borrow_mut() {
             (*s).as_mut().map(|state| f(state))
@@ -950,7 +950,7 @@ impl WndProc for MyWndProc {
                         error!("ResizeBuffers failed: 0x{:x}", res);
                     }
                 })
-                    .map(|_| 0)
+                .map(|_| 0)
             },
             WM_COMMAND => {
                 self.with_wnd_state(|s| s.handler.command(LOWORD(wparam as u32) as u32));
@@ -962,7 +962,7 @@ impl WndProc for MyWndProc {
                 unsafe {
                     let handled = self.with_wnd_state(|s| {
                         if let Some(event) =
-                        s.keyboard_state.process_message(hwnd, msg, wparam, lparam)
+                            s.keyboard_state.process_message(hwnd, msg, wparam, lparam)
                         {
                             // If the window doesn't have a menu, then we need to suppress ALT/F10.
                             // Otherwise we will stop getting mouse events for no gain.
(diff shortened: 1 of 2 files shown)


cmyr:
Thanks for the PR! Is there any additional context for this? Was it discussed somewhere, or was this just something you noticed was missing?

GeorgOhneH:
No additional context. I just wanted if for my own project und made a quick implementation. Also I'm unsure if disabling all windows is always the right choice. There are use cases where you only want to disable the direct parent.

sjoshid:
hi @GeorgOhneH 
Im working on adding child window support for Windows and this PR I feel is closely tied, if not completely, to it. I have a question and it is related to `Vec<HWND>` being stored as parents. 
I assume you are storing a vec so you can disable all non-foreground windows when a modal is displayed. But isnt it sufficient to disable _just_ the parent of modal window? If yes, then a `Vec` is not needed. Right?

GeorgOhneH:
It depends. It's mainly a design choice. There are use cases for both. I'm not sure how it's implemented for other platforms.
Secondly, it was just easier like this, because I don't know how to get the parent.

druid-shell/src/backend/windows/window.rs
```diff
@@ -1369,8 +1380,9 @@ impl WindowBuilder {
                         dwExStyle = 0;
                     }
                     WindowLevel::Modal => {
-                        dwStyle = WS_OVERLAPPED;
-                        dwExStyle = WS_EX_TOPMOST;
+                        dwStyle = WS_POPUPWINDOW | WS_CAPTION;
```
sjoshid, line 1383: Why would a modal be a `WS_POPUPWINDOW`? According to [this](https://docs.microsoft.com/en-us/windows/win32/winmsg/window-features#pop-up-windows) a modal window "appear outside an application's main window". In my mind, a modal window is a window that is at the **center** of the current foreground window while other windows are dimmed.

sjoshid: What Im trying to say is it should be a `WS_CHILD` window. Maybe you tried it already and didnt work?

sjoshid:
> I don't know how to get the parent

I dont know the best way to do this either but Im storing `WindowHandle` (that has `HWND` for Windows) directly in `WindowLevel` variants.

druid-shell/src/backend/windows/window.rs
GeorgOhneH: `WS_CHILD` has many restrictions, which are not really needed. For example I don't think a modal window should be confined by its parent.

But you are right `WS_POPUPWINDOW` is also not the right choice here. We should properly use something like this `WS_OVERLAPPED | WS_CAPTION | WS_SYSMENU | WS_THICKFRAME`  and also add an option to enable the minimize and maximize buttons.

sjoshid: Hmm....I have a different image of modals in my head then. I _thinnnnnkkk_ maybe @cmyr can help us here?? He said he was busy with other things on Zulip but I hope he sees this and nudges us in the right direction.

JAicewizard:
huh, this is how I hacked the GTK4 popups as well, just use all application windows as "parents" basically. I have code that changes the `WindowLevel` API to include a parent window for all non-main windows. Collin said he'd look more into that soon ("... soon" to be exact ;) ). We could merge this before that, or wait in case that is accepted and properly keep track of parents that way.

sjoshid:
@GeorgOhneH 
I went through your impl to understand your scenario better. I came up a different approach but still satisfies your case. 

How about we add this little piece of code to `AppState::build_native_window()`?

```
let result = builder.build(); 

if let Some(WindowLevel::Modal) = config.level {
    let h = something.unwrap(); //error check
    for win in self.inner.borrow_mut().windows.iter_mut() {
        win.handle.disable_if_not_same(&h); // <=== new method
    }
    Ok(h)
} else {
    result 
}
```

The main reason to come up with this is so you dont have to maintain `Vec<HWND>` anywhere. This info is already available to us. You need to do something similar to enable windows when modal is destroyed. 

Jaap and others (in zulip) have been talking about changing `WindowLevel` API to contain `WindowHandle`. And when (if :( ) that happens, you can easily use `WindowHandle` to get foreground window. So no need for `GetForegroundWindow` call. 

Let me know what you think.

cmyr:
At a high level, I'm worried that we're doing a bunch of hacking to get various child window features working, without having a really coherent idea of our goals and how best to achieve them in a consistent way.

There *is* a use case for a window that steals focus and is modal to the entire application. There is *also* a use case for a window that is modal to a single explicit parent window, and then there is a use-case for windows like drop-down menus that i also think of as modal, although they're slightly different.

What I would like to see is a unified design that can handle all of these cases in a reasonably principled way. I have sketched this out [very roughly here](https://gist.github.com/cmyr/df7545ec6598c72632fb6e7efa1a7154) but I haven't investigated the various platforms in depth, so it's very possible that I'm overlooking important things.


In any case: I'm reluctant to bring in new patches that are trying to solve one particular case, instead of trying to address the problem in a big picture way.

GeorgOhneH:
Makes sense. I guess I will close this PR

PR #1885, opened 2021-07-27
Remove Default impl for FlexParams

### LINKED ISSUE #1708: implementation of Default for FlexParams is invalid
#1691 added a check to prevent using a flex factor less than or equal to 0.0, but `FlexParams` has `#[derive(Default)]` which allows bypassing the check with `FlexParams::default()`.

Should we remove the implemenation or does it make sense to use a value like 1.0?

Maan2003:
Fixes #1708


cmyr (approved):
Thanks!

PR #1893, opened 2021-08-04
Fix word_range_for_pos

### LINKED ISSUE #1892: Textbox doesn't handle double-clicks properly
**Steps to reproduce:**

- Run the textbox example
- Write some text.
- Double-click on a word you wrote.

The expected behavior is that the entire word is selected. Instead it just acts as a simple click.

**Version:** 0.7.0 (commit dc17683deeb909a4ef20fb3174b1e3c1480dfb82)

Maan2003:
earlier for the last word pos..pos was returned
because peek will be None and unwraped to false

Also added some tests.

Fixes #1892


PoignardAzur (approved):
LGTM.

PoignardAzur:
It works, word selection now works as expected.

PR #1896, opened 2021-08-05
Rename variables named `inner` for clarity.

### LINKED ISSUE #269: standardize child/inner naming in widgets
@zaynetro was asking about this earlier: if a widget contains another widget, what should the name of the variable pointing to the contained widget be? Currently we use `inner` in some places and `child` in other places.

I think we should pick one and stick with it; I personally vote for `child`, which I find slightly more descriptive.

PoignardAzur:
In most cases, rename `inner` to `child` for consistency with existing 
code.
Rename `Inner` to `InnerAppState`.

Fixes #269

### DIFF
--- a/druid/src/tests/harness.rs
+++ b/druid/src/tests/harness.rs
@@ -47,13 +47,13 @@ pub(crate) const DEFAULT_SIZE: Size = Size::new(400., 400.);
 /// Also, timers don't work.  ¯\_(ツ)_/¯
 pub struct Harness<'a, T> {
     piet: Piet<'a>,
-    inner: Inner<T>,
+    mock_app: MockAppState<T>,
     window_size: Size,
 }
 
 /// All of the state except for the `Piet` (render context). We need to pass
 /// that in to get around some lifetime issues.
-struct Inner<T> {
+struct MockAppState<T> {
     data: T,
     env: Env,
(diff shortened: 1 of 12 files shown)


druid/src/tests/harness.rs
```diff
@@ -45,15 +45,15 @@ pub(crate) const DEFAULT_SIZE: Size = Size::new(400., 400.);
 /// if you want those functions run you will need to call them yourself.
 ///
 /// Also, timers don't work.  ¯\_(ツ)_/¯
-pub struct Harness<'a, T> {
+pub struct Harness<'a, T: Data> {
```
cmyr, line 48: what's the motivation for the `Data` bound?

cmyr (approved):
great, thanks! one small question, otherwise happy to see this go in.

druid/src/tests/harness.rs
PoignardAzur: Oh yeah, that's no really related to the rest of the PR.

Having the Data bound is useful to understand what `Harness` stores at a glance. It's not super important one way or another.

cmyr: I subscribe to the idea that bounds in struct declarations should be as minimal as possible, so I'd prefer to revert this.

PR #1898, opened 2021-08-05
Document `#[data(ignore)]` in book

### LINKED ISSUE #1362: document book: #[data(ignore)]
`#[data(ignore)]` has no documentation to be found and only 2 examples using it (one of which is not merged yet).

Maybe an idea is making a part in the book about `.update` with a section "when if .update called?" and reference `#[data(ignore)]` in there.

Once this is done the link in the cursor example should be updated so that it stays up-to-date.

PoignardAzur:
Fixes #1362


cmyr (approved):
thanks!

PR #1901, opened 2021-08-06
Update Lens documentation.

### LINKED ISSUE #431: Update lens module documentation to also document deriving Lens

PoignardAzur:
Update inline doc as well as the druid book.

Fixes #431


PoignardAzur:
Oh, this conflicts with the previous doc rewrite. Ugh. This isn't great.

cmyr (approved):
Thanks!

PR #1904, opened 2021-08-07
Add ui tests for Panoramix derive

### LINKED ISSUE #250: Add tests to Lens derive macro
Will be especially helpful for future expansions to Lens (#213).

PoignardAzur:
Fixes #250

### DIFF
--- a/druid-derive/tests/ui.rs
+++ b/druid-derive/tests/ui.rs
@@ -1,0 +1,21 @@
+// https://github.com/dtolnay/trybuild
+// trybuild is a crate that essentially runs cargo on the provided files, and checks the output.
+// Tests may suddenly fail after a new compiler release, and there's not much we can do about that.
+// If the test suite fails because of trybuild:
+// - Update your compiler to the latest stable version.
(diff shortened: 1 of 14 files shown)


druid-derive/tests/ui.rs
```diff
@@ -0,0 +1,14 @@
+use trybuild::TestCases;
```
cmyr, line 1: I would include a link to `trybuild` in the source here, and maybe a really quick summary of how it works. Basically: if some future update to rust brings slightly different error messages (breaking these tests) it would be nice if the person who came to look at this file to figure out the breakage had as easy a time as possible figuring out what they need to do.

cmyr (approved):
Nice, thanks.

CI failed on windows nightly, which seems to produce slightly different output. I'm not sure if that was some regression in a particular compiler version, or what? I've restarted CI and we'll see if it presists, otherwise we may need to go investigate upstream.

druid-derive/tests/ui.rs
PoignardAzur: Sure. Will look into it when I get back.

PR #1905, opened 2021-08-07
Update CONTRIBUTING.md

### LINKED ISSUE #1502: document steps involved in doing a release

PoignardAzur:
Fixes #1502

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -125,7 +125,7 @@ old versions of our sub-dependencies and so we shouldn't claim that the old vers
 
 An easy way to do this is to use the `cargo upgrade` tool available via [cargo-edit].
 
-```
+```sh
 cargo install cargo-edit
 ```
 
@@ -138,7 +138,7 @@ which are still [semver] compatible with what we have specified in our `Cargo.to
 
 If you just want to see what would happen you can add the `--dry-run` option.
 
-```
+```sh
 cargo update
 ```
 
@@ -148,7 +148,7 @@ It's crucial that we use `--to-lockfile` because without it `cargo upgrade` won'
 
 If you just want to see what would happen you can add the `--dry-run` option.
 
-```
+```sh
 cargo upgrade --workspace --to-lockfile
 ```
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -157,13 +157,44 @@ cargo upgrade --workspace --to-lockfile
...
+
+```sh
+cargo publish --manifest-path="druid-derive/Cargo.toml" --no-verify
+cargo publish --manifest-path="druid-shell/Cargo.toml"
+cargo publish --manifest-path="druid/Cargo.toml"
+```
+
+We need to run the commands in this order, because the druid crate depends on the two other crates. We use `--no-verify` for `druid-derive`, because the crate has a cyclic dependency on `druid` that might cause problem, since we haven't published the latest version of `druid` yet.
+
+<!--- TODO: check that the above is actually true. I'm not sure we really need `--no-verify` --->
```
cmyr, line 188: I think we do, since 'derive' depends on the version of 'druid' that we have not yet published.

cmyr (approved):
Thanks, one note on your comment.

CONTRIBUTING.md
PoignardAzur: The next person who publishes a new version needs to check.

I'm not sure you need `--no-verify`; because druid is only a dev-dependency, and I'm not sure the crate registry actually checks those.

Issue #1269, opened 2020-09-29
Rich text todos


cmyr:
Tracking issue for a few rough spots in #1255:

- layout may not be invalidated correctly if the env changes
- no functionality for editing the text; you have to recreate the whole object. I'm still not sure how this should interact with editing.
- default styles are confused; they currently live on the TextLayout, not on the RichText; this should probably change in some way, so that if you're using RichText we ignore anything set on the label itself.

Maan2003:
Edited to add a checkboxes

PoignardAzur:
I might work on this. I'm starting to look into widget styling anyway.

Maan2003:
I want to work of the first todo (invaliding the layout on env value change)

IWANABETHATGUY:
Any progress on this? I want to write a text editor using druid, I am wonder if druid could edit rich text for now.

PR #1907, opened 2021-08-09
RichText: Invalidate layout on Env change

Maan2003:
Related #1269

### DIFF
--- a/druid/src/text/attribute.rs
+++ b/druid/src/text/attribute.rs
@@ -190,6 +190,10 @@ impl AttributeSpans {
                 .fg_color
                 .iter()
                 .any(|span_attr| ctx.env_key_changed(&span_attr.attr))
+            || self
+                .font_descriptor
+                .iter()
+                .any(|span_attr| ctx.env_key_changed(&span_attr.attr))
     }
 }
 
--- a/druid/src/text/attribute.rs
+++ b/druid/src/text/attribute.rs
@@ -19,6 +19,8 @@ use std::ops::Range;
 use crate::piet::{Color, FontFamily, FontStyle, FontWeight, TextAttribute as PietAttr};
 use crate::{Command, Env, FontDescriptor, KeyOrValue};
 
+use super::EnvUpdateCtx;
+
 /// A clickable range of text with an associated [`Command`].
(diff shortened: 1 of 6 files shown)


druid/src/text/storage.rs
```diff
@@ -45,6 +54,24 @@ pub trait TextStorage: PietTextStorage + Data {
     }
 }
 
+/// Provides information about keys change for more fine grained invalidation
+pub struct EnvUpdateCtx<'a, 'b>(&'a UpdateCtx<'a, 'b>);
```
cmyr, line 58: Is there a reason not to just put these methods on `UpdateCtx` directly?

druid/src/text/attribute.rs
```diff
@@ -179,6 +181,16 @@ impl AttributeSpans {
...
     }
+
+    pub(crate) fn env_update(&self, ctx: &EnvUpdateCtx) -> bool {
+        self.size
+            .iter()
+            .any(|span_attr| ctx.env_key_changed(&span_attr.attr))
+            || self
+                .fg_color
+                .iter()
+                .any(|span_attr| ctx.env_key_changed(&span_attr.attr))
```
cmyr, line 192: do we also want to do `font_descriptor`?

cmyr (approved):
This looks straightforward, thanks. Not sure if we need a new type, but maybe there's a good reason I'm overlooking?

druid/src/text/storage.rs
Maan2003: Because `UpdateCtx` has lot of methods like `set_cursor` we don't want `TextStorage` to use.

druid/src/text/attribute.rs
Maan2003: missed that one

druid/src/text/storage.rs
cmyr: Okay. the `storage` mod isn't public, so you might want to reexport this from `druid::text`?

Maan2003: it is reexported already

cmyr: My work here is done 🤝

PR #1908, opened 2021-08-10
Make TextBox placeholder translatable

Swatinem:
This makes the `TextBox` placeholder a `LabelText` which can be initialized with a `LocalizedString` and thus translated.

The problem that I have is that the `TextBox` seems to be bound to a very specific editable text via `lens`, and that couples the `LabelText` to the same Data, which means I currently can’t use anything else from my main Data as placeholders of the `LocalizedString` (or `Dynamic` for that matter).

### DIFF
--- a/druid/examples/textbox.rs
+++ b/druid/examples/textbox.rs
@@ -34,7 +34,7 @@ const EXPLAINER: &str = "\
     This example demonstrates some of the possible configurations \
     of the TextBox widget.\n\
     The top textbox allows a single line of input, with horizontal scrolling \
-    but no scrollbars. The bottom textbox allows mutliple lines of text, wrapping \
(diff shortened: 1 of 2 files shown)


Maan2003 (approved):
Looks good to me. Thanks!

Maan2003:
> The problem that I have is that the TextBox seems to be bound to a very specific editable text via lens, and that couples the LabelText to the same Data, which means I currently can’t use anything else from my main Data as placeholders of the LocalizedString (or Dynamic for that matter).

that is a limitation of Lens. I don't think it is easily to fix

Swatinem:
I noted down the limitations in the commit message, otherwise the code is unchanged and was just rebased on master.

jneem:
This is good to go, right?

Swatinem:
@jneem yes, this should be good to go

PR #1910, opened 2021-08-10
User defined types in Env

Maan2003:


druid/src/env.rs
```diff
@@ -92,11 +92,8 @@ pub struct Key<T> {
     value_type: PhantomData<T>,
 }
 
-// we could do some serious deriving here: the set of types that can be stored
-// could be defined per-app
-// Also consider Box<Any> (though this would also impact debug).
 /// A dynamic type representing all values that can be stored in an environment.
-#[derive(Clone, Data, PartialEq)]
```
Maan2003, line 99: This removes `PartialEq`. it is not straight forward to implement with `Arc<dyn Any>`. I think Data should be sufficient

cmyr (approved):
Hmm okay, this does seem quite straightforward, let's give it a shot?


I am curious about the use-case, though!

Maan2003:
I want to store `Sender<Box<dyn Future>>` in `Env`. So I can spawn futures from any widget.

PR #1914, opened 2021-08-12
Fix example open_save.rs

minimal-state:
- The 'Open' button should use the open panel, not the save panel. #1914


jneem (approved):
Thanks!

Issue #1824, opened 2021-06-12
Do we need `WindowHandle::set_level`?


jneem:
Is it ever necessary to change the window level after it's already created? For windows that aren't yet created, we have `WindowBuilder::set_level`. There was some concern in #1785 whether `WindowHandle::set_level` will work on x11, but I don't know about any other platforms...

Maan2003:
`WindowHandle::set_level` will be useful for [WindowLevel::AlwaysOnTop](https://github.com/linebender/druid/pull/1818)

psychon:
Sounds like "always on top should not be a window level", but rather another property that can be controlled separately. And in window builder it has three states, "on", "off", "default", where "default" depends on the window level and implements the current behaviour.
(Does anyone/anything really need this flexibility? Am I over-engineering?)

jneem:
That sounds reasonable. But in addition, I think that always-on-top is a property that potentially makes sense to change after a window is displayed. Whereas (at least on windows and gtk/x11) the tooltip/dialog/whatever hints are things that can only be set once when the window is created.

JAicewizard:
In order to work well with GTK4/wayland the window level needs to be known at creation time.

I dont see a use of changing it after the fact anyways, why would you repurpose a window instead of creating a new one?
Or _when_ do you want to repurpose an existing window, without changing the child widgets? sounds very useless to me.

JAicewizard:
> Whereas (at least on windows and gtk/x11) the tooltip/dialog/whatever

Yes but that is diferent from always on top. Any window can be always on top in GTK(3/4), and this can be changed at any time using `gtk_window_set_transient_for` and `gtk_window_set_modal`.

psychon:
> Any window can be always on top in GTK(3/4), and this can be changed at any time using gtk_window_set_transient_for and gtk_window_set_modal.

Uhm. Correct me if I am wrong, but neither of these functions is "always on top". The first one is for "window `window` is a dialog for window `parent`" (and you need to provide a parent window). The second one marks a window as, well, modal, which means that this dialog must be closed before the user can continue working on the main window.

From the docs of `gtk_window_set_modal` (emphasis mine):

> **most** window managers will then disallow lowering the dialog below the parent.

This does not even talk about "always on top", at least not in the way I understood it.

I'd recommend `gtk_window_set_keep_above()` instead, but I also want to quote the docs on that one in the hope that something like this also ends up in druid's docs:

[...]

Issue #1757, opened 2021-05-03
calling `WindowDesc::set_position` has no effect on GTK when run on wayland


arifd:
The window always opens in the same part (close to top left) of my monitor every time, no matter what values for `Point` i feed it.

This is likely also the reason that `Dropdown` from the widget-nursery is broken, as they too exhibit the same behavior and use `.set_position()` to place their dropdown menu.

jneem:
Did this ever work on your system, or is it a regression?

arifd:
First time using it, so I can't tell you.

My git-fu is not very strong, but my computer is reasonably enough, and I think with git, you can actually go into the past and compile for previous versions, so if you can tell me what to throw into the terminal, I'll be happy to do any testing / answer this question for you / etc

jneem:
So I'd suggest working with the druid-widget-nursery repository first (and then we can try to nail down a druid commit later). I ran `git log src/dropdown.rs` to find that the first commit with dropdown support is `6c3e3512`. So first, checkout that commit to see if it worked then:

```sh
$ git checkout 6c3e3512
$ cargo run --example dropdown
```

If that one works, then you can use `git bisect` to find where it broke.

```sh
$ git bisect start
$ git bisect bad master # because the current version is broken
$ git bisect good 6c3e3512 # because the old version worked
```

Then git will automatically check out a commit for you. You test that commit, and type either `git bisect bad` or `git bisect good` depending on whether it worked or not. (If you're not sure, for example because that commit failed to compile, do `git bisect skip` instead.)

arifd:
Hi @jneem, thanks for that, git continues to impress me.

Well, I can confirm that
```
$ git checkout 6c3e3512
$ cargo run --example dropdown
```
produces a broken example also, although the behaviour is ever so slightly different.

- In `git checkout master`
    - If I never move the parent window
         - the menu appears to open inside the window (don't know if my design or by coincidence (- looks like correct behaviour)), but since clicking anywhere else (including the parent) doesn't close the dropdown menu, i can click the button to open up more menus and they tile one after the other.
    - If I move the parent
        - all dropdown menus open at that familiar top left coordinate
     
- In `git checkout 6c3e3512`
     - both moving the parent or not, dropdowns always exhibit the same behaviour, that is, they open up (and tile) at that top left position

jneem:
Ok, thanks for checking. One more thing: apparently Ubuntu 21.04 uses wayland by default; does anything change if you switch to x11?

[...]

PR #1919, opened 2021-08-14
make most window types have a child window

JAicewizard:
This does a couple of things.
It adds the notion of parent and child windows. all non-main windows have a parent window.
All non-main windows have their coordinate space shifted, so that the origin is the location of the main window.
This is usefull in almost all usecases.

This also removes the option to change the level after the fact. At least the GTK backend doesnt support this, and it would just be extra work to move all the widgets to the new window. I also dont think there is any usecase for this.

This solves positioning not working in wayland (#1757) and resolves the discussion over at #1824.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,8 @@ You can find its changes [documented below](#070---2021-01-01).
 - Move macOS only function to Mac extension trait ([#1863] by [@Maan2003])
 - x11: Only query atoms once instead of per window ([#1865] by [@psychon])
 - remove prefix from platform extension traits ([#1873] by [@Maan2003])
+- Remove `set_level` on windows ([#1919] by [@Maan2003])
+- Add parent windows to non-main windows. (Coordinate space is now from their origin) ([#1919] by [@Maan2003])
 
 ### Deprecated
 
@@ -775,6 +777,7 @@ Last release without a changelog :(
 [#1884]: https://github.com/linebender/druid/pull/1884
 [#1885]: https://github.com/linebender/druid/pull/1885
 [#1886]: https://github.com/linebender/druid/pull/1886
+[#1919]: https://github.com/linebender/druid/pull/1919
 
 [Unreleased]: https://github.com/linebender/druid/compare/v0.7.0...master
 [0.7.0]: https://github.com/linebender/druid/compare/v0.6.0...v0.7.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,8 +74,8 @@ You can find its changes [documented below](#070---2021-01-01).
 - Move macOS only function to Mac extension trait ([#1863] by [@Maan2003])
 - x11: Only query atoms once instead of per window ([#1865] by [@psychon])
 - remove prefix from platform extension traits ([#1873] by [@Maan2003])
-- Remove `set_level` on windows ([#1919] by [@Maan2003])
-- Add parent windows to non-main windows. (Coordinate space is now from their origin) ([#1919] by [@Maan2003])
+- Remove `set_level` on windows ([#1919] by [@JAicewizard])
+- Add parent windows to non-main windows. (Coordinate space is now from their origin) ([#1919] by [@JAicewizard])
 
 ### Deprecated
 
--- a/druid-shell/src/backend/mac/window.rs
+++ b/druid-shell/src/backend/mac/window.rs
@@ -1216,6 +1216,9 @@ impl WindowHandle {
 
     // Need to translate mac y coords, as they start from bottom left
     pub fn set_position(&self, mut position: Point) {
+        // TODO: Maybe collin can get this into a state where modal windows follow the parent?
+        // There is an API to do child windows, (https://developer.apple.com/documentation/appkit/nswindow/1419152-addchildwindow)
+        // but I have no good way of testing and making sure this works.
         if let Some(parent_state) = &self.parent {
             let pos = (*parent_state).get_position();
             position -= (pos.x, pos.y)
--- a/druid-shell/src/backend/mac/window.rs
+++ b/druid-shell/src/backend/mac/window.rs
@@ -89,9 +89,9 @@ mod levels {
         use WindowLevel::*;
         match window_level {
             AppWindow => NSNormalWindowLevel,
-            Tooltip => NSFloatingWindowLevel,
-            DropDown => NSFloatingWindowLevel,
-            Modal => NSModalPanelWindowLevel,
+            Tooltip(_) => NSFloatingWindowLevel,
+            DropDown(_) => NSFloatingWindowLevel,
+            Modal(_) => NSModalPanelWindowLevel,
         }
     }
 }
@@ -102,6 +102,15 @@ pub(crate) struct WindowHandle {
     /// a view. Also, this is better for hosted applications such as VST.
     nsview: WeakPtr,
     idle_queue: Weak<Mutex<Vec<IdleKind>>>,
+    parent: Option<Box<crate::WindowHandle>>,
+}
+
+impl std::fmt::Debug for WindowHandle {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        f.write_str("WindowHandle{\n")?;
+        f.write_str("}")?;
+        Ok(())
+    }
 }
 
 impl Default for WindowHandle {
@@ -109,6 +118,7 @@ impl Default for WindowHandle {
         WindowHandle {
             nsview: unsafe { WeakPtr::new(nil) },
             idle_queue: Default::default(),
+            parent: None,
         }
     }
 }
@@ -283,6 +293,7 @@ impl WindowBuilder {
             let mut handle = WindowHandle {
                 nsview: view_state.nsview.clone(),
                 idle_queue,
+                parent: None,
             };
 
             if let Some(window_state) = self.window_state {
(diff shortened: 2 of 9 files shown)


druid-shell/src/backend/mac/window.rs
```diff
@@ -102,13 +102,23 @@ pub(crate) struct WindowHandle {
     /// a view. Also, this is better for hosted applications such as VST.
     nsview: WeakPtr,
     idle_queue: Weak<Mutex<Vec<IdleKind>>>,
+    parent: Option<Box<crate::WindowHandle>>,
```
jneem, line 105: I think the `Box` is unnecessary.

jneem (approved):
Nice! I'd probably wait a bit to see if anyone else objects...

druid-shell/src/backend/mac/window.rs
JAicewizard: no because mac is different. It would be an infinitely big struct.

JAicewizard: All the others already have a reference in between the window handle and this option, mac doesnt.

Maan2003: why does mac need to store parent handle though?

JAicewizard: To change the coordinate space relative to the parent

JAicewizard:
Id actualy first like to do some discussion about popups, to popups also need to have a parent? and do they also need to be in the parents coordinate space? Thinking about this again it seems that popups maybe should have their own coordinate space.

JAicewizard:
nvm I forgot that we have `modal` as a 3rd window type, not `popup` :) Ignore my last message

druid-shell/src/backend/mac/window.rs
jneem: Oh, I see, it's because on mac you've put the parent directly in the `WindowHandle`. Why not put it in `ViewState`, so it's more similar to the other platforms?

JAicewizard: I dont exactly know how to get that from the `WindowHandle`

jneem: I can't test because I don't have a mac, but looking at `set_focused_text_field` gives a hint, I think:

```rust
// Here, self is a &WindowHandle
if let Some(view) = self.nsview.load().as_ref() {
    let state: *mut c_void = *view.get_ivar("viewState");
    let state = &mut (*(state as *mut ViewState));
}
```

JAicewizard: Oh yes I see, thanks! pushed the changes

CHANGELOG.md
```diff
@@ -74,6 +74,8 @@ You can find its changes [documented below](#070---2021-01-01).
 - Move macOS only function to Mac extension trait ([#1863] by [@Maan2003])
 - x11: Only query atoms once instead of per window ([#1865] by [@psychon])
 - remove prefix from platform extension traits ([#1873] by [@Maan2003])
+- Remove `set_level` on windows ([#1919] by [@Maan2003])
+- Add parent windows to non-main windows. (Coordinate space is now from their origin) ([#1919] by [@Maan2003])
```
Maan2003, line 78: heh, copy paste gone wrong
```suggestion
- Remove `set_level` on windows ([#1919] by [@JAicewizard])
- Add parent windows to non-main windows. (Coordinate space is now from their origin) ([#1919] by [@JAicewizard])
```

JAicewizard: :facepalm: typical me

Maan2003:
I believe x11 backend position translation is TODO?

JAicewizard:
X11 tooltips dont work, I could add the set_position conversion but it wouldn't do anything.

Maan2003:
no, I think they work now (at least the sub_window example does)

Maan2003:
uh oh time to debug refcells :pensive: .

I get and drag is working
image

JAicewizard:
For me it gets opened as a non-floating window, but that's fine-ish.

druid-shell/src/backend/mac/window.rs
```diff
@@ -1198,7 +1214,21 @@ impl WindowHandle {
     pub fn show_titlebar(&self, _show_titlebar: bool) {}
 
     // Need to translate mac y coords, as they start from bottom left
-    pub fn set_position(&self, position: Point) {
+    pub fn set_position(&self, mut position: Point) {
+        // TODO: Maybe collin can get this into a state where modal windows follow the parent?
+        // There is an API to do child windows, (https://developer.apple.com/documentation/appkit/nswindow/1419152-addchildwindow)
```
cmyr, line 1219: I'm not sure you can move the parent while a modal is presented?

druid-shell/src/backend/mac/window.rs
```diff
@@ -1198,7 +1214,21 @@ impl WindowHandle {
     pub fn show_titlebar(&self, _show_titlebar: bool) {}
 
     // Need to translate mac y coords, as they start from bottom left
-    pub fn set_position(&self, position: Point) {
+    pub fn set_position(&self, mut position: Point) {
+        // TODO: Maybe collin can get this into a state where modal windows follow the parent?
```
cmyr, line 1218:
```suggestion
        // TODO: Maybe @cmyr can get this into a state where modal windows follow the parent?
```

(@colin would also be fine, but this should be easier for people to understand)

cmyr (approved):
Okay as discussed I do have some longer-term ambitions here, but in the meantime this seems reasonable!

druid-shell/src/backend/mac/window.rs
JAicewizard: Tooltips and dropdowns suffer from the same issue, we expect them to be closed when moving a window but ATM that is upto the user to program. It might still be worth implementing for those? if not Ill remove the comment.

Maan2003: > ATM that is upto the user to program.

is it possible? We don't have any windowmove events

JAicewizard: uhh oh. If I am going to be implementing proper parent tracking, I can only do this for GTK and with some research probably X11. Others will have to help me on different platforms.

Maan2003: I think we should give windowmove events and then the user should do whatever they like (move with the window or close the popup)

JAicewizard: Ok, lets put that in another MR to not bloat this one.

jneem:
This can be merged, right? (modulo conflicts)

JAicewizard:
I fixed the conflicts, I think this should be ok to merge now.

CryZe:
This PR somehow killed DropDown entirely on Windows. It doesn't do anything anymore.

JAicewizard:
`WindowLevel::DropDown`? I didnt even change anything with those, the parent is completely ignored. Are you sure that it is this PR specifically?

CryZe:
Nvm, you are right. This was indeed already broken on master one commit before your PR got merged.

sjoshid:
Tooltip in sub-windows example definitely got worse in Windows. Reverting the latest commit on `sub_windows.rs` example does fix it, but maybe it breaks in Gtk?

druid/examples/sub_window.rs
```diff
@@ -145,11 +145,9 @@ impl<T, W: Widget<T>> Controller<T, W> for TooltipController {
                             WindowConfig::default()
                                 .show_titlebar(false)
                                 .window_size_policy(WindowSizePolicy::Content)
-                                .set_level(WindowLevel::Tooltip)
+                                .set_level(WindowLevel::Tooltip(ctx.window().clone()))
                                 .set_position(
-                                    ctx.window().get_position()
-                                        + window_pos.to_vec2()
-                                        + cursor_size.to_vec2(),
+                                    (window_pos.to_vec2() + cursor_size.to_vec2()).to_point(),
```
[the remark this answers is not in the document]
sjoshid: Apologies for being vague. Location of tooltip is way off target.
image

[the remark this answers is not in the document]
JAicewizard: Could it be some DPI scaling? I cant really test but ill check if I can spot something related to that.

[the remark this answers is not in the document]
sjoshid: The problem at least on Windows is that the cursor position received in `TooltipController` is in **window coords**. So we need to go from window coords to screen coords because screen coords is what Windows expect when you create a tooltip.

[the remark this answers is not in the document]
JAicewizard: Yes that conversion happens in win/window.rs. So that should be handled. Instead of adding the parent window size in the users code we do it in druid_shell.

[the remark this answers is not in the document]
sjoshid: Not sure if it's handled. Otherwise this should work??
Solution to this problem is simple
```
let tooltip_position_in_window_coordinates = (window_pos.to_vec2() + cursor_size.to_vec2()).to_point();
let final_position = self.window().get_position() + tooltip_position_in_window_coordinates .to_vec2();
```
Then pass `final_position` to sub window. This works for 100% scaling but not for other scaling factors. I was about to open a PR to fix all this. 

On a different topic, what should we do when we fix/enhance one platform but cant port changes to other platforms or test it? Ditto porting might not be even the right thing. I'll open topic in Zulip.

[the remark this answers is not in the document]
JAicewizard: No because those coordinates are in global space while they should be in window space. they are converted to global space by `druid_shell` not the client code. This was part of the change, so it should work. Maybe it cant get the parents state? that could be an issue.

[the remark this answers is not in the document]
sjoshid: The solution i gave in previous comment is basically the old code. That was dumb on my part. lol. The previous code didnt make much sense to me.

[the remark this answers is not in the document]
sjoshid: Where in shell is this conversion happening?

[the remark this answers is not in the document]
JAicewizard: https://github.com/linebender/druid/pull/1919/files#diff-1a1a9e4151b4aa7f5d4959110f78d6f99d7fa46ac42ce3d05ac2992ab3dfe6e9R1852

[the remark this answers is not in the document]
sjoshid: Check https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/Tooltip.20broken.20in.20Windows/near/254600089

PR #1923, opened 2021-08-16
Fix nightly compiler warnings

cmyr:
This fixes two small nightly warnings that were annoying me in CI.


druid/src/lib.rs
```diff
@@ -127,7 +127,7 @@
 //! [`image` crate]: https://crates.io/crates/image
 
 #![deny(
-    broken_intra_doc_links,
+    rustdoc::broken_intra_doc_links,
```
Maan2003, line 130: does not compile on rustc 1.51

cmyr: Is this an issue? It does work on 1.52, and that's already two versions behind current..

Maan2003: :shrug: I am not sure about bumping msrv just for this.

cmyr: no sweat, happy to just merge the other one

Maan2003 (approved):

PR #1929, opened 2021-08-20
WindowDesc::with_config and impl PartialEq for WindowConfig, WindowLevel

Maan2003:
needed for some experimentation in crochet. I am sure there can be more use cases and we can do this easily.


cmyr (approved):
Sounds good!

PR #1931, opened 2021-08-20
TextBox: handle SELECT_ALL 

### LINKED ISSUE #1854: `Ctrl+A` doesn't select all text in a `TextBox`
It seems like an implementation along those lines needs to be added, but I'm not 100% sure how to do this correctly. usize::MAX is certainly wrong and I'm not at all sure how IME works, which is the unused return type. Also I've seen that `EditSession` also just straight up has `do_action` which can handle `TextAction::SelectAll`, among all the others. So is there some duplication going on already with the other actions?!

Maan2003:
Fixes #1854

### DIFF
--- a/druid/src/widget/textbox.rs
+++ b/druid/src/widget/textbox.rs
@@ -363,7 +363,6 @@ impl<T: TextStorage + EditableText> TextBox<T> {
             key if HotKey::new(SysMods::Cmd, "y").matches(key) && cfg!(windows) => {
                 Some(sys::REDO.to(our_id))
             }
-            key if HotKey::new(SysMods::Cmd, "a").matches(key) => Some(sys::SELECT_ALL.to(our_id)),
             _ => None,
         }
     }
--- a/druid/src/widget/textbox.rs
+++ b/druid/src/widget/textbox.rs
@@ -363,6 +363,7 @@ impl<T: TextStorage + EditableText> TextBox<T> {
             key if HotKey::new(SysMods::Cmd, "y").matches(key) && cfg!(windows) => {
                 Some(sys::REDO.to(our_id))
             }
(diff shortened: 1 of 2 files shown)


druid/src/widget/textbox.rs
```diff
@@ -363,7 +363,6 @@ impl<T: TextStorage + EditableText> TextBox<T> {
             key if HotKey::new(SysMods::Cmd, "y").matches(key) && cfg!(windows) => {
                 Some(sys::REDO.to(our_id))
             }
-            key if HotKey::new(SysMods::Cmd, "a").matches(key) => Some(sys::SELECT_ALL.to(our_id)),
```
cmyr, line 366: hm, this *should* be working, so if it isn't it's probably that we just aren't handling this command? We'll need this anyway, since select-all can arrive from a menu (and hence a `Command`), and then it shouldn't be necessary to touch `simulate_input`.

cmyr (changes requested):
I don't think `simulate_input` is the right place to solve this, because iirc select_all is not necessarily something that is dispatched via IME.

To clarify: this is also currently broken on mac, even though mac does not use simulate input.

druid/src/widget/textbox.rs
Maan2003: yeah, we are not handling `SELECT_ALL`.

Maan2003:
Makes sense, I didn't have opinion on where this should be handled, just wanted this to be fixed :)

cmyr (approved):
Cool, this looks right. Oddly there's still an issue with the selection not causing a paint pass (even with a rebase), but we can worry about that afterwards. :)

Maan2003:
still having that issue :face_with_head_bandage:

PR #1937, opened 2021-08-25
Parameterize Container, SizedBox and Align on child type

PoignardAzur:
Add get_child method to all three types.

### DIFF
--- a/druid/src/widget/align.rs
+++ b/druid/src/widget/align.rs
@@ -20,65 +20,75 @@ use crate::{Data, Rect, Size, UnitPoint, WidgetPod};
 use tracing::{instrument, trace};
 
 /// A widget that aligns its child.
-pub struct Align<T> {
+pub struct Align<T, W> {
     align: UnitPoint,
-    child: WidgetPod<T, Box<dyn Widget<T>>>,
+    child: WidgetPod<T, W>,
     width_factor: Option<f64>,
     height_factor: Option<f64>,
 }
 
-impl<T> Align<T> {
+impl<T: Data, W: Widget<T>> Align<T, W> {
     /// Create widget with alignment.
     ///
     /// Note that the `align` parameter is specified as a `UnitPoint` in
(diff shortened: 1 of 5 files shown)


PoignardAzur:
Again, I'm not sure about this, discussion is welcome.

In particular, I have doubts about the semantics of [the `id()` method](https://github.com/linebender/druid/pull/1937/files#diff-ceed7bf9f161414c171139ac49a3e40c874757df2a869a48e97dcbee4c3550d8R205-R207), but I rewrote it as best as I could.

cmyr (approved):
Looks good, modulo the broken test. :)

PoignardAzur:
I'm still getting these test failures:

```
---- tests::layout_tests::flex_paint_rect_overflow stdout ----
thread 'tests::layout_tests::flex_paint_rect_overflow' panicked at 'assertion failed: `(left == right)`
  left: `(10.0, 10.0)`,
 right: `(-10.0, -10.0)`', druid/src/tests/layout_tests.rs:181:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- tests::layout_tests::simple_paint_rect stdout ----
thread 'tests::layout_tests::simple_paint_rect' panicked at 'assertion failed: `(left == right)`
  left: `(10.0, 10.0)`,
 right: `(10.0, -10.0)`', druid/src/tests/layout_tests.rs:108:9
```

The thing is, I'm not even sure whether I should even fix them, or just change the values in the test.

Best I can tell, these failures are due to me changing the way `SizedBox` stores its child, as a `WidgetPod` instead of a `Box<dyn Widget>`. But I'm not sure the values are wrong? I'm not sure what referential they're meant to be in, for starters. I'm going to try running examples to check what happens.

PoignardAzur:
I'm also getting `missed call to set_origin` warnings. Uuugh.

I'm closing this for now. I don't want to bother.

PR #1938, opened 2021-08-26
x11: Window Focus events

Maan2003:
related #167


druid-shell/src/backend/x11/window.rs
```diff
@@ -1142,6 +1144,14 @@ impl Window {
         Ok(())
     }
 
+    pub fn handle_enter_notify(&self) {
+        self.with_handler(|h| h.got_focus());
```
psychon, line 1148:
```rs
    /// Called when this window becomes the focused window.
    #[allow(unused_variables)]
    fn got_focus(&mut self) {}

    /// Called when this window stops being the focused window.
    #[allow(unused_variables)]
    fn lost_focus(&mut self) {}
```
Uhm... you are calling these when the mouse cursor enters/leaves the window, not when it is focused...?

Maan2003: oh so these are mouse cursor enter/leave events :facepalm: . What are the keyboard focus events then?

jneem:
This looks fine to me, but then I'm not the x11 expert... @psychon do you have time for another quick peek?

psychon:
Uhm... Feel free to merge this and see what breaks. I am not 100% satisfied, but I cannot come up with any scenario where this would break right now, so fine with me.

Hm... perhaps if you have some global hotkeys configured? E.g. Windows key does "something". That should cause some focus events with `mode` being equal to `Grab` or `Ungrab`. It might be best to ignore those, but dunno if that needs to be done now.

From https://www.x.org/releases/X11R7.6/doc/xproto/x11protocol.html#events:input_focus

----

When a keyboard grab activates (but before generating any actual KeyPress event that activates the grab), G is the grab-window for the grab, and F is the current focus:

- FocusIn and FocusOut events with mode Grab are generated (as for Normal above) as if the focus were to change from F to G.

When a keyboard grab deactivates (but after generating any actual KeyRelease event that deactivates the grab), G is the grab-window for the grab, and F is the current focus:

- FocusIn and FocusOut events with mode Ungrab are generated (as for Normal above) as if the focus were to change from G to F.

Maan2003:
I am not sure how to deal `mode = Grab, Ungrab`, but other window "grabbing the keyboard" still feels like a focus out to me

psychon:
I am using `i3`. When I press mod4+Shift+Space, the currently focused window becomes floating/not-floating. Running `/usr/bin/xev`, this key combination causes the following events right after each other:
```
FocusOut event, serial 35, synthetic NO, window 0x2a00001,
    mode NotifyUngrab, detail NotifyPointer

FocusIn event, serial 35, synthetic NO, window 0x2a00001,
    mode NotifyUngrab, detail NotifyAncestor
```
So, the window never "really" lost its focus. But yeah, that might not be a problem.

Hm.... when I open a new terminal (mod4+enter) while `xev` has the input focus, I get this:
```
FocusOut event, serial 35, synthetic NO, window 0x2e00001,
    mode NotifyGrab, detail NotifyAncestor

FocusOut event, serial 35, synthetic NO, window 0x2e00001,
    mode NotifyWhileGrabbed, detail NotifyNonlinear

FocusOut event, serial 35, synthetic NO, window 0x2e00001,
    mode NotifyUngrab, detail NotifyPointer

```
The above would cause the code from this PR to call `lost_focus()` three times in a row without any `got_focus()` in between. Could that be a problem?

Anyway, as I wrote above: Feel free to just merge this and see what comes. It should be good enough for now.

jneem (approved):

jneem:
@Maan2003 can we merge this? It seems like an improvement over the current state of things.

Maan2003:
Yes, lets merge this. We can deal with edges cases later if needed.

PR #1940, opened 2021-08-27
impl Clone for LabelText

Maan2003:
need this for [dropdown in nursery](https://github.com/linebender/druid-widget-nursery/blob/master/src/dropdown_select.rs#L61-L65)


cmyr (approved):
Makes sense!

PR #1946, opened 2021-08-30
Check if gtk is initialised in get_monitors before assuming so in `get_monitors`

### LINKED ISSUE #1800: Question: centering window and light theme
Hello,

Thank you for `druid`.

Is this a proper way to center a window?
```rust
use druid::widget::Label;
use druid::{AppLauncher, PlatformError, Point, Screen, Size, Widget, WindowDesc};

fn build_ui() -> impl Widget<()> {
    Label::new("")
}

    let window_size = Size {
        width: 500.0,
        height: 100.0,
    };

    if let Some(primary_display) = Screen::get_monitors()
        .iter()
        .find(|monitor| monitor.is_primary())
    {
        let main_window = WindowDesc::new(build_ui())
            .set_position(Point {
                x: primary_display.virtual_work_rect().center().x - window_size.width / 2.0,
                y: primary_display.virtual_work_rect().center().y - window_size.height / 2.0,
            })
            .window_size(window_size);

            .launch(())
            .expect("failed to launch");
    }

    Ok(())
}
```

And is there any way to set a light theme without setting each color?
By default it is a dark theme: https://github.com/linebender/druid/blob/99a8d6638833752a1a0f343af31f0da9804a1407/druid/src/theme.rs#L120-L185

### LINKED ISSUE #1941: Panic on Screen::get_monitors()
I recently started learning this GUI toolkit, amazing work! I've wanted to implement centering of the window on startup using the code from this issue: https://github.com/linebender/druid/issues/1800
But I'm getting a panic while using druid::Screen::get_monitors():
`GDK has not been initialized. Call gdk::init or gtk::init.`

Using git version of Druid.

I'm running Manjaro, I have gtk3 and gtk4 packages installed. What I'm missing?

JAicewizard:
fixes #1941

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,6 +98,7 @@ You can find its changes [documented below](#070---2021-01-01).
 - X11 backend now uses the platform locale ([#1756] by [@Maan2003])
 - `Either` and `Tab` widgets were still propagating events to hidden widgets ([#1860] by [@lisael])
 - RichText: Invalidate layout on Env change ([#1907] by [@Maan2003])
+- GTK: fix using gdk before initialising it ([#1946] by [@JAicewizard])
 
 ### Visual
 
@@ -782,6 +783,7 @@ Last release without a changelog :(
 [#1885]: https://github.com/linebender/druid/pull/1885
 [#1886]: https://github.com/linebender/druid/pull/1886
 [#1907]: https://github.com/linebender/druid/pull/1907
+[#1947]: https://github.com/linebender/druid/pull/1947
 
 [Unreleased]: https://github.com/linebender/druid/compare/v0.7.0...master
 [0.7.0]: https://github.com/linebender/druid/compare/v0.6.0...v0.7.0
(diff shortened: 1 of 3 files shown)


jneem:
Do you know why the CI is failing? Should it be `gtk::init`?

JAicewizard:
uhh, im not sure how that slipped past me. no its supposed to be GDK, but I needed to scope it first. Maybe a rebase mistake.

jneem:
Ok, because I *think* `gdk::init` exits on failure, whereas `gtk::init` doesn't. (I'm not sure, because the rust bindings don't seem to include `gdk_check_init`, and the docs for `gdk::init` are empty...). I suspect we don't want to exit on failure here, and even if we do, I think we need a better error message before exiting.

JAicewizard:
Yes `gtk::init` does not terminate the program on failure, instead it falls back to a textual interface setup presumable without GDK.

If we want to use `gdk::DisplayManager` we need to init gdk, if thats indirectly via gtk or not we somehow need to init gdk. Using the `gtk::init` might succeed in that call specifically (with an error), but later it will crash when trying to use `gdk::DisplayManager` anyways.

jneem:
I mean if `gtk::init` gives an error, then we can just print an informative error message and return `Vec::new()`, right? There's no need to create the `gdk::DisplayManager` if we know that it will crash...

jneem (approved):

jneem:
Oh, I think this fixes #1941, right, not #1800.

JAicewizard:
I am such a messy person sometimes, yeah you're right.

PR #1947, opened 2021-08-31
x11: invalidate_rect expand the rect properly

Maan2003:
the select example in nursery: 
### before
before

### after 
after

### DIFF
--- a/druid-shell/src/backend/x11/window.rs
+++ b/druid-shell/src/backend/x11/window.rs
@@ -909,8 +909,8 @@ impl Window {
     }
 
     fn add_invalid_rect(&self, rect: Rect) -> Result<(), Error> {
-        // expanding not needed here, because we are expanding at every use of invalid
-        borrow_mut!(self.invalid)?.add_rect(rect);
+        let scale = self.scale.get();
+        borrow_mut!(self.invalid)?.add_rect(rect.to_px(scale).expand().to_dp(scale));
         Ok(())
     }
 
--- a/druid-shell/src/backend/x11/window.rs
+++ b/druid-shell/src/backend/x11/window.rs
@@ -768,7 +768,7 @@ impl Window {
             let cairo_ctx = cairo::Context::new(&surface).unwrap();
             let scale = self.scale.get();
             for rect in invalid.rects() {
-                let rect = rect.to_px(scale);
+                let rect = rect.to_px(scale).round();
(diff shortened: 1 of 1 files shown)


druid-shell/src/backend/x11/window.rs
```diff
@@ -909,8 +909,8 @@ impl Window {
     }
 
     fn add_invalid_rect(&self, rect: Rect) -> Result<(), Error> {
-        // expanding not needed here, because we are expanding at every use of invalid
-        borrow_mut!(self.invalid)?.add_rect(rect);
+        let scale = self.scale.get();
+        borrow_mut!(self.invalid)?.add_rect(rect.to_px(scale).expand().to_dp(scale));
```
jneem, line 913: So clearly the comment was not correct (maybe the missing expand was the clipping stuff at the beginning of `render`?), but I'm not sure if this fix is correct either, because it might end up double-expanding in some situations.

jneem (changes requested):
So clearly the old comment was not correct,

druid-shell/src/backend/x11/window.rs
Maan2003: yeah the comment wasn't correct. there should be expand at https://github.com/linebender/druid/blob/fdc7562483188210971e13d71b1d400885ad9af6/druid-shell/src/backend/x11/window.rs#L770-L773 but still we will pass unexpanded invalid to druid https://github.com/linebender/druid/blob/fdc7562483188210971e13d71b1d400885ad9af6/druid-shell/src/backend/x11/window.rs#L786

now add_invalid_rect and render is same as gtk backend. 

> double-expanding in some situations

example?

jneem: There's an `expand` in `PresentData::present`, so with this change, the rectangle that gets presented gets converted to px, expanded, converted back to dp, converted back to px, and expanded again. Since there might be numerical error in the conversion-to-dp-and-back-to-px, the final presented rectangle could be larger than the one used while clipping the cairo context.

Maan2003: because before conversion-to-dp-and-back-to-px, it was aligned to px, `round` should fix the numerical errors. This is done by windows backend

jneem: Yup, as long as there's only one `expand` and the rest are `round`s, it should be fine.

jneem (approved):

PR #1952, opened 2021-09-02
Expand `log_to_console` doc comment

zhiburt:
Hey there,
I am not sure if it's any usefull but there it is.
You can close it if you take it to be useless.

Thanks for the druid.

### DIFF
--- a/druid/src/app.rs
+++ b/druid/src/app.rs
@@ -183,8 +183,8 @@ impl<T: Data> AppLauncher<T> {
     ///
     /// # Panics
     ///
-    /// Panics if the subscriber fails to initialize.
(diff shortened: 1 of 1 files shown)


PoignardAzur:
Usefulness aside, you added another description without removing the existing one.

This renders to:

```
Panics if the subscriber fails to initialize. It will raise a panic if a `tracing`/`tracing_wasm` global logger was already set.
```

Something like this might be better:

```
Panics if the subscriber fails to initialize; for instance, if a `tracing`/`tracing_wasm` global logger was already set.
```

(though I'm not sure we ought to be documenting tracing's panic causes)

jneem (approved):

PR #1953, opened 2021-09-03
Added strikethrough

jenra-uwu:
This pull request adds strikethrough to druid and updates the markdown example to use `~~this syntax~~` for ~~strikethrough~~.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -18,3 +18,4 @@ Maximilian Köstler
 Bruno Dupuis
 Christopher Noel Hesse
 Marcin Zając
+Laura Gallo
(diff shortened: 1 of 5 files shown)


Maan2003 (approved):
Thanks!

jenra-uwu:
> Thanks! Probably a good idea to add this to example as well.

Which example is that?

Maan2003:
markdown_preview.rs

cmyr:
Worth noting that I think this may only work on certain platforms, maybe even only linux? I don't recall exactly the state of this in piet.

jenra-uwu:
> Worth noting that I think this may only work on certain platforms, maybe even only linux? I don't recall exactly the state of this in piet.

Oh no. I have no way to test it on Windows/macOS and don't know how to test it on the web. I may send a pull request for piet if thats the case, since I really want strikethrough for my matrix client. Any idea how I can start tackling this?

cmyr:
Here's the tracking issue, without much info. It may be that this is now only missing on mac? in any case I'll post some more info here: https://github.com/linebender/piet/issues/301

jneem:
This can be merged, right?

jneem:
I'm going to take that as a "yes"

PR #1955, opened 2021-09-05
ExtEventSink::add_idle_callback

Maan2003:
Sometimes using commands to mutate data from ExtEventSink can be complicated and verbose.
This provides a simpler alternative.

### DIFF
--- a/druid/examples/async_event.rs
+++ b/druid/examples/async_event.rs
@@ -25,13 +25,8 @@ use instant::Instant;
 use std::thread;
 use std::time::Duration;
 
-use druid::widget::prelude::*;
-use druid::{AppLauncher, Color, Selector, Target, WidgetExt, WindowDesc};
-
-// If you want to submit commands to an event sink you have to give it some kind
-// of ID. The selector is that, it also assures the accompanying data-type is correct.
-// look at the docs for `Selector` for more detail.
-const SET_COLOR: Selector<Color> = Selector::new("event-example.set-color");
+use druid::widget::Painter;
+use druid::{AppLauncher, Color, RenderContext, Widget, WidgetExt, WindowDesc};
 
 pub fn main() {
     let window = WindowDesc::new(make_ui()).title("External Event Demo");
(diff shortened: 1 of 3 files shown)


druid/examples/async_event.rs
```diff
@@ -75,62 +70,22 @@ fn generate_colors(event_sink: druid::ExtEventSink) {
...
-        if event_sink
-            .submit_command(SET_COLOR, color.clone(), Target::Auto)
-            .is_err()
-        {
-            break;
-        }
+        let color_clone = color.clone();
+        // schedule idle callback to change the data
+        event_sink.add_idle_callback(move |data: &mut Color| {
+            *data = color_clone;
```
jneem, line 76: I think this is totally the right way to do this, but maybe the previous code had some pedagogical value? Sometimes you really do want to send a command instead of mutate the data...

druid/src/ext_event.rs
```diff
@@ -117,6 +118,25 @@ impl ExtEventSink {
         ));
         Ok(())
     }
+
+    /// Schedule an idle callback.
```
jneem, line 122: Could you add a few more details here? Like:
- the data must match the root data
- it will get run on the main thread

Maybe it's also worth copying or linking other details from the `druid_shell::IdleHandle` docs? I'd lean towards copying instead of linking, because `IdleHandle` seems to be mostly intended for druid itself, and not for druid users.

druid/src/ext_event.rs
```diff
@@ -117,6 +118,25 @@ impl ExtEventSink {
...
+    pub fn add_idle_callback<T: 'static + Data>(&self, cb: impl FnOnce(&mut T) + Send + 'static) {
+        let mut handle = self.handle.lock().unwrap();
+        if let Some(handle) = handle.as_mut() {
+            handle.add_idle(|win_handler| {
+                let win_handler = win_handler
+                    .as_any()
+                    .downcast_mut::<DruidHandler<T>>()
+                    .unwrap_or_else(|| {
+                        panic!(
```
jneem, line 131: Maybe `debug_panic!`?

jneem (changes requested):
This seems like a very sensible API to me.

druid/examples/async_event.rs
Maan2003: I feel blocking_function example is sufficient for that :man_shrugging:

druid/src/ext_event.rs
Maan2003: yeah makes sense to skip running the callback and don't panic the UI thread

jneem:
Thanks! CI failure seems spurious...

jneem (approved):

PR #1957, opened 2021-09-08
Update lens.ms -- fix typo

Jerboas86:
vaid <--> valid

### DIFF
--- a/docs/src/lens.md
+++ b/docs/src/lens.md
@@ -48,7 +48,7 @@ impl Lens<Container, String> for InnerLens {
 }
 ```
 
-This is a very simple case. All we need to do is project the function onto the field. Notice that this isn't the only vaid lens from `Container` to `String` we could have made - we could also project from `Container` to `another`. We made the choice how to transform `Container` into `String` when we implemented `Lens`.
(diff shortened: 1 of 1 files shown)


docs/src/lens.md
```diff
@@ -137,7 +137,7 @@ impl Lens<Container2, Name> for NameLens {
 }
 ```
 
-Now in addition to almost free `Clone`s, we also have cheap incremental updates to the data itself. In the case of names, this isn't that important, but if the vector had `1_000_000_000` elements, we could still make changes in only *O(log(n))* time (in this case the difference between `1_000_000_000` and `30` - pretty big!).
+Now in addition to almost free `Clone`s, we also have cheap incremental updates to the data itself. In the case of names, this isn't that important, but if the vector had `1_000_000_000` elements, we could still make changes in only *O(log(n))* time (in this case the difference between `1_000_000_000` and `9` - pretty big!).
```
Maan2003, line 140: mostly, we use log base 2? in any case, this should be a separate PR

Maan2003: some discussion: https://github.com/linebender/druid/pull/1444#discussion_r535253139

Jerboas86: Close the PR since everything is redundant with #1444.
Sorry for the inconvenience.

Maan2003: typo `valid` needs to be fixed

Jerboas86: I removed the last commit to keep only the relevant typo.

Maan2003 (approved):
Thanks!

PR #1960, opened 2021-09-11
Allow focus changes from Update

xarvic:
This PR adds the methods needed to change focus to `UpdateCtx`.


jneem:
What's the status of this? I don't see anything wrong with it, but was it discussed somewhere?

xStrom:
What's the status of this PR? @xarvic do you think we should still move forward with this?

If we should move forward, then please rebase on `master` and I will try to figure out if this change makes sense.
If you no longer think this is relevant then we can just close this PR.

PR #1961, opened 2021-09-11
Fix custom_widget example bounded logic, fixes #1959

### LINKED ISSUE #1959: Bounded/unbounded logic in custom_widget example makes no sense
I think the check should rather be `!bc.is_width_bounded() | !bc.is_height_bounded()`.

https://github.com/linebender/druid/blob/ebe2fb71ffb1d719e4d9efc1f67484717d942161/druid/examples/custom_widget.rs#L55-L72

timakro:


Maan2003 (approved):
Thanks!

PR #1964, opened 2021-09-11
svg link crashes if id isn't found.

### LINKED ISSUE #1963: svg links crash when link isn't found.
[this unwrap can cause crashes](https://github.com/linebender/druid/blob/master/druid/src/widget/svg.rs#L376)

I've attempted a fix using a null brush but complains about typing mismatch
```
usvg::Paint::Link(id) => match self.defs.find(id) {
    Some(v) => v,
    None => {
        trace!("svg link id requested but not found: {:?}", id);
        Rc::new(piet::NullBrush{})
    }
},
```

```
error[E0308]: mismatched types
   --> /home/.cargo/git/checkouts/druid-f71533f3d81c0bc8/393774a/druid/src/widget/svg.rs:381:33
    |
381 |                         Rc::new(piet::NullBrush{})
    |                                 ^^^^^^^^^^^^^^^^^ expected enum `piet_cairo::Brush`, found struct `NullBrush`
```

probably a quick fix for someone who knows the libraries involved. =D

james-lawrence:
svgs are messy and sometimes have invalid ids, when
such a case in encountered fallback to a transparent brush.

fixes #1963

### DIFF
--- a/druid/src/widget/svg.rs
+++ b/druid/src/widget/svg.rs
@@ -373,7 +373,13 @@ impl SvgRenderer {
                 let color = color_from_svg(*c, opacity);
                 Rc::new(ctx.solid_brush(color))
             }
-            usvg::Paint::Link(id) => self.defs.find(id).unwrap(),
+            usvg::Paint::Link(id) => match self.defs.find(id) {
+                Some(v) => v,
+                None => {
+                    trace!("svg link id requested but not found: {:?}", id);
+                    Rc::new(ctx.solid_brush(Color::TRANSPARENT))
+                }
+            },
         }
     }
 }


druid/src/widget/svg.rs
```diff
@@ -373,7 +373,13 @@ impl SvgRenderer {
                 let color = color_from_svg(*c, opacity);
                 Rc::new(ctx.solid_brush(color))
             }
-            usvg::Paint::Link(id) => self.defs.find(id).unwrap(),
+            usvg::Paint::Link(id) => match self.defs.find(id) {
+                Some(v) => v,
+                None => {
+                    trace!("svg link id requested but not found: {:?}", id);
```
Maan2003, line 379: warn or error?

james-lawrence: indifferent, I ran into this case on files in the wild. personally I just think its a fact of life and not really a error or issue.

james-lawrence: thinking on it more, I didn't look too closely at the svg... and looking at druids code it code have been due to another unimplemented node in the svg tree. ;) i think in that context probably should move it to a warn.

Maan2003: trace will go unnoticed during development. but warn can be spammy I believe :/

james-lawrence: I did confirm that none can occur due to unimplemented nodes in druid. as for spammy, svg stuff already spams like crazy due to missing implementation details.

anyways, logging can be adjusted at any time.  atm I think this is fine until SVG is more fully implemented. its more important that druid doesn't crash as a result.

james-lawrence:
closing in favor of #1970 which implements new svg functionality and will conflict with this change.

PR #1965, opened 2021-09-12
PointerEvent

jneem:
This is a very much WIP implementation of pointer events (i.e. allowing detection of touch, pens, pressure, etc.). It should build on windows, gtk, and X11, but gtk is the only one where most things are actually implemented.

The big question is how to handle code that doesn't care about anything other than the usual mouse events. On web, this is handled by first sending the pointer event, and then if the pointer is a primary pointer and nothing cared about the pointer event, sending an emulated mouse event. Do we need this same kind of two-stage event? On web, it's important for backwards compatibility, but for us it probably isn't.

Here are some alternative approaches for dealing with users that only want the "classic" kind of mouse interaction.
1. Require the user to check the `is_primary` field to filter out uninteresting pointer events. This is what I'm currently doing, but it seems like a bit of a footgun.
2. Have different enum members in `Event` for primary and secondary pointer events. Then users who want them will need to explicitly match on them. It does have the disadvantage of bloating up `Event` a bit.
3. Only propagate non-primary pointer events to widgets that explicitly request them (e.g. by calling something like `ctx.register_for_non_primary_pointer_events()`.

Of course, this is all assuming that for "classic" mouse interaction it's enough to just listen to the primary pointer events. Maybe I'm way off on that one...


jneem:
I had an attempt at option 3, but I don't think it's workable, for two reasons.

- In order to support events that only some widgets get, you need to route them using `Event::Internal`, which goes against some other expectations (for example, a container widget might think that it can block mouse events by simply not propagating them, but they'd secretly get propagated through `Event::Internal`).
- The propagation logic would live in `WidgetPod`, which means that some wrapper widgets without a `WidgetPod` might have the setting change unexpectedly.

I'm currently leaning towards option 2. It's the closest to the web model, which has `PointerEvent` for those who want all the events and `MouseEvent` for those who don't.

cmyr:
Super cool to see this @jneem I'll block out some time for a thorough review soon :)

jneem:
Thanks! "Thorough" might be premature, though, since I'm not even sure of the basic approach. (The current implementation is option 2 on my original post)

cmyr:
Okay I will amend that to "I will give this a high-level appraisal shortly". 🤓

cmyr:
propagation:

I would imagine this happening *per-phase*, that is you would need to request it each time a touch sequence is initiated. I'm not worried about the routing; parent widgets would be expected to forward the event to children exactly as they're currently expected to forward keyboard & mouse events?

hmm, I would expect some sort of 'touches began' event immediately after the first finger makes contact, and on a trackpad that has 'tap to click' enabled this has additional nuance.

jneem:
Ok, I think I understand the bit about "touches began". But what about a button that doesn't care about gestures and just wants to know about "mouse down" and "mouse up"? Wouldn't we want to send a "mouse down" fairly promptly?

cmyr:
yes, we will have to figure out exactly how the process of gesture recognition works. in the case of 'tap to touch' i would expect the OS to be doing a lot of this interpretation for us, and if we're doing it on our own we'll need to experiment.

Kiiyya:
Is there any thoughts on stylus input with pressure sensitivity? For example, in order to make a note-taking application or drawing app.

jneem:
Yes, definitely stylus pressure is part of the plan. The PR already implements it on the GTK backend.

The main open question is how to nicely handle "higher-level" input events. I have some thoughts but haven't had time to put them to a test. Hopefully in the next month or so as the semester winds down...

cmyr:
@jneem if this is ever at a point where you would like my feedback, please ping me!

irrenhaus:
This would be really great to have, especially regarding stylus support. Is there anything I can do to help to get this done & merged?

jneem:
Sorry, it's been slow going. I was planning to spend some more time polishing my proof of concept, but given that I'm having trouble finding time for it, let me just write it up now.

So the issue with this patch is that by adding lots of extra detail to events, it makes event-handling harder: any widgets that listen to events need to start being aware of the difference between, for example, primary and non-primary pointers. My previous attempts to make this easier involved trying to separate off the "high-level" events from the "low-level" events so that it would be easier to match on just the ones you want. But I think there's a better approach: build a higher-level abstraction *on top of* the `Event` enum. I call this higher-level abstraction a `Gesture`, and the main part of the API is:

```rust
pub trait Gesture {
    type A; // "active"
    type F; // "finished"
    fn event(&mut self, ctx: &mut EventCtx, ev: &Event) -> GestureState<Self::A, Self::F>;
}

pub enum GestureState<A, F> {
    NoChange,
    Begin,
    Update(A),
    End(F),
    Cancel,
}
```

Druid would provide various implementors of `Gesture`, such as `ClickGesture`, `DragGesture`, etc. A widget that wants to handle a mouse click would store a `ClickGesture` struct within the widget, and have an event function like this:

```rust
fn event(&mut self, ctx: &mut EventCtx, ev: &Event, data: &mut Data, env: &Env) {
    // self.click is of type ClickGesture
    if self.click(ctx, event).is_finished() {
        // handle the click
    }
}
```

Some key points about this proposed gesture API:
- the implementors of `Gesture` are supposed to encapsulate all the state involved in recognizing the gesture. Currently, widgets track this state manually.
- all the low-level event details are handled in `Gesture` implementors, so widget authors don't need to worry too much
- the API is supposed to be general enough to handle other gestures, like pinch-to-zoom

The main thing about this that I haven't gotten around to investigating is that we might to extend the `ctx.is_handled` handing somehow: if some gesture decides that it has recognized something, other gestures that are considering that same event sequence need to be notified so that they can reset their internal state.

@cmyr does this seem like a reasonable approach?

xarvic:
I really like this design.

I have one question. What is the purpose of the `Gesture` trait? To me it seams like you cant use different Gestures interchangeably, therefore i don't understand what is the use in forcing them to have the same API?

jneem:
Good point, I forgot to explain that part. I can think of at least two advantages to having an abstract interface. One is to allow for combinators, like say

```rust
struct FirstOf<A, F> {
    gestures: Vec<Box<dyn Gesture<A=A, F=F>>
}

impl<A, F> Gesture for FirstOf<A, F> {
    type A = A;
    type F = F;
    // ...
}
```

The idea is that we could easily create a combined gesture that listens for multiple gestures and reports the first one that happens.

The other potential application is to have widgets or controllers parametrized over gestures. Like, instead of a `Click` controller you could have a `OnGesture` controller like this:

```rust
struct OnGesture<G> { ... }

impl<G: Gesture, T, W> Controller<T, W> for OnGesture<G> { ... }
```

There could also be others, I didn't really think too much. Note that both macOS and GTK have a fairly generic interface to gestures, so those are probably reasonable places to look for inspiration.

druid-shell/src/backend/windows/window.rs
```diff
@@ -1074,13 +1048,14 @@ impl WndProc for MyWndProc {
                     Some(0)
                 }
             }
-            WM_MOUSEMOVE => {
+            // FIXME: *maybe* this is the pointer version of mousemove?
```
doppioandante, line 1051: AFAIK, it is

PR #1967, opened 2021-09-12
Fix the ListIter implementation for Vector

xarvic:
This PR fixes the `ListIter` implementation for `Vector` and makes the implementation for `Arc<Vec<...>>` more efficient.

The old implementation used `iter_mut()` on `Vector` which makes a deep copy of the Vector, therefore `Data::same` returned always `false`


Maan2003 (approved):
Thanks! Makes sense

PR #1973, opened 2021-09-16
enhance gtk backend to allow semi-transparent backgrounds.

james-lawrence:
without this change users are unable to have a semi-transparent
background. its either fully transparent or fully opaque.

```
env.adding(druid::theme::WINDOW_BACKGROUND_COLOR, druid::Color::TRANSPARENT.with_alpha(0.8))
```
before:
screenshot-2021-09-16-093537

after:
screenshot-2021-09-16-093223


druid/src/window.rs
```diff
@@ -426,6 +426,11 @@ impl<T: Data> Window<T> {
         for &r in invalid.rects() {
             piet.clear(
                 Some(r),
+                // this if statement, and the transparent setting in general
+                // should be unnecessary, simply setting the alpha channel on the
+                // window background color < 1.0 should be sufficient to enable
+                // transparency. this logic remains until all backends
+                // support such a behavior.
                 if self.transparent {
```
Maan2003, line 434: I think this is not the correct way to implement this. I would just: 
```suggestion
                if self.transparent && env.get(crate::theme::WINDOW_BACKGROUND_COLOR).as_rgba8().3 == 255
```

james-lawrence: that won't work. I'm looking or semi transparency for any color which given the current implementation requires the backends to change.

james-lawrence: or example:
screenshot-2021-09-16-104353

Maan2003: that should work without backend change. just set background color with some alpha and `.transparent(true)`.

james-lawrence: I did complement changing that code to use the backendground color. but then the logic gets really convoluted.
as now you need to not only check for the transparency flag, but then you need to get the background color, check if its alpha is set, otherwise default its alpha to 0.

as a result I had the choice of:
1. writing more conditional logic
2. just removing code.

option 2 is clearly the better option.

james-lawrence: possibly, but I'd wait for that to actually a real issue before complicating the API on top of everything. blending layers is a pretty common path through compositors.

and if the performance issue ever did come up in a real use case, I'd push for the logic to still be pushed down to the backends, as this is still simpler both in terms of end user API, code complexity.

```
        let mut transparent = self.background.3 < 0xFF;
        if let Some(screen) = window.screen() {
            let visual = screen.rgba_visual();
            transparent = visual.is_some();
            window.set_visual(visual.as_ref());
        }
```

james-lawrence: oh and just to be clear, I'll happily change this to whatever you want in the end. just pointing out why I didn't just make the change you suggested originally. mainly due to the minor irritation caused by the surprise of needing to set an additional flag on top of the color itself.

just let me know which approach you want me to take. ;)

Maan2003: your approach sound reasonable. We should deprecate `set_transparent`(because it is useless now)  . if we see performance issues in futures, we can move background color to the backends

Maan2003: also: we can just `transparent = true` default. then user can set transparent false if they are having performance problems

Maan2003: and changing the default should be a simpler change

james-lawrence: @Maan2003 updated.

druid/src/window.rs
```diff
@@ -426,7 +426,9 @@ impl<T: Data> Window<T> {
         for &r in invalid.rects() {
             piet.clear(
                 Some(r),
-                if self.transparent {
+                // we are maintaining backwards compatibility here where setting
+                // transparent to true explicitly would make a fully transparent window.
+                if self.transparent && env.get(crate::theme::WINDOW_BACKGROUND_COLOR).as_rgba().3 == 255.0 {
```
Maan2003, line 431: first, this check is wrong, you need `.as_rgba8().3 == 255` (range of rgba() is 0.-1.)

after fixing the check, this makes all default windows fully transparent (because default window background has alpha 255).

I think it is ok backwards compatibility because `transparent` is not yet released to crates.io (only on git)

james-lawrence: ah, yeah if we don't care about backwards compat then good, makes it less messy.

james-lawrence: after playing with this logic today. it still requires people to set two flags, but instead of for transparent its now for opaque.

james-lawrence: and it breaks a bunch of tests =p

james-lawrence: and if its fine to break backwards compat rather just remove this method entirely and go with replacing the boolean with a color we check in the backend.

Maan2003: it is a good idea to handle background color in the backend. it might help when ui doesn't draw fast enough to keep up with resize.

druid/src/window.rs
```diff
@@ -426,11 +426,8 @@ impl<T: Data> Window<T> {
         for &r in invalid.rects() {
             piet.clear(
                 Some(r),
-                if self.transparent {
-                    Color::TRANSPARENT
-                } else {
-                    env.get(crate::theme::WINDOW_BACKGROUND_COLOR)
-                },
+                env.try_get(crate::theme::WINDOW_BACKGROUND_COLOR)
+                    .unwrap_or_else(|_| self.background.clone()),
```
Maan2003, line 430: now the background should drawn in backends?

druid-shell/src/common_util.rs
```diff
@@ -152,3 +155,8 @@ impl Default for ClickCounter {
         ClickCounter::new(MULTI_CLICK_INTERVAL, MULTI_CLICK_MAX_DISTANCE)
     }
 }
+
+#[cfg(not(target_arch = "wasm32"))]
```
Maan2003, line 159: why not wasm32?

druid-shell/src/common_util.rs
```diff
@@ -152,3 +155,8 @@ impl Default for ClickCounter {
...
+pub fn is_transparent(c: &piet::Color) -> bool {
+    (c.as_rgba().3 - 1.0).abs() > f64::EPSILON
```
Maan2003, line 161: just bypass the f64 conversion entirely by using `as_rgba8`

druid-shell/src/common_util.rs
james-lawrence: its unused. and allowing unused was giving me issues.

druid/src/window.rs
james-lawrence: outside of the scope. this PR is about ease of use and allowing semitransparent backgrounds. lets stick to that issue.

james-lawrence: beyond scope creep, I'm not comfortable doing that change personally ATM. due to it impacting all the backends and it isn't just a manual translation of logic. it'll blow up testing, debugging, and implementation work and personally I only need semitransparent windows. best to hold off until someone is in the backends for actual work. I will be doing a wayland backend in the future if no one beats me to it. I'll investigate actually painting the background at that time.

druid-shell/src/common_util.rs
james-lawrence: fixed up. had to do some rust reading around the error i was seeing last night. =)

druid/src/window.rs
james-lawrence: p.s. thanks for being patient with me. only my 3rd week of programming in rust. still working my way through the ecosystem and learning its idiosyncrasies.

Maan2003: makes sense

Maan2003:
I think we are making it very more complicated. 
We just want to use semi transparent colors in window background. It is fine to require `WindowDesc::transparent` for transparent windows. So i propose:

- We don't change druid-shell
- both `.transparent` and setting the window background in env is needed to get transparent window
- warn if `window.transparent ^ (window.background.rgba8().3 != 255)`  
- we can just clear with `window.background`

PR #1976, opened 2021-09-18
Scroll to view

xarvic:
This PR adds the `scroll_to_view` and `scroll_area_to_view` methods to `UpdateCtx`, `LifecycleCtx` and `EventCtx`, which can be used to bring specific parts of a widget into view.

There are still some things to discuss:
- Currently the methods send commands if they are called from `UpdateCtx` or `LifecycleCtx`, but i am not sure this is the right way to do it.
- Calling the methods produces warnings, since the Notifications are passed up to the root. Should we hard code that this is intended behavior for this Selector or search for a better way to do it, like a `supress_warnings` flag in Notification?

### DIFF
--- a/druid/src/command.rs
+++ b/druid/src/command.rs
@@ -341,7 +341,7 @@ pub mod sys {
     /// translated.
     ///
     /// [`scroll_to_view`]: crate::EventCtx::scroll_to_view()
-    /// [`scroll_area_to_view`]: crate::EventCtx::scroll_are_to_view()
+    /// [`scroll_area_to_view`]: crate::EventCtx::scroll_area_to_view()
     pub const SCROLL_TO_VIEW: Selector<Rect> = Selector::new("druid_builtin.scroll_to");
 
     /// A change that has occured to text state, and needs to be
--- a/druid/src/command.rs
+++ b/druid/src/command.rs
@@ -175,10 +175,7 @@ pub mod sys {
     use std::any::Any;
 
     use super::Selector;
-    use crate::{
-        sub_window::{SubWindowDesc, SubWindowUpdate},
-        FileDialogOptions, FileInfo, SingleUse, WidgetId, WindowConfig,
-    };
+    use crate::{sub_window::{SubWindowDesc, SubWindowUpdate}, FileDialogOptions, FileInfo, SingleUse, WidgetId, WindowConfig, Rect};
 
     /// Quit the running application. This command is handled by the druid library.
     pub const QUIT_APP: Selector = Selector::new("druid-builtin.quit-app");
@@ -327,6 +324,8 @@ pub mod sys {
     /// Text input state has changed, and we need to notify the platform.
     pub(crate) const INVALIDATE_IME: Selector<ImeInvalidation> =
         Selector::new("druid-builtin.invalidate-ime");
+
+    pub const SCROLL_TO: Selector<Rect> = Selector::new("druid_builtin.scroll_to");
 
     /// A change that has occured to text state, and needs to be
     /// communicated to the platform.
--- a/druid/src/command.rs
+++ b/druid/src/command.rs
@@ -103,6 +103,7 @@ pub struct Notification {
     symbol: SelectorSymbol,
     payload: Arc<dyn Any>,
     source: WidgetId,
+    route: WidgetId,
 }
 
 /// A wrapper type for [`Command`] payloads that should only be used once.
@@ -423,6 +424,7 @@ impl Command {
             symbol: self.symbol,
             payload: self.payload,
             source,
+            route: source,
         }
     }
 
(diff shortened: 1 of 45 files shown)


druid/src/scroll_component.rs
```diff
@@ -515,13 +516,63 @@ impl ScrollComponent {
...
 
+/// The default handling of the [`SCROLL_TO_VIEW`] notification for a scrolling container.
+///
+/// The [`SCROLL_TO_VIEW`] notification is send when [`scroll_to_view`] or [`scroll_area_to_view`]
+/// are called.
+///
+/// [`SCROLL_TO_VIEW`]: crate::commands::SCROLL_TO_VIEW
+/// [`scroll_to_view`]: crate::EventCtx::scroll_to_view()
+/// [`scroll_area_to_view`]: crate::EventCtx::scroll_area_to_view()
+pub fn default_scroll_to_view_handling(
```
Maan2003, line 527: should this be a method on `ViewPort`?

druid/src/scroll_component.rs
```diff
@@ -515,13 +516,63 @@ impl ScrollComponent {
...
+    ctx: &mut EventCtx,
+    port: &mut Viewport,
+    global_highlight_rect: Rect,
+) {
+    let global_content_offset = ctx.window_origin().to_vec2() - port.view_origin.to_vec2();
```
Maan2003, line 532: probably not a good idea to use widget's origin here

druid/src/scroll_component.rs
```diff
@@ -515,13 +516,63 @@ impl ScrollComponent {
...
+}
+
 #[cfg(test)]
 mod tests {
     use float_cmp::approx_eq;
 
-    use super::*;
     use crate::kurbo::Size;
 
+    use super::*;
```
Maan2003, line 574: again accidental reordering

druid/src/widget/mod.rs
```diff
@@ -14,6 +14,54 @@
 
 //! Common widgets.
 
+pub use added::Added;
```
Maan2003, line 17: this reordering was accidental?

druid/src/widget/scroll.rs
```diff
@@ -14,11 +14,13 @@
 
 //! A container that scrolls its contents.
 
+use tracing::{instrument, trace};
```
Maan2003, line 17: again accidental reordering

druid/src/core.rs
```diff
@@ -968,6 +976,11 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
                     if let Some(change) = this_changed {
                         self.state.has_focus = change;
                         extra_event = Some(LifeCycle::FocusChanged(change));
+
+                        if change {
+                            //TODO: decide whether this should be done manually
```
Maan2003, line 981: I vote for manual

Maan2003 (commented):
Some minor nits for now. haven't carefully reviewed `default_scroll_to_view_handling` logic yet. hope to get that done tomorrow

druid/src/scroll_component.rs
xarvic: Good idea :)

druid/src/command.rs
```diff
@@ -328,6 +328,22 @@ pub mod sys {
...
+    ///
+    /// Widgets which hide their children, should always call `ctx.set_handled()` in response to
+    /// avoid unintended behaviour from widgets further down the tree.
+    /// If possible the widget should move its children to bring the area into view and then submit
+    /// a new notification with the region translated by the amount, the child it contained was
+    /// translated.
+    ///
+    /// [`scroll_to_view`]: crate::EventCtx::scroll_to_view()
+    /// [`scroll_area_to_view`]: crate::EventCtx::scroll_are_to_view()
+    pub const SCROLL_TO_VIEW: Selector<Rect> = Selector::new("druid_builtin.scroll_to");
```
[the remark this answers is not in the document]
xarvic: Ok

druid/src/scroll_component.rs
xarvic: What is the problem with using the widget's origin here?

druid/src/core.rs
xarvic: I couldn't come up with a case, where you don't want to do it. It feels natural to display the focused widget, but I don't have a strong opinion on this

druid/src/scroll_component.rs
```diff
@@ -515,13 +516,63 @@ impl ScrollComponent {
...
+    }
+    if content_highlight_rect.y0 < view_rect.y0
+        || content_highlight_rect.size().height > port.view_size.height
+    {
+        //Prefer the upper over the lower side if the scroll_to content is bigger than the view_size
+        new_origin.y = content_highlight_rect.y0;
+    } else if content_highlight_rect.y1 > view_rect.y1 {
+        new_origin.y = content_highlight_rect.y1 - port.view_size.height;
+    }
+
```
Maan2003, line 556: reuse `Clipbox::pan_to_visible` here?, also now this makes sense as a method on `Clipbox`

xarvic:
On Desktop this should be fine, but it will probably cause problems on mobile. For an example focusing a textbox, will first update the scroll positions and then request the virtual-keyboard, which results in a layout change and therefore the text box is might be hidden again.
Do you have an idea how to solve this or is this a problem should not care about now?

druid/src/scroll_component.rs
xarvic: Agreed

Maan2003:
I think the fundamental problem is "the focused should stay in view" or "the focused widget should come in view once". maybe Scroll widget keeps track of a "focused" rect that will stay in view unless manually scroll away

Maan2003:
for this PR, I would just add `scroll_to_view` and not do it automatically on focus change *yet*

xarvic:
Ok

xarvic:
> I think the fundamental problem is "the focused should stay in view" or "the focused widget should come in view once". maybe Scroll widget keeps track of a "focused" rect that will stay in view unless manually scroll away

A focused region that stays in in view until scrolled away sounds good, but i think keeping the rect wont be enough. If we have a two `Scroll`s in a `Flex` which again is wrapped in a `Scroll`, it could be possible that both of the inner `Scroll`s believe they contain the focused region and send `Notification`s to the outer `Scroll`.

Maan2003 (approved):
Thanks! Looks good

PR #1978, opened 2021-09-18
Route id

xarvic:
Adds `Notification::route`

### DIFF
--- a/druid/src/command.rs
+++ b/druid/src/command.rs
@@ -103,6 +103,7 @@ pub struct Notification {
     symbol: SelectorSymbol,
     payload: Arc<dyn Any>,
     source: WidgetId,
+    route: WidgetId,
 }
 
 /// A wrapper type for [`Command`] payloads that should only be used once.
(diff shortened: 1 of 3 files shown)


druid/src/command.rs
```diff
@@ -528,6 +530,34 @@ impl Notification {
...
+
+    /// Change the route id
+    pub(crate) fn with_route(mut self, widget_id: WidgetId) -> Self {
+        self.route = widget_id;
+        self
+    }
+
+    /// The [`WidgetId`] of the last [`Widget`] that this [`Notification`] was passed through.
+    ///
+    /// #Example
```
Maan2003, line 542:
```suggestion
    /// # Example
```

xarvic:
> so route only has id of the immediate child, not the full route. can you explain the usecase and how this will help? Also why not use Vec<WidgetId> for the widgets it traveled

The immediate child ID is useful, since it is easily accessible from the current widget.
For an example, if we want to implement the correct scroll_to behavior for enum_switcher, we need to know whether the notification came from within the active widget, if not we should discard it. 
In the linked issue are more examples.

Maan2003 (approved):
Thanks! makes sense. 

It is nice to update the `notifications` test but not a blocker.

PR #1982, opened 2021-09-21
Windows tooltip changes

sjoshid:
Tooltips were introduced in #1919. But they weren't working correctly Windows. Some issues (seen in `sub_window.rs` example) were:
* Tooltips opened up in a different location. 
* The closable/drag-able window was broken as well.
This PR fixes everything related to tooltip on Windows. Handles scaling issue as well. 

Related Zulip [topic](https://xi.zulipchat.com/#narrow/stream/147926-druid/topic/Tooltip.20broken.20in.20Windows).

### DIFF
--- a/druid-shell/src/backend/windows/window.rs
+++ b/druid-shell/src/backend/windows/window.rs
@@ -1387,12 +1387,18 @@ impl WindowBuilder {
                         dwStyle = WS_POPUP;
                         dwExStyle = WS_EX_NOACTIVATE | WS_EX_TOOLWINDOW;
                         focusable = false;
-                        let scaled_sub_window_point = WindowBuilder::scale_sub_window_position(
-                            self.position,
-                            parent_window_handle.get_scale(),
-                        );
-                        pos_x = scaled_sub_window_point.x as i32;
-                        pos_y = scaled_sub_window_point.y as i32;
+                        if let Some(point_in_window_coord) = self.position {
+                            let screen_point = parent_window_handle.get_position()
+                                + point_in_window_coord.to_vec2();
+                            let scaled_tooltip_point = WindowBuilder::scale_sub_window_position(
+                                Some(screen_point),
+                                parent_window_handle.get_scale(),
+                            );
+                            pos_x = scaled_tooltip_point.x as i32;
+                            pos_y = scaled_tooltip_point.y as i32;
+                        } else {
+                            warn!("No position provided for tooltip!");
+                        }
                     }
                     WindowLevel::DropDown(_) => {
                         dwStyle = WS_CHILD;
--- a/druid-shell/src/backend/windows/window.rs
+++ b/druid-shell/src/backend/windows/window.rs
@@ -1353,6 +1353,8 @@ impl WindowBuilder {
                 present_strategy: self.present_strategy,
             };
 
+            // TODO: pos_x and pos_y are only scaled for windows with parents. But they need to be
+            // scaled for windows without parents too.
             let (mut pos_x, mut pos_y) = match self.position {
                 Some(pos) => (pos.x as i32, pos.y as i32),
                 None => (CW_USEDEFAULT, CW_USEDEFAULT),
--- a/druid-shell/src/backend/windows/window.rs
+++ b/druid-shell/src/backend/windows/window.rs
@@ -1391,7 +1391,7 @@ impl WindowBuilder {
                             let screen_point = parent_window_handle.get_position()
                                 + point_in_window_coord.to_vec2();
                             let scaled_tooltip_point = WindowBuilder::scale_sub_window_position(
-                                Some(screen_point),
+                                screen_point,
                                 parent_window_handle.get_scale(),
                             );
                             pos_x = scaled_tooltip_point.x as i32;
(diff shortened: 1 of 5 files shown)


sjoshid:
Although this PR fixes tooltip on Windows only, it looks like this issue happens in Mac as well. That I can say by just reading Mac code. So not sure. 
Since I dont have access to MacOS, Im not going to make mac specific changes. Anyone with access to one can easily make them I assume.

druid-shell/src/backend/windows/window.rs
```diff
@@ -1382,15 +1380,25 @@ impl WindowBuilder {
...
+                            let screen_point = parent_window_handle.get_position()
+                                + point_in_window_coord.to_vec2();
+                            let scaled_tooltip_point = WindowBuilder::scale_sub_window_position(
+                                Some(screen_point),
+                                parent_window_handle.get_scale(),
+                            );
+                            pos_x = scaled_tooltip_point.x as i32;
+                            pos_y = scaled_tooltip_point.y as i32;
+                        } else {
+                            warn!("No position provided for tooltip!");
```
jneem, line 1400: This warning sounds reasonable to me, but maybe there should be something in the `WindowBuilder` API docs about tooltips expecting to have a position?

druid-shell/src/backend/windows/window.rs
```diff
@@ -1355,7 +1353,7 @@ impl WindowBuilder {
                 present_strategy: self.present_strategy,
             };
 
-            let (pos_x, pos_y) = match self.position {
+            let (mut pos_x, mut pos_y) = match self.position {
```
jneem, line 1356: Don't these also need to be scaled?

druid-shell/src/backend/windows/window.rs
```diff
@@ -1507,6 +1513,30 @@ impl WindowBuilder {
             Ok(handle)
         }
     }
+
+    /// When creating a sub-window, we need to scale its position with respect to its parent.
+    /// If there is any error while scaling, log it as a warn and show sub-window in top left corner of screen/window.
+    fn scale_sub_window_position(
+        un_scaled_sub_window_position: Option<Point>,
```
jneem, line 1520: You're only using this in one place, and you've already checked that there is a position, so the `Option` is redundant, no?

jneem (changes requested):
Looks good, just a few minor things. (Also I didn't test it because testing on windows is pretty inconvenient for me right now...)

druid-shell/src/backend/windows/window.rs
sjoshid: Yes there should be. I have added sub window specific doc to `WindowBuilder::set_position()`.

druid-shell/src/backend/windows/window.rs
sjoshid: Sub window position need to be scaled because they are shown in relation to their parent. But for app/parent windows, _mayyyyyybe_ it's OK to now scale them because you wont notice much difference?? I do feel I need to scale these too but left it as-is because it is not related to this tooltip PR.

druid-shell/src/backend/windows/window.rs
sjoshid: Agree. Made this much simpler.

sjoshid:
> 
> 
> Looks good, just a few minor things. (Also I didn't test it because testing on windows is pretty inconvenient for me right now...)

No problem. I have tested on Windows and Ubuntu. I feel (from looking at Mac code) that tooltips are broken in MacOS. But I cant test it there.

druid-shell/src/window.rs
```diff
@@ -492,8 +480,11 @@ impl WindowBuilder {
 
     /// Sets the initial window position in [display points], relative to the origin of the
     /// virtual screen.
+    /// For sub-windows like [tooltips] and [dropdowns], this position must be in window co-ordinates.
```
jneem, line 483: Could you also change the previous sentence? I don't think it's correct anymore...

druid-shell/src/backend/windows/window.rs
jneem: I think it really does need to be scaled, but I'd be happy with a `TODO` comment for now.

sjoshid: I'll add a `TODO`.

druid-shell/src/window.rs
sjoshid: Isnt it correct for app/parent windows? Their position must be relative to virtual screen right? Because they dont have a parent.

jneem: It's true for app windows, but it doesn't make clear (IMO) that it's only for app windows. I'd suggest something more like "Sets the initial window position in display points. For windows with a parent, the position is relative to the parent. For windows without a parent, it is relative to the origin of the virtual screen. See also [set_level]."

sjoshid: Sounds good. Made the change.

druid/src/app.rs
```diff
@@ -53,7 +53,7 @@ pub enum WindowSizePolicy {
 
 /// Window configuration that can be applied to a WindowBuilder, or to an existing WindowHandle.
 /// It does not include anything related to app data.
-#[derive(Debug, PartialEq)]
+#[derive(PartialEq)]
 pub struct WindowConfig {
```
jneem, line 57: Ah, sorry, just noticed this: why no `Debug`?

jneem:
Thanks! Just noticed a removed `Debug` that I didn't understand, but I think this is basically ready.

jneem (approved):

druid/src/app.rs
sjoshid: It was a `Debug` because `WindowLevel` was a `Debug`. And `WindowLevel` was a `Debug` just so we could warn "WindowBuilder::set_level({:?}) is not implemented". So basically once all the window levels are implemented we would not warn so we would not need `Debug`. 
Instead of saying ""WindowBuilder::set_level({:?}) is not implemented".. im saying "WindowLevel::DropDown and WindowLevel::Modal are not implemented in Windows".

druid-shell/src/backend/windows/window.rs
sjoshid: Oops. I said I'd add a `TODO` but I forgot. Sorry. I'll add it now.

giannissc:
How does not implementing the set_level() affect the dropdown example in the nursery? And what would be the difficulty for somebody else to fix WindowLevel::Dropdown and WindowLevel::Modal.

Thank you for fixing this btw on Windows cause it has been bugging me for a while :)

sjoshid:
Not implementing `set_level` means they are shown as a `WS_OVERLAPPEDWINDOW` window.

There is one thing I remember that I spend a lot of time on with dropdowns (`WS_CHILD`). It has to do with invalidation. It's in a Zulip group. I'll tag you there. Other than that, tooltips and dropdowns have a lot in common I think.

As for `WindowLevel::Modal` I honestly havent looked much in to it. Colin has few things in mind regarding how to handle them (I'll find the link with his thoughts and add here later)

sjoshid:
This link https://gist.github.com/cmyr/df7545ec6598c72632fb6e7efa1a7154

giannissc:
@sjoshid Yeah I do think that dropdowns and tooltips are pretty similar. And it very weird to see the dropdown shown as a separate appplication window on Windows.

Also thanks for the links :)

sjoshid:
@jneem when can we get this in? Assuming you dont have any other concerns.

PR #1984, opened 2021-09-23
Make set_window_state accept &self

Quant1um:
This PR makes `WindowHandle::set_window_state` accept `&self` instead of `&mut self` for consistency with other `set_*` functions (which in turn allows setting window state from `Widget::event`)


Quant1um:
Ok apparently that does not work for all platforms if im correct. Closing

PR #1989, opened 2021-09-26
wayland: small fixes

Maan2003:
These are needed because druid moved forward from when wayland backend was intially written.

### DIFF
--- a/druid-shell/src/backend/wayland/window.rs
+++ b/druid-shell/src/backend/wayland/window.rs
@@ -459,6 +459,7 @@ impl WindowData {
     /// - `size` is the physical size in pixels we are drawing.
     /// - `force` means draw the whole frame, even if it wasn't all invalidated.
     pub(crate) fn paint(&self, size: RawSize, buf: &mut [u8], force: bool) {
+        self.handler.borrow_mut().prepare_paint();
         //tracing::trace!("Paint call");
         //self.data.borrow().assert_size();
         if force {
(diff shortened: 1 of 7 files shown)


Maan2003:
Ah clippy, I will get clippy and other errors fixed tomorrow. We should also setup CI for wayland

jneem:
I'm not at all familiar with this code, but let's worry about proper reviews when we're thinking about merging the wayland branch...

druid-shell/src/backend/wayland/window.rs
```diff
@@ -448,7 +459,8 @@ impl WindowData {
     /// - `size` is the physical size in pixels we are drawing.
     /// - `force` means draw the whole frame, even if it wasn't all invalidated.
     pub(crate) fn paint(&self, size: RawSize, buf: &mut [u8], force: bool) {
-        //log::trace!("Paint call");
+        self.handler.borrow_mut().prepare_paint();
```
james-lawrence, line 462: what issue is this attempting to fix? I didn't find this necessary in my fork, only painting problem I was running into was around animations.

Maan2003: See the docs of `WinHandler::prepare_paint`. window handler is supposed to invalidate regions in prepare_paint. Without invalidation, there will be no repaints.

james-lawrence: I see. this doesn't seem to change any behavior rendering though. not against the change btw, was just curious because I didn't find it necessary and was curious what it fixes.

Maan2003: oh I recall, this fixes invalidate example

Issue #1986, opened 2021-09-23
Tooltip broken in Windows


sjoshid:
Tooltips were introduced in #1919. But they aren't working correctly in Windows specifically when you set scaling to something other than 100%. Some issues (seen in sub_window.rs example) are:

* Tooltips open up in a different location.
* The closable/drag-able window is broken as well.

giannissc:
In case you can't tell the mouse pointer is over the label widget. After the update the full screen hovering works great and the tooltip is right under the mouse pointer however for the unmaximised windows, the tooltip more the further away the further you more from the screen's left corner

giannissc:
But weren't these issues supposed to be solved with https://github.com/linebender/druid/pull/1982?

sjoshid:
Not really. There was no set definition of what `set_position` took: you could pass any type of `Point`.

With the PR, definition of `set_position` is clear: if creating a window with a parent window (tooltips or dropdowns) you _must_ pass a point in parent window coord.

giannissc:
Does it work out of the box with tooltip code or did you have to make some changes? Cause it doesn't work right for me..

PR #1992, opened 2021-09-27
Fixes to readme.md

winksaville:
For me hello_web, web and sub_window didn't work see readme.md.

### DIFF
--- a/druid/examples/readme.md
+++ b/druid/examples/readme.md
@@ -61,7 +61,6 @@ For more info and prerequistes see [druid/examples/hello_web/README.md](druid/ex
 ```
 cd druid/examples/hello_web
 wasm-pack build --out-dir pkg --out-name hello_web
-http
 ```
 [View at http://localhost:8000](http://localhost:8000].
 
@@ -74,7 +73,6 @@ For more info and prerequistes see [druid/examples/web/README.md](druid/examples
 ```
 cd druid/examples/web
 wasm-pack build --out-dir pkg --out-name web
-http
 ```
 [View at http://localhost:8000](http://localhost:8000].
 
--- a/druid/examples/readme.md
+++ b/druid/examples/readme.md
@@ -3,7 +3,7 @@
 There are several different kind of examples, some demonstrate one particular
 Druid concept, some are tools used for testing and debugging, and
 others are more complete examples of how to tie everything together.
-The latter are listed separately under [showcases](##Showcases).
+The latter are listed separately under [showcases](#Showcases).
 
 ## Anim
 ```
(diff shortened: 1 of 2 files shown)


druid/examples/readme.md
```diff
@@ -138,18 +163,13 @@ cargo run --example split_demo
 An example of how to split a widget in 2 in various ways. This also includes having the user drag the border!! 
 
 ## Sub Window
+Not working, no sub-window seen?
```
sjoshid, line 166: Are any of the issues part of this #1986?

winksaville: I'm on Linux so I don't think so, it also could be I don't know what a "sub window" is.. I was expecting some obvious, like a.new window to pop up. It would be great if you or some one who actually knows what to expect could try it!

druid/examples/readme.md
```diff
@@ -55,12 +55,31 @@ cargo run --example hello
 This shows some of the basics of druid. If you need a start of how to build an application with a text-box and some labels this is where to start.
 
 ## Hello_web
+Not working, browser is blank after going to http://localhost:8000?
+
+For more info and prerequistes see [druid/examples/hello_web/README.md](druid/examples/hello_web/README.md).
 ```
 cd druid/examples/hello_web
 wasm-pack build --out-dir pkg --out-name hello_web
+http
```
JAicewizard, line 64: I think we should not have this in the examples. How the user serves and loads this example is up to them not us.

JAicewizard:
Thanks for the contribution! This was a very dragged out pull request, seems like some things slipped through.

druid/examples/readme.md
sjoshid: That example works correctly in Linux. I dont know the exact definition of sub window but think of tooltips. When you hover over ""Hello World! It's mine!, you'll see a tooltip. Or when you click button "Make sub window" it'll show a sub window you can drag around. Tooltips, new windows that your main app window spawns or a dropdown (not yet available in druid) are all sub windows.

winksaville: @sjoshid, I do see the "Tips! Are good" ToolTip when I hover over "Hello World! It's mine!".

I now see the "Drag me" sub_window. The problem is that the `set_position(Point::new(1000.0, 500.0))` causes
it to be "off screen" with my dual monitor setup. As you can see from the screenshot I have two monitors one horizontal and one vertical and for some reason the "Drag me" sub_window is not visible. But If I remove the call `set_position` then "Drag me" appears in the middle of the 400 x 400 main screen:

image

So I'm also going to add that modification to this PR.

druid/examples/readme.md
winksaville: Done

druid/examples/readme.md
```diff
@@ -55,12 +55,31 @@ cargo run --example hello
 This shows some of the basics of druid. If you need a start of how to build an application with a text-box and some labels this is where to start.
 
 ## Hello_web
+Not working, browser is blank after going to http://localhost:8000?
```
winksaville, line 58: Does this work for other people, it does not seem to work for me?

druid/examples/readme.md
```diff
@@ -55,12 +55,31 @@ cargo run --example hello
...
 cd druid/examples/hello_web
 wasm-pack build --out-dir pkg --out-name hello_web
+http
 ```
+[View at http://localhost:8000](http://localhost:8000].
+
 This is an example of how to get almost any druid application can be used on the web. This is just the hello_world example but should work for all of them.
 
+## Web
+Not working, browser is blank after going to http://localhost:8000?
```
winksaville, line 71: Does this work for other people, as with the /web example it does not seem to work for me?

winksaville (commented):
What should we do with `hello_web` and `web`?

druid/examples/readme.md
winksaville: I don't know what's different but it's working for me now!

druid/examples/readme.md
winksaville: As Hello_web above, it works for me now.

winksaville:
> What should we do with `hello_web` and `web`?

I retested and they are working now. I've removed the "Not working ..." sentences in readme.md and thing this PR is ready to merge.

cmyr (approved):
Thanks!

PR #1993, opened 2021-09-28
Build on OpenBSD

klemensn:
Just like Linux, OpenBSD is sufficiently supported using GTK+3.

No platform specific changes required to successfully run the examples
and real-world applications like Psst (spotify GUI client).

The pre-push hook mentioned in CONTRIBUTING.md runs sucessfully on
OpenBSD/amd64 7.0 (after pointing druid to a ported version of piet in
.cargo/config.toml to allow building in the first place), i.e. no lint
or build failure.

### DIFF
--- a/druid-shell/build.rs
+++ b/druid-shell/build.rs
@@ -7,7 +7,9 @@ fn main() {
     use std::env;
     use std::path::PathBuf;
 
-    if env::var("CARGO_CFG_TARGET_OS").unwrap() != "linux" {
+    if env::var("CARGO_CFG_TARGET_OS").unwrap() != "linux"
+        || env::var("CARGO_CFG_TARGET_OS").unwrap() != "openbsd"
+    {
         return;
     }
 
--- a/druid-shell/build.rs
+++ b/druid-shell/build.rs
@@ -8,7 +8,7 @@ fn main() {
     use std::path::PathBuf;
 
     if env::var("CARGO_CFG_TARGET_OS").unwrap() != "linux"
-        || env::var("CARGO_CFG_TARGET_OS").unwrap() != "openbsd"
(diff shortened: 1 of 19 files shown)


klemensn:
I committed the hunks produced by `cargo fmt --all`.
I am not sure what to do about CI's failing Ubuntu workflows, though.

druid-shell/build.rs
```diff
@@ -7,7 +7,9 @@ fn main() {
     use std::env;
     use std::path::PathBuf;
 
-    if env::var("CARGO_CFG_TARGET_OS").unwrap() != "linux" {
+    if env::var("CARGO_CFG_TARGET_OS").unwrap() != "linux"
+        || env::var("CARGO_CFG_TARGET_OS").unwrap() != "openbsd"
```
Maan2003, line 11:
```suggestion
    if env::var("CARGO_CFG_TARGET_OS").unwrap() != "linux"
        && env::var("CARGO_CFG_TARGET_OS").unwrap() != "openbsd"
```

cmyr (approved):
Thanks!

klemensn:
Great, now I'm just waiting for a new proper release so software like `psst` can fetch the crate from the registry rather than a specific git checkout.

This is less important for development on personal machines, but environments such as those for official OS/distribution packages often prefetch things prior to build (which has no network) access, and OpenBSD's package infrastructure only knows how to fetch crates from the registry.

I've been working on fetching from git, which already works, but that also has implications to the build process, which are a bit trickier.

tl;dr:  it would be nice to see a new release sooner than later to ease downstream development.

xStrom:
> Great, now I'm just waiting for a new proper release

Good news @klemensn, [Druid v0.8.2 is now out](https://crates.io/crates/druid).

PR #1996, opened 2021-10-01
Do not panic in Application::try_global if Application is not created

### LINKED ISSUE #1995: X11: calling Screen::get_monitors panics
```rust
use druid::widget::SizedBox;
use druid::{AppLauncher, Screen, WindowDesc};

pub fn main() {
    Screen::get_monitors();
    let main_window = WindowDesc::new(SizedBox::empty());
        .log_to_console()
        .launch(())
        .expect("Failed to launch application");
}
```
error:
```
'Main thread assertion failed 1 != 0', druid-shell/src/util.rs:39:9
```

useful parts of backtrace 
```
   2: druid_shell::util::assert_main_thread
             at /src/druid/druid-shell/src/util.rs:39:9
   3: druid_shell::application::Application::try_global
             at /src/druid/druid-shell/src/application.rs:124:9
   4: druid_shell::backend::x11::screen::get_monitors
             at /src/druid/druid-shell/src/backend/x11/screen.rs:38:37
   5: druid_shell::screen::Screen::get_monitors
             at /src/druid/druid-shell/src/screen.rs:87:9
```

Maan2003:
`Application::try_global` called `assert_main_thread` but before Application is created the main thread is unclaimed.
I think there was no way for `Application::try_global` to return None.

Added `assert_main_thread_or_main_unclaimed` that does not panic if main thread is unclaimed.

Fixes #1995

### DIFF
--- a/druid-shell/src/util.rs
+++ b/druid-shell/src/util.rs
@@ -32,10 +32,27 @@ fn current_thread_id() -> u64 {
 /// # Panics
 ///
 /// Panics when called from a non-main thread.
+#[allow(dead_code)]
 pub fn assert_main_thread() {
     let thread_id = current_thread_id();
     let main_thread_id = MAIN_THREAD_ID.load(Ordering::Acquire);
     if thread_id != main_thread_id {
+        panic!(
+            "Main thread assertion failed {} != {}",
+            thread_id, main_thread_id
+        );
+    }
+}
+
+/// Assert that the current thread is the registered main thread or main thread is not claimed.
+///
+/// # Panics
+///
+/// Panics when called from a non-main thread and main thread is claimed.
+pub fn assert_main_thread_or_main_unclaimed() {
+    let thread_id = current_thread_id();
+    let main_thread_id = MAIN_THREAD_ID.load(Ordering::Acquire);
+    if thread_id != main_thread_id && main_thread_id != 0 {
         panic!(
             "Main thread assertion failed {} != {}",
             thread_id, main_thread_id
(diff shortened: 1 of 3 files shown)


druid-shell/src/util.rs
```diff
@@ -43,6 +44,22 @@ pub fn assert_main_thread() {
     }
 }
 
+/// Assert that the current thread is the registered main thread or main thread is not claimed.
+///
+/// # Panics
+///
+/// Panics when called from a non-main thread and main thread is claimed.
+pub fn assert_main_thread_or_main_unclaimed() {
```
cmyr, line 52: does this need to be public?

druid-shell/src/util.rs
```diff
@@ -32,6 +32,7 @@ fn current_thread_id() -> u64 {
 /// # Panics
 ///
 /// Panics when called from a non-main thread.
+#[allow(dead_code)] // can be unused on some backend
```
cmyr, line 35: I'd prefer to use either use [`cfg_atr`](https://doc.rust-lang.org/reference/conditional-compilation.html#the-cfg_attr-attribute) to turn this on per-platform, or use `cfg` to only compile this on platforms where it's used.

(I don't like blanket `allow`s like this because you can't tell from lints when something is actually not used anymore)

cmyr (approved):
This makes sense, a few little thoughts inline!

druid-shell/src/util.rs
Maan2003: all other functions in util are also public but util is not public. I will make all these `pub(crate)`

PR #2002, opened 2021-10-05
allow specifying a default stroke color

james-lawrence:
usvg defaults to black with no option to override it.

allow users to specify either globally or per svg the default
stroke color.

### DIFF
--- a/druid/src/widget/svg.rs
+++ b/druid/src/widget/svg.rs
@@ -20,6 +20,7 @@ use tracing::{instrument, trace};
 use crate::{
     kurbo::BezPath,
     piet::{self, GradientStop, LineCap, LineJoin, LinearGradient, RadialGradient, StrokeStyle},
+    theme,
     widget::common::FillStrat,
     widget::prelude::*,
     Affine, Color, Data, Rect, UnitPoint,
@@ -29,6 +30,7 @@ use crate::{
 pub struct Svg {
     svg_data: SvgData,
     fill: FillStrat,
+    default_color: Option<Color>,
 }
 
 impl Svg {
@@ -39,12 +41,19 @@ impl Svg {
         Svg {
             svg_data,
             fill: FillStrat::default(),
+            default_color: None,
         }
     }
 
     /// Builder-style method for specifying the fill strategy.
     pub fn fill_mode(mut self, mode: FillStrat) -> Self {
         self.fill = mode;
+        self
+    }
+
+    /// Builder-style method for specifying the default stroke color.
+    pub fn default_stroke_color(mut self, color: Color) -> Self {
+        self.default_color = Some(color);
         self
     }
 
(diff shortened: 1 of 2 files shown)


druid/src/widget/svg.rs
```diff
@@ -427,8 +446,13 @@ fn transform_to_affine(t: usvg::Transform) -> Affine {
     Affine::new([t.a, t.b, t.c, t.d, t.e, t.f])
 }
 
-fn color_from_svg(c: usvg::Color, opacity: usvg::Opacity) -> Color {
-    Color::rgb8(c.red, c.green, c.blue).with_alpha(opacity.value())
+fn color_from_svg(c: usvg::Color, opacity: usvg::Opacity, default: &Color) -> Color {
+    let computed = Color::rgb8(c.red, c.green, c.blue);
+    match computed {
+        Color::BLACK => default.clone(), // usvg uses black as its default.
```
jneem, line 452: I'm confused by this, because it looks as though you'll end up replacing everything black by the default color, even if black is specified in the svg. Does `usvg` not give a way to tell the difference between, say, `<rect x="10" y="10" width="100" height="100" stroke="black"/>` and `<rect x="10" y="10" width="100" height="100"/>`?

Also, you called the method `default_stroke_color`, but it seems to affect fills also?

james-lawrence: > Does usvg not give a way to tell the difference between, say, <rect x="10" y="10" width="100" height="100" stroke="black"/>

correct. this is less than an ideal solution.

> Also, you called the method default_stroke_color, but it seems to affect fills also?

true, happy to change the name, its basically the default painting color I guess. =)

james-lawrence: for [reference](https://github.com/RazrFalcon/resvg/blob/d2ead624b23c995a966251531e734aed95c368a5/usvg/src/style.rs#L157)

jneem: Hm, is it possible for you to work around this another way, for example by just changing the svg files? I'm not super keen to add a feature that we know we can't implement correctly...

james-lawrence: not really no. I'm loading applications icons, which I don't have direct color over.

I don't think it'd be that hard to work with resvg to fix the underlying problem to allow us to remove the black check. I'm just going to have minimal free time soon so I can't spear head it.

this feature is also opt in. it has no change in current behavior unless you explicitly set the values.

does your objection disappear if usvg receives a fix?

jneem: If this really only substituted missing colors, I wouldn't have any objections.

Maan2003: I also need this, a solution would be to let user provide the `color_from_svg` function

jneem:
It looks like the plan is to support this using CSS upstream? Then I'll close this for now; feel free to re-open.

PR #2011, opened 2021-10-14
Fix missing import in gtk/window.rs

klemensn:
Cherry-picked from jpochyla/druid@a9522f5 .

Otherwise building jpochyla/psst on Linux or OpenBSD fails:
```
error[E0433]: failed to resolve: use of undeclared type `XcbHandle`
   --> /home/kn/src/druid/druid-shell/src/backend/gtk/window.rs:118:30
    |
118 |         RawWindowHandle::Xcb(XcbHandle::empty())
    |                              ^^^^^^^^^ not found in this scope
    |
```

This code was introduced in 2da53f040625b57e3e7741ceba1d261e19b07359 which
seems to target Windows and Mac but not Linux, so I assume that is why it
went unnoticed.

### DIFF
--- a/druid-shell/src/backend/gtk/window.rs
+++ b/druid-shell/src/backend/gtk/window.rs
@@ -44,7 +44,7 @@ use instant::Duration;
 use tracing::{error, warn};
 
 #[cfg(feature = "raw-win-handle")]
-use raw_window_handle::{HasRawWindowHandle, RawWindowHandle};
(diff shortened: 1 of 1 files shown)


Maan2003 (approved):
Thanks! I think CI doesnot test with raw-win-handle feature

klemensn:
Pardon the noise, I meant put this comment https://github.com/linebender/druid/pull/1993#issuecomment-967223703 here.

PR #2013, opened 2021-10-16
Fix Container2 function signatures in docs example

kingosticks:
It doesn't really matter but these had the wrong type in their signatures.


kingosticks:
I did this pr quickly through github Web interface, if I need to add a changelog entry etc I can do that when at a computer (tomorrow)

cmyr (approved):
Thanks!

PR #2017, opened 2021-10-23
Add set window topmost function.

zemelLeong:
This adds a function to set the window topmost.


CryZe:
This somehow breaks child windows for me. They don't refresh anymore at all (so one frame gets painted, but it then never changes)

zemelLeong:
> This somehow breaks child windows for me. They don't refresh anymore at all (so one frame gets painted, but it then never changes)

What's your code look like? I test with example `sub_windows`. Modify the main window text, and then the sub-windows text changed too.
image
image

xStrom:
What's the status here? Are we still moving this forward @zemelLeong?

If so, then please rebase on `master` to resolve the merge conflicts.

PR #1189, opened 2020-09-07
fix layout of image when either width or height is bounded, but not the other

JAicewizard:
This PR fixes #1184 with a more considerate layout function never taking +infinite in size.

PR #2018, opened 2021-10-23
Fix Fill options for non Fixed width options

fishrockz:
Hi, long time since I last contributed but I have been making a little GUI for something I was working on and notice that the image demo's filling seemed broken.

I have not go my head round the layout function so any feed back on how this should work would be appreciated :)

### DIFF
--- a/druid/src/widget/image.rs
+++ b/druid/src/widget/image.rs
@@ -197,12 +197,34 @@ impl<T: Data> Widget<T> for Image {
     ) -> Size {
         bc.debug_check("Image");
 
-        // If either the width or height is constrained calculate a value so that the image fits
-        // in the size exactly. If it is unconstrained by both width and height then use the fill
-        // strategy to determine the size.
+        // Let the fill strat determine the size of the widget as a function of the bc.max()
+        // and the original image size. But catch cases were they would return INFINITY
+        // and try to make a sensible area as a function of the fill strat.
         let max = bc.max();
         let image_size = self.image_size();
-        let affine = self.fill.affine_to_fill(max, image_size).as_coeffs();
+        let mut max_width = bc.max().width;
+        let mut max_height = bc.max().height;
+        let mut fill_strat = self.fill;
+        if max_width == f64::INFINITY && max_height == f64::INFINITY {
+            // If we are in a scroll box for width and height then fall back to
+            // original image size
+            max_height = image_size.height;
+            max_width = image_size.width;
+        } else if max_width == f64::INFINITY || max_height == f64::INFINITY {
+            // If we are in a scroll box for width or heigh but not both
+            // then give self.fill a box the size of the scroll box and let it decide how to
+            // fill it. Unless that would result in requesting INFINITY in which case fall back
+            // to contain.
+            fill_strat = match fill_strat {
+                FillStrat::FitWidth if max_width == f64::INFINITY => FillStrat::Contain,
+                FillStrat::FitHeight if max_height == f64::INFINITY => FillStrat::Contain,
+                FillStrat::Cover | FillStrat::Fill => FillStrat::Contain,
+                _ => fill_strat,
+            }
+        }
+        let affine = fill_strat
+            .affine_to_fill(Size::new(max_width, max_height), image_size)
+            .as_coeffs();
         // The first and forth elements of the affine are the x and y scale factor.
         // So just multiply them by the original size to get the ideal area based on `self.fill`.
         let mut width = affine[0] * image_size.width;
@@ -211,6 +233,7 @@ impl<T: Data> Widget<T> for Image {
         // would be over scaled to keep AR fixed then we just clip back to the `bc.max()`
         width = width.min(max.width);
         height = height.min(max.height);
+
         let size = Size::new(width, height);
         trace!("Computed size: {}", size);
         size
@@ -437,8 +460,10 @@ mod tests {
             2,
         );
 
-        let image_widget =
-            Scroll::new(Container::new(Image::new(image_data).fill_mode(FillStrat::Fill)).with_id(id_1)).vertical();
+        let image_widget = Scroll::new(
+            Container::new(Image::new(image_data).fill_mode(FillStrat::Fill)).with_id(id_1),
+        )
+        .vertical();
 
         Harness::create_simple(true, image_widget, |harness| {
             harness.send_initial_events();
@@ -465,8 +490,10 @@ mod tests {
             2,
         );
 
-        let image_widget =
-            Scroll::new(Container::new(Image::new(image_data).fill_mode(FillStrat::Fill)).with_id(id_1)).horizontal();
+        let image_widget = Scroll::new(
+            Container::new(Image::new(image_data).fill_mode(FillStrat::Fill)).with_id(id_1),
+        )
+        .horizontal();
 
         Harness::create_simple(true, image_widget, |harness| {
             harness.send_initial_events();


druid/src/widget/image.rs
```diff
@@ -190,19 +190,22 @@ impl<T: Data> Widget<T> for Image {
...
+        // The first and forth elements of the affine are the x and y scale factor.
+        // So just multiply them by the original size to get the ideal area based on `self.fill`.
+        let mut width = affine[0] * image_size.width;
+        let mut height = affine[3] * image_size.height;
+        if bc.is_width_bounded() {
+            width = width.min(max.width);
+        }
+        if bc.is_height_bounded() {
+            height = height.min(max.height);
+        }
```
JAicewizard, line 207: I think the strategy to use `affine_to_fill` is probably good. However this doesnt take into account when both sides are unbounded.

In my if else chain the first 2 should remain the same, since this takes into account being inside unbounded parents. This code should only apply if its bounded on both sides, since `affine_to_fill` is not made for unbounded constraints. In the last case where all sides are unbounded it should use the `bc.constrain(self.image_data.size())` code (Which could be simplified to `self.image_data.size()` since bc is unbounded). 

The reason the code is the way it is now, is so that we dont return an infinite size when unbounded in either direction. Since your code doesnt take into account being unbounded on either side we still need these safeguards using a different implementation.

druid/src/widget/image.rs
```diff
@@ -190,19 +190,22 @@ impl<T: Data> Widget<T> for Image {
         bc.debug_check("Image");
 
         // If either the width or height is constrained calculate a value so that the image fits
```
JAicewizard, line 192: `Use the fillstrat to determine the size of the image so that the the image will fit perfectly using this filstrat. If it is unconstrained by both width and height take the size of the image.`

JAicewizard (changes requested):

JAicewizard:
I made some comments, I think I understand the problem now. Thanks for your comments.

Your patch indeed fixes the problems outlined by you, but it doesnt keep working in the problems outlined by me. So I suggested some changes so that it will keep working in cases I had problems with when I changed the code.

fishrockz:
@JAicewizard  and @cmyr i really don't understand "However this doesnt take into account when both sides are unbounded." or "Since your code doesnt take into account being unbounded on either side we still need these safeguards using a different implementation."

Screenshot from 2021-11-01 13-51-35
Screenshot from 2021-11-01 13-51-57

Here are two screen shots of the PR with `scale down` behaving exactly as i think you want it to behave and the second with my use case not taking up `infinite space` but instead taking up the space left over in the window. In that use case the image adjusts to fit the window.

If I am still miss understanding can you please give me details of how to recreate the issue?

druid/src/widget/image.rs
fishrockz: The problem is that some of the fill strats do take the size of the image but other strategies do not so your alternative is self contradicting.

druid/src/widget/image.rs
fishrockz: > Since your code doesnt take into account being unbounded on either side we still need these safeguards using a different implementation.

But this code never returns a unbounded result. and some of the fill strategies call for more than the image. So letting `affine_to_fill` determine the result dose all the work you are asking for when you want it and other things for the other fill strategies.

`self.fill.affine_to_fill(max, image_size)` takes the max size the widget may be to not change the window size, this gives it a `max` to work with. then it also knows the original image size, `image_size`. so for `no scale` or `scale down` the max it asks for the smallest of original image or the space available.

`affine_to_fill` is doing the guarding its self in the `non-constrained` case and the `fully constrained` case

fishrockz: I was looking at this code and realised I can simplify it without changing the behaviour of the demo at all. Also by having less branching it makes it simpler for the cpu so might be a tiny tiny bit faster.

I have added some comments that might make it easer to under stand as well.

JAicewizard:
> Here are two screen shots of the PR with scale down behaving exactly as i think you want it to behave and the second with my use case not taking up infinite space but instead taking up the space left over in the window. In that use case the image adjusts to fit the window.

This is because the examples constrains the image within the window. If you dont do that by using something scroll able it will give errors. You can check if the given bounds are constrained using `bc.is_{height/width}_bounded`, I am using that literal definition.

See https://github.com/linebender/druid/pull/1189 for fix and issue.

druid/src/widget/image.rs
JAicewizard: oh I see, I mean the size of the underlying image data.

druid/src/widget/image.rs
JAicewizard: >But this code never returns a unbounded result

It does, when the constraints are unbounded the resulting size of the image could be infinitely big. Since there are no checks for this this infinity will traverse it's way through the calculations into the final size.

> or the space available.

that's the problem, the space available could be infinite in either direction.

fishrockz:
> > Here are two screen shots of the PR with scale down behaving exactly as i think you want it to behave and the second with my use case not taking up infinite space but instead taking up the space left over in the window. In that use case the image adjusts to fit the window.
> 
> This is because the examples constrains the image within the window. If you dont do that by using something scroll able it will give errors. You can check if the given bounds are constrained using `bc.is_{height/width}_bounded`, I am using that literal definition.
> 
> See #1189 for fix and issue. (the one where I changed the code to what it is now)

That makes sense, thanks for explaining the case. yes i can see that the original didnt do a good job in that case but this new code will work according to the flutter guidance in that case AFAICT.

I don't know what flutter dose in that use case with the fill strategy but the new code should behave sensibly for `Contain`,   (`FitHeight`, or `FitWidth` depending on which scroll bars are used.) and `None` and  `ScaleDown,`

But the fill strategies  `Cover` `Fill` and `FitHeight` `FixWidth` depending on scroll bars dont really make sense for that use case so it makes sense that they don't work nicely.

I think its a shame to rule out a hole class of UI because some of the fill strats dont' make sense in that use case. If the image gets too far out of AR or toooo big then we could clip them?

Now I finally under stand your use case, i might be able think of something else..

fishrockz:
This is what i meant by

> I have not go my head round the layout function so any feed back on how this should work would be appreciated :)

I have not seen a case were bc.max() did not give a non integer or non sensible value. Even when non contained it still gives sensible values most of the time. it would be easy to check `if width == F64::inf` and then just fall back to width = image.width() or something like that etc. that would fix your use case and give flexiblity to others.

@cmyr / @JAicewizard  are there any other `supprising` values of bc.max()?

cmyr:
Okay, I wrote a thing here but I think it's actually not the most useful response.

Instead, I want to just point to the flutter docs for [`BoxFit`](https://api.flutter.dev/flutter/painting/BoxFit-class.html) and [`FittedBox`](https://api.flutter.dev/flutter/widgets/FittedBox-class.html). There are some other useful links from there.

Some things to note:

- there is some confusion in the code about the relationship between the box constraints and the fill strategy. When do we just use the box constraints? When do we use the fill strategy? etc.
- importantly, the fill strategy should not change the size of the widget. It should only effect how the image is fit within the space that has been chosen.
- One way to think about this: first we pick a size, and then we apply the fill start.
- in flutter, the fill strat is applied by a distinct widget. It may be useful for us to consider this approach, since it clarifies the relationship between fillstrat and layout? Basically the `FittedBox` widget first passes through the layout constraints to the child, and then it applies the fillstrat to the result, and generates a transform that is applied to the child. This is a bit complicated for us since we don't support transforming arbitrary widgets, since it gets weird if they accept input?


Anyway: my basic position would be to emulate what flutter is doing, as closely as possible. Without adding a new widget, this means:

- first just call `bc.constrain(image.size())` to get the output size
- have a function like [`applyBoxFit`](https://api.flutter.dev/flutter/painting/applyBoxFit.html) that takes image size and the calculated widget size and returns two rects describing the mapping of source->destination
- use this to determine what to paint. You can find the `paintImage` method in the flutter docs


What is important here is that we separate the applying of the constraints from the determination of the widget size. **It is not the responsibility of the image widget to choose a size, outside of ensuring that it satisfies the input constraints**. If you need to do fancier things to grow or shrink your widget in a given scenario, you should be accomplishing that with some other layout widget.

fishrockz:
Now I understand the case were you can have infinite bc.max() I think i have caught this edge case and do something sensible for it in the latest version of this PR.

But i had not understood the subtlety of flutter and druids layout so while this now restores the old functionality while also addressing the infinite edge case I'm not sure that the original behaviour was right.

But @cmyr how would you ever get the image to be big? even if it was in a layout widget the current code in master would always just return the image's max? would a layout widget just ignore the image widget and force the image bigger? what are the rules for the images layout function in that case? dose one of the examples cover something like this? even with a completely different widget?

fishrockz:
Thinking about this a bit more. 

Its worth noting that the code in master dose `force` the image larger if the widget is `constrained`. To me constrain dose not mean force to this size but no bigger, so in that case we should have this layout just always return the minimum between bc.max() and image.size()?

cmyr:
The parent layout widget would only influence the child by modifying the constraints. For instance if you had received constraints that were (0,0) (500,500) you could output 'tight' constraints that were (500,500), (500,500). This is achievable by built in widgets, and if you needed more complicated conditional logic you could have a custom container widget that implements it. 


I definitely agree that the original (and all subsequent!) impls of the image widget, in druid, have had incomplete implementations of the layout fn.

I think the best approach is to step back and try to just think this through systematically, instead of trying to achieve some specific set of layout objectives.

fishrockz:
> The parent layout widget would only influence the child by modifying the constraints. For instance if you had received constraints that were (0,0) (500,500) you could output 'tight' constraints that were (500,500), (500,500). This is achievable by built in widgets, and if you needed more complicated conditional logic you could have a custom container widget that implements it.

I don't understand this. Are the responsibilities of the layout fn documented some were, the trait docs were a bit light IIRC..

cmyr:
Sorry, I was composing that on my phone and I wasn't done, I guess it posted it after some timeout? Not sure.

The way that a parent widget influences a child's layout is by modifying the `BoxConstraints`, which are represented as a pair of (min, max) sizes. If these are equal, then we call the constraints "tight"; that is, ther is only one size that fits them. 

For a good overview of how this works, check out the flutter docs for [`BoxConstraints`](https://api.flutter.dev/flutter/rendering/BoxConstraints-class.html).

JAicewizard:
> Now I finally under stand your use case, i might be able think of something else..

You dont have to, your code works perfectly fine when everything is constrained and we can use it in that case. The only case it doesnt cover is when either sidie (or both ofc) is unconstrained. This is what my code handles, so you can combine them as I described in my review.

> It is not the responsibility of the image widget to choose a size, outside of ensuring that it satisfies the input constraints.

Could you maybe explain a bit more? The I think the widget should chose a size it will actually use, aka not take more space than actually needed. And this being an image widget I would also say that it is usefull to take the maximum size we will end up using (with accordance to the fillstrat).

druid/src/widget/image.rs
```diff
@@ -189,20 +189,44 @@ impl<T: Data> Widget<T> for Image {
...
-        } else if bc.is_height_bounded() && !bc.is_width_bounded() {
-            let ratio = max.height / image_size.height;
-            Size::new(ratio * image_size.width, max.height)
-        } else {
-            bc.constrain(self.image_data.size())
-        };
+        let mut max_width = bc.max().width;
+        let mut max_height = bc.max().height;
+        let mut fill_strat = self.fill;
+        if max_width == f64::INFINITY && max_height == f64::INFINITY {
```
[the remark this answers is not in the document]
fishrockz: These detect times when it would note be safe to use the max so they work quite differently.

druid/src/widget/image.rs
```diff
@@ -189,20 +189,44 @@ impl<T: Data> Widget<T> for Image {
...
+            // If we are in a scroll box for width or heigh but not both
+            // then give self.fill a box the size of the scroll box and let it decide how to
+            // fill it. Unless that would result in requesting INFINITY in which case fall back
+            // to contain.
+            fill_strat = match fill_strat {
+                FillStrat::FitWidth if max_width == f64::INFINITY => FillStrat::Contain,
+                FillStrat::FitHeight if max_height == f64::INFINITY => FillStrat::Contain,
+                FillStrat::Cover | FillStrat::Fill => FillStrat::Contain,
+                _ => fill_strat,
+            }
```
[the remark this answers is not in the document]
fishrockz: I don't believe this does behave very much like your code out side of the options that fall back to Contain.

fishrockz:
@cmyr If i understand correctly https://github.com/fishrockz/druid/commit/a123167d2c3c82ae2f89fb1cfc35871b464e872c would be much more idiomatic of fluter. BUT it will cause issues with the srollbar/srollbox with some fill options in paint (paint will crash as @JAicewizard  points out). Am I right in thinking that flutter would error out at the layout stage in this case? https://flutter.dev/docs/testing/common-errors#vertical-viewport-was-given-unbounded-height this seems to imply that flutter just refuses to layout a widget if something silly was going to happen? Our layout dose not provide a option to return a error so I'm not sure what to do with that case, I presume you dont want the layout to panic? I can detect the issue at paint time and paint a big error message instead of crashing?

I think the most pragmatic option is to do the min/max thing but with some logic to detect `is_infinite()` and expand as sensibly as possible in that case? `were sensible is what we think is least bad rather than idiomatic?`

druid/src/widget/image.rs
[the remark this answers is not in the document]
JAicewizard: Ah yes I see I missed 2 cases.

cmyr:
if the input constraints are not bounded, the image widget should ignore `FillStrat` completely and just choose its 'natural' size.

It is very possible in druid to produce a layout that is not meaningful/is poorly formed, like when you `expand()` inside of a scrollview. In these cases we tend to log an error and continue. In the future it would definitely be possible to have more debug tools available during layout for diagnosing these sorts of problems.

fishrockz:
> It is very possible in druid to produce a layout that is not meaningful/is poorly formed, like when you `expand()` inside of a scrollview. In these cases we tend to log an error and continue. In the future it would definitely be possible to have more debug tools available during layout for diagnosing these sorts of problems.

Right but AFAICT the reason that @JAicewizard updated this widget last was that druid was not `logging a error and continuing` but crashing horribly. In that case what does `logging a error and continuing` look like for a image? The edge case i am trying to deal with in my mind is `image in a size.expand() in a vertiacal scroll bar` this dose not make sense but we dont want druid to crash during the paint so what would you like to happen?

if layout returns the minimal box constraint which in that case would be fixed width but inf hieight, then at paint we detect the inf and just paint everything red or just return early without doing anything to the context? Would we log some errors in the paint function or the layout function? do you have a example of this?

fishrockz:
Also expanding on the `image` inside a expanded `size` in side a vertical scroll bar.

If FillSrat is `Fill` then clearly the right answer is to gracefully error. but if the FillStrat is to FillWidth,Contain,etc. then i think it makes much more sence to return a layout size that makes sense and allow for the image to fill the width and enough of the inf height to display the image correctly.

I don't really understand flutter enough to know what the write answer is.

Also in general:

If I'm only ever going to display so much of the screen because I have `ScaleDown` or `None` for the fill strat should the image's layout function respect the min of the tight constraint and return a size that is bigger than it will ever paint or just return the max that it will ever paint at layout time?

xStrom:
What's the status of this PR? Are we still moving foward with it @fishrockz?

If so, then please rebase on `master` to resolve the merge conflicts.

fishrockz:
@xStrom I'm happy to fix this, I got blocked as @cmyr wants this to be as much like flutter as possible but I got blocked on actually testing flutter as the docs were not comprehensive enough.

I'm out atm but I'll have a look when I get home

xStrom:
Excellent, having this rebased will help me test the code too.

I'll read through the earlier messages here tomorrow to better understand the history of concerns.

fishrockz:
@xStrom i have rebased to master

If you check out master and run `cargo run --example image --features="image png"` if you untick the fixed width then no mater what fill strat you pick you can not get the image to fill the space. This is a fetcher i find helpful.

But if you checkout this branch and run `cargo run --example image --features="image png"` and untick the fixed width then you can get the old behaviour from before with None or scale down but you can also use the other fill strategies effectively and how they worked when they were fist merged before JAicewizard changed the behaviour in PRs after i introduced this widget.

It seems cmyr wanted the flutter behaviour than rather than ether of our behaviours.

I did try to get closer to the flutter behaviour, i even have another non pushed branch trying to get there but without a working flutter system to test ageist i was struggling to understand exactly what flutter did.

PR #2020, opened 2021-10-27
Make SvgData size methods public

rosehuds:
I found them useful 🤷

also moved the private method downwards to preserve "public methods at the top"

### DIFF
--- a/druid/src/widget/svg.rs
+++ b/druid/src/widget/svg.rs
@@ -144,22 +144,8 @@ impl SvgData {
         }
     }
 
-    /// Calculates the transform that should be applied first to the svg path data, to convert from
-    /// image coordinates to piet coordinates.
-    fn inner_affine(&self) -> Affine {
-        let viewbox = self.viewbox();
-        let size = self.size();
-        // we want to move the viewbox top left to (0,0) and then scale it from viewbox size to
-        // size.
-        // TODO respect preserveAspectRatio
-        let t = Affine::translate((viewbox.min_x(), viewbox.min_y()));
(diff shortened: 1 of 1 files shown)


druid/src/widget/svg.rs
```diff
@@ -144,22 +144,8 @@ impl SvgData {
...
-        // TODO respect preserveAspectRatio
-        let t = Affine::translate((viewbox.min_x(), viewbox.min_y()));
-        let scale =
-            Affine::scale_non_uniform(size.width / viewbox.width(), size.height / viewbox.height());
-        scale * t
-    }
-
     /// Get the viewbox for the svg. This is the area that should be drawn.
-    fn viewbox(&self) -> Rect {
+    pub fn viewbox(&self) -> Rect {
```
cmyr, line 148: I would call these methods `svg_viewbox` and `svg_size` respectively, and clarify in the docs that these are the sizes reported for the underlying svg image, not for the widget.

cmyr (approved):
Thanks, this seems useful.

A suggestion inline, but otherwise I'm happy to see this merged :)

druid/src/widget/svg.rs
rosehuds: I should clarify this - `SvgData` seems to me to only deal with literally the data of the SVG it was created from, so it is fairly clear that "size" refers to the SVG - the type serves no other purpose. All widget interactions are done by `Svg`.

cmyr: thanks for the clarification!

PR #2023, opened 2021-10-30
add method label_left/right/up/down for trait WidgetExt

remme123:
for easily creating labeled widgets


cmyr (commented):
Thanks for the PR! I appreciate the, but I don't think this is a great fit for `WidgetExt`; the methods here are generally things that  are frequently wanted on all widgets, and adding a label feels too specific.

remme123:
Ok! I will close this PR.

PR #2027, opened 2021-11-04
Prevent reentrancy in Windows keyboard event handling

### LINKED ISSUE #2026: Reentrancy when handling keyboard events on Windows
Steps to reproduce:

1. On Windows 10, start the Narrator screen reader.
2. Run any program that uses druid-shell. I've tested with the Druid calc example and the druid-shell shello example.
3. Press Alt+Tab.

The following error is displayed in the console window:

```
druid_shell::backend::windows::window: failed to borrow WndState at druid-shell\src\backend\windows\window.rs:805:22
```

Diagnosis: Because of Narrator's keyboard event preprocessing, the application doesn't receive the key-down event for Alt until the Tab key is pressed. So Windows sends the `WM_KILLFOCUS` message while the application is handling the `WM_KEYDOWN` message for Alt. Then, `druid_shell::backend::windows::keyboard::is_last_message` calls `PeekMessageW`, and during that call, Windows handles the pending `WM_KILLFOCUS` message.

mwcampbell:
fixes #2026

### DIFF
--- a/druid-shell/src/backend/windows/keyboard.rs
+++ b/druid-shell/src/backend/windows/keyboard.rs
@@ -89,11 +89,14 @@ const SCAN_MASK: LPARAM = 0x1ff_0000;
 /// When this function returns `false`, there is another message in the queue
 /// with a matching scan code, therefore it is reasonable to stash the data
 /// from this message and defer til later to actually produce the event.
-unsafe fn is_last_message(hwnd: HWND, msg: UINT, lparam: LPARAM) -> bool {
+pub(super) unsafe fn is_last_message(hwnd: HWND, msg: UINT, lparam: LPARAM) -> bool {
     let expected_msg = match msg {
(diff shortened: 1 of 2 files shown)


druid-shell/src/backend/windows/keyboard.rs
```diff
@@ -547,17 +550,17 @@ impl KeyboardState {
     /// is likely low, though.
     pub(crate) unsafe fn process_message(
         &mut self,
-        hwnd: HWND,
         msg: UINT,
         wparam: WPARAM,
         lparam: LPARAM,
+        is_last: bool,
```
raphlinus, line 556: I'm wondering if "is_last_key" might be a better name, because at this place the context that it's keyboard specific is lost.

raphlinus (approved):
This looks like a good fix, thanks!

PR #2035, opened 2021-11-10
remove dead/unused code.

james-lawrence:


james-lawrence:
as an aside, outside of druid APIs the wayland code is fairly functional. mostly keyboard (repeating keys) and output selection remains.

druid-shell/src/backend/wayland/mod.rs
```diff
@@ -22,7 +22,6 @@ mod events;
 pub mod keyboard;
 pub mod menu;
 pub mod pointers;
-pub mod screen;
```
Maan2003, line 25: wait you can't just remove screen, druid shell depends on this and will not compile

james-lawrence: except I did, and it does.... =)

james-lawrence: ahh i lied. I was building on a different branch. will readd the method. =(

Maan2003: I think it is fine to leave it as stub.

james-lawrence: fixed up and pulled in 1 more small change that fixes an issue with the global id truly being a constant. (w/ a value of 1)

PR #2048, opened 2021-11-20
Upgrade tracing-subscriber

jplatte:
I use `tracing-subscriber` directly in my project and don't like duplicate dependencies.

### DIFF
--- a/druid-shell/Cargo.toml
+++ b/druid-shell/Cargo.toml
@@ -99,7 +99,7 @@ features = ["Window", "MouseEvent", "CssStyleDeclaration", "WheelEvent", "KeyEve
 piet-common = { version = "=0.5.0", features = ["png"] }
 static_assertions = "1.1.0"
 test-log = { version = "0.2.5", features = ["trace"], default-features = false }
-tracing-subscriber = "0.2.15"
+tracing-subscriber = { version = "0.3.2", features = ["env-filter"] }
 unicode-segmentation = "1.7.0"
 
 [build-dependencies]
--- a/druid-shell/Cargo.toml
+++ b/druid-shell/Cargo.toml
@@ -98,7 +98,7 @@ features = ["Window", "MouseEvent", "CssStyleDeclaration", "WheelEvent", "KeyEve
 [dev-dependencies]
 piet-common = { version = "=0.5.0", features = ["png"] }
 static_assertions = "1.1.0"
-test-env-log = { version = "0.2.5", features = ["trace"], default-features = false }
+test-log = { version = "0.2.5", features = ["trace"], default-features = false }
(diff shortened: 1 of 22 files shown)


ngugcx:
Is there any special reason why Druid depends on tracing?

jplatte:
It's a convenience thing, `AppLauncher::log_to_console` which uses `tracing-subscriber` on for native targets and `tracing-subscriber` + `tracing-wasm` on wasm.

I wouldn't personally mind if it was instead removed though 🤷🏼

sjoshid:
Im assuming the problem here is that you are using version 0.3.2 in your project while druid uses version 0.2.15. So the PR makes druid use version 0.3.2. Now both use version 0.3.2 and problem solved. 
But wouldnt you run into same issue if druid uses version > 0.3.2 in future?

jplatte:
There's not really a big *problem* here, it's just that build times get longer when you have duplicates of the same crate in the dependency tree. Yes of course druid should upgrade tracing-subscriber again in the future (when there's a breaking change release).

sjoshid:
The point that I was trying to make was Druid is not going to update to latest and greatest version to satisfy a downstream client￼. 
I empathize with your dependency issues but the real fix would be to completely remove it.

druid-shell/Cargo.toml
```diff
@@ -98,8 +98,8 @@ features = ["Window", "MouseEvent", "CssStyleDeclaration", "WheelEvent", "KeyEve
 [dev-dependencies]
 piet-common = { version = "=0.5.0", features = ["png"] }
 static_assertions = "1.1.0"
-test-env-log = { version = "0.2.5", features = ["trace"], default-features = false }
-tracing-subscriber = "0.2.15"
+test-log = { version = "0.2.5", features = ["trace"], default-features = false }
+tracing-subscriber = { version = "0.3.2", features = ["env-filter"] }
```
Maan2003, line 102: why did you add `env-filter` ?

jplatte: It's required by test-log. On 0.2.x, it was a default feature, now it isn't anymore: https://github.com/tokio-rs/tracing/releases/tag/tracing-subscriber-0.3.0

jneem:
We can certainly discuss removing the dependency, but at least since it's there for now, we may as well keep it up to date. (Also, annoying that `test-env-log` broke the build with a semver-compatible update, but :shrug:)

jneem (approved):

jplatte:
> We can certainly discuss removing the dependency, but at least since it's there for now, we may as well keep it up to date. (Also, annoying that `test-env-log` broke the build with a semver-compatible update, but :shrug:)

It only broke things due to `-D warnings`.

PR #2050, opened 2021-11-22
Support selected menu items with GTK backend

CryZe:
This simply constructs a `CheckMenuItem` instead of a `MenuItem` in case the menu item is selected.

### DIFF
--- a/druid-shell/Cargo.toml
+++ b/druid-shell/Cargo.toml
@@ -98,7 +98,7 @@ features = ["Window", "MouseEvent", "CssStyleDeclaration", "WheelEvent", "KeyEve
 [dev-dependencies]
 piet-common = { version = "=0.5.0", features = ["png"] }
 static_assertions = "1.1.0"
-test-env-log = { version = "0.2.5", features = ["trace"], default-features = false }
+test-log = { version = "0.2.8", features = ["trace"], default-features = false }
 tracing-subscriber = "0.2.15"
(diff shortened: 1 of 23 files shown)


jneem:
Did you push the right commit? I'm not seeing GTK changes...

CryZe:
Yes, there's 2 commits. The second fixes a CI issue.

jneem:
Ah, sorry. For some reason, I managed to get github to only show me the changes of the second commit...

jneem:
Hm, I'm playing around with the multiwin example and the behavior of unselected items is a bit strange: sometimes they show an empty checkbox and sometimes they show nothing. It seems like to match the expected GTK behavior, the druid API will need to use a 3-state enum instead of a bool, right? I mean, GTK distinguishes between: selected (box with a tick), selectable but not selected (empty box), and not selectable (nothing).

CryZe:
Mmh I haven't seen any empty checkboxes, is there a way to consistently reproduce it? Also yeah technically having 3 cases would be ideal.

jneem:
This is what it looks like when I run the multiwin example 

https://user-images.githubusercontent.com/1448113/143067063-98e69a00-63a6-45a7-98e3-5478f1ca1de5.mp4

CryZe:
I believe this just go superseded by: https://github.com/linebender/druid/pull/2251

PR #2051, opened 2021-11-26
Update set_position and get_position docs.

jneem:
This doesn't yet fix the fact that they're using absolute positions on windows (so as not to conflict with #2042).


Maan2003 (approved):
:+1:

PR #2057, opened 2021-11-30
wayland: attempt to create a working clipboard

james-lawrence:
while these changes don't fully implement a working copy/paste.
I have confirmed that when Clipboard::get_string is invoked the
contents from the system clipboard are returned.

### DIFF
--- a/druid-shell/src/backend/shared/xkb/keycodes.rs
+++ b/druid-shell/src/backend/shared/xkb/keycodes.rs
@@ -8,6 +8,7 @@ use super::xkbcommon_sys::*;
 pub fn map_key(keysym: u32) -> Key {
     use Key::*;
     match keysym {
+        XKB_KEY_v => Key::Character("v".to_string()),
         XKB_KEY_BackSpace => Backspace,
         XKB_KEY_Tab | XKB_KEY_KP_Tab | XKB_KEY_ISO_Left_Tab => Tab,
         XKB_KEY_Clear | XKB_KEY_KP_Begin | XKB_KEY_XF86Clear => Clear,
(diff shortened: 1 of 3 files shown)


james-lawrence:
as an aside the reason I think paste isn't working entirely with this change is because of which key codes are being returned... just havent had a chance to dig into it yet to confirm.

Maan2003 (approved):
Thanks!

druid-shell/src/backend/wayland/clipboard.rs
```diff
@@ -13,45 +13,285 @@
...
+        }
+
+        return None;
+    }
+}
+
+impl TryFrom<(&wl::Display, &wl::GlobalManager)> for Manager {
+    type Error = waylanderr::Error;
+
+    fn try_from(gm: (&wl::Display, &wl::GlobalManager)) -> Result<Self, Self::Error> {
```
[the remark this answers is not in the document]
Maan2003: It doesn't make sense conceptually, it is not trying to **convert** a `Display` and `GlobalManager`, it is creating a new clipboard manager for a Display using a Global Manager

druid-shell/src/backend/shared/xkb/keycodes.rs
```diff
@@ -8,6 +8,7 @@ use super::xkbcommon_sys::*;
 pub fn map_key(keysym: u32) -> Key {
     use Key::*;
     match keysym {
+        XKB_KEY_v => Key::Character("v".to_string()),
```
james-lawrence, line 11: so this fixes paste..... but it implies deeper issues in the keymapping logic for x11/wayland

Maan2003: We can have keep this for now

PR #2064, opened 2021-12-02
Allow dynamic LabelText closure to return different types

jplatte:
This should decrease copying (especially if the user-provided closure returns `Arc<str>` or something that can be converted to it without copying) and make comparisons in `resolve` faster.


jplatte:
CI failure looks unrelated.

Maan2003:
rebase should fix CI

Maan2003 (approved):
Thanks

jplatte:
Oh, CI failure *is* related. I'll update the examples tomorrow.

jpochyla:
I think one downside is that `dynamic`s will often return `format!()`ted strings, and that would have to reallocate.

Maan2003:
> I think one downside is that dynamics will often return format!()ted strings, and that would have to reallocate.

This already happens right? We always store data as `Arc<str>`

jplatte:
Updated the one example affected by this and added a changelog entry.

PR #2068, opened 2021-12-03
Remove unsound Send & Sync impls

### LINKED ISSUE #2067: Clippy warns about unsound `Send` implementation
Nightly clippy raises the following warning:

```
warning: this implementation is unsound, as some fields in `WindowState` are `!Send`
    --> druid-shell/src/backend/gtk/window.rs:1271:1
     |
1271 | unsafe impl Send for WindowState {}
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: `#[warn(clippy::non_send_fields_in_send_ty)]` on by default
note: the type of field `window` is `!Send`
    --> druid-shell/src/backend/gtk/window.rs:173:5
     |
173  |     window: ApplicationWindow,
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^
     = help: use a thread-safe type that implements `Send`
note: the type of field `drawing_area` is `!Send`
    --> druid-shell/src/backend/gtk/window.rs:180:5
     |
180  |     drawing_area: DrawingArea,
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^
     = help: use a thread-safe type that implements `Send`
note: the type of field `surface` is `!Send`
    --> druid-shell/src/backend/gtk/window.rs:191:5
     |
191  |     surface: RefCell<Option<Surface>>,
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     = help: use a thread-safe type that implements `Send`
note: the type of field `handler` is `!Send`
    --> druid-shell/src/backend/gtk/window.rs:196:5
     |
196  |     pub(crate) handler: RefCell<Box<dyn WinHandler>>,
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     = help: use a thread-safe type that implements `Send`
note: the type of field `deferred_queue` is `!Send`
    --> druid-shell/src/backend/gtk/window.rs:201:5
     |
201  |     deferred_queue: RefCell<Vec<DeferredOp>>,
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     = help: use a thread-safe type that implements `Send`
note: the type of field `parent` is `!Send`
    --> druid-shell/src/backend/gtk/window.rs:206:5
     |
206  |     parent: Option<crate::WindowHandle>,
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     = help: use a thread-safe type that implements `Send`
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#non_send_fields_in_send_ty
```

jplatte:
The unsound impls weren't really being relied upon because it was ensured that `WindowState` was never truly moved or accessed across thread boundaries, but this PR changes things such that they're no longer required, preventing the impls from causing issues in the future.

Fixes #2067.


jplatte:
I made sure that the hello example still starts up successfully on Linux. Should I test anything else to make sure the panic is not hit?

jplatte:
Oh, seems like I broke something again. Will look into it in a bit.

Maan2003:
> I made sure that the hello example still starts up successfully on Linux. Should I test anything else to make sure the panic is not hit?

try timer, sub window, slow function and async_event examples

jplatte:
Okay, seems like this approach doesn't work because the idea is really that any thread can register a callback with `IdleHandle::add_idle_callback`, and that callback will end up being called on the UI thread. I think the `unsafe impl`s could potentially be removed by splitting `WindowState` into a `Send` part and a non-`Send` part, I'll try that in the coming days.

jneem:
I'm a bit confused by this clippy lint, to be honest. Isn't the whole point of `unsafe impl Send` to say "I know that some of the fields are `!Send`, but I promise to use them safely"? Is there any usage of `unsafe impl Send` that doesn't trigger the lint?

jplatte:
Makes sense. I was also a bit confused, but at the same time I thought that this impl is rather dangerous. I'll close this for now since I won't have time to try the alternative approach I mentioned above in the near future.

PR #2070, opened 2021-12-03
wayland: enable buffer recreation to always be allowed.

james-lawrence:
- fixes a seg fault when resizing the current window.
- fixes a bug where even if no seg fault happens we skip painting.


Maan2003 (approved):
:tada:

Issue #933, opened 2020-05-15
X11 lacks menu support


Finnerale:
As mentioned in #860 we will likely have to implement menus our selves.

psychon:
From https://github.com/xi-editor/druid/pull/599

> menus: I think we're going to have to draw these ourselves for X11 (???).
> There may be some other way to do it by requesting the window manager to do it
> for you, but I didn't dig too deep into it yet. Suggestions welcome here.

Nope, you have to implement menus yourself in X11. Menus are windows with their `override-redirect` property set to `true`. This means that the window manager completely ignores these windows. Additionally, one usually grabs all input with a `GrabPointer` request, but I am not actually completely sure.

(File dialog support falls in a similar category: BYOFD - bring your own file dialog)

fxdave:
Another thing to consider:
 - https://gitlab.com/vala-panel-project/vala-panel-appmenu/

Menus should be hidden when there is a global app menu like on mac or gnome-hud, unity, etc... on Linux.

derekdreery:
The wayland backend will need menus as well.

cmyr:
the wasm backend also needs a menus solution.

PR #2071, opened 2021-12-04
wayland: cleanup popup implementation refactor.

james-lawrence:
only piece I have left at this point is the screen implementation which i need to make some improvements to before I open a PR it.

after that I'll do a rebase against master for the wayland branch. I *think* at that point it'll be close enough to mainline. PRs after the screen stuff should be minor enhancements, bug fixes (I have 1 or 2 things I might fix), or code cleanup.

still a few small odds an ends to make it feature complete (but i havent needed them so... here we are. =)).
- bunch of the window builder API stuff needs to be wired up, titles, resizable, etc. these are small though and should be easily implemented as needed.
- tooltips, and modal window levels need to be implemented. but they should be identical to the dropdowns with a different positioning configuration.
- menus/dialog stuff. these can be shared with x11 i think?
- keymapping issues (copy/paste not generating the correct keystroke of x11/wayland).

should I create issues for these things?

### DIFF
--- a/druid-shell/src/backend/wayland/application.rs
+++ b/druid-shell/src/backend/wayland/application.rs
@@ -181,13 +181,7 @@ impl Application {
         let (outputsaddedtx, outputsaddedrx) = calloop::channel::channel::<Output>();
 
         let globals = wl::GlobalManager::new_with_cb(&attached_server, {
-            move |event, registry, data| {
-                tracing::debug!(
-                    "global manager event received {:?}\n{:?}\n{:?}",
-                    event,
-                    registry,
-                    data
-                );
+            move |event, registry, _ctx| {
                 match event {
                     wl::GlobalEvent::New {
                         id,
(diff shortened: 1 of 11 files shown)


Maan2003:
I also think this is ready to merge into master! It is pretty close to the x11 backend.

>  bunch of the window builder API stuff needs to be wired up, titles, resizable, etc. these are small though and should be easily implemented as needed.

yes should create an issue with `missing` `shell/wayland` labels

>  tooltips, and modal window levels need to be implemented. but they should be identical to the dropdowns with a different positioning configuration.

the positioning is exactly same :thinking: 

> menus/dialog stuff. these can be shared with x11 i think?

we can just update the title/labels for #933 and #936

> keymapping issues (copy/paste not generating the correct keystroke of x11/wayland).

Yes should create an issue. I think it is for all `ctrl` shortcuts

james-lawrence:
> the positioning is exactly same

oh? then its really easy. ;)

PR #2079, opened 2021-12-06
wayland: rebased and merge into master

james-lawrence:
pull all the changes from wayland into master since we're down to just a few missing pieces in wayland (mostly pieces also missing within the x11 backend) time to let it soak in master.


Maan2003:
We also want CI for wayland backend, probably just need to copy
https://github.com/linebender/druid/blob/2d80ca04312746567219ebed2aacc0ee5d72b6c7/.github/workflows/ci.yml#L113-L153

james-lawrence:
agreed, can do it as part of the next PR if you want. CI env is very different than my environment, which isn't an issue but don't want to think about it for this one. =)

Maan2003:
I can do it tomorrow. wanted it because `cargo clippy --features wayland --no-default-features` generates 107 warnings :sweat_smile:

Maan2003:
can you add the following commits:
Maan2003@33b31ba08429a2c6963372ae3c7a866ee396a471 - add CI for wayland
Maan2003@3fe57bb23da4126ad243e77e396e7f4bfeebdf40 - fixes bug in build.rs
Maan2003@8a5884c6eaabc94a5f428175ae41c8d8605bcad2 - disable ApplicationExt static assertion for wayland
Maan2003@3c9591c47a73e3b3c059b27d03d00ccb5cd9dbb2 - fixes most of clippy warnings

Maan2003:
also squash 11d72d8f027dd6a2fcea23691a3a3267fcb88ae7 and c3d7fdbe592c8cde6695ff346e034c5b14457222

james-lawrence:
@Maan2003 all green! =)

emersion (changes requested):
With my wlroots and Wayland maintainer hat on, I'm NACKing this pull request. Usage of wlr-output-management protocol is not suitable for Wayland clients like druid. We _will_ break your client if you go down this path.

james-lawrence:
I talked with @emersion the path hes referring to is strictly how we're accessing screen dimensions, there is no alternative path but its not critical and the code will degrade gracefully. basically only screens output will be impacted. not really worried about it.

emersion:
Sorry, it seems like there is a misunderstanding. Please remove all wlr-output-management usage. Regular clients like druid must not use it.

> the code will degrade gracefully

No. In the future, the compositor may disconnect the client with a fatal error if it tries to bind to the wlr-output-management interface.

james-lawrence:
@emersion degrade gracefully means 'check if manager object exists' if it doesn't move on. leave it alone. not your project and I already have plans to remove it. just not in this PR.

emersion:
> degrade gracefully means 'check if manager object exists' if it doesn't move on

Then be aware that your client _will_ break, ie. will be disconnected with a fatal error. We will advertise the manager global and kill any client that tries to use it.

I will not leave it alone, because it's a _very bad_ thing that regular clients use our privileged protocols.

james-lawrence:
@emersion we talked, your concerns were understood. using unstable protocols by default comes with that risk. I'd expect the same potential risks from any of the unstable protocols. this isn't news to anyone. same with the xdg_output, or layershell.

emersion:
No, it has nothing to do about the protocols being "unstable". "unstable" protocols (in wayland-protocols parlance) just means that a separate protocol may supersede it at some point, but "unstable" protocols never see backwards-incompatible changes. See the wayland-protocols README.

In other words, even if you use an "unstable" protocol, your client will never be broken by a protocol update.

What I'm talking about here is very different. I'm saying that I'll make it so your clients won't even launch on purpose, to avoid having other clients be tempted to use privileged protocols they must not use.

> this isn't news to anyone

Please take time to understand how Wayland works before making your points.

> if it breaks its entirely not your problem.

Yes, it _is_ my problem, because it causes more work for me. I'm already spending a good chunk of time trying to convince you and explain that this pull request is doing the wrong thing.

kennylevinsen:
We have talked it out on IRC, and we're all good. :)

In the wlroots team we're mainly worried about special protocols ending up used by normal clients when such use will be blocked in the future. @james-lawrence explained that this part is going away in the next PR, so while we find it odd that its included here, I guess that's just people having different git workflows.

james-lawrence:
@Maan2003 this is ready to go, I pulled in changes from the next branch I was going to PR to resolve @emersion concerns.

makes the PR larger than I wanted and makes it no longer a pure rebase but ./shrug. at this point I think with there will be 1 final small PR around layershell reinitializing unless I encountered something when working on my applications.

I'll be opening up issues around the remaining work once the final PR is in for documentation/disccusion purposes for whomever takes up the torch next. ;)

james-lawrence:
oh and of course I'll be avail if issues crop up.

druid-shell/src/backend/wayland/display.rs
```diff
@@ -0,0 +1,215 @@
...
+    pub(super) display: wlc::Display,
+    pub(super) registry: wlc::GlobalManager,
+    pub(super) xdg_base: wlc::Main<xdg_wm_base::XdgWmBase>,
+    pub(super) queue: std::rc::Rc<std::cell::RefCell<wlc::EventQueue>>,
+    dispatcher: std::sync::Arc<Dispatcher>,
+}
+
+// because we have the global environment we need to mark these as safe/send.
+// strictly speaking we should probably guard the access to the various fields
+// behind a mutex, but in practice we are not actually accessing across threads.
```
Maan2003, line 125: can you explain this in more detail:
- where is this Send and Sync is needed? 
- how can we be sure that we are not using it across threads?

druid-shell/src/backend/wayland/display.rs
```diff
@@ -0,0 +1,215 @@
...
+
+pub trait GlobalEventConsumer {
+    fn consume(
+        &self,
+        event: &wlc::GlobalEvent,
+        registry: &wlc::Attached<wl_registry::WlRegistry>,
+        ctx: &wlc::DispatchData,
+    );
+}
+
```
Maan2003, line 73: I don't understand the purpose of this trait, and `GlobalEventSubscription` .

why can't `Dispatcher` just `BTreeMap<u64, Arc<dyn Fn(.......)>>` (using a type alias for `Arc<dyn ....>` for making it look readable)

druid-shell/src/backend/wayland/outputs/output.rs
```diff
@@ -0,0 +1,212 @@
...
+use super::super::error;
+use super::super::outputs;
+use wayland_client as wlc;
+use wayland_client::protocol::wl_output;
+use wayland_client::protocol::wl_registry;
+use wayland_protocols::unstable::xdg_output::v1::client::zxdg_output_manager_v1;
+use wayland_protocols::unstable::xdg_output::v1::client::zxdg_output_v1;
+
+pub trait Consumer {
+    fn consume(
```
Maan2003, line 11: this trait is no longer useful? There is only impl and just a single call to `consume`. that too in this module

druid-shell/src/backend/wayland/surfaces/idle.rs
```diff
@@ -0,0 +1,55 @@
...
+pub(super) enum Kind {
+    Callback(Box<dyn IdleCallback>),
+    Token(window::IdleToken),
+}
+
+impl std::fmt::Debug for Kind {
+    fn fmt(&self, format: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        match self {
+            Kind::Callback(_) => format.debug_struct("Idle(Callback)").finish(),
+            Kind::Token(_) => format.debug_struct("Idle(Token)").finish(),
```
Maan2003, line 14: also print the token id here? `IdleToken` impls `Debug`

druid-shell/src/backend/wayland/surfaces/surface.rs
```diff
@@ -0,0 +1,627 @@
...
+impl Outputs for Dead {
+    fn removed(&self, _: &outputs::Meta) {}
+
+    fn inserted(&self, _: &outputs::Meta) {}
+}
+
+impl Popup for Dead {
+    fn surface<'a>(
+        &self,
+        _: &'a wlc::Main<wayland_protocols::xdg_shell::client::xdg_surface::XdgSurface>,
```
Maan2003, line 569: some `use`s? very long types :sweat_smile:

druid-shell/src/backend/wayland/window.rs
```diff
@@ -0,0 +1,618 @@
...
+
+impl surfaces::Popup for WindowHandle {
+    fn surface<'a>(
+        &self,
+        popup: &'a wayland_client::Main<
+            wayland_protocols::xdg_shell::client::xdg_surface::XdgSurface,
+        >,
+        pos: &'a wayland_client::Main<
+            wayland_protocols::xdg_shell::client::xdg_positioner::XdgPositioner,
+        >,
```
Maan2003, line 72: again some `use`s, huge types :grimacing:

druid-shell/src/backend/wayland/window.rs
```diff
@@ -0,0 +1,618 @@
...
+        self.inner.outputs.removed(o)
+    }
+
+    fn inserted(&self, o: &outputs::Meta) {
+        self.inner.outputs.inserted(o)
+    }
+}
+
+impl surfaces::Popup for WindowHandle {
+    fn surface<'a>(
```
Maan2003, line 65: lifetime is not needed?

Maan2003 (commented):
Sorry for such a late review :/

Maan2003:
Also, choice of using `Rc` or `Arc` feels arbitrary. please explain the reasoning

james-lawrence:
@Maan2003 all good, its holiday season. =)

james-lawrence:
@Maan2003 generally speaking I recall some comment about moving to arc be default within druid but i could be misremembering.

I'm fairly indifferent to the usage of either outside of what rust itself enforces or performance related issues.

most (all?) of the RC usage is from the original branch I started this work off of. it seems slapdash because why I defaulted to using arcs and original branch defaulted to using RC.

Maan2003:
> I recall some comment about moving to arc be default within druid.

that is probably for druid only, we mostly use `Rc` in druid shell

james-lawrence:
ah then it was a misunderstanding on my end. is the arc/rc stuff a blocker for this PR? would be easier for me to tackle after this is merged to limit the diverging from my private branch.

druid-shell/src/backend/wayland/pointers.rs
```diff
@@ -0,0 +1,394 @@
...
+    // there is likely a better default based on research to use.
+    // during experimentation this allowed one to get to around 4 clicks.
+    // but likely heavily dependent on the machine.
+    const THRESHOLD: std::time::Duration = std::time::Duration::from_millis(500);
+
+    fn reset(ts: std::time::Instant, btn: mouse::MouseButton) -> Self {
+        Self(ts, 1, btn)
+    }
+
+    fn debounce(&mut self, current: MouseEvtKind) -> MouseEvtKind {
```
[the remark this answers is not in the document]
james-lawrence: interesting. also not sure of the correct behavior. can fix if something crops up.

druid-shell/src/backend/wayland/pointers.rs
```diff
@@ -0,0 +1,394 @@
...
+                    }
+                };
+                let winhandle = match winhandle.data() {
+                    Some(w) => w,
+                    None => {
+                        tracing::warn!("dropping mouse events, no window available");
+                        appdata.pointer.queued_events.borrow_mut().clear();
+                        return;
+                    }
+                };
```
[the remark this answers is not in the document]
james-lawrence: will make this change but there are some issues with this code in general w/ respect to mutliple windows.

druid-shell/src/backend/wayland/display.rs
james-lawrence: entirely unnecessary now. will remove.

druid-shell/src/backend/wayland/outputs/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+    pub subpixel: wl_output::Subpixel,
+    pub transform: wl_output::Transform,
+    pub make: String,
+    pub model: String,
+    pub scale: f64,
+    pub enabled: bool,
+    pub position: Position,
+}
+
+impl Default for Meta {
```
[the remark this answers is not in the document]
james-lawrence: can switch it to using new if you want. primarily using it for instantiation.

druid-shell/src/backend/wayland/display.rs
james-lawrence: original intention was to allow you to not rely on the particular implemention when processing events. I can probably remove it and collapse everything.

Maan2003:
> is the arc/rc stuff a blocker for this PR? 

We can fix it later :)

Maan2003 (approved):
LGTM

druid-shell/examples/empty_window.rs
```diff
@@ -0,0 +1,136 @@
+// Copyright 2018 The Druid Authors.
```
Maan2003, line 1: Do we need this example? It is very similar to shello example

druid-shell/src/backend/mod.rs
```diff
@@ -35,11 +35,30 @@ pub use x11::*;
 #[cfg(all(feature = "x11", any(target_os = "linux", target_os = "openbsd")))]
 pub(crate) mod shared;
 
-#[cfg(all(not(feature = "x11"), any(target_os = "linux", target_os = "openbsd")))]
+#[cfg(all(feature = "wayland", target_os = "linux"))]
+mod wayland;
+#[cfg(all(feature = "wayland", target_os = "linux"))]
+pub use wayland::*;
+#[cfg(all(feature = "wayland", target_os = "linux"))]
+pub(crate) mod shared;
```
Maan2003, line 43: also add openbsd
```suggestion
#[cfg(all(feature = "wayland", any(target_os = "linux", target_os = "openbsd")))]
mod wayland;
#[cfg(all(feature = "wayland", any(target_os = "linux", target_os = "openbsd")))]
pub use wayland::*;
#[cfg(all(feature = "wayland", any(target_os = "linux", target_os = "openbsd")))]
pub(crate) mod shared;
```

druid-shell/examples/empty_window.rs
james-lawrence: nope not really. will purge it.

james-lawrence:
@Maan2003 bump =) weekend is coming up have some bug fixes in my branch I'm planning on opening another PR for =)

Maan2003:
Sorry, I forgot to merge this.

Issue #1067, opened 2020-06-29
No command sent when selecting multiple directories in SHOW_OPEN_PANEL


TheNeikos:
A bug I ran into just now:

I currently have:

```
    let dialog_options = FileDialogOptions::new()
        .select_directories()
        .multi_selection();

    let window = WindowDesc::new(build_schauglas)
        .menu(MenuDesc::new(LocalizedString::new("menubar")).append(
            MenuDesc::new(LocalizedString::new("file").with_placeholder("File")).append(
                MenuItem::new(
                    LocalizedString::new("open").with_placeholder("Open"),
                    Command::new(SHOW_OPEN_PANEL, dialog_options),
                ),
            ),
        ));
```

And in my delegate, I am simply printing out whatever command gets in:

```
      -> bool {
        debug!("Cmd: {:?}", cmd);
        if let Some(file_info) = cmd.get(OPEN_FILE) {
            return true;
        }

        false
    }
```

this gives me the following outputs:

- When selecting a single directory: `Cmd: Command { symbol: "druid-builtin.open-file-path", payload: Any }`
- When selecting multiple: <Nothing>

Info:
- Windows 10
- Druid 0.6

Finnerale:
Now this one is interesting.
It seems like the options for selecting multiple files / directories were added in #616, yet they have never been implemented.
Enabling `multi_selection` does set the flag for the dialog, but does not handle the result correctly (It just expects a single file).
On Windows it apparently returns nothing, on GTK it just returns the uppermost selected file, not sure what the effect on macOS is.

cmyr:
Huh yea this is an oversight. Agree with @Finnerale about the required API changes.

raphlinus:
Since we want get rid of the sync call to deal with #1068, maybe we can roll up both API changes at the same time.

cmyr:
I also think we might want to unify file opening to use whatever API we end up with for modals alerts; see #1039.

atlanticaccent:
Hi, it'd be great if this were to be implemented for Windows too.

I do think it should be called out in the `FileDialogOption` docs that this is not implemented on Windows yet.

PR #2081, opened 2021-12-07
Implemented multiple selection for gtk backend

Psykopear:
As noted in #1067 the multiple selection is not actually implemented in the backends.
This PR implements the multiple selection dialog for the GTK backend.

~~I think I might have broken other backends, but~~ I'm opening a PR so we can discuss if this is the right solution and if needed implement it for other platforms.

The CI passed, so I suppose I didn't breake other backends

### DIFF
--- a/druid-shell/src/backend/gtk/dialog.rs
+++ b/druid-shell/src/backend/gtk/dialog.rs
@@ -98,23 +98,15 @@ pub(crate) fn get_file_dialog_path(
     let result = dialog.run();
 
     let result = match result {
-        ResponseType::Accept => {
-            if dialog.filenames().is_empty() {
-                Err(anyhow!("No path received for filename"))
-            } else {
-                if !options.multi_selection && dialog.filenames().len() > 1 {
-                    // This really shouldn't happen, unless there is a bug
-                    // when setting the dialog options
-                    Err(anyhow!("More than one path received for single selection"))
-                } else {
-                    Ok(dialog
-                        .filenames()
-                        .into_iter()
-                        .map(|p| p.into_os_string())
-                        .collect())
-                }
+        ResponseType::Accept => match dialog.filenames() {
+            filenames if filenames.is_empty() => Err(anyhow!("No path received for filename")),
+            // This really shouldn't happen, unless there is a bug
+            // when setting the dialog options
+            filenames if filenames.len() > 1 && !options.multi_selection => {
+                Err(anyhow!("More than one path received for single selection"))
             }
-        }
+            filenames => Ok(filenames.into_iter().map(|p| p.into_os_string()).collect()),
+        },
         ResponseType::Cancel => Err(anyhow!("Dialog was deleted")),
         _ => {
             tracing::warn!("Unhandled dialog result: {:?}", result);
--- a/druid-shell/src/backend/gtk/dialog.rs
+++ b/druid-shell/src/backend/gtk/dialog.rs
@@ -37,7 +37,7 @@ pub(crate) fn get_file_dialog_path(
     window: &Window,
     ty: FileDialogType,
     options: FileDialogOptions,
-) -> Result<OsString, Error> {
+) -> Result<Vec<OsString>, Error> {
     // TODO: support message localization
 
     let (title, action) = match (ty, options.select_directories) {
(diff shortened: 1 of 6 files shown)


druid-shell/src/backend/gtk/window.rs
```diff
@@ -907,17 +907,28 @@ impl WindowState {
...
                         FileDialogType::Open,
                         options,
                     )
-                    .ok()
-                    .map(|s| FileInfo {
-                        path: s.into(),
-                        format: None,
+                    // If result is an error, return an empty vector,
+                    // otherwise return a vec of FileInfos
+                    .map_or(vec![], |f| {
```
jneem, line 919: I know this is the same behavior as before, but it does seem like the errors from `get_file_dialog_path` should get logged, at least...

druid/src/win_handler.rs
```diff
@@ -80,6 +80,8 @@ struct DialogInfo {
     id: WindowId,
     /// The command to send if the dialog is accepted.
     accept_cmd: Selector<FileInfo>,
+    /// The command to send if the dialog is accepted with multiple files.
+    accept_multiple_cmd: Option<Selector<Vec<FileInfo>>>,
```
jneem, line 84: Is it necessary for this to be an `Option`? I mean, I see that it's not used for save dialogs, but then save dialogs shouldn't get configured for multi-selection anyway...

druid-shell/src/backend/gtk/window.rs
Psykopear: Makes sense, I'll add some log here

druid/src/win_handler.rs
Psykopear: It wasn't an `Option` in my first implementation, I made it so to avoid having the field in save dialogs as you noticed.
If you think this isn't a problem though, having it always set makes the code nicer elsewhere, so I'm going to change this.

druid-shell/src/backend/gtk/dialog.rs
```diff
@@ -98,9 +98,14 @@ pub(crate) fn get_file_dialog_path(
...
 
     let result = match result {
-        ResponseType::Accept => match dialog.filename() {
-            Some(path) => Ok(path.into_os_string()),
-            None => Err(anyhow!("No path received for filename")),
+        ResponseType::Accept => match dialog.filenames() {
+            filenames if filenames.is_empty() => Err(anyhow!("No path received for filename")),
+            // This really shouldn't happen, unless there is a bug
+            // when setting the dialog options
+            filenames if filenames.len() > 1 && !options.multi_selection => {
```
jneem, line 105: One last corner case that I noticed because of the `unwrap` below: if they pass `FileDialogType::Save` and `options.multi_selection = true` (which they shouldn't but I don't think we prevent it), then we correctly handle that combination on line 62 (by ignoring multi selection) but we don't check here that we only got one file back. Ok, maybe that's a bit paranoid, but I don't think it's any more paranoid than the existing checks that we have here...

druid-shell/src/backend/gtk/window.rs
```diff
@@ -927,7 +940,9 @@ impl WindowState {
                     )
                     .ok()
                     .map(|s| FileInfo {
-                        path: s.into(),
+                        // `get_file_dialog_path` should make sure one and only one
+                        // file is passed
```
jneem, line 944: Let's be confident here :) "`get_file_dialog_path` guarantees that save dialogs only return one path", or something like that.

Psykopear: Ok, I'm going to change the comment

druid-shell/src/backend/gtk/dialog.rs
Psykopear: I'm not sure I understand what you mean:

> if they pass FileDialogType::Save and options.multi_selection = true, then we correctly handle that combination on line 62  but we don't check here that we only got one file back

We do check that the we only got one file back (`filenames.is_empty` and `filenames.len() > 1`), and return an `Err` if that is not true.

Do you mean that since we ignored the option in the case of a Save dialog, we should not return an error, but the whole `filenames` vec (essentially avoid this check)?

jneem: What I meant was: in the case of a Save dialog, we should return an error if there's more than one filename, even if `options.multi_selection` is true. (This is what the `unwrap` at line `945` assumes)

Psykopear: Ah, ok, so instead of checking for `&& !options.multi_selection` we should do:
```rust
            filenames if filenames.len() > 1 && action == FileChooserAction::Save => {
```

to uphold what we write later in the other comment, right?

jneem: Yes, but you will need to check both the action *and* `options.multi_selection` because we still need to handle the case of open dialogs that expect a single file.

Psykopear: Ok, fixing it now

jneem (approved):

jneem:
Thanks!

Psykopear:
> Thanks!

Thank you for the review :)

PR #2084, opened 2021-12-09
Fix click event handling of switch widget

remme123:
In the mouse click event handling of widget switch:
```rust
            Event::MouseUp(_) => {
                if !ctx.is_disabled() {
                    if self.knob_dragged {
                        // toggle value when dragging if knob has been moved far enough
                        *data = self.knob_pos.x > switch_width / 2.;
                    } else if ctx.is_active() {
                        // toggle value on click
                        *data = !*data;
                    }
                }
```
which checks the active state of ctx.

### DIFF
--- a/druid/src/widget/switch.rs
+++ b/druid/src/widget/switch.rs
@@ -112,7 +112,7 @@ impl Widget<bool> for Switch {
                     if self.knob_dragged {
                         // toggle value when dragging if knob has been moved far enough
                         *data = self.knob_pos.x > switch_width / 2.;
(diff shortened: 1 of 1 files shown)


jneem:
Thanks for the PR!

I'm not sure that this is the right fix, though, because `Click` can be used in many places, not just as a wrapper around `Switch`. I suspect that the right thing is to make `ControllerHost` wrap its child in a `WidgetPod`. That way `Click` and whatever it wraps will have their own active state.

PR #2087, opened 2021-12-13
Dropdown in Windows

sjoshid:
I managed to successfully F up PR #2042. (Patting my back)

This PR adds ability to let dropdowns cross their parent boundary.

SubWindow

### DIFF
--- a/druid-shell/src/backend/windows/window.rs
+++ b/druid-shell/src/backend/windows/window.rs
@@ -237,6 +237,7 @@ struct WindowState {
     // Is the window focusable ("activatable" in Win32 terminology)?
     // False for tooltips, to prevent stealing focus from owner window.
     is_focusable: bool,
+    window_level: RefCell<Option<WindowLevel>>,
 }
 
 impl std::fmt::Debug for WindowState {
@@ -1396,6 +1397,7 @@ impl WindowBuilder {
             let mut dwExStyle: DWORD = 0;
             let mut focusable = true;
             let mut parent_hwnd = None;
+            let window_level = RefCell::new(self.level.clone());
             if let Some(level) = self.level {
                 match level {
                     WindowLevel::AppWindow => (),
@@ -1441,6 +1443,7 @@ impl WindowBuilder {
                 handle_titlebar: Cell::new(false),
                 active_text_input: Cell::new(None),
                 is_focusable: focusable,
+                window_level,
             };
             let win = Rc::new(window);
             let handle = WindowHandle {
@@ -1885,19 +1888,28 @@ impl WindowHandle {
 
     pub fn set_position(&self, position: Point) {
         self.defer(DeferredOp::SetWindowState(window::WindowState::Restored));
-        let position = unsafe {
-            if !GetWindow(self.get_hwnd().unwrap(), GW_OWNER).is_null() {
-                // Has owned window. Convert point from window coords to screen coords.
-                self.get_position() + position.to_vec2()
+        if let Some(w) = self.state.upgrade() {
+            if let Some(window_level) = w.window_level.borrow().as_ref() {
+                match window_level {
+                    WindowLevel::Tooltip(parent_window_handle)
+                    | WindowLevel::DropDown(parent_window_handle)
+                    | WindowLevel::Modal(parent_window_handle) => {
+                        // Has owned window. Convert point from window coords to screen coords.
+                        let screen_position =
+                            parent_window_handle.get_position() + position.to_vec2();
+                        self.defer(DeferredOp::SetPosition(screen_position));
+                    }
+                    _ => {
+                        // Dead code!! Not sure if app window will ever be owned but we can be defensive.
+                        self.defer(DeferredOp::SetPosition(position));
+                    }
+                }
             } else {
-                position
+                self.defer(DeferredOp::SetPosition(position));
             }
-        };
-        self.defer(DeferredOp::SetPosition(position));
-    }
-
-    pub fn set_level(&self, _level: WindowLevel) {
-        warn!("Window level unimplemented for Windows!");
+        } else {
+            self.defer(DeferredOp::SetPosition(position));
+        }
     }
 
     // Gets the position of the window in virtual screen coordinates
--- a/druid-shell/src/backend/windows/window.rs
+++ b/druid-shell/src/backend/windows/window.rs
@@ -1342,9 +1342,11 @@ impl WindowBuilder {
 
     pub fn set_level(&mut self, level: WindowLevel) {
         match level {
-            WindowLevel::AppWindow | WindowLevel::Tooltip(_) => self.level = Some(level),
+            WindowLevel::AppWindow | WindowLevel::Tooltip(_) | WindowLevel::DropDown(_) => {
+                self.level = Some(level)
+            }
             _ => {
-                warn!("WindowLevel::Modal and WindowLevel::DropDown is currently unimplemented for Windows backend.");
+                warn!("WindowLevel::Modal is currently unimplemented for Windows backend.");
             }
         }
     }
@@ -1393,29 +1395,28 @@ impl WindowBuilder {
             let mut dwStyle = WS_OVERLAPPEDWINDOW;
             let mut dwExStyle: DWORD = 0;
             let mut focusable = true;
+            let mut parent_hwnd = None;
             if let Some(level) = self.level {
                 match level {
                     WindowLevel::AppWindow => (),
-                    WindowLevel::Tooltip(parent_window_handle) => {
+                    WindowLevel::Tooltip(parent_window_handle)
+                    | WindowLevel::DropDown(parent_window_handle) => {
+                        parent_hwnd = parent_window_handle.0.get_hwnd();
                         dwStyle = WS_POPUP;
                         dwExStyle = WS_EX_NOACTIVATE | WS_EX_TOOLWINDOW;
                         focusable = false;
                         if let Some(point_in_window_coord) = self.position {
                             let screen_point = parent_window_handle.get_position()
                                 + point_in_window_coord.to_vec2();
-                            let scaled_tooltip_point = WindowBuilder::scale_sub_window_position(
+                            let scaled_point = WindowBuilder::scale_sub_window_position(
                                 screen_point,
                                 parent_window_handle.get_scale(),
                             );
-                            pos_x = scaled_tooltip_point.x as i32;
-                            pos_y = scaled_tooltip_point.y as i32;
+                            pos_x = scaled_point.x as i32;
+                            pos_y = scaled_point.y as i32;
                         } else {
-                            warn!("No position provided for tooltip!");
+                            warn!("No position provided for subwindow!");
                         }
-                    }
-                    WindowLevel::DropDown(_) => {
-                        dwStyle = WS_CHILD;
-                        dwExStyle = 0;
                     }
                     WindowLevel::Modal(_) => {
                         dwStyle = WS_OVERLAPPED;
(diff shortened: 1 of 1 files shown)


druid-shell/src/backend/windows/window.rs
```diff
@@ -1488,7 +1489,7 @@ impl WindowBuilder {
                 pos_y,
                 width,
                 height,
-                0 as HWND,
+                parent_hwnd.unwrap_or_else(|| 0 as HWND),
```
jneem, line 1492: I think `unwrap_or` here would be simpler than `unwrap_or_else`

jneem (commented):
Thanks! Just one very minor suggestion.

druid-shell/src/backend/windows/window.rs
sjoshid: Done

druid-shell/src/backend/windows/window.rs
```diff
@@ -1882,9 +1883,17 @@ impl WindowHandle {
         self.defer(DeferredOp::ShowTitlebar(show_titlebar));
     }
 
-    // Sets the position of the window in virtual screen coordinates
     pub fn set_position(&self, position: Point) {
         self.defer(DeferredOp::SetWindowState(window::WindowState::Restored));
+        let position = unsafe {
+            if !GetWindow(self.get_hwnd().unwrap(), GW_OWNER).is_null() {
```
sjoshid, line 1889: @jneem 
I think it would be better to store `WindowLevel` in `WindowHandle`. 
We would then be consistent with what is done in `build()`. ie convert point from window coords to screen coords only when this subwindow is `WindowLevel::Tooltip` or `WindowLevel::Dropdown`. 

What do you think?

sjoshid: There's a `set_level` for `WindowHandle` which is unimpl today. Maybe above suggestion will fix that too.

sjoshid:
I need to fix the changes I made in set_position(). 
Please hold off on merging this.

druid-shell/src/backend/windows/window.rs
jneem: Hm, I can see the argument for including `WindowLevel` in `WindowHandle`, but it will require cross-platform changes and I don't think there's a *big* advantage, is there? I mean, you can query the parent easily enough from the windows API (and I think on other platforms too).

Also, I thought we had agreed in #1824 to remove `set_level` from `WindowHandle`.

sjoshid: Few points:
* My point is consistency between `build()` and `set_position()`. Since `build()` calculates positions based on `WindowLevel` so should `set_position`. And not make a platform call which many might not know straight away. (it also negates the need for an extra `unsafe` call if that matters)
* Also there's a bug in `set_position`. I cant use `self.get_position()`. I need parent's position not its own. We can get that if `WindowLevel` (which internally gives us `WindowHandle`) is stored in `WindowHandle`
* To further my previous point, in gtk, parent handle is stored in `WindowHandle` for exactly that purpose. Only difference is _I think_ it's better to store `WindowLevel` (again we get window handle internally) for the sake of first point above. 

_I keep saying store `WindowLevel` in `WindowHandle` but what I actually mean is store it within one of its fields. Like in Windows can store level in `WindowState` which is field of `WindowHandle`_

> Also, I thought we had agreed in #1824 to remove `set_level` from `WindowHandle`.

Ohh. You are right. I can remove it as part of this?

jneem: > I keep saying store `WindowLevel` in `WindowHandle` but what I actually mean is store it within one of its fields.

Ah, got it. I thought you were proposing to put it in the actual cross-platform `WindowHandle` struct, which would then require more invasive changes...

Can you explain the benefit of storing the `WindowLevel` as opposed to just the parent? My understanding of the position calculation was that it should only depend on whether there's a parent or not.

jneem: > Ohh. You are right. I can remove it as part of this?

Please!

sjoshid: > Can you explain the benefit of storing the WindowLevel as opposed to just the parent?

So instead of doing this:
```
if let Some(p) = self.parent {
    // use p to calculate position
}
```

We can do this:
```
if WindowLevel::Tooltip(p) || WindowLevel::Dropdown(p) {
    // use p to calculate position
}
```

`WindowLevel` check is consistent with `build()`

jneem: ...but I think you also want to include `WindowLevel::Modal(p)` there, because the only difference between those 3 levels is their z-order. They should all treat positioning in the same way.

sjoshid: True. But at least in Windows it is unimplemented/unfinished. I didnt include it because I that change should be included when modal feature is implemented.

sjoshid: Not sure if you agree to my previous point but Im going to resolve this conversation. It's been a while. If you disagree, please let me know and I'll add `WindowLevel::Modal(p)`.

jneem: Oh, sorry, I had forgotten about this, but I don't understand the part about modal not being implemented in windows. The only difference between (druid terminology) tooltip, modal, and dropdown is that they have different z-order. They should have exactly the same behavior in all other respects, and so there's no reason to support, say, dropdown without also supporting modal.

sjoshid: Where is the best place in code to see different z-orders in action? In Windows at least, z-order is same for tooltips and dropdown (and maybe for modal) ie 1 + max of existing z-orders.

jneem: It seems like the mac backend is the only one that handles `WindowLevel` according to the documentation (see `as_raw_window_level`), but I don't have a mac so I can't test it. The gtk backend does something different. I guess what happened was that `WindowLevel` was created according to the mac behavior, and then because the variants have suggestive-sounding names it was implemented differently on gtk (and partially on windows).

But I think the right way forward is the following: let's first make window parent/child positioning work. After that, we can worry about z-order and any other behavioral changes that different subwindows can have. We'll need to figure out what we can reliably support cross-platform.

sjoshid: Sounds good!!
I have made changes as per this conversation.

sjoshid:
^^
This is ready to merge (assuming there are no more code review changes)

druid-shell/src/backend/windows/window.rs
```diff
@@ -237,6 +237,7 @@ struct WindowState {
     // Is the window focusable ("activatable" in Win32 terminology)?
     // False for tooltips, to prevent stealing focus from owner window.
     is_focusable: bool,
+    window_level: RefCell<Option<WindowLevel>>,
```
jneem, line 240: This never changes, right? So you don't need the `RefCell`. And I think `AppWindow` is a good default so we don't need the `Option` either.

druid-shell/src/backend/windows/window.rs
```diff
@@ -1342,9 +1343,11 @@ impl WindowBuilder {
 
     pub fn set_level(&mut self, level: WindowLevel) {
         match level {
-            WindowLevel::AppWindow | WindowLevel::Tooltip(_) => self.level = Some(level),
+            WindowLevel::AppWindow | WindowLevel::Tooltip(_) | WindowLevel::DropDown(_) => {
```
jneem, line 1346: Can you add `Modal` here also?

druid-shell/src/backend/windows/window.rs
```diff
@@ -1393,30 +1396,30 @@ impl WindowBuilder {
...
             let mut dwExStyle: DWORD = 0;
             let mut focusable = true;
+            let mut parent_hwnd = None;
+            let window_level = RefCell::new(self.level.clone());
             if let Some(level) = self.level {
                 match level {
                     WindowLevel::AppWindow => (),
-                    WindowLevel::Tooltip(parent_window_handle) => {
+                    WindowLevel::Tooltip(parent_window_handle)
+                    | WindowLevel::DropDown(parent_window_handle) => {
```
jneem, line 1405: Maybe `Modal` here too? In principle, this "regresses" the topmost behavior of modal, but I'm pretty sure the current code is dead code anyway, because `set_level` doesn't currently set modal.

druid-shell/src/backend/windows/window.rs
```diff
@@ -1882,14 +1886,30 @@ impl WindowHandle {
...
+                    WindowLevel::Tooltip(parent_window_handle)
+                    | WindowLevel::DropDown(parent_window_handle)
+                    | WindowLevel::Modal(parent_window_handle) => {
+                        // Has owned window. Convert point from window coords to screen coords.
+                        let screen_position =
+                            parent_window_handle.get_position() + position.to_vec2();
+                        self.defer(DeferredOp::SetPosition(screen_position));
+                    }
+                    _ => {
+                        // Dead code!! Not sure if app window will ever be owned but we can be defensive.
```
jneem, line 1903: Ah ok, I see what the `Option<WindowLevel>` is doing now. I think I prefer getting rid of the option still, because then the types will get rid of this dead code.

druid-shell/src/backend/windows/window.rs
```diff
@@ -1882,14 +1886,30 @@ impl WindowHandle {
...
+                        self.defer(DeferredOp::SetPosition(position));
+                    }
+                }
+            } else {
+                self.defer(DeferredOp::SetPosition(position));
+            }
+        } else {
```
jneem, line 1910: I don't think there's a need for this else. If the state doesn't upgrade, it should mean that the window was closed.

druid-shell/src/backend/windows/window.rs
sjoshid: Dang! Forgot to revert this after removing `set_level`. Thanks.

druid-shell/src/backend/windows/window.rs
sjoshid: Added `Modal` and removed the `match` statement.

druid-shell/src/backend/windows/window.rs
sjoshid: Default as `AppWindow` made this easier. Fixed this and your suggestion below. Thanks.

druid-shell/src/backend/windows/window.rs
sjoshid: Added `Modal`

jneem (approved):
Thanks!

sjoshid:
^ Thank you too. 
Im going to pick your brain one last time.
Recently installed linux to understand more about gtk because it always comes up but Im seeing.........bizarre behavior to say the least. 
My assumption was subwindow positioning was impl recently in gtk meaning it was working.
Is the assumption wrong?

https://user-images.githubusercontent.com/6751090/147617220-d240169c-d9e6-469b-9d28-f5981dbb042b.mp4

https://user-images.githubusercontent.com/6751090/147617185-547673eb-81d4-4e63-834e-b2b1d150c52e.mp4

Maan2003:
@sjoshid can you check the x11 backend (`cargo run --example subwindow --features x11 --no-default-features`)

jneem:
Thanks, I'll look into those. It's highly likely that some things didn't get properly updated when we changed to relative positions.

sjoshid:
@Maan2003 I cant. :( Im running into build errors one after another. Sorry. 

@jneem Thanks.

sjoshid:
I know Im late....but I fixed build errors (like `libclang` not found, etc.) for x11. 
Happy to report x11 is working as expected (to Windows). The tooltip position is off in `sub_window` example but other than that it looks good.

Maan2003:
nice :). please create an issue about tooltip positioning with the desktop enviroment version, screenshots, ... I tried fixing positioning but it is hard to fix for all enviroments on linux.

sjoshid:
Done. #2100

jneem:
@Maan2003 I think I found a little bug with DP scaling and get_position (#2101).

PR #2101, opened 2021-12-30
[GTK] Do scaling correctly for child window position.

jneem:
The parent's position is in display points, so we need to be in display
points before subtracting it.


Maan2003 (approved):
Looks good!

sjoshid:
Im on latest but I still see the issue. Same issue happens in x11 (which contradicts what I said in #2087). In x11, title bar drag works but moving the window does not. 

https://user-images.githubusercontent.com/6751090/149960098-652b7897-4988-408d-9dd8-a76804438603.mp4

PR #2106, opened 2022-01-03
RFC: svg cleanup.

james-lawrence:
removes most of the SVG code from the codebase. relying instead of resvg's implementation. opened as a PR as a proof of concept. it renders more SVGs correctly, downside of course is the additional optional dependencies.


druid/src/widget/svg.rs
```diff
@@ -1,568 +1,195 @@
-// Copyright 2019 The Druid Authors.
```
ratmice, line 1: Confused by this '- }' in the diff, since it doesn't appear in the original file (wondering if github diff interface is mangling it).

Otherwise, could you explain the removal of the copyright?

james-lawrence: this is a widget I implemented in my own codebase to get essentially fully functioning SVG rendering. vs the hodge podge druid currently has. its meant as an example implementation for druid to consider. just slapped it in place.

benefits:
1. correctly renders essentially all svgs
2. removal of the font loading for every svg dramatically speeds up rendering.
3. simple caching to prevent unnecessary re-rendering.
4. no svg rendering code is owned by druid.

downsides:
1. dependency on tiny-skia.

ratmice: Marking as resolved, I no longer see the diff weirdness (Which was in the in-tree sources not matching what was actually in tree), and your latest push leaves copyright header intact thanks.

james-lawrence:
@xStrom this implementation is standalone it completely replaced the current implementation.

james-lawrence:
@xStrom - main differences between the implementations:

- removed upfront font loading. massive performance hit having it vs allowing people to opt into font loading. the performance hit was primarily due to loading it for every image vs having a cached value somewhere.
- overall support for rendering svgs properly is much better in this branch.

fixing up most of the errors.

james-lawrence:
we should be gtg. only have the ability to test linux backends personally they all opened and rendered the svg example.

```
cargo run --features="svg" --example svg # gtk
cargo run --features="svg wayland" --example svg
cargo run --features="svg x11" --example svg
```

xStrom:
I tried the example on Windows and the tiger looks low-res. I'm guessing that there's some rasterization happening at the resolution specified in dp (which the Druid widgets operate with), while thinking that is the final physical pixel value.

I wonder if this is a straightforward fix. Maybe we can just ask for a higher res rasterization.

Any thoughts? I'll peek at the code a bit myself too.

xStrom:
I sort of fixed the scale problem in my local branch. I'll see how easily I could make it mergeable.

xStrom:
Unfortunately looks like `Scale` (*the info*) isn't piped from internals to the widgets yet. The fix is rather easy, and I have it confirmed with a hardcoded 1.75 scale factor as that matches my setup.

I think the move here is to just accept the regression for now and merge. I'll fix up the scale later myself.

xStrom (approved):
As mentioned earlier, my review here is rather surface level.

* The example still runs fine.
* There is a scale issue, which needs further Druid internals work and I'll document separately.
* I defer judgement of superiority here and just accept the claim.

As such, I think we can merge.

PR #2108, opened 2022-01-04
Add Arch installation instruction

ModProg:


Maan2003:
I am not sure if this is needed, because on arch the dev package is not different from main gtk package. And gtk is everywhere.

ModProg:
That is true, and if you don't have it that is probably an active decision.

I would be fine with leaving thid out.

-------- Original-Nachricht --------

PR #2115, opened 2022-01-08
Add methods WindowHandle::is_resizable etc

RagibHasin:
Add 3 methods to `WindowHandle` for querying if the window:
1. is resizable,
2. has title bar,
3. is transparent.

At this moment only Windows and GTK+ backends are implemented and only Windows backend is tested locally.

I am not familiar with any backend except Windows so I don't see myself implementing these for the other backends.

These functionalities, especially transparency testing, are necessary for proper acrylic and mica background support on Windows through widgets.

Also, I anticipate that it would be also necessary for macOS translucent background if somebody implements it sometime.


druid-shell/src/backend/wayland/window.rs
```diff
@@ -106,10 +106,25 @@ impl WindowHandle {
...
     }
 
+    pub fn is_resizable(&self) -> bool {
+        tracing::warn!("is_resizable is unimplemented on wayland");
+        true
+    }
+
+    pub fn is_transparent(&self) -> bool {
+        tracing::warn!("is_transparent is unimplemented on wayland");
+        false
```
james-lawrence, line 116: warning is unnecessary for transparent. it'll really is unnecessary for wayland. wayland allows transparency by default. depending on the intent of this method we may wish to return true.

james-lawrence (commented):
slightly curious about the point of these method. mostly its querying state that the user of druid has manually set. hats the usecase?

RagibHasin:
> slightly curious about the point of these method. mostly its querying state that the user of druid has manually set. hats the usecase?

These methods are internally used by widgets to determine their behavior without explicitly repeating these info upon creating them.

druid-shell/src/backend/wayland/window.rs
RagibHasin: Didn't know that! Would change to true always.

james-lawrence:
@RagibHasin yes, that part was obvious. =) I was hoping for a couple actual usecases for these. resize for example I'd probably just react to lifecycle size events. and I'm struggled to come up with a usecase for is_transparent that isn't resolved just by setting the background to transparent.

xStrom:
Alright we'll let this sit a bit more then, as I am very busy with 0.8 work as it is.

The code changes are simple and would be a fast review, but the concerns of purpose brought up by @james-lawrence would need deeper thinking, which I'll postpone right now.

Regarding sticking with this PR here or making one for [Glazier](https://github.com/linebender/glazier) - it really comes down to what you yourself want. Glazier is without a doubt the future of `druid-shell`. However it is unknown if and when Druid will migrate from `druid-shell` to `glazier`. So if you want to, in practice, have these features in Druid - then keeping this PR here open would be the fastest way.

PR #2117, opened 2022-01-10
fix dropdown position for gtk backend

maurerdietmar:
We need to position internal windows (dropdowns, tooltips) relative
to the content area.

### DIFF
--- a/druid-shell/src/backend/gtk/window.rs
+++ b/druid-shell/src/backend/gtk/window.rs
@@ -1009,25 +1009,33 @@ impl WindowHandle {
         }
     }
 
-    /// The GTK implementation of content_insets differs from, e.g., the Windows one in that it
-    /// doesn't try to account for window decorations. Depending on the platform, GTK might not
-    /// even be aware of the size of the window decorations. And anyway, GTK's `Window::resize`
-    /// function [tries not to include] the window decorations, so it makes sense not to include
-    /// them here either.
-    ///
-    /// [tries not to include]: https://developer.gnome.org/gtk3/stable/GtkWidget.html#geometry-management
     pub fn content_insets(&self) -> Insets {
         if let Some(state) = self.state.upgrade() {
             let scale = state.scale.get();
             let (width_px, height_px) = state.window.size();
             let alloc_px = state.drawing_area.allocation();
-            let window = Size::new(width_px as f64, height_px as f64).to_dp(scale);
-            let alloc = Rect::from_origin_size(
-                (alloc_px.x as f64, alloc_px.y as f64),
-                (alloc_px.width as f64, alloc_px.height as f64),
-            )
-            .to_dp(scale);
-            window.to_rect() - alloc
+            let menu_height_px = height_px - alloc_px.height;
+
+            if let Some(window) = state.window.window() {
+                let frame = window.frame_extents();
+                let (pos_x, pos_y) = window.position();
+                Insets::new(
+                    (pos_x - frame.x) as f64,
+                    (pos_y - frame.y + menu_height_px) as f64,
+                    (frame.x + frame.width - (pos_x + width_px)) as f64,
+                    (frame.y + frame.height - (pos_y + height_px)) as f64,
+                )
+                .to_dp(scale)
+                .nonnegative()
+            } else {
+                let window = Size::new(width_px as f64, height_px as f64).to_dp(scale);
+                let alloc = Rect::from_origin_size(
+                    (alloc_px.x as f64, alloc_px.y as f64),
+                    (alloc_px.width as f64, alloc_px.height as f64),
+                )
+                .to_dp(scale);
+                window.to_rect() - alloc
+            }
         } else {
             Insets::ZERO
         }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,7 @@ You can find its changes [documented below](#070---2021-01-01).
 - Tabs: allow getting and setting the tab index of a Tabs widget ([#2082] by [@rjwittams]
 - `RangeSlider` and `Annotated` ([#1979] by [@xarvic])
 - Add `Checkbox::from_label` constructor ([#2111] by [@maurerdietmar])
+- fix content_insets for gtk backend ([#2117] by [@maurerdietmar])
 
 ### Changed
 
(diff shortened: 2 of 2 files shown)


maurerdietmar:
I am quite unsure if we should add this logic into set_position(). Instead, we could use different get_position methods, i.e:

get_window_position: Returns the position of the window decoration top-left corner.
get_position: Returns the position of the content window top-left corner.

Also, I have only tested this on Debian/Linux.

jneem:
This is tricky, unfortunately, and will definitely need testing on wayland because window positioning works differently there.

Note that we currently have the `content_insets` method that, when combined with `get_position` and `set_position`, should allow for handling the various cases. The gtk implementation of `content_insets` doesn't currently account for the window decorations, but maybe that should be changed? (I found [this](https://stackoverflow.com/questions/35908676/gtk-window-how-to-get-window-decoration-sizes) hint for how to calculate the decoration sizes.)

I think one reasonable choice for `druid_shell` would be to just say that `set_position` and `get_position` always work with the top-left coordinate of the window decorations (for both the child and the parent), and that `content_insets` on the parent window can be used if you want to position a child relative to the content of the parent.

If we want a friendly API for positioning child windows relative to the content of the parent, we could handle the insets in druid's `new_sub_window` function. I'd prefer this approach, because it's easier to keep cross-platform behavior sane and in sync if the druid-shell API is kept simple and orthogonal.

maurerdietmar:
> If we want a friendly API for positioning child windows relative to the content of the parent, we could handle the insets in druid's `new_sub_window` function. I'd prefer this approach, because it's easier to keep cross-platform behavior sane and in sync if the druid-shell API is kept simple and orthogonal.

Tried that, but how do I call  `content_insets()` from inside `new_sub_window`? I have no idea how to  get the `WindowHandle` there?

maurerdietmar:
> `self.state.window` or `self.window()`, where `self` here is the `EventCtx` (or `UpdateCtx` or `LifeCycleCtx`)

Thanks for the hint! Just updated the patch and implemented it as you suggested.

I haven't updated the doc comments for `contents_insets()` so far, because I am unsure what to write ...

druid/src/contexts.rs
```diff
@@ -443,12 +443,25 @@ impl_context_method!(EventCtx<'_, '_>, UpdateCtx<'_, '_>, LifeCycleCtx<'_, '_>,
...
+        mut window_config: WindowConfig,
         widget: W,
         data: U,
         env: Env,
     ) -> WindowId {
         trace!("new_sub_window");
+
+        if matches!(
+            window_config.level,
+            Some(WindowLevel::DropDown(_) | WindowLevel::Tooltip(_))
```
jneem, line 455: According to the `WindowLevel` docs the difference between various levels is only in their z-order, so I don't think you should change behavior based on `WindowLevel`.

druid-shell/src/backend/gtk/window.rs
```diff
@@ -1021,13 +1021,27 @@ impl WindowHandle {
...
+                let frame = window.frame_extents();
+                let (pos_x, pos_y) = window.position();
+                Insets::new(
+                    (pos_x - frame.x) as f64,
+                    (pos_y - frame.y) as f64,
+                    (frame.x + frame.width - (pos_x + width_px)) as f64,
+                    (frame.y + frame.height - (pos_y + height_px)) as f64,
+                )
+                .to_dp(scale)
+                .nonnegative()
```
jneem, line 1035: Does this take into account the menu bar? We want the insets to give the location of the drawing area: the menu should be excluded, just like the window decorations are.

jneem (commented):
I think you can just delete the doc on `content_insets`, because after this change there won't be anything special about our GTK implementation of them anymore.

druid/src/contexts.rs
maurerdietmar: > According to the `WindowLevel` docs the difference between various levels is only in their z-order, so I don't think you should change behavior based on `WindowLevel`.

Well, the docs also say: 

> Describes the purpose of a window and should be mapped appropriately to match platform conventions.

WindowLevel is also used to set the WM type hint, so this is by sure more than a Z-order.

Anyways, I cant see any other property which indicates the window type, so what exactly do you have in mind?

jneem: That's a good point, although it isn't particularly consistent with the previous sentence. I guess we need to figure out (and document!) what else `WindowLevel` is actually supposed to mean.

But I still think it would be surprising if window positioning worked differently for different kinds of windows: why should it be that tooltips are relative to the parent's content but modals are relative to the parent's window decorations?

maurerdietmar: > But I still think it would be surprising if window positioning worked differently for different kinds of windows: why should it be that tooltips are relative to the parent's content but modals are relative to the parent's window decorations?

Agreed. That is why is suggested to use two different `get_position` methods (`get_window_position` and `get_position`).

druid-shell/src/backend/gtk/window.rs
maurerdietmar: > Does this take into account the menu bar? 

Sorry, just pushed a fix for that.

druid/src/contexts.rs
maurerdietmar: > Agreed. That is why is suggested to use two different `get_position` methods (`get_window_position` and `get_position`).

Please ignore that last argument - it does not really makes any difference.

> > But I still think it would be surprising if window positioning worked differently for different kinds of windows: why should it be that tooltips are relative to the parent's content but modals are relative to the parent's window decorations?

Thought a bit more about that, and it is quite obvious that dropdowns and tooltips need to be positioned relative to the content root.

druid-shell/src/backend/gtk/window.rs
jneem: Don't forget to delete the obsolete doc comment!

druid/src/contexts.rs
jneem: I agree that it's very natural for dropdowns and tooltips to be positioned relative to the content. But there are also some disadvantages:
- it's undocumented
- it differs from established APIs, like `gtk_window_move` and (I think) `SetWindowPos` on windows, which are always relative to the window frame
- if we support other type hints in the future, we will need to decide (and document!) how they're treated

The first point is the big one: if we're going to go down this road, it needs to be clearly documented.

maurerdietmar:
> I agree that it's very natural for dropdowns and tooltips to be positioned relative to the content. But there are also some disadvantages

Ok, so I removed the code in `make_sub_window` - instead, the caller can easily handle that (I will send a patch for `dropdown.rs` once this gets applied)

jneem (approved):
Thanks for your patience!

PR #2119, opened 2022-01-11
Send the `WindowHandle` with `AppDelegate::window_added`.

### LINKED ISSUE #2116: Improving Access to `WindowHandle`
This discussion originally started [in the Zulip chat](https://xi.zulipchat.com/#narrow/stream/255910-druid-help/topic/Getting.20a.20WindowHandle.20with.20a.20WindowId), but since it seems like a proper issue I figured I'd actually open one.

I started with Druid fairly recently, and so far I've found two problems that both have the same root cause - insufficient access to `WindowHandle`:
1. From `AppDelegate::window_added`, there doesn't seem to be a way to get a `WindowHandle`. This prevents me from doing OS-specific window initialization stuff in the place where it makes most logical sense, the `window_added` function.
2. When opening a modal, the `WindowLevel::Modal` enum takes a `WindowHandle` as an argument. Unfortunately, I'm trying to open this modal upon receiving a command (in `AppDelegate::command`), and I don't have access to any window-related information. I know the `command` function isn't tied to a specific window, but in this case I would like to open the modal with the handle to the main application window. I already store the `WindowId` of the main application, so all I need in this case is the ability to use the stored `WindowId` to get the associated `WindowHandle`.

Both of these pain points I've encountered could be solved by providing the ability to lookup a `WindowHandle` with a `WindowId`. Actually, when I first saw the `WindowId` type, that's more or less what I expected it was for to begin with.

I don't mind taking a crack at implementing this myself, but as I'm fairly new to Druid I'm not exactly sure of the implementation details. I'm also not 100% sure there isn't just something I'm missing.

zedseven:
Closes  #2116.

Mostly just applying @sjoshid's recommendations and updating the one example that used `window_added`.

This is a breaking change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,6 +87,7 @@ You can find its changes [documented below](#070---2021-01-01).
 - Add parent windows to non-main windows. (Coordinate space is now from their origin) ([#1919] by [@JAicewizard])
 - `ListIter` implementations for `Arc<Vec<T>>`, `(S, Arc<Vec<T>>)`, `Arc<VecDequeue<T>>` and `(S, Arc<VecDequeue<T>>)` ([#1967] by [@xarvic])
 - Closures passed to `Label::new` can now return any type that implements `Into<ArcStr>` ([#2064] by [@jplatte])
+- `AppDelegate::window_added` now receives the new window's `WindowHandle`. ([#2119] by [@zedseven])
 
 ### Deprecated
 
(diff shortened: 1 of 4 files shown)


jneem:
The CI failures are [clippy's fault](https://github.com/rust-lang/rust-clippy/issues/8045), not yours.

jneem (approved):

PR #2121, opened 2022-01-12
FIX: fix example code at version `0.7.0`

duyixian1234:


jneem:
Thanks for the PR! As I wrote in the issue, these examples are supposed to be for master, not 0.7.0. I'd definitely accept a PR to make that more clear, though.

PR #2130, opened 2022-01-18
[GTK] Fix override-redirect.

jneem:
Override-redirect needs to be set after the window is realized.

This used to work, but got broken in some refactor.


Maan2003:
Is this mentioned in gtk docs somewhere? I can't find it.

jneem:
The issue is that the G*d*k window doesn't exist until the widget is
realized, so the 'if let' was failing

Maan2003 (approved):
:+1:

PR #2141, opened 2022-02-07
Updated Scroll_to_View

xarvic:
This PR adds that `ClipBox` and `Tabs` can handle `scroll_to_view`.

- ClipBox discards scroll_to_view_requests for hidden areas unless `managed()` was called.
- Tabs discards all scroll_to_view requests which originate from hidden widgets.
- `Notification` now has a `known_target` flag. When set to false there will be no warning even if the notification reaches the root widget.

### DIFF
--- a/druid/src/command.rs
+++ b/druid/src/command.rs
@@ -347,7 +347,7 @@ pub mod sys {
     /// a new `SCROLL_TO_VIEW` notification with the same region relative to the new child position.
     ///
     /// When building a new widget using ClipBox take a look at [`ClipBox::managed`] and
-    /// [`ClipBox::default_scroll_to_view_handling`].
+    /// [`Viewport::default_scroll_to_view_handling`].
     ///
     /// [`scroll_to_view`]: crate::EventCtx::scroll_to_view()
     /// [`scroll_area_to_view`]: crate::EventCtx::scroll_area_to_view()
--- a/druid/src/command.rs
+++ b/druid/src/command.rs
@@ -352,7 +352,7 @@ pub mod sys {
     /// [`scroll_to_view`]: crate::EventCtx::scroll_to_view()
     /// [`scroll_area_to_view`]: crate::EventCtx::scroll_area_to_view()
     /// [`ClipBox::managed`]: crate::widget::ClipBox::managed()
-    /// [`ClipBox::default_scroll_to_view_handling`]: crate::widget::ClipBox::default_scroll_to_view_handling()
+    /// [`Viewport::default_scroll_to_view_handling`]: crate::widget::Viewport::default_scroll_to_view_handling()
     pub const SCROLL_TO_VIEW: Selector<Rect> = Selector::new("druid-builtin.scroll-to");
 
     /// A change that has occured to text state, and needs to be
--- a/druid/src/command.rs
+++ b/druid/src/command.rs
@@ -104,7 +104,7 @@ pub struct Notification {
     payload: Arc<dyn Any>,
     source: WidgetId,
     route: WidgetId,
-    known_target: bool,
+    warn_if_unused: bool,
 }
 
 /// A wrapper type for [`Command`] payloads that should only be used once.
@@ -435,7 +435,7 @@ impl Command {
             payload: self.payload,
             source,
             route: source,
-            known_target: true,
+            warn_if_unused: true,
         }
     }
 
@@ -558,17 +558,17 @@ impl Notification {
         self.source
     }
 
-    /// If set to false this notification will not produce a warning when reaching the root widget.
+    /// Builder-style method to set warn_if_unused.
     ///
     /// The default is true.
-    pub fn known_target(mut self, known_target: bool) -> Self {
-        self.known_target = known_target;
+    pub fn warn_if_unused(mut self, warn_if_unused: bool) -> Self {
+        self.warn_if_unused = warn_if_unused;
         self
     }
 
-    /// Returns whether this notification was sent to a specific widget.
-    pub fn has_known_target(&self) -> bool {
-        self.known_target
+    /// Returns whether there should be a warning when no widget handles this notification.
+    pub fn warn_if_unused_set(&self) -> bool {
+        self.warn_if_unused
     }
 
     /// Change the route id
--- a/druid/src/command.rs
+++ b/druid/src/command.rs
@@ -104,6 +104,7 @@ pub struct Notification {
     payload: Arc<dyn Any>,
     source: WidgetId,
     route: WidgetId,
+    known_target: bool,
 }
 
 /// A wrapper type for [`Command`] payloads that should only be used once.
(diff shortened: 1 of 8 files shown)


druid/src/widget/tabs.rs
```diff
@@ -579,7 +580,14 @@ impl<TP: TabsPolicy> TabsBody<TP> {
 impl<TP: TabsPolicy> Widget<TabsState<TP>> for TabsBody<TP> {
     #[instrument(name = "TabsBody", level = "trace", skip(self, ctx, event, data, env))]
     fn event(&mut self, ctx: &mut EventCtx, event: &Event, data: &mut TabsState<TP>, env: &Env) {
-        if event.should_propagate_to_hidden() {
+        if let Event::Notification(notification) = event {
+            if notification.is(SCROLL_TO_VIEW)
+                && Some(notification.route()) != self.active_child(data).map(|w| w.id())
+            {
+                // Ignore SCROLL_TO_VIEW requests from every widget except the active.
```
jneem, line 587: Maybe this could be a property of notifications? Like, say `Notification::ignore_if_hidden`? I'm just trying to think of a way to avoid special-casing `SCROLL_TO_VIEW` in every widget that hides children.

jneem (commented):
Could you explain the motivation for having a `ClipBox` that *isn't* managed? My initial reaction is that if you're clipping your children and ignoring `SCROLL_TO_VIEW` then that's a bug, and we don't need to be working around it in `ClipBox`.

druid/src/command.rs
```diff
@@ -552,6 +558,19 @@ impl Notification {
         self.source
     }
 
+    /// If set to false this notification will not produce a warning when reaching the root widget.
+    ///
+    /// The default is true.
+    pub fn known_target(mut self, known_target: bool) -> Self {
```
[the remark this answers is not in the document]
xarvic: Sounds good

druid/src/widget/tabs.rs
xarvic: I am not sure about this. Then instead of checking whether a `Notification` is `SCROLL_TO_VIEW`, every widget which hides children has to check whether `Notification::ignore_if_hidden` is set. 
Also stopping `Notification`s for hidden widgets is probably pretty unique to `SCROLL_TO_VIEW`.

xarvic:
> Could you explain the motivation for having a `ClipBox` that _isn't_ managed? My initial reaction is that if you're clipping your children and ignoring `SCROLL_TO_VIEW` then that's a bug, and we don't need to be working around it in `ClipBox`.

I can't think of a specific usecase. But you can use `ClipBox` without creating a custom widget and in this case it could result in pretty strange behavior.
Maybe we should print a warning when a `SCROLL_TO_VIEW` notification was stopped?

druid/src/widget/tabs.rs
jneem: I guess it's unique for now, but I could imagine it not staying that way. But ok, we can punt on this until it actually becomes an issue.

jneem:
Sorry for the slow reply. I've been thinking a bit more about this, and I think what worries me is that both the mananged and unmanaged scenarios have the potential for silently causing unexpected behavior. I agree that if you use `ClipBox` without handling `SCROLL_TO_VIEW` (for example, because you just use it directly in your widget tree without a custom widget or a controller) then you might see surprising behavior without warning. But on the other hand if you *are* handling `SCROLL_TO_VIEW` but you forget to make the `ClipBox` managed then you'll also see surprising behavior without warning.

I don't see a great way out of this situation, but I'm still of the opinion that if you have a `ClipBox` you should handle `SCROLL_TO_VIEW`. So maybe it makes sense to have `managed` be the default, and the alternative could be called by a very explicit name like `clip_scroll_notifications`?

xarvic:
> I don't see a great way out of this situation, but I'm still of the opinion that if you have a `ClipBox` you should handle `SCROLL_TO_VIEW`. So maybe it makes sense to have managed be the default, and the alternative could be called by a very explicit name like clip_scroll_notifications?

Sounds good.

> I think what worries me is that both the mananged and unmanaged scenarios have the potential for silently causing unexpected behavior.

In this case maybe we should not have a default value, but force the user always to decide which option would fit? Either by introducing `ClipBox::managed` and `ClipBox::unmanaged` instead of `ClipBox::new` or adding a flag to the current constructor?
Although this would be a breaking change.

xarvic:
@jneem what do you think about this. I am in favor of the breaking change. But your suggestion would also be fine for me.

jneem:
Oh, sorry, for some reason in my mind I had already replied. I don't have a strong preference between having a default or not having a default. I agree that defaulting to `managed` has the advantage of not being breaking, but I don't think that's a dealbreaker. I do prefer having named methods over having flags (because the resulting code is easier to read).

xarvic:
Have I understood this correctly, you are in favor of the `ClipBox::managed` and `ClipBox::unmanaged` approach?

jneem:
I'm ok with two approaches
- `ClipBox::new` defaulting to managed, together with a `ClipBox::unmanaged(self) -> Self` builder method. (Or maybe a more explicit name like `ClipBox::clip_notifications`?); or
- `ClipBox::managed` and `ClipBox::unmanaged`.

I don't have a strong preference between these two.

druid/src/widget/clip_box.rs
```diff
@@ -237,13 +302,37 @@ impl<T, W> ClipBox<T, W> {
 
 impl<T, W: Widget<T>> ClipBox<T, W> {
     /// Creates a new `ClipBox` wrapping `child`.
-    pub fn new(child: W) -> Self {
+    ///
+    /// This method should only be used when creating your own widget, which uses ClipBox
+    /// internally.
+    ///
+    /// `ClipBox` will forward [`SCROLL_TO_VIEW`] notifications to its parent unchanged.
+    /// In this case the parent has to handle said notification itself. By default the ClipBox will
```
jneem, line 310: There's no more "by default", right? The choice is now explicit.

druid/src/widget/clip_box.rs
```diff
@@ -237,13 +302,37 @@ impl<T, W> ClipBox<T, W> {
...
             constrain_horizontal: false,
             constrain_vertical: false,
             must_fill: false,
+            managed: true,
+        }
+    }
+
+    /// Creates a new unmanaged `ClipBox` wrapping `child`.
+    ///
+    /// This method should be used when you are using ClipBox in the widget-hierachie directly.
```
jneem, line 327: hierarchy

Maybe also add a sentence (or a link to `managed`) saying what's the actual difference in behavior.

jneem (approved):
Thanks for the changes! Just a couple documentation clarifications, and I think you also need to fix the doc link for CI to be satisfied.

druid/src/widget/clip_box.rs
xarvic: Yes

druid/src/widget/clip_box.rs
xarvic: Ok

xarvic:
Thanks for your time :)

druid/src/window.rs
```diff
@@ -290,11 +290,15 @@ impl<T: Data> Window<T> {
...
             }
 
+            ctx.notifications.retain(|n| n.warn_if_unused_set());
             if !ctx.notifications.is_empty() {
                 info!("{} unhandled notifications:", ctx.notifications.len());
                 for (i, n) in ctx.notifications.iter().enumerate() {
                     info!("{}: {:?}", i, n);
                 }
+                info!(
+                    "if this was intended use EventCtx::submit_notification_unknown_target instead"
```
Fenex, line 300: fix #2210

PR #2145, opened 2022-02-24
Cache layout

xarvic:
This PR changes the implementation of layout for widgets with multiple children. Now these widgets only call `layout` on children which need a layout call. Either because `WidgetPod::requested_layout` is set or because their `BoxConstrains` have changed.


jneem (approved):
Cool!

jpochyla:
Could this be used to avoid the full repaint currently performed after each layout pass?

jneem:
In principle, yes, I think so. It would need some more work though; it won't just happen automatically from this.

xarvic:
The problem is [this](https://github.com/linebender/druid/blob/b684dd65687dd3b6920bfecc45eb6e895bff1733/druid/src/window.rs#L392) line.
But it should be possible get the old and new paint rects of every widget which was moved or resized during layout.

PR #2149, opened 2022-03-08
Unified origin

xarvic:
This PR removes the `viewport_offset` property of `WidgetPod` and changes the `ClipBox` implementation to use `WidgetPod::set_origin` instead. This will fix incorrect hot state when scrolling.

TODOs:
- [x] find a way to merge `WidgetPod::set_origin` and `WidgetPod::set_origin_dyn`
- [x] decide when hot state should change. I think the current implementation, which is doing it during layout is wrong, since we don't know the new origin in layout.
- [x] change TextBox implementation to use `scroll_to_view` and `LifeCycle::ViewContextChanged` instead of manually managing its `Scroll` widget

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -107,6 +107,7 @@ You can find its changes [documented below](#070---2021-01-01).
 - Replace `info_span!` with `trace_span!` ([#2203] by [@NickLarsenNZ])
 - `WidgetPod::event` propagates handled mouse events to active children ([#2235] by [@xarvic])
 - changing hot state is now done in `Lifecycle::ChangeViewState` instead of `layout` ([#2149] by [@xarvic])
+- `WidgetPod::set_origin` no longer takes `data` and `env` as parameters. ([#2149] by [@xarvic])
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,7 +74,7 @@ You can find its changes [documented below](#070---2021-01-01).
 - Added `compute_max_intrinsic` method to the `Widget` trait, which determines the maximum useful dimension of the widget ([#2172] by [@sjoshid])
 - Windows: Dark mode support for the title bar ([#2196] by [@dristic])
 - `ZStack` widget ([#2235] by [@xarvic])
-- `Lifecycle::ViewStateChanged`, `InternalLifecycle::RouteViewStateChanged`, `ChangeCtx`, and `RequestCtx` ([2149] by [@xarvic])
+- `Lifecycle::ViewStateChanged`, `InternalLifecycle::RouteViewStateChanged`, `ChangeCtx`, and `RequestCtx` ([#2149] by [@xarvic])
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,7 @@ You can find its changes [documented below](#070---2021-01-01).
 - Added `compute_max_intrinsic` method to the `Widget` trait, which determines the maximum useful dimension of the widget ([#2172] by [@sjoshid])
 - Windows: Dark mode support for the title bar ([#2196] by [@dristic])
 - `ZStack` widget ([#2235] by [@xarvic])
+- `Lifecycle::ViewStateChanged`, `InternalLifecycle::RouteViewStateChanged`, `ChangeCtx`, and `RequestCtx` ([2149] by [@xarvic])
 
 ### Changed
 
@@ -105,6 +106,7 @@ You can find its changes [documented below](#070---2021-01-01).
 - `RadioGroup` widgets are now constructed with new `row()`, `column()`, and `for_axis()` methods ([#2157] by [@twitchyliquid64])
 - Replace `info_span!` with `trace_span!` ([#2203] by [@NickLarsenNZ])
 - `WidgetPod::event` propagates handled mouse events to active children ([#2235] by [@xarvic])
+- changing hot state is now done in `Lifecycle::ChangeViewState` instead of `layout` ([#2149] by [@xarvic])
 
 ### Deprecated
 
@@ -853,6 +855,7 @@ Last release without a changelog :(
 [#2117]: https://github.com/linebender/druid/pull/2141
 [#2145]: https://github.com/linebender/druid/pull/2145
 [#2148]: https://github.com/linebender/druid/pull/2148
+[#2149]: https://github.com/linebender/druid/pull/2149
 [#2151]: https://github.com/linebender/druid/pull/2151
 [#2157]: https://github.com/linebender/druid/pull/2157
 [#2158]: https://github.com/linebender/druid/pull/2158
--- a/druid/src/contexts.rs
+++ b/druid/src/contexts.rs
@@ -51,6 +51,17 @@ macro_rules! impl_context_method {
     };
 }
 
+/// A macro for implementing context traits for multiple contexts.
+macro_rules! impl_context_trait{
+    ($tr:ty => $ty:ty,  { $($method:item)+ } ) => {
+        impl<'b> $tr for $ty { $($method)+ }
+    };
+    ($tr:ty => $ty:ty, $($more:ty),+, { $($method:item)+ } ) => {
+        impl_context_trait!($tr => $ty, { $($method)+ });
+        impl_context_trait!($tr => $($more),+, { $($method)+ });
+    };
+}
+
 /// Static state that is shared between most contexts.
 pub(crate) struct ContextState<'a> {
     pub(crate) command_queue: &'a mut CommandQueue,
@@ -130,7 +141,7 @@ pub(crate) struct ZOrderPaintOp {
 /// commands.
 pub struct PaintCtx<'a, 'b, 'c> {
     pub(crate) state: &'a mut ContextState<'b>,
-    pub(crate) widget_state: &'a WidgetState,
+    pub(crate) widget_state: &'a mut WidgetState,
     /// The render context for actually painting.
     pub render_ctx: &'a mut Piet<'c>,
     /// The z-order paint operations.
@@ -140,6 +151,176 @@ pub struct PaintCtx<'a, 'b, 'c> {
     /// The approximate depth in the tree at the time of painting.
     pub(crate) depth: u32,
 }
+
+/// The state of a widget and its global context.
+pub struct State<'a, 'b> {
+    #[allow(dead_code)]
+    pub(crate) state: &'a mut ContextState<'b>,
+    pub(crate) widget_state: &'a mut WidgetState,
+}
+
+/// trait for accessing state and widget_state of the context.
+pub trait AnyCtx<'b> {
+    /// Get the state of a widget.
+    ///
+    /// This method can be used to perform merge_up of the widget state with a generic ctx
+    /// This method is intended to be used only by the framework.
+    fn state<'a>(&'a mut self) -> State<'a, 'b>;
+}
+
+/// Convenience trait for code generic over contexts.
+///
+/// Methods to do with commands and timers.
+/// Available to all contexts but PaintCtx.
+pub trait CommandCtx<'b>: AnyCtx<'b> {
+    /// Submit a [`Command`] to be run after this event is handled. See [`submit_command`].
+    ///
+    /// [`submit_command`]: EventCtx::submit_command
+    fn submit_command(&mut self, cmd: impl Into<Command>);
+    /// Returns an [`ExtEventSink`] for submitting commands from other threads. See ['get_external_handle'].
+    ///
+    /// [`get_external_handle`]: EventCtx::get_external_handle
+    fn get_external_handle(&self) -> ExtEventSink;
+    /// Request a timer event. See [`request_timer`]
+    ///
+    /// [`request_timer`]: EventCtx::request_timer
+    fn request_timer(&mut self, deadline: Duration) -> TimerToken;
+}
+
+/// Convenience trait for invalidation and request methods available on multiple contexts.
+///
+/// These methods are available on [`EventCtx`], [`LifeCycleCtx`], and [`UpdateCtx`].
+pub trait RequestCtx<'b>: CommandCtx<'b> {
+    /// Request a [`paint`] pass. See ['request_paint']
+    ///
+    /// ['request_paint']: EventCtx::request_paint
+    fn request_paint(&mut self);
+    /// Request a [`paint`] pass for redrawing a rectangle. See [`request_paint_rect`].
+    ///
+    /// [`request_paint_rect`]: EventCtx::request_paint_rect
+    /// [`paint`]: Widget::paint
+    fn request_paint_rect(&mut self, rect: Rect);
+    /// Request a layout pass. See [`request_layout`].
+    ///
+    /// [`request_layout`]: EventCtx::request_layout
+    fn request_layout(&mut self);
+    /// Request an animation frame. See [`request_anim_frame`].
+    ///
+    /// [`request_anim_frame`]: EventCtx::request_anim_frame
+    fn request_anim_frame(&mut self);
+    /// Indicate that your children have changed. See [`children_changed`].
+    ///
+    /// [`children_changed`]: EventCtx::children_changed
+    fn children_changed(&mut self);
+    /// Create a new sub-window. See [`new_sub_window`].
+    ///
+    /// [`new_sub_window`]: EventCtx::new_sub_window
+    fn new_sub_window<W: Widget<U> + 'static, U: Data>(
+        &mut self,
+        window_config: WindowConfig,
+        widget: W,
+        data: U,
+        env: Env,
+    ) -> WindowId;
+    /// Change the disabled state of this widget. See [`set_disabled`].
+    ///
+    /// [`set_disabled`]: EventCtx::set_disabled
+    fn set_disabled(&mut self, disabled: bool);
+    /// Indicate that text input state has changed. See [`invalidate_text_input`].
+    ///
+    /// [`invalidate_text_input`]: EventCtx::invalidate_text_input
+    fn invalidate_text_input(&mut self, event: ImeInvalidation);
+    /// Scrolls this widget into view.
+    ///
+    /// [`scroll_to_view`]: EventCtx::scroll_to_view
+    fn scroll_to_view(&mut self);
+    /// Scrolls the area into view. See [`scroll_area_to_view`].
+    ///
+    /// [`scroll_area_to_view`]: EventCtx::scroll_area_to_view
+    fn scroll_area_to_view(&mut self, area: Rect);
+}
+
+impl_context_trait!(
+    AnyCtx<'b> => EventCtx<'_, 'b>, UpdateCtx<'_, 'b>, LifeCycleCtx<'_, 'b>, LayoutCtx<'_, 'b>, PaintCtx<'_, 'b, '_>,
+    {
+        fn state<'a>(&'a mut self) -> State<'a, 'b> {
+            State {
+                state: &mut *self.state,
+                widget_state: &mut *self.widget_state,
+            }
+        }
+    }
+);
+
+impl_context_trait!(
+    CommandCtx<'b> => EventCtx<'_, 'b>, UpdateCtx<'_, 'b>, LifeCycleCtx<'_, 'b>, LayoutCtx<'_, 'b>,
+    {
+
+        fn submit_command(&mut self, cmd: impl Into<Command>) {
+            Self::submit_command(self, cmd)
+        }
+
+        fn get_external_handle(&self) -> ExtEventSink {
+            Self::get_external_handle(self)
+        }
+
+        fn request_timer(&mut self, deadline: Duration) -> TimerToken {
+            Self::request_timer(self, deadline)
+        }
+    }
+);
+
+impl_context_trait!(
+    RequestCtx<'b> => EventCtx<'_, 'b>, UpdateCtx<'_, 'b>, LifeCycleCtx<'_, 'b>,
+    {
+        fn request_paint(&mut self) {
+            Self::request_paint(self)
+        }
+
+        fn request_paint_rect(&mut self, rect: Rect) {
+            Self::request_paint_rect(self, rect)
+        }
+
+        fn request_layout(&mut self) {
+            Self::request_layout(self)
+        }
+
+        fn request_anim_frame(&mut self) {
+            Self::request_anim_frame(self)
+        }
+
+        fn children_changed(&mut self) {
+            Self::children_changed(self)
+        }
+
+        fn new_sub_window<W: Widget<U> + 'static, U: Data>(
+            &mut self,
+            window_config: WindowConfig,
+            widget: W,
+            data: U,
+            env: Env,
+        ) -> WindowId {
+            Self::new_sub_window(self, window_config, widget, data, env)
+        }
+
+        fn set_disabled(&mut self, disabled: bool) {
+            Self::set_disabled(self, disabled)
+        }
+
+        fn invalidate_text_input(&mut self, event: ImeInvalidation) {
+            Self::invalidate_text_input(self, event)
+        }
+
+        fn scroll_to_view(&mut self) {
+            Self::scroll_to_view(self)
+        }
+
+        fn scroll_area_to_view(&mut self, area: Rect) {
+            Self::scroll_area_to_view(self, area)
+        }
+    }
+);
+
 
 // methods on everyone
 impl_context_method!(
--- a/druid/src/contexts.rs
+++ b/druid/src/contexts.rs
@@ -321,7 +321,6 @@ impl_context_trait!(
     }
 );
 
-
 // methods on everyone
 impl_context_method!(
     EventCtx<'_, '_>,
--- a/druid/src/contexts.rs
+++ b/druid/src/contexts.rs
@@ -343,7 +343,7 @@ impl_context_method!(EventCtx<'_, '_>, UpdateCtx<'_, '_>, {
 });
 
 //methods on event, update and layout.
-impl_context_method!(EventCtx<'_, '_>, UpdateCtx<'_, '_>, {
+impl_context_method!(EventCtx<'_, '_>, UpdateCtx<'_, '_>, LayoutCtx<'_, '_>, {
     /// Indicate that your view_context has changed.
     ///
     /// Widgets must call this method after changing the clip region of thier children.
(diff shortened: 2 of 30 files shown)


jpochyla:
Hi @xarvic, do you think this is ready to be merged?

xarvic:
I am still waiting for awnser, how to make `WidgetPod::set_origin` callable from multiple contexts. TextBox for an example updates its scroll position when it recieves `LifeCycle::FocusChanged`. Which would otherwise be imposible.

xarvic:
Apart from that the PR is ready for review.

jpochyla:
I think #1611 was missing a use-case inside Druid, and this is a compelling one!

xarvic:
@jpochyla the PR is now ready for review.

druid/src/core.rs
```diff
@@ -389,34 +361,34 @@ impl<T, W: Widget<T>> WidgetPod<T, W> {
     ///
     /// The provided `child_state` should be merged up if this returns `true`.
     fn set_hot_state(
```
xStrom, line 363: This comment needs to be rewritten given that both `rect` and `child_state` are no longer parameters.

druid/src/event.rs
```diff
@@ -14,12 +14,13 @@
 
 //! Events.
 
-use crate::kurbo::{Rect, Shape, Size, Vec2};
+use crate::kurbo::{Rect, Size};
+use std::ops::{Add, Sub};
```
xStrom, line 18: The (currently undocumented) standard of ordering imports is std/external/internal. Given that, please move the `use crate::kurbo::{Rect, Size};` line below, next to the other `use crate::...` lines. Yes it was wrong before too.

druid/src/event.rs
```diff
@@ -303,6 +304,8 @@ pub enum LifeCycle {
     ///
     /// [`EventCtx::is_focused`]: struct.EventCtx.html#method.is_focused
     FocusChanged(bool),
+    ///
+    ViewContextChanged(ViewContext),
```
xStrom, line 308: Missing comment.

druid/src/event.rs
```diff
@@ -332,7 +335,7 @@ pub enum InternalLifeCycle {
     /// Used to route the `DisabledChanged` event to the required widgets.
     RouteDisabledChanged,
     /// The parents widget origin in window coordinate space has changed.
-    ParentWindowOrigin,
+    RouteViewContextChanged(ViewContext),
```
xStrom, line 338: Is this comment still correct for the new `RouteViewContextChanged(ViewContext)`?
Even if so, perhaps you could be so kind as to fix the grammar `parents widget -> parent widget's`. Thanks!

druid/src/event.rs
```diff
@@ -361,54 +364,22 @@ pub enum InternalLifeCycle {
     DebugInspectState(StateCheckFn),
 }
 
-impl Event {
-    /// Transform the event for the contents of a scrolling container.
+/// Information about the widgets surroundings.
```
xStrom, line 367: *widgets* -> *widget's*

druid/src/event.rs
```diff
@@ -361,54 +364,22 @@ pub enum InternalLifeCycle {
...
+#[derive(Debug, Copy, Clone)]
+pub struct ViewContext {
+    /// The origin of this widget's parent relative to the window.
+    pub parent_window_origin: Point,
+
+    /// The last position the cursor was on, relative to the widget.
```
xStrom, line 373: I think it should be: *the cursor was ~on~ at*

druid/src/event.rs
```diff
@@ -361,54 +364,22 @@ pub enum InternalLifeCycle {
...
-                    mouse_event.pos += offset;
-                    Some(Event::Wheel(mouse_event))
-                } else {
-                    None
-                }
-            }
-            _ => Some(self.clone()),
-        }
-    }
+    /// The area may be larger than the widgets paint_rect.
```
xStrom, line 378: 1) *widgets* -> *widget's*
2) If `paint_rect` is a variable put it in quotes.

druid/src/event.rs
```diff
@@ -466,7 +437,32 @@ impl LifeCycle {
...
-            | LifeCycle::BuildFocusChain => false,
+            | LifeCycle::BuildFocusChain
+            | LifeCycle::ViewContextChanged { .. } => false,
+        }
+    }
+
+    /// Returns an event for a widget which maybe is overlapped by another widget.
+    ///
+    /// When ignore is set to `true` the widget will set its hot state to `false` even if the cursor
+    /// is inside its bounds.
```
xStrom, line 448: Given that `ignore` is a variable, put it in quotes.

druid/src/event.rs
```diff
@@ -485,14 +481,26 @@ impl InternalLifeCycle {
...
+            InternalLifeCycle::RouteViewContextChanged { .. } => false,
             InternalLifeCycle::DebugRequestState { .. }
             | InternalLifeCycle::DebugRequestDebugState { .. }
             | InternalLifeCycle::DebugInspectState(_) => true,
         }
     }
 }
 
+impl ViewContext {
+    /// Transforms the `ViewContext` into the coordinate_space of its child.
```
xStrom, line 493: I think its best to remove the underscore here from *coordinate_space*.

druid/src/widget/clip_box.rs
```diff
@@ -82,6 +87,18 @@ impl Viewport {
         }
     }
 
+    /// Sets the component selected by axis of viewport origin to `pos`, while trying to keep the
+    /// view rectangle inside the content rectangle.
+    ///
+    /// Returns true if the position changed. Note that the valid values for the viewport origin
+    /// are constrained by the size of the child, and so the origin might not get set to exactly
+    /// `pos`.
```
xStrom, line 95: Lets put quotes around `axis` and `true`.

druid/src/widget/clip_box.rs
```diff
@@ -335,72 +354,80 @@ impl<T, W: Widget<T>> ClipBox<T, W> {
...
+        env: &Env,
+        delta: Vec2,
+    ) -> bool {
+        self.with_port(ctx, data, env, |_, port| {
+            port.pan_by(delta);
+        })
     }
 
-    /// Changes the viewport offset on the specified axis to 'position'.
+    /// Pans the minimal distance to show the target rect.
```
xStrom, line 377: Is this *rect* referencing the `region` variable? If so, lets change it to `region`.

druid/src/widget/scroll.rs
```diff
@@ -54,23 +55,42 @@ impl<T, W: Widget<T>> Scroll<T, W> {
...
+        &mut self,
+        ctx: &mut C,
+        data: &T,
+        env: &Env,
+        delta: Vec2,
+    ) -> bool {
+        self.clip.pan_by(ctx, data, env, delta)
     }
 
     /// Scroll the minimal distance to show the target rect.
```
xStrom, line 68: Let's change this *rect* to `region` as well.

xStrom (commented):
Thanks for taking this on and sorry for letting it sit so long!

Given that this changes quite a bit of code it's going to take several rounds of review. I'm going to start off with a simple surface level review of small stuff that stuck out as I went over the changes.

Next I will dig deeper to understand what is even going on here. 😄

druid/src/core.rs
```diff
@@ -359,7 +367,7 @@ impl<T, W: Widget<T>> WidgetPod<T, W> {
     ///
     /// Returns `true` if the hot state changed.
     ///
-    /// The provided `child_state` should be merged up if this returns `true`.
+    /// The WidgetPod should merge up its state when `true` is returned.
     fn set_hot_state(
```
xStrom, line 371: This method's comment is still stale. The very first sentence talks about `rect` which is no longer a parameter.

druid/src/event.rs
```diff
@@ -351,7 +357,7 @@ pub enum InternalLifeCycle {
     /// For testing: request the `DebugState` of a specific widget.
     ///
     /// This is useful if you need to get a best-effort description of the
-    /// state of this widget and its children. You can dispatch this event,
+    /// state of this widget and its children. You can dispatch this event,set_hot
```
xStrom, line 360: What's going on with this change, is this *set_hot* a mistaken paste?

druid/src/event.rs
```diff
@@ -364,18 +370,25 @@ pub enum InternalLifeCycle {
     DebugInspectState(StateCheckFn),
 }
 
-/// Information about the widgets surroundings.
+/// Information about the widget's surroundings.
+///
+/// The global origin is also saved in the widget state.
+///
+/// When the ViewContext of widget changes it receives a `ViewContextChanged` event.
```
xStrom, line 377: Let's put `ViewContext` into quotes and add an indefinite article before *widget* - *When the `ViewContext` of **a** widget .....*

xStrom (commented):
Thanks for investing your time into making the adjustments. I have reviewed the latest commit and have a few more minor notes.

I've also been playing around with the code a bit, but a more thorough review of the substance here isn't ready yet.

druid/src/contexts.rs
```diff
@@ -505,21 +507,28 @@ impl_context_method!(EventCtx<'_, '_>, UpdateCtx<'_, '_>, {
     /// Clear the cursor icon.
     ///
     /// This undoes the effect of [`set_cursor`] and [`override_cursor`].
-    ///
+    ///[`PaintCtx`]
```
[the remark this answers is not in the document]
xarvic: Jup, slipped in there, i just removed it :)

druid/src/core.rs
```diff
@@ -125,7 +128,12 @@ pub struct WidgetState {
     /// Some of our children have the `view_context_changed` flag set.
     pub(crate) children_view_context_changed: bool,
 
+    /// indicate that the [`ViewContext`] changed.
     ///
+    /// When this flag is set, the associated widget will receive a `LifeCycle::ViewContextChanged
+    /// event`.
```
xStrom, line 134: `LifeCycle::ViewContextChanged event` -> `LifeCycle::ViewContextChanged` event

druid/src/core.rs
```diff
@@ -272,23 +279,21 @@ impl<T, W: Widget<T>> WidgetPod<T, W> {
...
-            data,
-            env,
-        ) {
-            ctx.widget_state.merge_up(&mut self.state);
+
+        if origin != self.state.origin {
+            self.state.origin = origin;
+            self.state.view_context_changed = true;
+            // identical to calling merge up but faster!
+            ctx.state().widget_state.children_view_context_changed = true;
```
xStrom, line 296: Definitely not *identical* as for example it doesn't do invalidation region handling. Anyway `merge_up()` will be called higher up the stack anyway right? Generally speaking you would call `set_origin()` from somewhere like `layout()` and when that stack unwinds `WidgetPod` will call `merge_up()`. Is there a reasonable scenario where you can call `set_origin()` outside of such a context?

If not then I'm thinking we can just delete this line of code and the `merge_up()` higher up the stack will deal with propagation just fine.

druid/src/core.rs
```diff
@@ -272,23 +279,21 @@ impl<T, W: Widget<T>> WidgetPod<T, W> {
...
-    pub fn set_origin(&mut self, ctx: &mut LayoutCtx, data: &T, env: &Env, origin: Point) {
-        self.state.origin = origin;
+    pub fn set_origin<'a>(
+        &mut self,
+        ctx: &mut impl ChangeCtx<'a>,
+        _data: &T,
+        _env: &Env,
+        origin: Point,
+    ) {
+        //TODO: decide whether we should keep data and env for compatibility or do a breaking change
```
xStrom, line 289: I say go for the breaking change. This is a significant change under the hood and we don't need to keep pretending it didn't happen. Especially as this change is really easy to deal with - you just delete some parameters from your call sites.

What's more, I think we might be able to also delete the `ctx` parameter, see my other comment about `merge_up()` in this method.

xStrom (commented):
Okay, here's my first pass on the code. One interesting thing is that it's looking like there's increasingly fewer reasons to have the context trait changes. We'll see about that, for now can just look over the stuff I commented in this batch.

druid/src/window.rs
```diff
@@ -182,8 +182,13 @@ impl<T: Data> Window<T> {
             );
         }
 
-        if self.root.state().needs_window_origin && !self.root.state().needs_layout {
-            let event = LifeCycle::Internal(InternalLifeCycle::ParentWindowOrigin);
+        if self.root.state().children_view_context_changed && !self.root.state().needs_layout {
+            let event =
+                LifeCycle::Internal(InternalLifeCycle::RouteViewContextChanged(ViewContext {
+                    window_origin: Point::ORIGIN,
```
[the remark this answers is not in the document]
xarvic: The WidgetPod already adds its own origin to `window_orign` before propagating the event. If we changed the value to `self.root.state().origin`, the `window_origin` of the `root_widget` would be two times its origin.

druid/src/contexts.rs
```diff
@@ -142,6 +152,169 @@ pub struct PaintCtx<'a, 'b, 'c> {
     pub(crate) depth: u32,
 }
 
+/// The state of a widget and its global context.
+pub struct State<'a, 'b> {
+    #[allow(dead_code)]
```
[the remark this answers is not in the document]
xarvic: Currently the only method using `State` is `set_origin`, which does not the context_state, but it felt wrong to only add the bits of state i needed to this struct.

druid/src/core.rs
```diff
@@ -1013,10 +964,30 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
...
+
+                        return;
+                    } else if self.state.children_view_context_changed {
+                        extra_event = Some(LifeCycle::Internal(
+                            InternalLifeCycle::RouteViewContextChanged(
+                                view_context.for_child_widget(self.state.origin),
+                            ),
+                        ));
+
+                        self.state.view_context_changed = false;
```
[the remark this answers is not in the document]
xarvic: Oh, right. This is probably a bit confusing...

druid/src/core.rs
xarvic: It is identical in its effect on the parent's state, since `view_context_changed` is the only part of the state which changed and merged up.

> Anyway merge_up() will be called higher up the stack anyway right?

Let's again say we have the following hierarchy of widgets: A is parent of B and B is parent of C. 
Now A calls `B.layout(...)`, in this call B calls `C.layout(...)` in the end the states of B and C are merged.
Then B calls `C.set_origin(...)`. 
Then B.layout(...) ends and B and A are merged, which does not take C's state into account.
Therefore if we don't pass `ctx` in `C.set_origin`, there is no way to merge up the state, since there is no guarantee, that any method of C is called again before paint.

druid/src/core.rs
xarvic: > I say go for the breaking change. This is a significant change under the hood and we don't need to keep pretending it didn't happen. Especially as this change is really easy to deal with - you just delete some parameters from your call sites.

Agreed :)

druid/src/core.rs
```diff
@@ -1013,10 +964,30 @@ impl<T: Data, W: Widget<T>> WidgetPod<T, W> {
...
+                InternalLifeCycle::RouteViewContextChanged(view_context) => {
+                    if self.state.view_context_changed {
+                        self.lifecycle(
+                            ctx,
+                            &LifeCycle::ViewContextChanged(*view_context),
+                            data,
+                            env,
+                        );
+                        self.state.view_context_changed = false;
+                        self.state.children_view_context_changed = false;
```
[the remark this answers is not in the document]
xarvic: 1) You are right, they are redundant.

2) Lets say we have a widget A which has a child widget B. 
When A receives a `Lifecycle::ViewContextChnaged` event, it can first call B.set_origin and then propagate the event.
This works and there are widgets were this is needed.
But doing it the other way round results in inconsistent state.
In the first case the the state is correctly handled and we don't need to set the flags to `true`.
And in the second case setting them to true doesn't help either, since the widget isn't getting another `ViewContextChanged` event, before the next event.

I probably should add this to the docs of `set_origin`.

druid/src/core.rs
xStrom: Thanks for the explanation. I was wrong and you are correct, we need this. (*I was fooled by examples still seeming to work, but that was a mirage.*)

I still don't like the *identical* claim and let me explain a bit more of why. So it's not even so much about what is happening today - it is about how brittle this comment is. When someone will update `merge_up()` in the future to do more they probably won't update this comment. Which means that there will be an out-of-sync comment, if you will.

The better approach is to have the comments describe things that are actually happening in this section, and not be dependant on some more distant piece of code not changing.

On the topic of brittleness and this being a performance trick, in the past we've actually deliberately used `merge_up()` while knowing it has a performance cost. The reason being that the win is a more robust function that won't break as easily as new functionality is added.

Now where exactly to draw the line is subjective of course. I personally have a tolerance for more performance tweaks than just always calling `merge_up()` which is also why I think it's fine if we keep this trick in. However I would prefer to change the wording of the comment.

Perhaps something like:
```rust
// Instead of calling merge_up we directly propagate the flag we care about, to gain performance
```

druid/src/window.rs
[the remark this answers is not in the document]
xStrom: Okay if this is not the same as the root's origin, maybe we can add a short comment for others who might be confused like me. 
```rust
window_origin: Point::ORIGIN, // not the same as the root widget's origin
```

Would this accurately represent the situation?

druid/src/contexts.rs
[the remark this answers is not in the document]
xStrom: That feels wrong to me. YAGNI comes to mind. We are paying a real cost both in terms of compile time and in terms of mental space when reading the code with the extra lifetime names. Yet we aren't getting any value *right now* out of this.

I understand the idea of this being maybe useful in the future, yet this type of stuff usually ends up being a premature thing to do.

I would rather not include this as live code. We could leave it in as a commented-out line with a comment mentioning that this could be enabled when there is an actual need. However the lifetimes of the `State` struct would still need to be redone.

druid/src/window.rs
```diff
@@ -182,8 +182,13 @@ impl<T: Data> Window<T> {
             );
         }
 
-        if self.root.state().needs_window_origin && !self.root.state().needs_layout {
-            let event = LifeCycle::Internal(InternalLifeCycle::ParentWindowOrigin);
+        if self.root.state().children_view_context_changed && !self.root.state().needs_layout {
```
xStrom, line 185: I think `self.root.state().children_view_context_changed` needs to be `widget_state.children_view_context_changed` instead.

When you resize the window `self.root.state().view_context_changed` changes to `true` but `self.root.state().children_view_context_changed` doesn't. The `widget_state` here is a temporary state acting as the virtual parent of the root, and thus has its `children_view_context_changed` as `true`.

For `self.root.state().needs_layout` it's less important and both should work the same, but for consistency I'd change that too to `widget_state.needs_layout`.

CHANGELOG.md
```diff
@@ -105,6 +106,7 @@ You can find its changes [documented below](#070---2021-01-01).
 - `RadioGroup` widgets are now constructed with new `row()`, `column()`, and `for_axis()` methods ([#2157] by [@twitchyliquid64])
 - Replace `info_span!` with `trace_span!` ([#2203] by [@NickLarsenNZ])
 - `WidgetPod::event` propagates handled mouse events to active children ([#2235] by [@xarvic])
+- changing hot state is now done in `Lifecycle::ChangeViewState` instead of `layout` ([#2149] by [@xarvic])
```
xStrom, line 109: Let's add another changelog line under *Changed* about `set_origin`. Perhaps:
```markdown
- `WidgetPod::set_origin` no longer takes `data` and `env` as parameters. ([#2149] by [@xarvic])
```

druid/src/window.rs
```diff
@@ -527,7 +524,7 @@ impl<T: Data> Window<T> {
         let mut ctx = PaintCtx {
             render_ctx: piet,
             state: &mut state,
-            widget_state: &widget_state,
+            widget_state: &mut widget_state,
```
xStrom, line 527: What's the reason for this `mut` here? Seems not needed, the code certainly compiles without it.

druid/src/core.rs
```diff
@@ -81,6 +81,10 @@ pub struct WidgetState {
     /// `size` these constitute the child's layout rect.
     origin: Point,
     /// The origin of the parent in the window coordinate space;
+    //TODO: decide whether we should remove this. With this PR (https://github.com/linebender/druid/pull/2149)
+    // the position of a widget can change during event, but its global position only gets updated
+    // after layout while calling some_pod.layout_rect() gives you the correct local position every
+    // time.
     pub(crate) parent_window_origin: Point,
```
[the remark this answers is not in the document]
xStrom: The comment seems wrong though. It says that the global position only gets updated after layout, but it actually gets updated during `LifeCycle::ViewContextChanged`. So for example when I scroll, the event handler sets a new origin and then `LifeCycle::ViewContextChanged` gets sent from the window, but there is never any layout.

So to me it seems that `parent_window_origin` is stale only temporarily.
- `event()` happens, `set_origin()` gets called
    - `parent_window_origin` is now stale
- `LifeCycle::ViewContextChanged` gets sent, updates `parent_window_origin`
    - `parent_window_origin` is now synced again


Am I misunderstanding something? Because it seems to me right now, that the staleness is only during the specific event that triggers `set_origin()`, after `set_origin()` was called.

druid/src/window.rs
xarvic: Good catch :)

druid/src/window.rs
xarvic: Jup. I forgot to change that back.

druid/src/window.rs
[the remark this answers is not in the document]
xarvic: Yes, that's right.

druid/src/contexts.rs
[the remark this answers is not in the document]
xarvic: Sounds good :)

druid/src/core.rs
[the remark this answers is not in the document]
xarvic: My bad. Yes, when no widget needs layout LifeCycle::ViewContextChanged is sent directly after event and again after update if needed. Im not sure though if this is good. Its probably best to just send it once directly before paint.

> Am I misunderstanding something? Because it seems to me right now, that the staleness is only during the specific event that triggers set_origin(), after set_origin() was called.

Yeah, you are right. And the more i think about it, it seems unlikely, that in the same event / update cycle a widget changes the origin of an widget and another widget further down the hierarchy uses the global origin.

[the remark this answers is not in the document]
xarvic: My bad. Yes, when no widget needs layout LifeCycle::ViewContextChanged is sent directly after event and again after update if needed. Im not sure though if this is good. Its probably best to just send it once directly before paint.

> Am I misunderstanding something? Because it seems to me right now, that the staleness is only during the specific event that triggers set_origin(), after set_origin() was called.

Yeah, you are right. And the more i think about it, it seems unlikely, that in the same event / update cycle a widget changes the origin of an widget and another widget further down the hierarchy uses the global origin.

druid/src/core.rs
xarvic: merge_up() is great for methods which call into the widget trait. The difference here is imo that it is obvious which parts of the state are getting changed by calling `set_origin()`, but i'm completely fine with using `merge_up` instead for consistency. 

I will add the comment for now.

druid/src/window.rs
```diff
@@ -182,21 +182,26 @@ impl<T: Data> Window<T> {
...
+                    last_mouse_position: self.last_mouse_pos,
+                    clip: self.size.to_rect(),
+                }));
             self.lifecycle(queue, &event, data, env, false);
         }
 
         // Update the disabled state if necessary
         // Always do this before updating the focus-chain
-        if self.root.state().tree_disabled_changed() {
+        if widget_state.tree_disabled_changed() {
```
xStrom, line 197: For the record this looks like a bugfix as it has the same *children* tracking logic that view context has. 👍

druid/src/core.rs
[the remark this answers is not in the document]
xStrom: I think that sending `LifeCycle::ViewContextChanged` quickly after event is fine, it reduces the chance of developers using stale data. The performance cost should be negligible.

Overall I think it's fine to keep `parent_window_origin` functioning as it is right now with this PR. Sure it could be improved in some way, but I think that can be proposed and discussed in separate PR if anyone has a desire for it. The problem isn't big enough to be required to solve here. Especially as removing it completely would remove some public API methods too.

What we should do is improve the documentation a bit around this issue. There are two target demographics if you will. Internal Druid developers and external Druid developers.

For the external ones, I think we should add a note to the docs of `set_origin()`.
```rust
    /// The changed origin won't be fully in effect until [`LifeCycle::ViewContextChanged`] has
    /// finished propagating. Specifically methods that depend on the widget's origin in relation
    /// to the window will return stale results during the period after calling `set_origin` but
    /// before [`LifeCycle::ViewContextChanged`] has finished propagating.
```

For the internal ones, let's change the comment of `parent_window_origin`.
```rust
    /// The origin of the parent in the window coordinate space.
    /// 
    /// Note that this value can be stale after `set_origin` was called
    /// but before `LifeCycle::ViewContextChanged` has fully propagated.
```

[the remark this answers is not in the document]
xStrom: I think that sending `LifeCycle::ViewContextChanged` quickly after event is fine, it reduces the chance of developers using stale data. The performance cost should be negligible.

Overall I think it's fine to keep `parent_window_origin` functioning as it is right now with this PR. Sure it could be improved in some way, but I think that can be proposed and discussed in separate PR if anyone has a desire for it. The problem isn't big enough to be required to solve here. Especially as removing it completely would remove some public API methods too.

What we should do is improve the documentation a bit around this issue. There are two target demographics if you will. Internal Druid developers and external Druid developers.

For the external ones, I think we should add a note to the docs of `set_origin()`.
```rust
    /// The changed origin won't be fully in effect until [`LifeCycle::ViewContextChanged`] has
    /// finished propagating. Specifically methods that depend on the widget's origin in relation
    /// to the window will return stale results during the period after calling `set_origin` but
    /// before [`LifeCycle::ViewContextChanged`] has finished propagating.
```

For the internal ones, let's change the comment of `parent_window_origin`.
```rust
    /// The origin of the parent in the window coordinate space.
    /// 
    /// Note that this value can be stale after `set_origin` was called
    /// but before `LifeCycle::ViewContextChanged` has fully propagated.
```

xStrom (approved):
Thanks for your continued effort on this! I think this is ready for merge.

xarvic:
And thank you for the thorough review :)

jpochyla:
Thanks a lot Kaur!

PR #2151, opened 2022-03-12
Allow to use Keys in SizedBox methods

### LINKED ISSUE #2144: Configurable SizedBox size
I want `SizedBox` width/height to be configurable by user. I tried to do this by putting values to env and then calling `.fix_width()` with `Key<f64>`, but it only accepts plain `f64` values.

How to achieve non-static fixed widget size?

GoldsteinE:
Closes #2144


xarvic (approved):
Looks good, thank you :)

PR #2152, opened 2022-03-17
Creating lens from closures

hkoosha:
Hi! (Rust newbie here)

I find it easier in my project to create Lenses from closures. Here's my attempt to simplify things. I have either:
- Came up with a good idea.
- Have got Lenses totally wrong :smile:

Please review and advise. Thanks!

PS: the reason I need closures for my lenses, are that the state struct contains fields, whose types are structs from other traits. And the actual state for a widget is deeply nested in those fields.

### DIFF
--- a/druid/src/lens/lens.rs
+++ b/druid/src/lens/lens.rs
@@ -279,6 +279,42 @@ where
     }
 }
 
+struct LensAdapter<GETTER, SETTER>
+{
+    pub getter: GETTER,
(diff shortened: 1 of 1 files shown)


xarvic:
You understood lenses correctly, but druid already has `lens::Map` which does the same thing :)

hkoosha:
Oh! darn! :D
I'll close this PR. Thanks!

PR #2157, opened 2022-03-25
Implement RadioGroup::{row,column,for_axis} to allow layout customization

twitchyliquid64:
Trivial change to allow the user to set the Axis on which Radio's are laid out. Breaking change in that `RadioGroup::new` is now `RadioGroup::column`.

Builder methods `column()`, `row()`, `for_axis()` are consistent with the `Flex` widget.


jneem (commented):
Thanks for the PR! I'd be inclined to copy the `Flex` naming for consistency, even if it is a breaking change. (e.g. `RadioGroup::column`, `RadioGroup::row`, and `RadioGroup::for_axis`) What do you think about that?

twitchyliquid64:
Sure thing! I'll get on it this weekend :)

twitchyliquid64:
Done! please take a look :)

jneem:
Looks great, thanks! We'll definitely want a changelog entry too (which I can do today if you don't have time).

(Note that the links in the changelog aren't automatic; they're all defined at the bottom)

twitchyliquid64:
Done, thanks!

jneem (approved):
Thanks for your patience! I have to wait for CI to finish, and then I'll merge.

PR #2158, opened 2022-03-26
Fix Controller links for Click

yrns:
Fixed some links in the docs.


jneem (approved):

jneem:
Thanks! Technically, we can now even have rustdoc check the links for us, using something like
```rust
//! [`Controller`]: crate::widget::Controller
```
I'd be happy to merge a PR for that too :)

PR #2162, opened 2022-04-05
Adjust README code example

zpervan:
Looking into the example, the README code snippet contains compilation errors and cannot be run. This PR contains minimal changes to make the example runnable and usable.


jneem:
Thanks for the PR! I think this is intentional, though: the examples in the README are intended to build with the master branch, not with the 0.7 release. There's a link in the README to an example that does build with 0.7.

PR #2169, opened 2022-04-21
Silence clippy.

jneem:
I'm not 100% sure about the naming of `to_string`, which seems pretty overloaded. But `from_wide` was definitely a little confusing...


Maan2003 (approved):
Thanks

Issue #2160, opened 2022-03-28
OpenUI-esque extensible widgets proposal


nickdbush:
I would like to propose that we re-evaluate the base widgets that Druid provides developers at the get-go to ensure that they provide sensible defaults for a good getting-started experience, while also enabling more precise thematic control in an extensible structure.

The [OpenUI project](https://open-ui.org/) is tackling similar concerns for the web platform, and I feel that there is much to learn from them with regard to creating these extensible widgets. While not every component would have an analogous form in Druid, as some are designed specifically for document navigation (such as breadcrumbs), I feel there is a large overlap between what a Druid standard component library should provide and what a web standard component library should provide. For instance, the [editors draft](https://open-ui.org/components/select) for the `<Select />` element is instructive in how we might go about providing styling capabilities without developers needing to build their own:

<img width="800" alt="image" src="https://user-images.githubusercontent.com/10998201/160486483-61d7b0c3-fe70-4808-83d0-4e98cfde6c9b.png">

It's certainly a fair point to argue that perhaps Druid should not provide extensible widgets at this time, and arguments made that this should sit in a third-party library instead. My response would be that having a well-rounded set of highly customisable and extendable components would allow us to move people towards the happy path more effectively in a number of ways. Firstly, for developers just picking up Druid, it would serve as instructive learning material for how robust components should be designed. Secondly, it would allow community contributions to continually improve and fix components across all Druid apps, without the need for complex migration processes from users. I see this being especially important with regards to potential future assistive technology support -- if a button is just a customised version of Druid's platform button, then when that support is introduced, large amounts of that legwork can be adopted easily and is more likely to exist by default in the future.

I'm happy to start taking a look at this in more detail and start drawing up specifications for the components in general (the contract of which will be important if we are to improve components down the line without incurring unexpected behaviour in existing apps). I'm also very happy to pass it on to one of the maintainers if they would rather manage this instead (or if this seems out of scope for the time being).

mwcampbell:
I'm not a Druid maintainer, only a contributor. However, if you have time to prototype this design, I'd like to see how it interacts with accessibility support. I have [a branch of Druid](https://github.com/mwcampbell/druid/tree/accesskit) that implements basic accessibility (only on Windows so far) using my [AccessKit](https://github.com/accesskit/AccessKit) project.

black7375:
Projects that can be referenced in relation to the design system

- https://www.fast.design/

Issue #2159, opened 2022-03-28
Buttons cannot be customized: Can we make them generic over a containing widget?


twitchyliquid64:
As it stands, the current implementation of `widget::Button` cannot be meaningfully customized. Some things that would be nice:

 - Containing other visual cues, such as a icon
 - Increasing the size for buttons that are meant to be prominent

Proposal:

The current implementation nests a `Label<T>` but aside from `Button::dynamic()` (which we can keep as its a builder method) doesn't do anything label-specific in layout/paint. What if we made `Button` generic over its contained widget?

That would allow us to keep the button-specific logic we know and love (the hovering, rounded edges, click visuals etc) but enable custom contents as well as some control over sizing.

Happy to send a PR if this works.

twitchyliquid64:
I agree, it would be nice to have a clearer strat on what the next-gen improvements & patterns are across our widget base. Agree that discussion might make more sense as a supplementary issue. Do you have something in mind?

At least for this issue around buttons, I'm fairly confident that this fits the existing model of composing widgets within other widgets through a builder method, delegating the non-core events/logic down to descendant widgets while handling whats important for itself (in a buttons case, thats the hover effect, the rounded corners, and the `on_click` handler).

nickdbush:
I've popped the [comment above](https://github.com/linebender/druid/issues/2159#issuecomment-1081007534) into a new issue (#2160) – if you have any thoughts regarding the above, do add them there as well!

PR #2170, opened 2022-04-21
Separation of Concern by separating presentation from event handling widgets

vprimachenko:
# Preface

after reading #2159 and #2160 i felt compelled to share some of the ideas gathered in my own GUI experiments

# The Idea

is basically separation of concern. Instead of having a `Button` widget which handels presentation *and* logic this PR proposes a way to attach specific behaviors to  "purely presentation" widgets, thereby allowing for customizability of appearance and interaction.

The issue presented in #2159 would thus be solved with the proposed architecture as follows:

```rust
Row::new((
    Icon::new(Symbol::PlusOnes),
    Label::new("increment"),
))
.handle_event::<LeftClick, _>(|data| {
    *data += 1;
})
```

similarly a OpenUI-like dropdown would wrap widgets handling presentation of the dropdown button, the suggestion list etc in ` Interactive::<ClickController>::new(...)`, ` Interactive::<ScrollController>::new()` and the like.

# Future direction

another idea which emerged from my experiments is to further separate the visual layer by providing a `impl Theme` as a type parameter to the Widgets, where the `Theme` would be responsible for the actual visual presentation, thus superseding current `env` based way. The base `trait Theme` would then provide a default `impl` of a basic, minimally styled widget presentations. (c.f. [Iced per widget `render` traits](https://github.com/iced-rs/iced/blob/0.3/native/src/widget/button.rs#L253-L278))

```rust
Column::new((
    Label::<FancyTheme>::new("i am fancy"),
    Label::<SeriousTheme>::new("i am serious"),
))
```

I am very much looking for feedback on these ideas.

P.S.

i apologise for this PR not being formatted properly, it was not clear to me how PRs for research branches are to be prepared.

### DIFF
--- a/xilem/examples/counter.rs
+++ b/xilem/examples/counter.rs
@@ -13,6 +13,10 @@
 // limitations under the License.
 
 use xilem::{button, v_stack, Adapt, App, AppLauncher, LayoutObserver, Memoize, View};
+use xilem::interactive::{Interactive, ClickController, ClickEvents, InteractiveExt, LeftClick};
+use xilem::label::Label;
+use xilem::EventResult;
+
 
 #[derive(Default)]
 struct AppData {
@@ -25,6 +29,26 @@ fn count_button(count: u32) -> impl View<u32> {
 
 fn app_logic(data: &mut AppData) -> impl View<AppData> {
     v_stack((
+        Interactive::<ClickController>::new(
+            Label::new(format!("this is a clickable label: {}", data.count)),
+            |event, _controller, data: &mut AppData| {
+                data.count += 1;
+                match event {
+                    ClickEvents::LeftClick => println!("LeftClick"),
+                    ClickEvents::MiddleClick => println!("MiddleClick"),
+                    ClickEvents::RightClick => println!("RightClick"),
+                };
+                EventResult::Nop
+            }
+        ),
+
+        format!("this is another clickable label: {}", data.count)
+            .handle_event::<LeftClick, _>(|data: &mut AppData| {
+                data.count += 1;
+                println!("simple click handler");
+                EventResult::Nop
+            }),
+
         format!("count: {}", data.count),
         button("reset", |data: &mut AppData| data.count = 0),
         Memoize::new(data.count, |count| {
--- a/xilem/src/lib.rs
+++ b/xilem/src/lib.rs
@@ -35,3 +35,7 @@ pub use view::vstack::v_stack;
 pub use view::View;
 pub use widget::align::{AlignmentAxis, AlignmentProxy, HorizAlignment, VertAlignment};
 pub use widget::Widget;
+
+pub use view::interactive;
+pub use event::EventResult;
+pub use view::label;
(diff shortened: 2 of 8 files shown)


raphlinus:
Intriguing ideas and I will continue to think about them. I'm not inclined to merge for now (ie I'm not applying the same "optimistic merge" strategy as crochet) because there's still a fair amount of iteration on the core to be done and I want to keep it minimal until I figure out exactly what that should look like. But I definitely appreciate having a concrete proposal (with code) to look at.

I will say this: based on my explorations so far, I think that following the decisions made by SwiftUI should be the default, in other words we should understand those decisions, and if we end up doing something different, we should be prepared to argue persuasively why the approach is better. The documentation for [Button](https://developer.apple.com/documentation/swiftui/button) is a good read - there's a lot going on there, including a trail of breadcrumbs leading through [ButtonStyle](https://developer.apple.com/documentation/swiftui/buttonstyle) as a way of customizing the theme.

I feel like I understand the low-level data flows through idiopath pretty well now, but styling is still a big question mark. As a default, I'm tempted to retain what Druid does now (in the service of getting something working), but I also see the potential to do something better.

vprimachenko:
thank you for the feedback, i am totally on board as keeping this PR as a conversation piece instead of merging.

as to the SwiftUI example you mention: SwiftUI's `Button` is kinda what i envisioned `Interactive::<ClickController>` to be, no dictating to the developer what exactly the `Label` generic should be.

It might be better to change the wording and naming of this structure, for example 

```rust
Behavioural::<ButtonBehaviour>::new(
   Label::new("click me"),
   |_, _, _|{ }
)
```

furthermore, utilizing factory functions in lieu of 
```rust 
#[allow(non_snake_case)]
fn Button<State>(action: impl Fn(&mut State) -> (), label: impl View) -> impl View {
    Behavioural::<ButtonBehaviour>::new(label, move |_, _, data| { action(data) })
}
#[allow(non_snake_case)]
fn Label(text: impl Into<String>) -> impl View;
```

one can achieve a developer experience similar to SwiftUI, sans named parameters and trailing closure syntax

```rust 
Button(frobulate, Label("perform frobulation"))
```

leontepe:
> ```rust
> #[allow(non_snake_case)]
> fn Button<State>(action: impl Fn(&mut State) -> (), label: impl View) -> impl View {
>     Behavioural::<ButtonBehaviour>::new(label, move |_, _, data| { action(data) })
> }
> #[allow(non_snake_case)]
> fn Label(text: impl Into<String>) -> impl View;
> ```
> 
> one can achieve a developer experience similar to SwiftUI, sans named parameters and trailing closure syntax
> 
> ```rust
> Button(frobulate, Label("perform frobulation"))
> ```

I think this is going a little bit too far, at least in that specific form (not adhering to naming conventions is confusing and slapping `#[allow(non_snake_case)]` onto every function is tiresome). Although I do sympathize greatly with trying to get the syntax as comfortable as SwiftUI's.

xStrom:
Xilem now lives in [its own repository](https://github.com/linebender/xilem). Thus pull requests related to Xilem no longer make sense here.

For further discussion open an issue there, or talk about in [Zulip](https://xi.zulipchat.com/#narrow/stream/354396-xilem).

PR #2172, opened 2022-04-24
Introducing intrinsics

sjoshid:
This PR adds `compute_max_intrinsic` function to `Widget` trait. Along with the usual `layout` params, this method also takes an `Axis` param that controls _which_ intrinsic is expected. 

`Axis::Horizontal` = Max **width** intrinsic is expected. 
`Axis::Vertical` = Max **height** intrinsic is expected. 

Each widget gets to decide what its intrinsic max width/height is. 

For some non-container widgets like `Label` it is the width the label would take without any line breaks. 

For container widget like `AspectRatioBox` it depends on its width/height irrespective of child's intrinsic width/height. Child's intrinsic is calculated only if `AspectRatioBox` width/height is infinite. 

~~*There are some `todo`s Im planning to fix before I mark this ready for review. But I wanted to get other main points across with this draft PR.*~~
Addressed all `todo`s.


Zarenor (commented):
Here's a review of this PR; once we've talked through these suggestions, I'll be happy to review again and approve for merging. I'm excited to have this functionality.
I do wonder - is 'max' redundant in the function names? As we've defined it, the intrinsic width is width a widget wishes to be if it has no width constraint - max or min, given the height constraint it has, and vice-versa for height... I'm not sure it's a maximum or a minimum?

druid/src/widget/intrinsic_width.rs
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    fn update(&mut self, ctx: &mut UpdateCtx, old_data: &T, data: &T, env: &Env) {
+        self.child.update(ctx, old_data, data, env);
+    }
+
+    fn layout(&mut self, ctx: &mut LayoutCtx, bc: &BoxConstraints, data: &T, env: &Env) -> Size {
+        let mut bc1 = *bc;
+        //sj_todo make height bounded (height of screen?) while width unbounded
+        bc1.set_max_height(700.);
```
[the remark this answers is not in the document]
sjoshid: That's right. Im going to remove this from draft.

sjoshid:
@Zarenor I agree with your naming suggestions. Actually, anything than what I have makes sense. Im horrible at naming.

druid/src/box_constraints.rs
```diff
@@ -270,6 +270,16 @@ impl BoxConstraints {
...
+
+    /// Set max width to max_width
+    pub fn set_max_width(&mut self, max_width: f64) {
+        self.max.width = max_width;
+    }
+
+    /// Set max height to max_height
+    pub fn set_max_height(&mut self, max_height: f64) {
+        self.max.height = max_height;
+    }
```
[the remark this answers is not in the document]
sjoshid: ^ made me notice there arent any other `&mut` in BC. I added those two, so it is easier for me to _just_ change maxes without changing anything else. I can of course create `Size`s and then create new BC but having the ability to just change maxes makes it so much cleaner.

druid/src/widget/container.rs
```diff
@@ -139,6 +139,21 @@ impl<T: Data> Container<T> {
     pub(crate) fn border_is_some(&self) -> bool {
         self.border.is_some()
     }
+
+    /// max_intrinsic_size = child.max_intrinsic_size + container's border size * 2
+    fn compute_intrinsic<F>(&mut self, bc: &BoxConstraints, env: &Env, f: F) -> f64
+    where
+        F: FnOnce(&BoxConstraints, &mut Box<dyn Widget<T>>) -> f64,
```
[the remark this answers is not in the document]
sjoshid: I made a change to make this clear.

druid/src/widget/flex.rs
```diff
@@ -625,6 +626,116 @@ impl<T: Data> Flex<T> {
...
+            BoxConstraints,
+            &mut WidgetPod<T, Box<dyn Widget<T>>>,
+        ) -> f64,
+        F2: FnMut(
+            f64,
+            &mut LayoutCtx,
+            BoxConstraints,
+            &mut WidgetPod<T, Box<dyn Widget<T>>>,
+        ) -> f64,
+    {
```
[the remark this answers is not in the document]
sjoshid: > instead of do_opposite, can we be a little more descriptive?

Made the change. `measure_cross_intrinsic` is much better than what I had. 

> In the same vein as the earlier comment, f1 and f2 don't tell you much. A more descriptive name would be nice

Correct. Made the change. 

> Also: is it necessary to have different closure types? They look to have the same type constraint?

Right. That's what I had the first time. But compiler says `no two closures, even if identical, have the same type`. They have same signature but what they _do_ differ.

druid/src/widget/flex.rs
```diff
@@ -625,6 +626,116 @@ impl<T: Data> Flex<T> {
...
+    }
+
+    /// This function calculates intrinsic size on main axis. Ie. it calculate intrinsic width of
+    /// [`Axis::Horizontal`] or intrinsic height of a [`Axis::Vertical`].
+    ///
+    /// It returns sum of intrinsic sizes on main axis of each child.
+    fn do_same<F>(&mut self, env: &Env, mut f: F) -> f64
+    where
+        F: FnMut(&mut WidgetPod<T, Box<dyn Widget<T>>>) -> f64,
+    {
```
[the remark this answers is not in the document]
sjoshid: Made the change.

sjoshid:
I forgot to address this point
> I do wonder - is 'max' redundant in the function names? As we've defined it, the intrinsic width is width a widget wishes to be if it has no width constraint - max or min, given the height constraint it has, and vice-versa for height... I'm not sure it's a maximum or a minimum?

`max` is used just so we can, down the road, introduce a `min`. Also, Flutter (inspiration behind this PR) has both.

druid/src/widget/widget.rs
```diff
@@ -222,6 +222,46 @@ pub trait Widget<T> {
             ..Default::default()
         }
     }
+
+    /// Max intrinsic/preferred width is the width a widget would take given infinite width constraint.
```
jneem, line 226: This is a little different from how intrinsic width is [defined in Flutter](https://api.flutter.dev/flutter/rendering/RenderFractionallySizedOverflowBox/computeMaxIntrinsicWidth.html), where max intrinsic width is *"the smallest width beyond which increasing the width never decreases the preferred height"*. Is the difference intentional, and if so could you elaborate on why?

druid/src/widget/flex.rs
```diff
@@ -942,6 +1056,99 @@ impl<T: Data> Widget<T> for Flex<T> {
...
+        bc: &BoxConstraints,
+        data: &T,
+        env: &Env,
+    ) -> f64 {
+        match self.direction {
+            Axis::Horizontal => {
+                let f = |widget: &mut WidgetPod<T, Box<dyn Widget<T>>>| {
+                    widget
+                        .widget_mut()
+                        .compute_max_intrinsic_width(ctx, bc, data, env)
```
jneem, line 1072: Have you considered introducing a `Widget::compute_max_intrinsic(axis: Axis, ctx, bc, data, env)` method? It seems like it might get rid of some duplication here and maybe even eliminate the closure parameters to `measure_main_intrinsic` and `measure_cross_intrinsic`?

druid/src/box_constraints.rs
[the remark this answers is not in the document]
jneem: One way to have the convenience without breaking the mutability convention would be to have `BoxConstraints::with_max_width(&self, max_width) -> BoxConstraints` functions.

But I'm actually wondering if there's a better API that will also ensure that the constraints remain valid (because with `set_max_width` you could break the invariant that the minimum constraint is smaller than the maximum constraint). What about something like:
```rust
/// Sets the max on a given axis to infinity.
pub fn unbound_max(&self, axis: Axis) -> BoxConstraints;
// and maybe, for convenience
pub fn unbound_max_width(&self) -> BoxConstraints;
pub fn unbound_max_height(&self) -> BoxConstraints;

/// Shrinks the max dimension on the given axis.
/// (TODO: specify what should happen if this shrinks the max to less than the min. Decrease the min?
/// Just set the max to the min?)
pub fn shrink_max_to(&self, axis: Axis, dim: f64) -> BoxConstraints;
// and maybe, for convenience
pub fn shrink_max_width_to(&self, dim: f64) -> BoxConstraints;
pub fn shrink_max_height_to(&self, dim: f64) -> BoxConstraints;
```

Then instead of
```rust
    bc.set_max_width(f64::INFINITY);
    bc.set_max_height(available);
```
you can say
```rust
    let bc = bc.unbound_max_width().shrink_max_height_to(available);
```

druid/src/widget/flex.rs
sjoshid: Hmmm. `Axis` makes sense only for flex.

druid/src/box_constraints.rs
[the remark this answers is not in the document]
sjoshid: I like it. BC is nothing but `Sizes` that are cheap to clone. I was just being overly cautions of clones. 
I'll make the change.

druid/src/widget/widget.rs
sjoshid: I honestly feel it makes the definition a little convoluted. For its intents and purposes, what I have is an easier version of that.

druid/src/widget/flex.rs
jneem: Flex is definitely the big one, but I think there are others. In `Container` and `Padding`, `compute_max_intrinsic_height` is basically a copy-paste of `compute_max_intrinsic_width`, and I'd sort of expect that to be the most common case, would you not?

sjoshid: It isnt a copy-paste. One computes height while other computes width. They _look_ the same. My biggest problem when I impl this was that those computes are converse of each other. I guess that's what you are alluding to?
Biggest pain was flex. I could have separated them out (not have closures) but it'd have been a nightmare maintenance wise. For simpler things like `Padding` I choose not to use closures because too much abstraction is difficult to follow.

sjoshid: I still feel maybe Im missing your point. lol. Sorry. .If I did please elaborate or create a gist.

jneem: What I meant was: if you had only one function but it had an `axis: Axis` parameter, then the implementation for `Padding` would just be the following function, with no need to have two functions that look almost the same.
```rust
fn compute_max_intrinsic(
        &mut self,
        axis: Axis,
        ctx: &mut LayoutCtx,
        bc: &BoxConstraints,
        data: &T,
        env: &Env,
    ) -> f64 {
        let inset_size = self.insets.resolve(env).size();
        let child_bc = bc.shrink(inset_size);
        let child_max_intrinsic = self
            .child
            .widget_mut()
            .compute_max_intrinsic(axis, ctx, &child_bc, data, env);
        child_max_intrinsic + axis.major(inset_size)
    }
```

sjoshid: Why always `axis.major(size)` at the end? FYI Im computing not just max width of a `Flex::row` or max height of a `Flex::column`. Im also computing max height of `Flex::row` and max width of a `Flex::column`. When sizing and direction are opposite, you'd need `axis.cross(size)`. No?

jneem: The example above was for `Padding`, not `Flex`.

Yes, `Flex` would be more complicated and involve the cross-axis also. Still, I think passing the axis and using `major` and `cross` would be easier to follow than passing two callbacks.

jneem: No, no: the axis parameter specifies whether we're interested in the max intrinsic width or the max intrinsic height. It's not related to the axis of the Flex. I'm proposing to replace
```rust
fn compute_max_intrinsic_width(&mut self, ...) -> f64;
fn compute_max_intrinsic_height(&mut self, ...) -> f64;
```
by
```rust
fn compute_max_intrinsic(&mut self, Axis::Horizontal, ...) -> f64;
fn compute_max_intrinsic(&mut self, Axis::Vertical, ...) -> f64;
```
because for many widgets I suspect that `compute_max_intrinsic_width` and `compute_max_intrinsic_height` will mostly be copy-pastes of one another.

sjoshid: OK I'll give this a try.

jneem: I'm glad it worked out!

sjoshid: Added another commit.

druid/src/widget/widget.rs
```diff
@@ -222,6 +223,37 @@ pub trait Widget<T> {
...
     }
+
+    /// Computes max intrinsic/preferred dimension of a widget on the provided axis.
+    /// Max intrinsic/preferred dimension is the dimension the widget could take, provided infinite
+    /// constraint on that axis.
+    ///
+    /// If axis == Axis::Horizontal, widget is being asked to calculate max intrinsic width.
+    /// If axis == Axis::Vertical, widget is being asked to calculate max intrinsic height.
+    ///
+    /// Box constrains must be honored in intrinsics computation.
```
jneem, line 234: constraints

druid/src/widget/widget.rs
```diff
@@ -222,6 +223,37 @@ pub trait Widget<T> {
             ..Default::default()
         }
     }
+
+    /// Computes max intrinsic/preferred dimension of a widget on the provided axis.
```
jneem, line 227: Putting a newline after the first short sentence will make the rustdoc output look better, I think

druid/src/widget/flex.rs
```diff
@@ -942,6 +943,113 @@ impl<T: Data> Widget<T> for Flex<T> {
...
+    fn compute_max_intrinsic(
+        &mut self,
+        axis: Axis,
+        ctx: &mut LayoutCtx,
+        bc: &BoxConstraints,
+        data: &T,
+        env: &Env,
+    ) -> f64 {
+        match (self.direction, axis) {
+            (Axis::Vertical, Axis::Horizontal) | (Axis::Horizontal, Axis::Vertical) => {
```
jneem, line 956: Maybe `if self.direction == axis` would be clearer?

jneem (commented):
Nice! I only have nitpicking things left. I think @Zarenor was also taking a look so I'm not marking as "Approved" just yet, but it looks fine to me.

druid/src/widget/flex.rs
sjoshid: 1000%

druid/src/widget/widget.rs
```diff
@@ -222,6 +223,38 @@ pub trait Widget<T> {
...
+    /// Box constraints must be honored in intrinsics computation.
+    ///
+    /// AspectRatioBox is an example where constraints are honored. If height is finite, max intrinsic
+    /// width is *height * ratio*.
+    /// Only when height is infinite, child's max intrinsic width is calculated.
+    ///
+    /// Intrinsic is a *could-be* value. It's the value a widget *could* have given infinite constraints.
+    /// This does not mean the value returned by layout() would be the same.
+    ///
+    /// Make sure this function doesn't return an infinite value.
```
xStrom, line 244: To be precise it's a *method* not just a *function*. We could say `This method **must** return a finite value.`

xStrom (commented):
I did a quick review for some polish. I didn't really do deep analysis of the algorithms though.

Most of my comments are about `use` import ordering. Sorry about that. I would love to have this automated in the future. There has been a [rustfmt proposal](https://github.com/rust-lang/rustfmt/issues/2979) for this since 2018 but nobody has invested the time to get it fully done yet. There does seem to be *some progress* in that rustfmt now has a [group_imports](https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=#group_imports) option, but it's still unstable. I'll see if we can have a better solution to this, or at least some sort of style guide document. But yeah, for now just remember that the general grouping/ordering of imports should be:

```rust
use std::foo;

use some_extrnal_crate::foo;

use crate::my_foo;
```

CHANGELOG.md
```diff
@@ -71,6 +71,7 @@ You can find its changes [documented below](#070---2021-01-01).
 - `EventCtx::submit_notification_without_warning` ([#2141] by [@xarvic])
 - `WidgetPod::requested_layout` ([#2145] by [@xarvic])
 - Make `Parse` work better with floats and similar types ([#2148] by [@superfell])
+- Introduced intrinsics ([#2172] by [@sjoshid])
```
xStrom, line 74: Let's make this a bit more explicit and descriptive. I would suggest something like:
```
`compute_max_intrinsic` method to the `Widget` trait, which determines the maximum useful dimension of the widget ([#2172] by [@sjoshid])
```

We also need to define the `[#2172]` link at the bottom of the changelog where other issue links are defined.

xStrom:
I'll also add for the record, that for me personally `compute_max_intrinsic` seems a bit verbose of a name, specifically the *compute* part. For example, we don't have `compute_layout` - we just have `layout`. However I recongize that this is based on the Flutter method [computeMaxIntrinsicWidth](https://api.flutter.dev/flutter/rendering/RenderFractionallySizedOverflowBox/computeMaxIntrinsicWidth.html) and historically we've had the rule of thumb that when in doubt, do what Flutter does.

So I don't think this is a blocker to get this merged. We can bikeshed over the name later, perhaps as part of the Xilem API rework.

xStrom (approved):
Thanks for your continued effort to get this in!

sjoshid:
Any time. Thank you, @Zarenor  and @jneem for spending time and giving valuable feedback. :)

PR #2176, opened 2022-04-29
Fix readme

ashigirl96:


LoipesMas:
I think it's supposed to be like this, because that's how it works in current master branch. [docs.rs has an example working with 0.7.0](https://docs.rs/druid/latest/druid/#examples), if that's what you are looking for.

Maan2003:
See #1826 #2162

PR #2179, opened 2022-05-03
Narrow the scope of the unsafe block

peamaeq:
Can shrink unsafe block for the Application impl


xStrom (changes requested):
Please run `cargo fmt` to fix the formatting issues.

PR #2183, opened 2022-05-07
Xilem architecture prototype

raphlinus:
I am writing a [blog post on the Xilem architecture](https://github.com/raphlinus/raphlinus.github.io/pull/71), and want that to refer to a PR, as that is the most stable identifier for a branch; the PR can be closed and the branch deleted, and still resolve.

I will also continue to explore and develop in this branch. We might rethink that if we start developing more collaboratively.

Also note, I do the same with the idiopath_py and idiopath_async branches. Those are shorter-term explorations, and basically I just want to capture what I did. It's likely the ideas for both will get merged into this branch.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ members = [
     "druid/examples/web",
     "druid/examples/hello_web",
     "druid/examples/value_formatting",
+    "idiopath",
 ]
 default-members = [
     "druid",
--- a/druid-shell/Cargo.toml
+++ b/druid-shell/Cargo.toml
@@ -16,7 +16,15 @@ default-target = "x86_64-pc-windows-msvc"
 [features]
 default = ["gtk"]
 gtk = ["gdk-sys", "glib-sys", "gtk-sys", "gtk-rs"]
-x11 = ["x11rb", "nix", "cairo-sys-rs", "bindgen", "pkg-config"]
+x11 = [
+    "ashpd",
+    "bindgen",
+    "cairo-sys-rs",
+    "futures",
+    "nix",
+    "pkg-config",
+    "x11rb",
+]
 wayland = [
     "wayland-client",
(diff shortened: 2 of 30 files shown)


jeffparsons:
Two months on, the peanut gallery wonders what impressions/experiences Druid developers and/or users have had with this prototype. Is the mood optimistic? Have unexpected challenges arisen?

I hope this doesn't come across as "are we there yet" — I'm actually just maddeningly curious about what people think about it! 😅

jplatte:
I didn't get the impression that this is in the state where it can be tried out easily, since there was no call for testing or similar.

dhardy:
> The most innovative aspect of Xilem is event dispatching based on an id path, at each stage providing mutable access to app state.

Nice to see that this idea caught on. You can copy [my code for compressed paths](https://github.com/kas-gui/kas/blob/master/crates/kas-core/src/core/widget_id.rs) if you like, though they may be less useful for Xilem since you have separate `Id` and `Path`.

In general I like this idea quite a lot. Some comments:

- Most elements have to implement both `View` and `Widget`
- What's the relationship between `View::event` and `Widget::event`? Aha, the `Widget` impl leverages Druid's existing machinery for handling events, then triggers a "view event". Is this a temporary relationship or could all event handling code be pushed to the "view" side?
- Is it desirable that an app always has one "app data"? What if, instead, an app consisted of any number of "modules", each with its own data and view, with each module's view optionally being a window root and optionally being embedded within another view?
- `EventResult::Nop` — is that "no-op" or "nope"? :laughing: 

Xilem has inspired me to consider something vaguely similar built on top of KAS, which could go some way to solving its biggest weaknesses (difficulty expressing dynamic layout, and not requiring manual updates of widgets dependant on app state). But if we're honest, the Rust UI space is a lot more crowded today than it was when these projects started.

PR #2189, opened 2022-05-18
impl Sync for IdleHandle

raphlinus:
Make IdleHandle implement the Sync bound on all platforms. This is required to wrap IdleHandle in any Arc, motivated in particular by ArcWaker which is in turn motivated by async Xilem experiments (see #2184)


jneem:
Could you also add `Sync` to the assertion at the end of `druid-shell/src/window.rs`?

Issue #2089, opened 2021-12-16
Allow change window visibility programmatically


wspl:
For the visibility of the window needs to be extended:
- Allows to describe in `WindowDesc` whether the window is visible by default or not.
- `WindowHandle::show` should be allowed to be called repeatedly.
- Add `WindowHandle::hide` method to be able to hide the window.

sjoshid:
It doesnt make sense to create a new, hidden window. So by default, a new window must always be visible. 
You can very well hide/show it after its creation. I dont know your exact requirements but I guess you want to run your app in the background. So I think this is a valid request. 
If you have fixed this please open a PR. If not, I can take it up.

newcomb-luke:
I may try to work on this, as it can be useful for very strange projects. One of which I have.

xStrom:
With #2191 merged this should be at least mostly supported now.

If there are specific issues with repeated calls to `WindowHandle::show` or such, that should get a new issue.

PR #2191, opened 2022-05-21
Show/Hide Window Programmatically

newcomb-luke:
This pull request closes issue  #2089.
You can now hide and show a window programmatically. This is done by a new hide() method on WindowHandle, and by a new command Selector called `HIDE_WINDOW`. This has been implemented on the x11, GTK+, Windows, and macOS backends. Wayland does not yet seem to support hiding/showing windows through their own protocol, and window managers seem to be currently just falling back to the x11 protocol, so Wayland implementation was held off.
`HIDE_WINDOW` leaves the macOS only `HIDE_APPLICATION` Selector alone in that `HIDE_WINDOW` does what it says, and only hides a specific window.
The Window can be shown again by the command `SHOW_WINDOW` which calls the WindowHandle's bring_to_front_and_focus method, which still works as intended when the window is already shown, but when it is hidden, it shows the window again, and brings it to the front and focuses it.


newcomb-luke:
Seems to still work well with all of the new code changes that have been going on.

yim7:
Thanks, I used your code.

`hide` feature works perfectly, but i can't show window after hiding it.
The deferred_queue cannot be processed because the window does not receive events after hiding. We should call the window display function directly.
And `ShowWindow` don't bring the window into foreground. We can use `SetForegroundWindow` to implement `bring_to_front_and_focus`

newcomb-luke:
> Thanks, I used your code.
> 
> `hide` feature works perfectly, but i can't show window after hiding it. The deferred_queue cannot be processed because the window does not receive events after hiding. We should call the window display function directly. And `ShowWindow` don't bring the window into foreground. We can use `SetForegroundWindow` to implement `bring_to_front_and_focus`

What version of Windows are you trying to run it on? Maybe it is being treated differently. I've tested it on Windows 10.

How did you try to show the window? When I spawn a thread and run `handle.submit_command(druid::commands::SHOW_WINDOW, (), Target::Window(window_id))`, it works perfectly, and shows the window again. The program would crash when I tried to `ShowWindow` in a non-deferred way. 

I will implement `SetForegroundWindow` however, to make it bring the window to the front.

newcomb-luke:
I've added SetForegroundWindow, and everything should be up to date with master.

newcomb-luke:
Synced with master.

xStrom (approved):
This looks reasonable, thanks for your effort!

PR #2195, opened 2022-06-03
Preserve the aspect ratio of a clipped region in an Image

barsae:
Previously, Image always used the full (non-clipped) image for computing
the aspect ratio to maintain for FillStrats like Contain.  This could
cause clipped regions to be unexpectedly stretched.

This also adds support to the Image example for clipping.


derekdreery (approved):
I've looked through this and it looks good. AFAICT preserving aspect ratio in this situation is universally what you would want/expect druid to do.

PR #2196, opened 2022-06-04
Add dark mode support for Windows.

dristic:
Small edit to let Windows theme the title bar with dark mode support. I followed the instructions at the bottom of this page: https://docs.microsoft.com/en-us/windows/apps/desktop/modernize/apply-windows-themes

Here is how it looks:
DarkMode

First time contributing so let me know if I missed anything. Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,7 @@ You can find its changes [documented below](#070---2021-01-01).
 - `WidgetPod::requested_layout` ([#2145] by [@xarvic])
 - Make `Parse` work better with floats and similar types ([#2148] by [@superfell])
 - Added `compute_max_intrinsic` method to the `Widget` trait, which determines the maximum useful dimension of the widget ([#2172] by [@sjoshid])
+- Windows: Dark mode support for the title bar ([#2196] by [@dristic])
 
 ### Changed
 
@@ -554,6 +555,7 @@ Last release without a changelog :(
 [@superfell]: https://github.com/superfell
 [@GoldsteinE]: https://github.com/GoldsteinE
 [@twitchyliquid64]: https://github.com/twitchyliquid64
+[@dristic]: https://github.com/dristic
 
 [#599]: https://github.com/linebender/druid/pull/599
 [#611]: https://github.com/linebender/druid/pull/611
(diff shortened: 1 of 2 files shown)


jneem:
Thanks for the PR! Given that our theme is dark, this makes total sense. But can I also check my understanding of the situation with you? (I don't use windows much, so I might be way off.)

So the page you linked to describes a way of determining whether the theme is "dark" or "light," and the ideal would be to adapt druid's default colors according to that. This PR does *not* do anything about changing druid's theme, but it does tell windows that druid supports dark themes, so that windows can draw a dark titlebar when a dark theme is enabled. This is a little misleading but it's still an improvement over the status quo (and we'll want this code anyway, once we start actually supporting dark/light modes).

dristic:
Yes you are correct. This allows Windows to restyle the title bar to the system-preferred mode. This is disabled by default for backwards compatibility reasons. It does not allow the application to respond to the system-preferred mode. This would probably take a bit more work to pipe the events up to the application layer through the shell and write the code to detect it. I could look into it in a future change when I have some more time!

dristic:
It looks like the function to query theme color is in the windows-rs library so I would have to pull this in as a dependency to add the theme mode detection functions: https://github.com/microsoft/windows-rs/blob/cbc54970034445e0e50ae548ab8d93768548a4fb/crates/libs/windows/src/Windows/UI/ViewManagement/mod.rs#L3677

jneem:
Got it, thanks! I think there was some discussion in the past about moving to windows-rs, but I don't remember the conclusion. And anyway it's definitely out of scope for this PR!

Would you mind adding an entry to the changelog? Other than that, I think this is good to go.

xStrom:
Yes `windows-rs` is most likely a future dependency. However we've been postponing it because in its current state it causes slow compiles and also links the final binary to libraries  that we don't actually need. Microsoft has said that they're working on addressing those issues.

dristic:
Add a changelog entry, thanks!

jneem:
Sorry, I forgot to tell you: the PR and author links in the changelog aren't magic: they're all defined at the bottom of the file...

dristic:
Fixed, thanks!

jneem (approved):

xStrom:
Finally got around to testing this and as far as I can tell this doesn't actually have any automatic theme switching by the OS. Now Druid apps always have a dark title bar.

@dristic have you seen the title bar be light after this change?

dristic:
When testing locally I see the same thing. It looks like I have to add the color detection hook in to see if we need to set the title bar to dark or not. I will get this fixed and open another PR. Good catch!

Issue #2190, opened 2022-05-18
Dark or light mode detection


marcelohdez:
Hello, is there any way to detect light or dark mode on the system an application is running on to change themes respectively?

After looking for a while I can't find any mention of it in Druid, and no Druid example seems to adhere to the dark/light mode setting (at least on MacOS) despite its [theme containing keys](https://docs.rs/druid/latest/druid/theme/index.html#constants) for [dark](https://docs.rs/druid/latest/druid/theme/constant.BACKGROUND_DARK.html) and [light](https://docs.rs/druid/latest/druid/theme/constant.BACKGROUND_LIGHT.html) colors?

Should I just rely on the [dark-light](https://crates.io/crates/dark-light) crate for now and check periodically?

<img width="1139" alt="Screen Shot 2022-05-18 at 10 48 09 AM" src="https://user-images.githubusercontent.com/76508651/169070587-a95125a1-954e-403d-8b0b-c11dc3f46f8e.png">

xStrom:
#2196 and #2204 made some first steps in implementing the dark/light detection for Windows in `druid-shell`. It's done once though and is not really reactive to runtime changes, nor does it expose this to `druid` apps themselves. That work remains to be done in the future.

Those issues also mention the potential of adding a dependency to `windows-rs` to get better access to this information. Other options might include `WM_THEMECHANGED`, `WM_SYSCOLORCHANGE`, and `RegNotifyChangeKeyValue`.

PR #2204, opened 2022-06-12
Add detection for dark or light theme.

dristic:
From https://github.com/linebender/druid/pull/2196#issuecomment-1152968392

The windows API does not actually detect dark mode but forces it on. This is the simplest way to detect dark mode without depending on UWP APIs that require windows-rs. If it fails to read the registry it defaults to a light theme.

I tested this locally and the app will use the dark theme if the user's OS is set to dark otherwise it uses a light theme.

It does not detect this if the settings change while running the app but without UWP APIs I believe I would have to poll the registry which does not feel like a great solution.

Thanks!

### DIFF
--- a/druid-shell/src/backend/windows/window.rs
+++ b/druid-shell/src/backend/windows/window.rs
@@ -42,6 +42,7 @@ use winapi::um::unknwnbase::*;
 use winapi::um::uxtheme::*;
 use winapi::um::wingdi::*;
 use winapi::um::winnt::*;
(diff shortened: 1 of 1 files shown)


vprimachenko:
Consider checking whether a windows event signals the change. `WM_THEMECHANGED` and `WM_SYSCOLORCHANGE` come to mind.

xStrom:
@vprimachenko thanks for the specific pointers, I wrote those down in #2190 which tracks progress on this.

This PR here was just a quick follow-up to #2196 and thus limited in scope on purpose.

PR #2206, opened 2022-06-14
feat(druid-shell): added `window` mode on wasm target

Fenex:
A `window` mode has been added: restricted area by canvas size will be
use to draw an application on a web page. The mode will be enabled if
`canvas` element has both attributes: `height` and `width`.

A `fullscreen` mode will be on by default (`height` or\and `width`
attrs are missed)

изображение

### DIFF
--- a/druid-shell/src/backend/web/window.rs
+++ b/druid-shell/src/backend/web/window.rs
@@ -171,7 +171,7 @@ impl WindowState {
         let dpr = w.device_pixel_ratio();
 
         match self.canvas_size {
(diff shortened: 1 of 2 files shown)


Fenex:
Not sure, is correct name "window mode" for that or not )

jneem:
Thanks for the PR and sorry for the slow reponse! The CI is complaining because (1) it expects the code to be formatted with rusfmt and (2) there were some issues with the master branch. I've rebased and formatted, so hopefully it will succeed now.

jneem (approved):

PR #2208, opened 2022-06-20
Update nix to 0.24.0

cmyr:
I haven't tested this, so if anyone running linux wants to +1 that would be great. :)


xarvic:
I just tested it. It works on Linux.

i509VCB:
We could disable the default features and enable what we use, but most crates do not do that yet...

cmyr:
I'm going to close this, someone should do this who is actually developing on linux. 😅

PR #2213, opened 2022-06-29
Fix double-release bug in macOS save dialog.

### LINKED ISSUE #2209: segfault from `cargo run --example open_save`
afbe73b92e7f54a8c9e0800a1c440252b80ae0ba with rustc 1.61.0 (fe5b13d68 2022-05-18), try saving a file, after using the default name and location it segfaulted.

```
     Running `target/debug/examples/open_save`
2022-06-21T15:46:21.929670Z DEBUG druid::localization: available locales [], current en-US
2022-06-21T15:46:21.929897Z DEBUG druid::localization: resolved: [en-US]
2022-06-21T15:46:21.974916Z DEBUG druid::window: TextFieldToken(1) added
Segmentation fault: 11
```

xStrom:
This PR fixes a double-release bug in `druid-shell` in the macOS file save dialog code path.

It's not completely clear to me when an object needs manual releasing. The problematic object in question here is a string returned by [`stringByAppendingPathExtension`](https://developer.apple.com/documentation/foundation/nsstring/1412501-stringbyappendingpathextension) and the Apple documentation for that mentions nothing.

The best I found was an [old comment on StackOverflow](https://stackoverflow.com/a/4354184/138826) which claims that *The convention is that a method should autorelease any object it returns. The only exception (AFAIK) is for constructors*. Looking at our existing macOS code, both `alloc` and `new` seem to require releasing. Are there others? Would be nice to know for future code reviews.

Right now though, it seems clear that `stringByAppendingPathExtension` sets up `autorelease` by itself, so we don't need to do it.

Before the change in this PR, on macOS 10.15.4, I get `objc[484]: NSPathStore2 object 0x7f852fb2ed50 overreleased while already deallocating; break on objc_overrelease_during_dealloc_error to debug`. After this change, the warning disappears.

I am going to assume, based on the stack trace provided, that this will also resolve the segfault reported in #2209. @SamuelMarks could you please test if this resolves the segfault you're seeing?

Fixes #2209.

PS. This PR also contains a pre-emptive robustification by replacing a few `NSString` allocations/releasings with the util function we already have for that.

### DIFF
--- a/druid-shell/src/backend/mac/dialog.rs
+++ b/druid-shell/src/backend/mac/dialog.rs
@@ -20,9 +20,7 @@ use std::ffi::OsString;
 
 use cocoa::appkit::NSView;
 use cocoa::base::{id, nil, NO, YES};
-use cocoa::foundation::{
-    NSArray, NSAutoreleasePool, NSInteger, NSPoint, NSRect, NSSize, NSString, NSURL,
-};
+use cocoa::foundation::{NSArray, NSAutoreleasePool, NSInteger, NSPoint, NSRect, NSSize, NSURL};
 use objc::{class, msg_send, sel, sel_impl};
 
 use super::util::{from_nsstring, make_nsstring};
@@ -222,9 +220,7 @@ unsafe fn file_format_popup_button(allowed_types: &[crate::FileSpec], popup_fram
     let popup_button: id = msg_send![class!(NSPopUpButton), alloc];
     let _: () = msg_send![popup_button, initWithFrame:popup_frame pullsDown:false];
     for allowed_type in allowed_types {
-        let title = NSString::alloc(nil)
-            .init_str(allowed_type.name)
-            .autorelease();
+        let title = make_nsstring(allowed_type.name);
         msg_send![popup_button, addItemWithTitle: title]
     }
     let _: () = msg_send![popup_button, setTag: FileFormatPopoverTag];
@@ -237,7 +233,7 @@ unsafe fn file_format_label() -> (id, NSSize) {
     let _: () = msg_send![label, setDrawsBackground:false];
     // FIXME: As we have to roll our own view hierachy, we're not getting a translated
     // title here. So we ought to find a way to translate this.
-    let title = NSString::alloc(nil).init_str("File Format:").autorelease();
+    let title = make_nsstring("File Format:");
     let _: () = msg_send![label, setStringValue: title];
     let _: () = msg_send![label, sizeToFit];
     (label.autorelease(), label.frame().size)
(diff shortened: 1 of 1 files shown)


cmyr (approved):
Yes, this is annoying.

In the olden days, before objective-c had autorelease, every object was init'd with,

`[[MyType alloc] myInitMethod: myArgs]`; (e.g. (`MyObject::alloc().init(my_args)`), and this would return a pointer that had manual reference counting semantics, and you would be obliged to call `[object release]` when you were done with it, to free the memory.

Sometime before Swift, the objective-c runtime was updated to support automatic reference counting, which is what the 'autorelease' call is opting us into.

We should assume that all platform Cocoa APIs are using automatic reference counting, and also that the foundation APIs (which are written in C, not Objective-C) are not.

So to the best of my understanding:

- anytime we create an object with a manual `alloc` call, we should call `autorelease` exactly once.
- any time we make an object through a constructor  (without calling `alloc` ourselves) then we can assume that it is autoreleased,
- *unless* it is a foundation type, (which will have a `Cf` or `Ct` or similar prefix) in which case the ownership semantics follow the Core Foundation [ownership policy](https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFMemoryMgmt/Concepts/Ownership.html#//apple_ref/doc/uid/20001148-CJBEJBHH).

SamuelMarks:
Thanks, good drill-down. Solved the issue.

PS: I built everything from the tree at your PR, and got:
```
error[E0432]: unresolved import `raw_window_handle::macos`
  --> druid-shell/src/backend/mac/window.rs:43:25
   |
43 | use raw_window_handle::{macos::MacOSHandle, HasRawWindowHandle, RawWindowHandle};
   |                         ^^^^^ could not find `macos` in `raw_window_handle`

error[E0432]: unresolved import `crate::dialog::FileDialogType`
  --> druid-shell/src/backend/mac/dialog.rs:27:40
   |
27 | use crate::dialog::{FileDialogOptions, FileDialogType};
   |                                        ^^^^^^^^^^^^^^ no `FileDialogType` in `dialog`

error[E0432]: unresolved import `crate::dialog::FileDialogType`
  --> druid-shell/src/backend/mac/window.rs:58:40
   |
58 | use crate::dialog::{FileDialogOptions, FileDialogType};
   |                                        ^^^^^^^^^^^^^^ no `FileDialogType` in `dialog`

error[E0599]: no variant or associated item named `MacOS` found for enum `RawWindowHandle` in the current scope
    --> druid-shell/src/backend/mac/window.rs:1422:26
     |
1422 |         RawWindowHandle::MacOS(handle)
     |                          ^^^^^ variant or associated item not found in `RawWindowHandle`
```

Issue #2211, opened 2022-06-28
Can't build druid-shell on Mac with error: `*const Class` cannot be sent between threads safely


Steve-xmh:
I was using stable rust compiler `stable-x86_64-apple-darwin - rustc 1.61.0 (fe5b13d68 2022-05-18) (from rustc 1.61.0 (fe5b13d68 2022-05-18))` and try to build my program with druid (Commit 7d08672), but some errors occured.

I've already checked #176 which had the same problem, but upgrading rustc doesn't work for me as I'm using Github Action.

Error output:
```
error[E0277]: `*const Class` cannot be sent between threads safely
--> /Users/runner/.cargo/git/checkouts/druid-f71533f3d81c0bc8/7d08672/druid-shell/src/backend/mac/application.rs:160:1
|
160| /lazy_static! {
161| |    static ref APP_DELEGATE: AppDelegate = unsafe {
162| |        let mut decl = ClassDecl::new("DruidAppDelegate", class!(NSObject))
163| |            .expect("App Delegate definition failed");
...|
176| |    };
177| |}
| |_^`*const Class` cannot be sent between threads safely
|
= help: within `AppDelegate`, the trait `Send` is not implemented for `*const Class`
note: required because it appears within the type `AppDelegate`
--> /Users/runner/.cargo/git/checkouts/druid-f71533f3d81c0bc8/7d08672/druid-shell/src/backend/mac/application.rs:157:8
|
157| struct AppDelegate(*const Class);
| ^^^^^^^^^^^
= note: required because of the requirements on the impl of `Sync` for `spin::once::Once<AppDelegate>`
= note: required because it appears within the type `lazy_static::lazy::Lazy<AppDelegate>`
= note: shared static variables must have a type that implements `Sync`
= note: this error originates in the macro `__lazy_static_create` (in Nightly builds, run with -Z macro-backtrace for more info)
error[E0277]: `*const Class` cannot be sent between threads safely
--> /Users/runner/.cargo/git/checkouts/druid-f71533f3d81c0bc8/7d08672/druid-shell/src/backend/mac/window.rs:339:1
|
339| /lazy_static! {
340| |    static ref VIEW_CLASS: ViewClass = unsafe {
341| |        let mut decl = ClassDecl::new("DruidView", class!(NSView)).expect("View class defined");
342| |        decl.add_ivar::<*mut c_void>("viewState");
...|
524| |    };
525| |}
| |_^`*const Class` cannot be sent between threads safely
|
[...]

PR #2214, opened 2022-07-01
Fix build error on Mac when spin_no_std is enabled by dependents of lazy_static. (#2211)

Steve-xmh:
Fix build error on Mac when spin_no_std is enabled by dependents of lazy_static. (#2211)

### DIFF
--- a/druid-shell/src/backend/mac/application.rs
+++ b/druid-shell/src/backend/mac/application.rs
@@ -156,6 +156,7 @@ impl DelegateState {
 
 struct AppDelegate(*const Class);
 unsafe impl Sync for AppDelegate {}
+unsafe impl Send for AppDelegate {}
 
 lazy_static! {
     static ref APP_DELEGATE: AppDelegate = unsafe {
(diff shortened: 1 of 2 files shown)


xStrom:
Thanks! We may end up going with this solution, or we may switch out `lazy_static` for `once_cell` as briefly discussed in #2211. I will look into this myself soon.

Steve-xmh:
> Thanks! We may end up going with this solution, or we may switch out `lazy_static` for `once_cell` as briefly discussed in #2211. I will look into this myself soon.

That's ok, take your own time!

Anyway, switch to `once_cell` can be a good choice too. But we can merge it first if just need to temporarily fix this problem.

xStrom:
I think clippy will fail, as I'm still working on that PR myself. I'll run the tests to see.

xStrom:
Okay, `master` now has clippy issues solved thanks to #2215. You can rebase this on `master` and then the tests should succeed.

Steve-xmh:
Already merged.

xStrom (approved):
Thanks! This does resolve the issue, so I'll approve this as the quick fix to have a working `master`.

Later we might switch out `lazy_static` completely, but that will be a separate PR.

PR #2215, opened 2022-07-01
Satisfy clippy 1.62

xStrom:
Lints addressed:
* [`format_push_string`](https://rust-lang.github.io/rust-clippy/master/index.html#format_push_string)
* [`let_unit_value`](https://rust-lang.github.io/rust-clippy/master/index.html#let_unit_value)

### DIFF
--- a/druid-shell/src/backend/gtk/window.rs
+++ b/druid-shell/src/backend/gtk/window.rs
@@ -76,11 +76,11 @@ const SCALE_TARGET_DPI: f64 = 96.0;
 /// Taken from <https://gtk-rs.org/docs-src/tutorial/closures>
 /// It is used to reduce the boilerplate of setting up gtk callbacks
 /// Example:
-/// ```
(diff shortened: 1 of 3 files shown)


jneem (approved):

xStrom:
Looks like some pseudocode in GTK backend comments is failing, will need to look into that a bit more.

jneem:
Those doc comments have been there since forever. Strange that `rustdoc` only just decided to try compiling them...

xStrom:
I don't know for sure, but this line in the changelog might explain it:

`cargo test` now passes `--target` to `rustdoc` if the specified target is the same as the host target. [#10594](https://github.com/rust-lang/cargo/pull/10594)

xStrom:
The tests require the code to be full, even if most of it is hidden. Given that it is only one macro's docs, I'll just set the ignore tag for it.

PR #2218, opened 2022-07-04
Update nix to 0.24.0

Maan2003:
#2208 revived


cmyr (approved):
Thanks!

CryZe:
What about reducing the feature set?

cmyr:
@CryZe can you elaborate?

CryZe:
nix 0.24 introduces lots of features. They are enabled by default, but you probably want to actually make use of this and reduce it to only what is actually used.

Issue #1711, opened 2021-04-10
On master branch, Cmd+Q doesn't quit app, but emits warning instead


TonalidadeHidrica:
I recently switched from `v0.7.0` to `master`.  I recognized that Cmd+Q does not quit the app anymore.  This happens not only in my project but also in the example (at least, `flex` example).  Instead, it emits a warning every time I submit the keystroke, apparently related to OS layer:

`cargo run --example flex`
```
2021-04-10 13:37:22.456 flex[74291:7719784] +[NSView doCommandBySelector:]: unrecognized selector sent to class 0x7fff885efa08
```

This happens only in `master`.  In `v0.7.0`, I can quit the app with Cmd+Q without any warning.

# Environment

- druid master (2c5d73c8426f90d2a1ede9fdd3b5cb7bee57f108)
- macOS Big Sur 11.2.3 (Intel)

I can provide further environmental information if needed.

TonalidadeHidrica:
I ran `git bisect` and found the (possible) cause of the commit: 8e09a6d5847f83b194dced8c5b9cfa3381b80371.

```
# bad: [2c5d73c8426f90d2a1ede9fdd3b5cb7bee57f108] Add missing derived lens documentation (#1696)
# good: [ed4f9ef0e763d8396ef2fb7facd8ea4ba541c41e] Prepare for 0.7.0
git bisect start '2c5d73c' 'v0.7.0'
# good: [3b902a993623d5ed4e96693d4efbc0ece1fd2333] Fixup changelog links
git bisect good 3b902a993623d5ed4e96693d4efbc0ece1fd2333
# good: [9c361872050cac1aa055bce2e908ad23adc9341d] Relax constraints on who can change focus
git bisect good 9c361872050cac1aa055bce2e908ad23adc9341d
# bad: [0e88d99692ff0163cb0f0838e915b29a25732a77] Fix another clippy lint
git bisect bad 0e88d99692ff0163cb0f0838e915b29a25732a77
# good: [a281a814d1bc61a5169491991367ecb53fd0f383] Unify druid & druid-shell selection types
git bisect good a281a814d1bc61a5169491991367ecb53fd0f383
# bad: [75e83ae5a73fb1da2407eb157580145d833e1caf] updating changelog for on_added WidgetExt method (#1668)
git bisect bad 75e83ae5a73fb1da2407eb157580145d833e1caf
# good: [dde1587267ef0f0f8f4275276aef9288397f8678] Unify druid & druid-shell Movement types
[...]

PR #2221, opened 2022-07-05
[macos] Fix issue with missing default menu

### LINKED ISSUE #2008: [mac] How do I fully exit an app?
In druid demo running on macOS, click exit button, the window will disappered, but the icon is still on the dock. the app cannot active again.

cmyr:
This has been an issue since the menu refactor, apparently, but i hadn't
bothered to track it down until now. I think it might have also been surfaced
by an OS release?

With this fix, macos windows will have a standard looking menu, which
will provide functionality like cmd-Q to quit. This likely closes a
number of open issues.

- closes #2008
- references #1711 (which this may fix?)


jneem (approved):
I'm not really qualified to review this, but I'll give you the green button anyway

PR #2228, opened 2022-07-19
Support open/save dialogs on wayland.

jneem:
There's one little to-do left: the parenting isn't right because the window-identifier business in `ashpd` isn't working right now because of wayland crate version mismatches.

### DIFF
--- a/druid-shell/src/backend/wayland/surfaces/layershell.rs
+++ b/druid-shell/src/backend/wayland/surfaces/layershell.rs
@@ -4,6 +4,8 @@ use wayland_protocols::xdg_shell::client::xdg_surface;
 
 use crate::kurbo;
 use crate::window;
+use crate::FileDialogOptions;
+use crate::FileDialogToken;
 
 use super::super::error;
 use super::super::outputs;
@@ -367,6 +369,14 @@ impl Handle for Surface {
 
     fn invalidate_rect(&self, rect: kurbo::Rect) {
         return self.inner.wl_surface.borrow().invalidate_rect(rect);
+    }
+
+    fn open_file(&self, options: FileDialogOptions) -> Option<FileDialogToken> {
+        self.inner.wl_surface.borrow().open_file(options)
+    }
+
+    fn save_as(&self, options: FileDialogOptions) -> Option<FileDialogToken> {
+        self.inner.wl_surface.borrow().save_as(options)
     }
 
     fn remove_text_field(&self, token: crate::TextFieldToken) {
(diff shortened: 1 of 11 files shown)


druid-shell/src/backend/wayland/surfaces/layershell.rs
```diff
@@ -369,6 +371,14 @@ impl Handle for Surface {
         return self.inner.wl_surface.borrow().invalidate_rect(rect);
     }
 
+    fn open_file(&self, options: FileDialogOptions) -> Option<FileDialogToken> {
```
james-lawrence, line 374: minor question: do we need to add these to every surface or can we just use the already implemented data() method to access the implementations directly vs expanding the api surface of handle.

druid-shell/src/backend/wayland/surfaces/surface.rs
```diff
@@ -469,6 +478,26 @@ impl Data {
         self.schedule_deferred_task(DeferredTask::Paint);
     }
 
+    fn open_file(&self, options: FileDialogOptions) -> Option<FileDialogToken> {
+        // FIXME: current ashpd has issues with wayland versions
```
james-lawrence, line 482: is there a link to the upstream issue we could put here as a reference?

druid-shell/src/backend/wayland/surfaces/layershell.rs
jneem: I was kinda wondering that too, but it seemed better to just follow what all the other methods were doing... I'd be happy to change it, though

druid-shell/src/backend/wayland/surfaces/surface.rs
jneem: I guess [this](https://github.com/bilelmoussaoui/ashpd/pull/76). Which is merged, but there's still an issue with `asphd`'d wayland-client version not matching ours (and [depending on a pre-release](https://github.com/bilelmoussaoui/ashpd/pull/79), which is a bit of a pain). I think the sanest thing to do is to wait until version 0.3 of the wayland-client libraries.

Also, it looks like getting an identifier requires a round-trip to the compositor, and so the new function is async...

druid-shell/src/backend/wayland/surfaces/layershell.rs
james-lawrence: the PR is fine strictly speaking as is; since you're basically following the pattern every backend is doing.... its just a nit i have with druids internals. these methods shouldn't need to be nested as deeply into the backends as they are. a backend really should not care about opening file dialogs. i've covered the topic in detail in other issues.

looks like you're only using the idle handle. which is already exposed, and long term the wayland id. I'd try to keep the file stuff as lightly coupled as possible personally. but rather see functionality implemented than worry about architectural issues that impact every backend currently =)

PR #2230, opened 2022-07-22
Added handle_titlebar support on GTK backend

Steve-xmh:
I also modified the transparency example and added drag area to that circle shape to see the effect.


jneem (approved):
Thanks for the PR! I have to admit that I'm not in love with this `handle_titlebar` API, which I think was mostly copied from the windows API without a lot of thought about how well it would work cross-platform. But since we have it, I guess we may was well have it in GTK too :shrug:

PR #2233, opened 2022-07-25
Windows: fix the raw-win-handle feature

sidit77:
The raw-win-handle feature is currently broken on windows because it was not properly updated to `raw-window-handle` 0.4 when that dependency was bumped.

### DIFF
--- a/druid-shell/src/backend/windows/window.rs
+++ b/druid-shell/src/backend/windows/window.rs
@@ -48,7 +48,7 @@ use winapi::Interface;
 use wio::com::ComPtr;
 
 #[cfg(feature = "raw-win-handle")]
-use raw_window_handle::{windows::WindowsHandle, HasRawWindowHandle, RawWindowHandle};
+use raw_window_handle::{HasRawWindowHandle, RawWindowHandle, Win32Handle};
 
 use piet_common::d2d::{D2DFactory, DeviceContext};
 use piet_common::dwrite::DwriteFactory;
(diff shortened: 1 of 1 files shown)


druid-shell/src/backend/windows/window.rs
```diff
@@ -185,18 +185,18 @@ impl Eq for WindowHandle {}
...
     fn raw_window_handle(&self) -> RawWindowHandle {
         if let Some(hwnd) = self.get_hwnd() {
-            let handle = WindowsHandle {
-                hwnd: hwnd as *mut core::ffi::c_void,
-                hinstance: unsafe {
-                    winapi::um::libloaderapi::GetModuleHandleW(0 as winapi::um::winnt::LPCWSTR)
-                        as *mut core::ffi::c_void
-                },
-                ..WindowsHandle::empty()
+            // Using empty + set fields instead Win32Handle{...} to shut up a unexplainable compiler error
```
jneem, line 188: I wouldn't say it's "unexplainable." `Win32Handle` is marked as `#[non_exhaustive]`, which is explained [here](https://doc.rust-lang.org/reference/attributes/type_system.html)

jneem (changes requested):
Thanks! It looks like we really do need to be running this feature in CI -- I'll try to make a PR for that tomorrow. In the meantime, just a minor quibble about the comment...

druid-shell/src/backend/windows/window.rs
sidit77: Ah, that explains it. Thanks.

jneem:
Thanks again for the PR! But since #2238 fixes this and also bumps raw-window-handle to 0.5, I'm going to close this one and merge that one instead.

PR #2235, opened 2022-07-28
Add ZStack widget

xarvic:
Adds a `ZStack` widget which stacks widgets on top of each other and handles the hot state of overlapping widgets correctly.
Adds the obstructed flag to `MouseEvent` which is used by `WidgetPod` to determine if the widget should be hot.


jneem:
I think it definitely makes sense to have a z-stack widget; it seems like a basic building block. I'm unsure about the "obstructed" business, though. Is the `set_handled` mechanism not good enough?

xarvic:
> I'm unsure about the "obstructed" business, though. Is the set_handled mechanism not good enough?

I changed the implementation. This makes is_handled a bit more complex, but overall i think it is a better design.

jneem (approved):
Sorry for the slow response...

Issue #2237, opened 2022-08-01
"module unix is private" when building druid shell with raw-win-handle


MDeiml:
When building [psst](https://github.com/jpochyla/psst), compiling druid-shell fails as it uses the private `unix` module from `raw-window-handle`. Defined here (link to version 0.4.2 which is currently used):
https://github.com/rust-windowing/raw-window-handle/blob/62cbf83e8667a646bc4b53ccd8dd9c01812add23/src/lib.rs#L28

The relevant import is here
https://github.com/linebender/druid/blob/235d7210ac255ff79da4b867f4c0d7a4b4e63823/druid-shell/src/backend/gtk/window.rs#L46-L47

Issue #2225, opened 2022-07-16
New raw-window-handle release


i509VCB:
There is a new breaking version of raw-window-handle which appears to split between display and window handles: https://docs.rs/raw-window-handle/0.5.0/raw_window_handle/

PR #2238, opened 2022-08-01
Update raw window handle and fix raw-win-handle feature

Azorlogh:
Fixes https://github.com/linebender/druid/issues/2237, resolves https://github.com/linebender/druid/issues/2225

Updates raw-window-handle from 0.4.2 to 0.5.0 and fixes the feature which is currently broken on all platforms.

I tested it on all platforms except macOS.

Small note: for the `x11` backend, the xcb connection was previously stored in the `RawWindowHandle`, but since 0.5.0 decoupled the `WindowHandle`s and `DisplayHandles`, it is no longer present. It was not used anywhere by druid however.

### DIFF
--- a/druid-shell/src/backend/x11/window.rs
+++ b/druid-shell/src/backend/x11/window.rs
@@ -1581,6 +1581,7 @@ impl IdleHandle {
 #[derive(Clone, Default)]
 pub(crate) struct WindowHandle {
     id: u32,
+    #[allow(dead_code)] // Only used with the raw-win-handle feature
     visual_id: u32,
     window: Weak<Window>,
 }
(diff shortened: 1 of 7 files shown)


druid-shell/src/backend/x11/window.rs
```diff
@@ -1581,6 +1581,7 @@ impl IdleHandle {
 #[derive(Clone, Default)]
 pub(crate) struct WindowHandle {
     id: u32,
+    visual_id: u32,
```
jneem, line 1584:
```suggestion
    #[allow(dead_code)] // Only used with the raw-win-handle feature
    visual_id: u32,
```

jneem (changes requested):
Thanks for the PR! You'll need to appease the CI before it will let us merge, though.

jneem (approved):

Issue #2120, opened 2022-01-12
Codes in Readme and  Examples are outdated for version `0.7.0`


duyixian1234:
[example in Readme](https://github.com/linebender/druid#example)
```rust
fn main() -> Result<(), PlatformError> {
    let main_window = WindowDesc::new(ui_builder());  // Should use `ui_builder` instead of `ui_builder()`
    let data = 0_u32;
    AppLauncher::with_window(main_window)
        .log_to_console() // Should use `use_simple_logger` instead of `log_to_console`
        .launch(data)
}
```

jneem:
These examples are for working with the master branch instead of 0.7.0. Under the example there's a comment about this, along with a link to examples that work in `0.7.0`. Maybe we can make this more prominent somehow?

PR #2250, opened 2022-08-30
fix: update `calc` example to version `0.7.0`

EstebanBorai:
Replaces `log_to_console` with `use_simple_logger` and also updates
function expected as parameter for `WindowDesc::new`.

Refer: https://github.com/linebender/druid/issues/2120

### DIFF
--- a/druid/examples/calc.rs
+++ b/druid/examples/calc.rs
@@ -246,7 +246,7 @@ fn build_calc() -> impl Widget<CalcState> {
 }
 
 pub fn main() {
-    let window = WindowDesc::new(build_calc())
+    let window = WindowDesc::new(build_calc)
         .window_size((223., 300.))
         .resizable(false)
         .title(
@@ -259,7 +259,7 @@ pub fn main() {
         in_num: false,
     };
     AppLauncher::with_window(window)
-        .log_to_console()
+        .use_simple_logger()
         .launch(calc_state)
         .expect("launch failed");
 }


xStrom:
Thanks for your effort, however the example code in the `master` branch is meant to be compatible with the code in the `master` branch, not `0.7.0`. That means the current `calc` example code is already correct.

EstebanBorai:
Thanks!

EstebanBorai:
@xStrom just to be clear, when I see [the `Cargo.toml` in `master` branch, for the `druid` crate](https://github.com/linebender/druid/blob/0ebb79905ec5365ae158e7e255f452ebd6eb41e8/druid/Cargo.toml#L3) I can see the version is `0.7.0` so I'm not sure where is the code for the example that belongs to the version `0.7.0` should be instead?

xStrom:
You can find it for example under the GitHub [releases page](https://github.com/linebender/druid/releases). There you can download either a compressed archive, or you can also find a link to tree at the [tagged commit](https://github.com/linebender/druid/tree/v0.7.0).

EstebanBorai:
> You can find it for example under the GitHub [releases page](https://github.com/linebender/druid/releases). There you can download either a compressed archive, or you can also find a link to tree at the [tagged commit](https://github.com/linebender/druid/tree/v0.7.0).
> 
> So specifically for the `calc` example you can find the code [here](https://github.com/linebender/druid/blob/v0.7.0/druid/examples/calc.rs).

Thanks for taking the time! Now I can see where the example lives!

Its a bit counter intuitive for me perhaps because I'm used to refer to the `master` /`main` branches as the "latest-stable" version.

Tysm!

xStrom:
In my experience `master` is never the latest stable version, and instead it is the development branch. Do you have any examples of projects where `master` is not the development branch?

EstebanBorai:
I think I expressed incorrectly sorry, basically every merged PR ends up in `master`. Whenever you want to release, you would create a tag from `master`.

What was confusing for me was, the fact that `master` looked "out of date" based on the latest release `0.7.0`. So perhaps it was developed on a different branch?

> Do you have any examples of projects where master is not the development branch?

Here I do the such approach: https://github.com/EstebanBorai/http-server

xStrom:
Yeah `master` can look _out of date_ but it is actually a newer version. A PR has been merged that changed the examples and there is no `0.8.0` yet - which will then match what is in `master` .. until another PR changes something.

We are on the same page now, so happy to get that resolved.

PR #2254, opened 2022-09-10
Wayland protocol objects versioning & XDG Shell initial size

PolyMeilex:
This PR introduces support for running on a wide variety of versions of core protocols.
Druid can now run on both Weston and Mutter.

While at it, I also fixed the initial size of top levels, they are no longer 1px x 1px, they now use the size provided by the user.

### DIFF
--- a/druid-shell/src/backend/wayland/application.rs
+++ b/druid-shell/src/backend/wayland/application.rs
@@ -171,11 +171,15 @@ impl Application {
                         let id = *id;
                         let version = *version;
 
-                        if !(interface.as_str() == "wl_seat" && version >= 7) {
+                        if interface.as_str() != "wl_seat" {
                             return;
                         }
+
                         tracing::debug!("seat detected {:?} {:?} {:?}", interface, id, version);
-                        let new_seat = registry.bind::<WlSeat>(7, id);
+
+                        // 7 is the max version supported by wayland-rs 0.29.5
+                        let version = version.min(7);
+                        let new_seat = registry.bind::<WlSeat>(version, id);
                         let prev_seat = weak_seats
                             .upgrade()
                             .unwrap()
@@ -185,6 +189,10 @@ impl Application {
                             prev_seat.is_none(),
                             "internal: wayland should always use new IDs"
                         );
+
+                        // TODO: This code handles only app startup, but seats can come and go on the fly,
+                        // so we have to handle that in the future
+
                         // Defer setting up the pointer/keyboard event handling until we've
                         // finished constructing the `Application`. That way we can pass it as a
                         // parameter.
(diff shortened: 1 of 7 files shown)


druid-shell/src/backend/wayland/application.rs
```diff
@@ -292,6 +300,9 @@ impl Application {
                         }
                         // Dont worry if they go away - we will just stop receiving events. If the
                         // capability comes back we will start getting events again.
+
+                        // TODO: Acording to the protocol the above comment is not true, and describes compositor bug,
```
james-lawrence, line 304: just change the comment if its wrong. ;)

james-lawrence (commented):
minor comment otherwise LGTM.

druid-shell/src/backend/wayland/application.rs
PolyMeilex: Right, it was a quick sketchy comment, will update it and add a reference while at it.

druid-shell/src/backend/wayland/surfaces/surface.rs
```diff
@@ -376,12 +379,22 @@ impl Data {
...
+                        rect.x0,
+                        rect.y0,
+                        rect.x1 - rect.x0,
+                        rect.y1 - rect.y0,
+                    );
+                } else {
+                    // We don't care about obscure pre version 4 compositors
+                    // and just damage the whole surface instead of
+                    // translating from buffer coordinates to surface coordinates
+                    self.wl_surface.borrow().damage(0, 0, i32::MAX, i32::MAX);
```
jneem, line 396: druid actually uses the damaged region in order to optimize its own drawing, so I think if you damage the whole surface here then you'll end up with a partially rendered window. Which is probably no big deal if all real compositors support this. But I think it at least deserves a warning...

jneem:
Thanks for the PR! It would be great if the wayland backend received consistent love :)

The CI is complaining that you need to run `cargo fmt`, and besides that I just have one small question...

druid-shell/src/backend/wayland/surfaces/surface.rs
PolyMeilex: Would `self.invalidate()` do the job? The same way we do it for `force == true`

jneem: Yep, I guess that should take care of telling druid that it needs to draw everything. I'm a little confused about that `invalidate()` though, because shouldn't we also be calling `self.wl_surface.borrow().damage(...)` in that branch?

PolyMeilex: I think it should be called, as far as I can tell what happens when no damage is sent is not specified in the protocol. So compositor may just do nothing in that case.

PolyMeilex: Done

jneem (approved):

PR #2255, opened 2022-09-10
Fixed pointer type for ARM64 Windows support

Steve-xmh:


raphlinus (approved):
Thanks for this! I probably would write a single cfg with an `or()` but that's a style nit.

PR #2259, opened 2022-09-13
pop-up window without a parent use config position

yim7:
My app has only one pop-up window, but it always shows in top left corner of the screen.

The reason is that the parent window is not considered empty and the calculation of the relative position of the child window fails.

```
WindowDesc::new(app_ui())
    .window_size(size)
    .set_level(WindowLevel::Tooltip(WindowHandle::default()))
```

### DIFF
--- a/druid-shell/src/backend/windows/window.rs
+++ b/druid-shell/src/backend/windows/window.rs
@@ -1401,18 +1401,20 @@ impl WindowBuilder {
                         dwStyle = WS_POPUP;
                         dwExStyle = WS_EX_NOACTIVATE | WS_EX_TOOLWINDOW;
                         focusable = false;
-                        if let Some(point_in_window_coord) = self.position {
-                            let screen_point = parent_window_handle.get_position()
-                                + point_in_window_coord.to_vec2();
-                            let scaled_point = WindowBuilder::scale_sub_window_position(
-                                screen_point,
-                                parent_window_handle.get_scale(),
-                            );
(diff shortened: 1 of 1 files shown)


jneem:
Thanks for the PR! I'm not on windows to test right now, but is this correct in the presence of scaling? It seems like the code in question adds an offset of the parent origin *and* does some DPI scaling, and I think without the parent you still want the scaling...

yim7:
> 
You're right, it would be better if the window without parents could be scaled correctly.
A popup window without the parent will be the same as the top-level window. And the position of the top window will not be scaled in the current widnows backend. 

I hope to improve this, but I'm not familiar with widnwos gui.

yim7:
> Thanks for the PR! I'm not on windows to test right now, but is this correct in the presence of scaling? It seems like the code in question adds an offset of the parent origin _and_ does some DPI scaling, and I think without the parent you still want the scaling...

You're right, it would be better if the window without parents could be scaled correctly.
A popup window without the parent will be the same as the top-level window. And the position of the window without  parents will not be scaled in current druid.
I hope to improve this, but I am not familiar with windows gui.

xStrom:
I'm not convinced that this is a valid premise. The primary window should definitely not be a tooltip. Use `WindowLevel::AppWindow` for your primary window. The `Tooltip`, `DropDown`, and `Modal` levels are meant for child windows.

Also, as it happens the code in question has been refactored in #2296 so we can't merge anyway.

PR #2260, opened 2022-09-20
Example and new convenience methods for druid menus

xarvic:


xStrom:
Please rebase on `master` so we can pass the CI.

Also, what is going on here? Please provide at least a short paragraph describing what the changes are supposed to be doing.

xarvic:
This PR introduces helper methods for working with selectable menu items. Menuitem::togle_data() turns a menu item with data of `bool` into a simple checkbox.
MenuItem::radio_item() makes the menu item selectable. It is selected if the data is equal to the provided value.

PR #2263, opened 2022-09-21
Replace lazy_static with once_cell

jplatte:


notgull (commented):
The only real issue I have with this is the inclusion of version 1.15.0 of `OnceCell`, which includes a dramatic MSRV bump to rustc 1.56. I'm not sure what the MSRV policy is for Druid, but if it's lower than 1.56 this could introduce an unnecessary bump.

This can be avoided by using `^1.14` instead.

jplatte:
I haven't seen an MSRV specified anywhere. Happy to reduce the minimum version of once_cell if maintainers prefer.

jplatte:
Rebased, and lowered the minimum version for `once_cell` because it doesn't really hurt.

cmyr (approved):
Sure let's try this, and we'll see if anything breaks 💁

PR #2265, opened 2022-09-23
always_on_top for WindowBuilder

mgalos999:
WindowBuilder now has a `.always_on_top(bool)` setting.
Currently I have only implemented this for windows and just have warnings for other backends.


xStrom:
Sorry for letting this wait so long. I can review it now, but I need you to rebase on `master` to fix clippy errors and run `cargo fmt` so we can pass CI.

PR #2267, opened 2022-10-01
pass build_ui function object to WindowDesc::new

H4estu:
`WindowDesc::new` expects a function object as its argument, not the value of the called object. If `build_ui()` is passed, we get the following compiler error:

```
expected a `std::ops::FnOnce<()>` closure, found `impl druid::Widget<()>`

expected an `FnOnce<()>` closure, found `impl druid::Widget<()>`
```

Passing `build_ui` (no parentheses) to `WindowDesc::new` fixes the error.

### DIFF
--- a/docs/src/get_started.md
+++ b/docs/src/get_started.md
@@ -26,7 +26,7 @@ fn build_ui() -> impl Widget<()> {
 }
 
 fn main() -> Result<(), PlatformError> {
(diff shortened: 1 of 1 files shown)


CryZe:
You are "wrong". This compiles just fine on the master branch, just not on the latest release.

H4estu:
Ah, good catch, thanks. Since I don't see a way to point this to the latest release, I'll close this.

PR #2276, opened 2022-10-14
Upgrade bindgen from 0.58 to 0.61

jplatte:
Removes the unmaintained `ansi_term` dependency from the dependency tree. The only breaking changes in 0.59 and 0.60 are MSRV bumps, according to their changelog.


jneem (approved):

jplatte:
0.61 was released yesterday, so I bumped again. Doesn't look like it contains any changes that would require druid-shell to adjust anything.

PR #2279, opened 2022-10-22
Removed unused dependency and move time to dev

hbina:
Signed-off-by: Hanif Bin Ariffin <hanif.ariffin.4326@gmail.com>


jneem:
It looks like `test_log` is actually used on the mac backend. I have to admit, I don't know why...

xStrom:
Given that the premise of this PR has turned out to be incorrect and there hasn't been any discussion from the author I'll close this.

PR #2282, opened 2022-10-26
Fix mouse enter/leave events on macos 13 fixes #2280

### LINKED ISSUE #2280: Mouse move events do not work properly on macOS13.
Mouse move events do not work properly on macOS13

liias:
macOS 13 introduces an issue that mouse events are not fired anymore. This is caused by `NSView.addTrackingArea` being called too soon and is fixed here to be done after `NSView.initWithFrame` has been called.

The docs at https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/EventOverview/TrackingAreaObjects/TrackingAreaObjects.html does say that you can create an `NSTrackingArea` instance and add it to a view at any point because successful creation does not depend on the view being added to a window. But it seems there are still some undocumented preconditions.

### DIFF
--- a/druid-shell/src/backend/mac/window.rs
+++ b/druid-shell/src/backend/mac/window.rs
@@ -290,6 +290,18 @@ impl WindowBuilder {
             let frame = NSView::frame(content_view);
             view.initWithFrame_(frame);
 
+            // The rect of the tracking area doesn't matter, because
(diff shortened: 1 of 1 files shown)


raphlinus (approved):
Thanks, from what I can tell this is a valid fix.

liias:
Thanks for such quick process!

PR #2286, opened 2022-11-20
Fixed window creation position at high-dpi screen

Steve-xmh:
Fixed windows position at high-dpi screen when creating window.

Because I've just noticed that when I need to make window center on high-dpi screen the window will shift to the upper left corner. I've just checked the code for a while and tried to simply fix it.

### DIFF
--- a/druid-shell/src/backend/windows/window.rs
+++ b/druid-shell/src/backend/windows/window.rs
@@ -737,7 +737,7 @@ impl WndProc for MyWndProc {
                     state.hwnd.set(hwnd);
                 }
                 
-                let origin_pos = self.handle.borrow().get_position());
+                let origin_pos = self.handle.borrow().get_position();
                 self.set_scale(scale);
                 self.handle_deferred(DeferredOp::SetPosition(origin_pos));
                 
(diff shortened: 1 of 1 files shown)


xStrom:
Thank you for your effort!

I haven't looked at the substance of this PR yet, but please do the following:
* Rebase on `master` so that we can run the tests (The errors seen here are fixed in `master`)
* Run `cargo fmt`

xStrom:
Thanks again for tackling this issue. This PR here does address the symptom but as I was looking into the problem I found that we need a deeper fix.

That fix is now available in #2296. In my own testing it solves this issue. Perhaps you could see if it solves it for you too?

Steve-xmh:
> Thanks again for tackling this issue. This PR here does address the symptom but as I was looking into the problem I found that we need a deeper fix.
> 
> That fix is now available in #2296. In my own testing it solves this issue. Perhaps you could see if it solves it for you too?

Thanks a lot! Your code also works on my code. So I think I should close this pr for your further modification since my patch is too simple (As I'm new at contributing code at Druid).

Anyway, thanks a lot for your work on this simple patch!

PR #2289, opened 2022-11-22
Clippy.

jneem:


cmyr (approved):
Looks good thanks! There's some clippy warnings on rust beta that we could consider fixing while we're here, but also happy to just get this merged!

jneem:
Good point, those look pretty easy and sensible. I'll merge this first, though...

PR #2290, opened 2022-11-25
Bump piet.

jneem:
This also required bumping cairo and therefore gtk.


jneem:
PS: I tried fixing the beta lints, but ran into trouble. There's a new [manual_clamp](https://github.com/rust-lang/rust-clippy/pull/9484/commits/b22118457245cdd074daaaaea950ec49397f8712) lint that over-aggressive (in my opinion) because `clamp` panics with `max < min`, and we have a couple of places where `max < min` is allowed. Allowing `clippy::manual_clamp` breaks the non-beta build because there's a lint against unknown lints. We could turn off that lint, but I think it's actually useful...

cmyr (approved):
looks good!

cmyr:
And fair on the lints, we can figure it out when they creep into stable. :)

PR #2291, opened 2022-11-25
ZStack fix

xarvic:
This PR makes it possible for a parent widget to use mouse events (for an example scrolling) while hovering over a `ZStack`.

TODOs:
- [x]  decide whether Event::Wheel is a follow up event when the widget is active.

### DIFF
--- a/druid/src/event.rs
+++ b/druid/src/event.rs
@@ -451,9 +451,9 @@ impl Event {
 
     pub fn is_pointer_event(&self) -> bool {
         matches!(
-                event,
-                Event::MouseDown(_) | Event::MouseUp(_) | Event::MouseMove(_)
-            )
+            event,
+            Event::MouseDown(_) | Event::MouseUp(_) | Event::MouseMove(_)
+        )
     }
 }
 
--- a/druid/src/event.rs
+++ b/druid/src/event.rs
@@ -448,6 +448,13 @@ impl Event {
             | Event::Zoom(_) => false,
         }
     }
+
+    pub fn is_pointer_event(&self) -> bool {
+        matches!(
+                event,
+                Event::MouseDown(_) | Event::MouseUp(_) | Event::MouseMove(_)
+            )
+    }
 }
 
 impl LifeCycle {
--- a/druid/src/event.rs
+++ b/druid/src/event.rs
@@ -449,10 +449,13 @@ impl Event {
         }
     }
 
+    /// Returns true if the event involves a cursor.
+    ///
+    /// These events interact with the hot state and
     pub fn is_pointer_event(&self) -> bool {
         matches!(
-            event,
-            Event::MouseDown(_) | Event::MouseUp(_) | Event::MouseMove(_)
+            self,
+            Event::MouseDown(_) | Event::MouseUp(_) | Event::MouseMove(_) | Event::Wheel(_)
         )
     }
 }
--- a/druid/src/widget/z_stack.rs
+++ b/druid/src/widget/z_stack.rs
@@ -98,7 +98,9 @@ impl<T: Data> Widget<T> for ZStack<T> {
                     ctx.set_handled();
                     layer.child.event(ctx, event, data, env);
                 } else {
-                    layer.child.event(ctx, &Event::Internal(InternalEvent::MouseLeave), data, env);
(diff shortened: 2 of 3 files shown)


druid/src/event.rs
```diff
@@ -448,6 +448,13 @@ impl Event {
             | Event::Zoom(_) => false,
         }
     }
+
+    pub fn is_pointer_event(&self) -> bool {
+        matches!(
+            event,
```
jneem, line 454: self, not event, right?

druid/src/event.rs
```diff
@@ -448,6 +448,13 @@ impl Event {
             | Event::Zoom(_) => false,
         }
     }
+
+    pub fn is_pointer_event(&self) -> bool {
```
jneem, line 452: Could you add a doc comment?

druid/src/widget/z_stack.rs
```diff
@@ -94,17 +91,22 @@ impl<T: Data> ZStack<T> {
...
-
+        let mut previous_hot = false;
         for layer in self.layers.iter_mut() {
-            layer.child.event(ctx, event, data, env);
-
-            if is_pointer_event && layer.child.is_hot() {
-                ctx.set_handled();
+            if event.is_pointer_event() && previous_hot {
+                if layer.child.is_active() {
+                    ctx.set_handled();
```
jneem, line 98: Do we still need this `set_handled`? Given your point about `set_handled` affecting the rest of the widget tree, it seems maybe better to avoid it?

jneem (commented):
Thanks! I guess figuring out the status of wheel isn't super critical, since mouse up/down are much more important.

This conversation has made me realize, though, that I don't really understand exactly when it's appropriate to `set_handled`...

druid/src/event.rs
xarvic: Yes, you are right

xarvic:
> Thanks! I guess figuring out the status of wheel isn't super critical, since mouse up/down are much more important.

Event::Wheel is now a pointer event, which is probably the easiest solution.

druid/src/widget/z_stack.rs
xarvic: Do mean replace it with the `obstructed` flag or just remove it?

jneem: There's a good chance I'm missing something, but I was thinking just remove it. Does that work?

I'm expecting that this child is active but not hot, and it will know how to correctly handle mouse events based on that, even if they aren't marked as handled.

xarvic: > I'm expecting that this child is active but not hot, and it will know how to correctly handle mouse events based on that, even if they aren't marked as handled.

The child only knows that it is not hot because it received a follow up event (Mouse event while active is set) which was marked as ´handled´.

jneem: Can't the child do `ctx.is_hot()`?

xarvic: It is about the `Pod` not the widget. The `Pod` sees a mouse event inside its bounds. Therefore it sets its `hot` state to `true`. The only exception is that `active` and `hot` are both set. In that case the `Pod` sets `hot` to `false` but still propagates the event.

jneem: You mean `handled` and not `active`, right? So in that case I understand why you need `set_handled` here, thanks for the explanation! I still think it isn't ideal because (as you explained in some other thread) it affects the global widget tree and not just the subtree under the zstack. But I don't think figuring it out needs to block this pr...

xarvic: Ups my mistake. I meant `active` and `set_handled`. In this case `hot` is set to `false` even if the curser is inside the widgets bounds

xarvic: > I still think it isn't ideal because (as you explained in some other thread) it affects the global widget tree and not just the subtree under the zstack. But I don't think figuring it out needs to block this pr...

I aggree. The only thing which is still brocken is scrolling while pressing a mousebutton. That is not the most impprtant usecase. I think we can merge this PR as it is.

jneem (approved):

PR #2292, opened 2022-11-26
Add Lapce

ShalokShalom:
Lapce is an advanced example, showing a real world application.


ratmice:
While it would be nice, I believe lapce uses a fork of druid which seems to have diverged with what looks like some not insignificant changes https://github.com/lapce/druid.  I feel like it'd be best if the README linked to projects with less risk of sending users down a rabbit hole only to find out something is specific to that variation of druid.

ShalokShalom:
Thanks a lot!

PR #2293, opened 2022-11-27
Fix meta not being a modifier.

lzhoucs:
We added super -> meta for GTK in https://github.com/linebender/druid/pull/331, however we didn't convert the modifier state. This has left inconsistency for super key events from GTK which is converted to a regular meta key by us. This causes issues in lapce when super key is used as a modifier key, e.g: when adding a key binding for save by pressing `super + s`, only `s` is captured as shown below:
before this fix - press `super + s`:
image
after this fix - press `super + s`:
image

We can also see the issue in the example app [event_viewer](https://github.com/linebender/druid/blob/master/druid/examples/event_viewer.rs):
before this fix - press `ctrl` key:
image

before this fix - press `super` key:
image

after this fix - press `super` key:
image


jneem (approved):

jneem:
Thanks, looks good!

Issue #2283, opened 2022-10-28
[BUG] Build failure on Wayland + FreeBSD


ngortheone:
Trying to build master branch on FreeBSD with this command
```
cargo build --no-default-features --features wayland
```

Gives the following error:

```
error: could not select an appropriate backend
  --> /home/ngor/.local/cargo/registry/src/github.com-1ecc6299db9ec823/piet-common-0.5.0/src/lib.rs:48:9
   |
48 |         compile_error!("could not select an appropriate backend");
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0432]: unresolved import `backend`
  --> /home/ngor/.local/cargo/registry/src/github.com-1ecc6299db9ec823/piet-common-0.5.0/src/lib.rs:52:9
   |
52 | pub use backend::*;
   |         ^^^^^^^ use of undeclared crate or module `backend`

For more information about this error, try `rustc --explain E0432`.
error: could not compile `piet-common` due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
```

I am doing something wrong or is this a bug?
Thanks

ngortheone:
This also happens when I follow example from druid book https://linebender.org/druid/get_started.html

ngortheone:
This seems to be caused by old version of `piet-common` dependecny .. changing piet-common to latest master branch fixes this issue, but now I have new issue:

```  
/opt/ngor/src/github.com/linebender/druid/druid-shell/wrapper.h:1:10: fatal error: 'xkbcommon/xkbcommon-compose.h' file not found
  thread 'main' panicked at 'Unable to generate bindings: ClangDiagnostic("/opt/ngor/src/github.com/linebender/druid/druid-shell/wrapper.h:1:10: fatal error: 'xkbcommon/xkbcommon-compose.h' file not found\n")', druid-shell/build.rs:53:10
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
  ```

Seems that build.rs is broken - it can't find the header file (which is present in the system)

**Comment by jneem** [comment]
Maybe we should be using the [clang_arg](https://stackoverflow.com/questions/42741815/setting-the-include-path-with-bindgen) function to provide the include path to `bindgen`. It looks like the include paths should be accessible to us from the calls to `probe_library`...

**Comment by ngortheone** [comment]
@jneem yes, my current diff to build.rs:
```
[...]

PR #2295, opened 2022-11-29
Add xkbcommon path to build script.

jneem:
This was suggested by @ngortheone in #2283. @Maan2003 do you know why we were blocklisting `_.*` before?


Maan2003 (approved):
LGTM!
But, I am not sure why we were blocking _.*

PR #2296, opened 2022-12-01
Respect scale when setting window position during creation.

xStrom:
As pointed out in #2286 by @Steve-xmh our current Windows backend doesn't respect scale when setting the window position during creation. Indeed there was even a TODO note to that effect in the code.

This PR addresses that shortcoming. I also refactored some of that position/size scaling code to be more unified. Previously there were different rounding strategies used in different code paths for example. This can lead to really tricky to debug issues.


xStrom:
I think in this case it's best to merge and see if someone who uses `master` will complain.

PR #2301, opened 2022-12-07
Various documentation-related fixes

PoignardAzur:
Move and expand doc on hot and active status.
Document crate options.
Add discontinuation notice.
Rewrite Druid book.
Remove crochet stuff.
Fix typos.
Fix broken link.
Fix outdated doc links.
Fix examples/readme.md.


docs/src/03_data.md
```diff
@@ -0,0 +1,97 @@
...
+```rust,no_run,noplaypen
+{{#include ../../druid/src/data.rs:same_fn}}
+```
+
+This method checks for equality, but allows for false negatives.
+
+
+## Performance
+
+It is important that your data be cheap to clone and cheap to compare; we encourage the use of reference counted pointers to allow cheap cloning of more expensive types. `Arc` and `Rc` have blanket `Data` impls that do pointer comparison, so if you have a type that does not implement `Data`, you can always just wrap it in one of those smart pointers.
```
xStrom, line 24:
```suggestion
It is important that your data is cheap to clone and cheap to compare; we encourage the use of reference counted pointers to allow cheap cloning of more expensive types. `Arc` and `Rc` have blanket `Data` impls that do pointer comparison, so if you have a type that does not implement `Data`, you can always just wrap it in one of those smart pointers.
```

Alternatively *are* if data is plural, but I think that would confuse more people so I went with singular data.

docs/src/03_data.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Collections
+
+`Data` is expected to be cheap to clone and cheap to compare, which can cause
+issues with collection types. For this reason, `Data` is not implemented for
+`std` types like `Vec` or `HashMap`.
+
+You can always put these types inside an `Rc` or an `Arc`, or if you're dealing with
+larger collections you can build Druid with the `im` feature, which brings in
+the [`im crate`], and adds a `Data` impl for the collections there. The [`im`
```
xStrom, line 34:
```suggestion
the [`im` crate], and adds a `Data` impl for the collections there. The [`im`
```
The link is defined this way.

docs/src/03_data.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+[`im` crate]: https://docs.rs/im
+
+
+## Mapping `Data` with lenses
+
+In Druid, most container widgets expect their children to have the same associated data. If you have a `Flex<Foobar>`, you can only append widgets that implement `Widget<Foobar>` to it.
+
+In some cases, however, you want compose widgets that operate on different subsets of the data. Maybe you want to add two widgets to the above Flex, one that uses the field `foo` and another that uses the field `bar`, and they might respectively implement `Widget<Foo>` and `Widget<Bar>`.
```
xStrom, line 61:
```suggestion
In some cases, however, you want to compose widgets that operate on different subsets of the data. Maybe you want to add two widgets to the above `Flex`, one that uses the field `foo` and another that uses the field `bar`, and they might respectively implement `Widget<Foo>` and `Widget<Bar>`.
```

xStrom (changes requested):
Please rebase this on `master` and resolve the conflicting `druid/src/contexts.rs` file.

I reviewed the other files and generally it looks like a good set of updates.

Thanks for taking the time!

PoignardAzur:
Sorry, haven't had too much bandwidth lately. I'll try to fix the PR this weekend.

xStrom:
No worries, I'm aware as you know 🙂 Whenever you find it reasonable.

PoignardAzur:
Alright, I've rebased and applied some of your suggestions.

xStrom (approved):
Excellent, I think it's good to merge! 🙂

Issue #803, opened 2020-04-05
Improve List documentation


Finnerale:
The [`List`](https://docs.rs/druid/0.5.0/druid/widget/struct.List.html) widget does not really have any documentation right now except for the example project.

It would be nice if the docs would at least explain
- Why it is needed
- How to use it (including what `ListIter` is)
- How to make it scroll

Valentine-Mario:
I was also concerned why the entire project lack proper documentation. This would be a total turn off for people looking for a GUI library to use.

derekdreery:
@Valentine-Mario this is a bit OT for here - use this for development and zulip for questions.

That said, the reason the docs are lacking is because it is early days and things are still changing. Someone could invest time in writing documentation, that is then out of date a week later. This development stage is a necessary part of creating an optimal solution in the long term, but if you were to come back later you would find things a lot more stable and well documented.

PRs adding docs are always welcome though. :)

PR #2302, opened 2022-12-08
Adding Examples and documentation for List Widget

ThomasMcandrew:
Hoping to resolve https://github.com/linebender/druid/issues/803 What is the protocol for working with issues?

Also Is there a way to link the docs to `im` Crate without going externally as I do. I could not find any examples of this.

### DIFF
--- a/druid/src/widget/list.rs
+++ b/druid/src/widget/list.rs
@@ -54,7 +54,7 @@ use crate::{
 /// # use druid::{ Data, Lens, Widget, WidgetExt};
 /// #[cfg(feature = "im")]
 /// use druid::im::Vector;
-/// use druid::widget::list;
+/// use druid::widget::List;
 ///
 /// #[derive(Clone, Data, Lens)]
 /// struct AppState {
@@ -87,7 +87,7 @@ use crate::{
 ///
 /// ```
 /// # use druid::widget::{ Label, List };
-/// # use druid::{ Data, Lens, Widget, WidgetExt};
+/// # use druid::{ Data, Lens, Widget, WidgetExt, Env};
 /// # use druid::im::Vector;
 ///
 /// # #[derive(Clone, Data, Lens)]
@@ -103,7 +103,7 @@ use crate::{
 ///
 /// fn list_item() -> impl Widget<String> {
 ///     let label = Label::new(|data: &String, _env: &Env|
-///         data.clone())
+///         data.clone());
 ///     label
 /// }
 /// ```
@@ -133,7 +133,7 @@ use crate::{
 /// List can be used with any complex widgets.
 /// ```
 /// # use druid::widget::{ Label, List };
-/// # use druid::{ Data, Lens, Widget, WidgetExt};
+/// # use druid::{ Data, Lens, Widget, WidgetExt, Env};
 /// # use druid::im::Vector;
 ///
 /// #[derive(Clone, Data, Lens)]
@@ -199,7 +199,7 @@ impl<T: Data> List<T> {
     /// }
     /// //This widget is the same as the two widgets above
     /// //combined.
-    /// fn combined -> impl Widget<AppState> {
+    /// fn combined() -> impl Widget<AppState> {
     ///     let list = List::new(||
     ///             Label::new(|data: &String, _env: &Env|
     ///                 data.clone()))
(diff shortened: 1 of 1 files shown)


druid/src/widget/list.rs
```diff
@@ -34,6 +34,126 @@ use crate::{
...
+///     list
+/// }
+///
+/// fn list_item() -> impl Widget<InnerState> {
+///     let label = Label::new(|data: &InnerState, _env: &Env|
+///         data.name.clone());
+///     label
+/// }
+/// ```
+/// [`im`]: https://docs.rs/druid/0.6.0/druid/im/index.html
```
ThomasMcandrew, line 153: This is the link I am referring to in the description. is there a way to do it like the flex crate below?

druid/src/widget/list.rs
```diff
@@ -34,6 +34,126 @@ use crate::{
 };
 
 /// A list widget for a variable-size collection of items.
+/// For a static sized collection of items use [`Flex`].
+/// The best data structure to use with the list widget is [`Vector`] from the [`im`] feature.
+/// To include the [`im`] feature, update Cargo.toml
+/// ```ignore
```
PoignardAzur, line 40:
```suggestion
/// ```toml
```
Marking the code as toml gets you better highlighting.

druid/src/widget/list.rs
```diff
@@ -34,6 +34,126 @@ use crate::{
...
+///         .lens(AppState::list_data);
+///     list
+/// }
+///
+/// fn list_item() -> impl Widget<String> {
+///     let label = Label::new(|data: &String, _env: &Env|
+///         data.clone());
+///     label
+/// }
+/// ```
```
PoignardAzur, line 100: This whole example seems way too big to me. It's the example equivalent of an integration test, where you'd want a unit test.

I'd recommend skipping the `AppState` data structure and the lensing, and just using `Vector<String>` in your example.

druid/src/widget/list.rs
```diff
@@ -34,6 +34,126 @@ use crate::{
...
+///         .lens(AppState::list_data);
+///     list
+/// }
+///
+/// fn list_item() -> impl Widget<InnerState> {
+///     let label = Label::new(|data: &InnerState, _env: &Env|
+///         data.name.clone());
+///     label
+/// }
+/// ```
```
PoignardAzur, line 152: I think you should remove that part, or at least make it way shorter. That fact that widgets can be compose doesn't need to be documented in every container widget.

druid/src/widget/list.rs
```diff
@@ -45,6 +165,41 @@ pub struct List<T> {
 impl<T: Data> List<T> {
     /// Create a new list widget. Closure will be called every time when a new child
     /// needs to be constructed.
+    ///
+    /// The list widget can have a function passed in as a closure, or an abstract closure can also
+    /// be provided.
+    ///
```
PoignardAzur, line 171: This part seems superfluous. I think you can assume the vast majority of Rust users will be familiar with how closures work. (And if they aren't, this isn't the place to explain them)

druid/src/widget/list.rs
```diff
@@ -45,6 +165,41 @@ pub struct List<T> {
...
+    /// //This widget is the same as the two widgets above
+    /// //combined.
+    /// fn combined() -> impl Widget<AppState> {
+    ///     let list = List::new(||
+    ///             Label::new(|data: &String, _env: &Env|
+    ///                 data.clone()))
+    ///         .lens(AppState::list_data);
+    ///     list
+    /// }
```
PoignardAzur, line 201: You can probably remove the `combined` function; the other two are descriptive enough on their own.

PoignardAzur (commented):
My general takeaway is "this should be a bit shorter". I'm usually in favor of docs that assume the user knows very little and tries to explain things from the ground up, but in this case I think having this level of explanation for a single widget may be confusing; it may give the impression this widget composes differently than other widgets.

If you think there should be more documentation explaining how composition works, I'd recommend adding it to the book, the widget trait, or the documentation root.

druid/src/widget/list.rs
PoignardAzur: From my tests, it seems that ```[`im`]``` alone works, but only if you build the docs with `--feature=im`. I wouldn't bother.

PR #2303, opened 2022-12-09
Fix clippy lints from 1.66

PoignardAzur:


PoignardAzur:
Ugh. The PR fails CI test because it mentions a lint that doesn't exist yet. I think the simplest solution is to keep that PR in reserve and retry CI when the next Rust version comes out.

jneem:
Yes, I encountered this in #2290. Also, be careful with the `clamp` lint, because following clippy's advice there can introduce panics.

druid/src/scroll_component.rs
```diff
@@ -244,7 +244,7 @@ impl ScrollComponent {
         let usable_space = viewport_major - major_padding;
 
         let length = (percent_visible * viewport_major).ceil();
-        let length = length.max(bar_min_size).min(usable_space);
+        let length = length.clamp(bar_min_size, usable_space);
```
xStrom, line 247: This seems like a potential panic. Usable space could be below the minimum bar size.

xStrom (commented):
Rust 1.66 has arrived so we can continue with this now.

This PR does not seem to resolve all clippy issues though, so we can't pass the CI.

druid/src/scroll_component.rs
jneem: I complained about this upstream, and it sounded like they were going to change the lint. Dunno if it made it into 1.66 though.

xStrom: The stable clippy still complains about this the same way. It has that *note: clamp will panic if max < min, min.is_nan(), or max.is_nan()* - which is something at least.

Feels like an overeager lint to me. Now that it has reached stable maybe some people will step on this rake and complain.

xStrom:
As this is blocking some other issues I took the initiative to get this done in #2317.

PR #2311, opened 2022-12-17
ViewContextChanged fix

### LINKED ISSUE #2306: `List` does not always propagate `ViewContextChanged`
If the list is reasonably dynamic and the elements inside change, then `ViewContextChanged` is not always propagated to all the elements. I have yet to figure out when / why this happens.

xarvic:
@CryZe, does this fix #2306?

### DIFF
--- a/druid/src/core.rs
+++ b/druid/src/core.rs
@@ -1263,7 +1263,7 @@ impl WidgetState {
             sub_window_hosts: Vec::new(),
             is_explicitly_disabled_new: false,
             update_focus_chain: false,
-            view_context_changed: false,
+            view_context_changed: true,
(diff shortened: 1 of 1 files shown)


CryZe:
Yeah that does indeed seem to fix the bug. Thank you 🎉

xStrom (commented):
I hope to take a deeper look into this real soon. For now, please rebase this on `master` so we can pass CI.

xStrom (approved):
I have throughly reviewed this to work. I reported my findings in [#2306](https://github.com/linebender/druid/issues/2306#issuecomment-1369755361).

Thanks!

PR #2313, opened 2022-12-20
L10n example and performance improvements

JasperDeSutter:
Adds an example showing how to use fluent to translate text, including a more advanced plural selection.
Fixes a couple unneeded String clones in the localization implementation by using a Cow, and prevents the bidi markers from being written by fluent so we don't need to remove them.


xStrom:
I'll try to take a look at this soon. Meanwhile please rebase on `master` to fix the clippy errors so we can pass the CI.

PR #2317, opened 2022-12-29
Satisfy clippy 1.66

xStrom:
This PR is a continuation of #2303.

* I explicitly disabled the clamp clippy check in the scroll case where it would case a panic.
* Resolved other missing clippy issues.


cmyr (approved):
Thanks!

PR #2319, opened 2022-12-31
Add a public constructor to StringCursor

benoitryder:
`StringCursor` fields being private is the only thing preventing a user to define its own `EditableText`, unless it directly wraps an existing one.

As an exemple, with this patch it is possible to wrap a `RichText`, implement few traits, and use it in a `TextBox`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,7 @@ You can find its changes [documented below](#070---2021-01-01).
 - Windows: Dark mode support for the title bar ([#2196] by [@dristic])
 - `ZStack` widget ([#2235] by [@xarvic])
 - `Lifecycle::ViewStateChanged`, `InternalLifecycle::RouteViewStateChanged`, `ChangeCtx`, and `RequestCtx` ([#2149] by [@xarvic])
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -75,6 +75,7 @@ You can find its changes [documented below](#070---2021-01-01).
 - Windows: Dark mode support for the title bar ([#2196] by [@dristic])
 - `ZStack` widget ([#2235] by [@xarvic])
 - `Lifecycle::ViewStateChanged`, `InternalLifecycle::RouteViewStateChanged`, `ChangeCtx`, and `RequestCtx` ([#2149] by [@xarvic])
+- Add a public constructor to `StringCursor` ([#2319] by [@benoitryder])
```
xStrom, line 78: Unfortunately the links aren't autogenerated. You need to manually define both `[#2319]` and `[@benoitryder]` at the bottom of the file.

xStrom (commented):
Looks reasonable, just need a slight change to the changelog.

xStrom (approved):
Thanks!

PR #2326, opened 2023-01-10
Satisfy clippy 1.67 beta.

xStrom:
Clippy lints addressed:
* [`uninlined_format_args`](https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args)


jneem (approved):
It looks like there are still some warnings in `druid/src/env.rs` and `druid-shell/examples/invalidate.rs`

xStrom:
Yeah there are even some more, which I'll also handle.

However the more serious issue is that there is a infinite loop of clippy with the beta toolchain.

```rust
warning: variables can be used directly in the `format!` string
   --> druid\src\widget\flex.rs:619:13
    |
619 | /             debug_assert!(
620 | |                 flex >= 0.0,
621 | |                 "flex value for space should be greater than equal to 0, received: {}",
622 | |                 flex
623 | |             );
    | |_____________^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args
    = note: `#[warn(clippy::uninlined_format_args)]` on by default

```
into

```rust
warning: panic message contains an unused formatting placeholder
   --> druid\src\widget\flex.rs:621:84
    |
621 |                 "flex value for space should be greater than equal to 0, received: {flex}"
    |                                                                                    ^^^^^^
    |
    = note: this message is not used as a format string when given without arguments, but will be in Rust 2021
    = note: `#[warn(non_fmt_panics)]` on by default
help: add the missing argument
    |
621 |                 "flex value for space should be greater than equal to 0, received: {flex}", ...
    |                                                                                           +++++
help: or add a "{}" format string to use the message literally
    |
621 |                 "{}", "flex value for space should be greater than equal to 0, received: {flex}"
    |                 +++++

```

I will investigate it tomorrow. Perhaps just updating the edition will solve the issue. However I feel clippy shouldn't recommend fixes that it itself will immediately complain about.

xStrom:
Okay, so with the beta clippy both `assert!` and `debug_assert!` are incorrectly handled with Rust 2019. This was already fixed in [rust-clippy#10055](https://github.com/rust-lang/rust-clippy/pull/10055) and has landed in nightly, but it's unclear to me if it will arrive with stable 1.67 or some future version.

I think we might be able to just upgrade to Rust 2021 and get around this issue. I'll see if I can make it happen.

xStrom:
Alright, this is ready for another look. It solves all clippy errors now.

jneem (approved):
:tada:

PR #2327, opened 2023-01-12
Update Rust edition to 2021.

xStrom:
It seems that we can rather easily upgrade to Rust 2021 and there is currently a direct benefit, so I think we should do it. Specifically clippy 1.67 (beta) is buggy with Rust 2018 as I found out in #2326. Upgrading to Rust 2021 will solve that issue for us.

---

There is an exception in `druid-derive` which will remain Rust 2018. We have macros that use syntax which is [no longer allowed in Rust 2021](https://doc.rust-lang.org/edition-guide/rust-2021/reserving-syntax.html).

> This is mostly relevant to macros. E.g. `quote!{ #a#b }` is no longer accepted.

```rust
quote! {
    #[doc = #struct_docs]
    #[allow(non_camel_case_types)]
    #[derive(Debug, Copy, Clone)]
    pub struct #field_name#lens_ty_generics(#(#phantom_decls),*); // <- Illegal in Rust 2021

    impl #lens_ty_generics #field_name#lens_ty_generics{ // <- Illegal in Rust 2021
        #[doc = #fn_docs]
        pub const fn new()->Self{
            Self(#(#phantom_inits),*)
        }
    }
}
```

As `druid-derive` doesn't have any `assert!` or `debug_assert!` usage that clippy complains about, staying at 2018 there isn't an issue for us right now.

### DIFF
--- a/docs/book_examples/Cargo.toml
+++ b/docs/book_examples/Cargo.toml
@@ -1,8 +1,11 @@
 [package]
 name = "book_examples"
 version = "0.1.0"
+license = "Apache-2.0"
+repository = "https://github.com/linebender/druid"
 authors = ["Colin Rofls <colin@cmyr.net>"]
-edition = "2018"
+edition = "2021"
+publish = false
 
 [dependencies]
 druid = { path = "../../druid", features = [ "im" ] }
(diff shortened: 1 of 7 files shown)


cmyr (approved):
thanks!

PR #2328, opened 2023-01-12
Input Region and Always On Top support

jaredoconnell:
This PR adds useful window features for specialized use cases. I am personally using it to create a better solution to a docking system than to have a bunch of individual windows, since Wayland doesn't allow setting Window positions. This allows specialized situations, and allows a better, less finicky option than the existing ways.

Always on top is implemented as having an input bool, allowing it to be handled by flags or levels on each system.
Input Region is implemented as having an input of a Region, where each platform converts that into a way that can be represented by each system.

All inputs are done directly in the window handler interface. I thought of the potential benefits to integrating this more into the Druid widget system, but that seems wasteful. In specialized use cases that need input region functionality, it can be handled by a custom root widget, which is how I show in the example file.

I added an example Window that demonstrates the changes and allows you to discover the system limitations for the changes on each platform.
Run it with `cargo run --example input_region`
image

If accepted, I'll also contribute these changes to Glazier, which looks to be a fork of druid-shell.

These changes are implemented on:
- GTK
- Wayland (Where possible)
- Windows (Using API calls that should be compatible back to Windows 2000, so in reality the lowest supported OS by Druid)
- Mac OS (Where necessary. Only always on top. Because a transparent frameless window will allow clicks through transparent parts)

Stubs are created for compiling on:
- Web (not possible to support this)
- X11 (I don't know enough about this to implement on this platform. But I left comments on the things that may be helpful)

I tested it on a lot. Here are the test results:
|OS | Arch | DE | API Used| Notes |
|---  | ---      | ---  | ---             | ---        |
| Windows 10 | AMD64 | N/A | Win32 | Works perfectly, minus transparent window border when input region is set |
| Windows 11 | ARM64 | N/A | Win32 | Same as W10 |
| Mac OS         | ARM64 | N/A | Cocoa | Always on top works perfectly. Region not implemented due to transparency doing that task |
| Ubuntu 20.04 LTS | ARM64 | Default |  GTK | Works perfectly |
| Fedora 37 | AMD64 | Gnome on Wayland| GTK | Input region works perfectly except when Windowed. Wayland doesn't allow always on top |
| Fedora 37 | AMD64 | Gnome on Wayland| Wayland | Same |
| Fedora 37 | AMD64 | Gnome on Wayland| X11 | It compiles but does nothing |
| Fedora 37 | ARM64 | Gnome on X11 | GTK | Same as on AMD64 on Wayland. Works perfectly with exceptions. No always on top |
| Fedora 37 | ARM64 | KDE Plama  | GTK | Input region works. You need to handle always on top and titlebar/frame in the Window menu from KDE.
| Fedora 37 | ARM64 | Cinnamon | GTK | Works perfectly. |
| Fedora 37 | ARM64 | XFCE | GTK | Works perfectly |
| Web           | wasm32 | Firefox | N/A | Compiles but does nothing to the window when run |

Let me know if anything needs to be changed. I definitely think these changes will be valuable in the upstream repository.


druid-shell/src/region.rs
```diff
@@ -71,6 +71,10 @@ impl Region {
         self.rects.iter().any(|r| r.intersect(rect).area() > 0.0)
     }
 
+    pub fn contains(&self, point: Point) -> bool {
```
xStrom, line 74: This function needs a comment.

druid/examples/input_region.rs
```diff
@@ -0,0 +1,194 @@
+use druid::widget::prelude::*;
```
xStrom, line 1: This file needs header comments (copyright/description), take a look at how other examples do it.

xStrom (commented):
Thank you for your efforts. From the surface the work looks comprehensive. Hopefully we can have a nice review cycle soon.

On a meta level, I'll note that due to the size of the changes this probably won't make it in time for the Druid 0.8 release which happening early next week hopefully. However we can land it afterwards and yes if it lands we can probably port it to glazier as a follow-up.

For now please do the following:
* Add a changelog entry to `CHANGELOG.md`
* Fix the clippy erros on Windows

```
error: using `clone` on type `*mut winapi::shared::minwindef::HRGN__` which implements the `Copy` trait
Error:    --> druid-shell\src\backend\windows\window.rs:735:54
    |
735 | ...                   let region_tmp = win32_region.clone();
    |                                        ^^^^^^^^^^^^^^^^^^^^ help: try removing the `clone` call: `win32_region`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#clone_on_copy
    = note: `-D clippy::clone-on-copy` implied by `-D warnings`

error: using `clone` on type `*mut winapi::shared::minwindef::HRGN__` which implements the `Copy` trait
Error:    --> druid-shell\src\backend\windows\window.rs:749:54
    |
749 | ...                   let region_tmp = win32_region.clone();
    |                                        ^^^^^^^^^^^^^^^^^^^^ help: try removing the `clone` call: `win32_region`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#clone_on_copy
```

druid/examples/input_region.rs
jaredoconnell: Added.

druid-shell/src/region.rs
jaredoconnell: Added.

jaredoconnell:
> Thank you for your efforts. From the surface the work looks comprehensive. Hopefully we can have a nice review cycle soon.
> 
> On a meta level, I'll note that due to the size of the changes this probably won't make it in time for the Druid 0.8 release which is happening early next week hopefully. However we can land it afterwards and yes if it lands we can probably port it to glazier as a follow-up.
> 
> For now please do the following:
> 
>     * Add a changelog entry to `CHANGELOG.md`
> 
>     * Fix the clippy erros on Windows
> 
> 
> ```
> error: using `clone` on type `*mut winapi::shared::minwindef::HRGN__` which implements the `Copy` trait
> Error:    --> druid-shell\src\backend\windows\window.rs:735:54
>     |
> 735 | ...                   let region_tmp = win32_region.clone();
>     |                                        ^^^^^^^^^^^^^^^^^^^^ help: try removing the `clone` call: `win32_region`
>     |
>     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#clone_on_copy
>     = note: `-D clippy::clone-on-copy` implied by `-D warnings`
> 
> error: using `clone` on type `*mut winapi::shared::minwindef::HRGN__` which implements the `Copy` trait
> Error:    --> druid-shell\src\backend\windows\window.rs:749:54
>     |
> 749 | ...                   let region_tmp = win32_region.clone();
>     |                                        ^^^^^^^^^^^^^^^^^^^^ help: try removing the `clone` call: `win32_region`
>     |
>     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#clone_on_copy
> ```

Changes addressed. Thank you for the prompt review!

CHANGELOG.md
```diff
@@ -76,6 +76,7 @@ You can find its changes [documented below](#070---2021-01-01).
 - `ZStack` widget ([#2235] by [@xarvic])
 - `Lifecycle::ViewStateChanged`, `InternalLifecycle::RouteViewStateChanged`, `ChangeCtx`, and `RequestCtx` ([#2149] by [@xarvic])
 - Add a public constructor to `StringCursor` ([#2319] by [@benoitryder])
+- Input Region and Always On Top support ([#2328] by [@jaredoconnell])
```
xStrom, line 79: Unfortunately we don't have automatic link generation. Please manually define both `[#2328]` and `[@jaredoconnell]` at the bottom of the file.

jaredoconnell: Ah, I see now. Fixed!

james-lawrence:
from wayland perspective LGTM. its propagating the same architectural issues as indicated by the number of warnings of unimplemented log messages. but otherwise fine.

druid-shell/src/backend/windows/window.rs
```diff
@@ -666,12 +722,99 @@ impl MyWndProc {
...
+                                if self.has_titlebar() {
+                                    let region_tmp = win32_region;
+                                    let header_rect = CreateRectRgn(0, 0, client_width, y_offset);
+                                    CombineRgn(win32_region, header_rect, region_tmp, RGN_OR);
+                                }
+
+                                let scale = self.scale();
+                                for rect in region.rects() {
+                                    // Create the region as win32 expects it. Round down for low coords and up for high coords
+                                    let region_part = CreateRectRgn(
```
xStrom, line 743: `CreateRectRgn` can fail.

Also the resulting `region_part` needs to be fed to `DeleteObject` after we're done combining it.

druid-shell/src/backend/windows/window.rs
```diff
@@ -666,12 +722,99 @@ impl MyWndProc {
...
+                                    if result == ERROR {
+                                        warn!("Error combining regions in SetRegion deferred op");
+                                    }
+                                }
+
+                                SetWindowRgn(hwnd, win32_region, 1);
+                            }
+                            None => {
+                                // Set it to have no region
+                                SetWindowRgn(hwnd, null_mut(), 1);
```
xStrom, line 765: Failure chance with this `SetWindowRgn` too.

druid-shell/src/backend/windows/window.rs
```diff
@@ -666,12 +722,99 @@ impl MyWndProc {
...
+            };
+            if GetClientRect(hwnd, &mut client_rect) == FALSE {
+                warn!(
+                    "failed to get client rect: {}",
+                    Error::Hr(HRESULT_FROM_WIN32(GetLastError()))
+                );
+            }
+            // Convert client rect to screen coords to match the window rect
+            let mut client_screen_offset_point = POINT { x: 0, y: 0 };
+            ClientToScreen(hwnd, &mut client_screen_offset_point);
```
xStrom, line 809: `ClientToScreen` can fail.

xStrom (commented):
Alright I've started playing around with this. I really like the example you've put together, it has helped me test and debug issues already.

The current Windows code has three memory leaks. Not just memory either, it leaks GDI resource handles which the OS keeps in limited supply. This means that if you just grab on to a corner of the window and keep resizing the window like a maniac, then after about 5-10 seconds you're out of resource handles. The process is sped up by a factor of two if you enable the titlebar.

In order to determine this I had to add a bunch of error checking and logging that the current code lacks. I have some inline comments about that. As a general rule, we want to err on the side of too much error handling in unsafe code.

druid/examples/input_region.rs
```diff
@@ -0,0 +1,219 @@
+// Copyright 2019 The Druid Authors.
```
xStrom, line 1: You can change the 2019 to 2022 or 2023, whichever year you created this file on.

jaredoconnell:
I'm glad you found the example useful!

Thanks for pointing out the error checks and memory leaks I overlooked! Those are definitely very important for good software.
I think I addressed all of the failure cases and memory leaks.

I tested this on Windows 10 and Windows 11.
On Windows 10, I moved the newest version around for several minutes without failure, with only a slight increase from around 9 to around 16 MB. But that appears to be the OS, not the program, since it does it with the input region feature disabled.
Windows 11 has a huge jump in memory when moving around the window like a maniac. It jumps from around 7 MB to around 135 MB. However, it drops to around 35 MB after a few seconds without resizing. I turned on and off the titlebar and it reset to around 13.5, similar to Windows 10. So this looks to be an OS problem, instead of an issue with this implementation. Though if there is a better way that someone finds in the future, that would be nice.

I also fixed an issue with lingering shadows in Mac OS.

CHANGELOG.md
```diff
@@ -57,71 +58,108 @@ Later support was widened in [#2254] by [@PolyMeilex].
...
+- X11: Support for changing cursors. ([#1755] by [@Maan2003])
+- X11: Support for custom cursors. ([#1801] by [@psychon])
+- X11: Support for window focus events. ([#1938] by [@Maan2003]
+- X11: Support for file dialogs. ([#2153] by [@jneem])
+- X11: Implemented keyboard layout detection. ([#1779] by [@Maan2003])
+- X11: Implemented `Screen::get_monitors`. ([#1804] by [@psychon])
+- X11: Implemented `Application::get_locale`. ([#1756] by [@Maan2003])
+- Web: Implemented `Application::get_locale`. ([#1791] by [@Maan2003])
+- OpenBSD: Support for building. ([#1993] by [@klemensn])
+- FreeBSD: Support for building. ([#2249] by [@nunotexbsd])
```
jaredoconnell, line 162: I think GitHub is a little confused from the merging into the older branch. The shown additions are already in the branch `master`. 
A local diff shows the correct lack of changes here.

druid-shell/src/region.rs
```diff
@@ -71,6 +71,11 @@ impl Region {
         self.rects.iter().any(|r| r.intersect(rect).area() > 0.0)
     }
 
+    /// Returns `true` if the given point is contained within any rectangle in the region.
```
xStrom, line 74:
```suggestion
    /// Returns `true` if the given `point` is contained within any rectangle in the region.
```

xStrom (commented):
Okay here is another round of comments, mostly doc changes.

Overall the changes continue to look good. Next steps for me are:
* Go over the Windows code once more just to be sure.
* Test the example in action on at least Windows and macOS

CHANGELOG.md
```diff
@@ -8,7 +8,7 @@ You can find its changes [documented below](#082---2023-01-27).
 ### Highlights
 
 ### Added
-
+- Input Region and Always On Top support ([#2328] by [@jaredoconnell])
```
xStrom, line 11:
```suggestion

- Input Region and Always On Top support. ([#2328] by [@jaredoconnell])
```

xStrom (commented):
I went over the Windows code again, and besides one comment it looks good.

jaredoconnell:
> I went over the Windows code again, and besides one comment it looks good.

Thanks for the review! I tested your change and it looks good. No issues.

xStrom (approved):
Thanks again for working on this! It's now ready for merging. :tada:

Issue #2140, opened 2022-02-07
Codebase uses approx_eq incorrectly


PoignardAzur:
The macro `approx_eq!(...)` is used in several tests in places where `assert_approx_eq!(...)` should probably be used instead. Because of this, several tests might not actually be checking anything.

Example: https://github.com/linebender/druid/blob/ac3e0a6601fabeefa56c1ca7f6e4f4096febf4b5/druid-derive/tests/with_lens.rs#L63

cmyr:
good catch!

PoignardAzur:
I noticed when I tried to tweak some tests in my fork, and noticed that the tests were passing even with absurd values.

xStrom:
This was solved by #2331.

PR #2331, opened 2023-01-15
Convert approx_eq! calls to assert_approx_eq!

cbondurant:
Resolves issue #2140

No other changes were needed, all tests still passed after this change.


xStrom:
Thanks for taking this on! Overall it looks good, but the changelog entry should be moved from the *Added* section to the *Maintenance* section, because this isn't actually a new feature that Druid users can use, but instead it involves internal tests.

cbondurant:
Gotcha. That amend should fix it.

xStrom (approved):
Thanks!

PR #2335, opened 2023-01-16
Add `Event::WindowScale` and `Ctx::scale` for easier scale handling.

### LINKED ISSUE #2333: SVG widget doesn't respect scale
With the merger of #2106 we regressed on respecting the scale of dp:px for SVG. This can be solved with a few lines, but requires the actual scale information in the widget code. That info isn't piped through right now, so the prerequisite step is to get that piped in.

Hopefully I can have this fixed soon, but just in case I'm creating this issue to track it.

xStrom:
`Event::WindowScale` wasn't fully connected, but the `druid-shell` part of it is somewhat functional, although not complete. Still, it's better to move this forward and surface the event to widgets.

The `Ctx::scale` method is added to every context and makes accessing the current window's scale as easy as accessing the widget's size. This is really a convenience method to improve ergonomics and doesn't enable new functionality.

Additionally I cleaned up the new SVG widget (from #2106) and it now makes use of this scale information to properly render the SVG at various scale factors.

Fixes #2333

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,8 +75,8 @@ You can find its changes [documented below](#070---2021-01-01).
 - Windows: Dark mode support for the title bar ([#2196] by [@dristic])
 - `ZStack` widget ([#2235] by [@xarvic])
 - `Lifecycle::ViewStateChanged`, `InternalLifecycle::RouteViewStateChanged`, `ChangeCtx`, and `RequestCtx` ([#2149] by [@xarvic])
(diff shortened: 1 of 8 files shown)


jneem (approved):
Whoever wrote `svg` originally didn't like imports :smile:

james-lawrence:
@jneem no i didn't. =)

james-lawrence:
@xStrom all looks good to me. =)

PR #2336, opened 2023-01-16
Increase crate versions in preparation for releasing v0.8.0

xStrom:
This is quite straightforward. I updated all the relevant links and bumped the versions of the following crates:

* `druid` from `0.7.0` to `0.8.0`
* `druid-shell` from `0.7.0` to `0.8.0`
* `druid-derive` from `0.4.0` to `0.5.0`

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,6 @@ fn ui_builder() -> impl Widget<u32> {
     Flex::column().with_child(label).with_child(button)
 }
 ```
-This example works with the `master` branch. Here's an example for 0.8.0: https://docs.rs/druid/0.8.0/druid/#examples
 
 Check out the [the examples folder] for a more comprehensive demonstration of
 Druid's existing functionality and widgets. Check [druid_widget_nursery](https://github.com/linebender/druid-widget-nursery) for more widgets.
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ fn ui_builder() -> impl Widget<u32> {
     Flex::column().with_child(label).with_child(button)
 }
 ```
-This example works with master. Here's an example for 0.7.0: https://docs.rs/druid/latest/druid/#examples
+This example works with the `master` branch. Here's an example for 0.8.0: https://docs.rs/druid/0.8.0/druid/#examples
 
 Check out the [the examples folder] for a more comprehensive demonstration of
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -71,7 +71,7 @@ fn ui_builder() -> impl Widget<u32> {
     Flex::column().with_child(label).with_child(button)
 }
 ```
-This example works with master. Here's an example for 0.7.0: https://docs.rs/druid/latest/druid/#examples
+This example works with the `master` branch. Here's an example for 0.8.0: https://docs.rs/druid/0.8.0/druid/#examples
```
jneem, line 74: Given that link doesn't exist, maybe this change should come after release? (And same for the various doc link changes)

xStrom: It needs to be before, because `cargo publish` will take the `README.md` file and use the link from it in the published crate's readme.

So the most reasonable thing here I think is to always change these links just before the release. Which is sort of where we are now. I can wait with merging this to limit the time window even more.

CryZe:
Is there a plan when to release the new version? I'd like to get in some PRs for my issues in time.

README.md
jneem: Hm, good point. Would it make sense to point the links at `docs.rs/druid/latest` instead?

cmyr: I think it's okay if the readme is broken for a day or two, pending the release.

xStrom: The rationale against `latest` is for cases where someone is viewing/using a release that isn't `latest`.

So for example if you look at the [`0.6.0` page on crates.io](https://crates.io/crates/druid/0.6.0), the readme links to the `0.6.0` docs (e.g. https://docs.rs/druid/0.6.0/druid/trait.Widget.html). If we were to use `latest`, then viewing historic versions would lead to wrong or even broken links. That is to say, the `Widget` trait might have changed, or even be removed/renamed.

For that reason, I think it's best to have links with the specific version in them that matches the version of the crate.

jneem: Makes sense, thanks!

jneem (approved):

jaredoconnell (approved):

xStrom:
I created a [milestone for Druid 0.8](https://github.com/linebender/druid/milestone/4) which tracks all the PRs that still need merging. Once those are approved and merged, this PR here will become unblocked.

PR #2337, opened 2023-01-17
Use standardized Druid naming convention.

xStrom:
The naming convention was agreed upon years ago:
* Druid is a proper name and always capitalized
    - Except when referencing the crate `druid` in which case we stylize it
    - Except when referencing the crate `druid-shell` in which case we stylise that too

This PR resolves most of the cases where this convention was not followed.


jaredoconnell (approved):

longmathemagician (approved):
Passes case-sensitive searches & a manual once-over, marking as approved.

PR #2338, opened 2023-01-17
Fix documentation links.

xStrom:
I went over all the documentation and fixed every broken link I could find. It's not perfect, there are still broken links that are caused by code that is not in this repo (`kurbo`, `piet`, `im`) but obviously those will need to be fixed elsewhere.

The strategy here was to resolve broken links by changing the URL link to a [rustdoc managed symbol based link](https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html). This method supports the link appearing at different levels of nesting (*which URL links never did*) and also has the benefit of sometimes being automatically detected when broken. Not every link was converted, but the remaining URL links still work.

Another thing to note is that when a symbol is in scope then the link doesn't have to be defined at all.
```rust
/// If the [`Widget`] trait is in scope, then that link will just work.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -149,6 +149,7 @@ You can find its changes [documented below](#070---2021-01-01).
 
 ### Docs
 
+- Fixed all the broken links that originated from Druid crates. ([#2338] by [@xStrom])
 - Fixed docs of derived Lens ([#1523] by [@Maan2003])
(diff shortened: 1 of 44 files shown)


jaredoconnell (approved):
Fixes and improvements to documentation are appreciated!
I tested this locally and it looks good in the files I tested.

longmathemagician (approved):
This is a huge improvement even though a few more fixes are needed elsewhere.

PR #2342, opened 2023-01-22
Rewrite changelog in preparation for releasing v0.8.0

xStrom:
I did a major rewrite of the changelog for 0.8 and also a bit of categorizing for 0.7. The strategy was consistency and grouping. Entries related to a similar thing are now either next to each other or even combined into a single entry.

Also:
* Added a bunch of missing entries to 0.8 and a few to 0.7.
* Moved incorrectly placed entries to the correct section.
* Fixed missing/wrong link definitions.


jaredoconnell (approved):
There are too many changes for me to verify each entry, but this looks like a good way to organize it. Thank you for fixing everyone's missed entries! Good documentation is appreciated.

xStrom:
> There are too many changes for me to verify each entry

Yeah that isn't going to be feasible. I think the process here would be to just have multiple people take a peek and see if anything pops out. Otherwise we can just move ahead with what hopefully is obviously an improvement.

longmathemagician (approved):

PR #2343, opened 2023-01-24
Fix `ScrollComponent` not handling `LifeCycle::HotChanged`.

xStrom:
This PR adds `LifeCycle::HotChanged` handling to `ScrollComponent` which allows it to hide the scroll bars when the widget loses hot state but doesn't receive any `MouseMove` events that would confirm this. This bug was hidden on GTK thanks to #552 but revealed again due to #2324.

The fix can be seen in action with the `scroll_colors` example as seen below:

before_leave
after_leave


jaredoconnell (approved):
Looks good to me. I just confirmed that the scroll bar gets stuck for the old versions, but not for your patch.

longmathemagician (approved):

PR #2346, opened 2023-01-27
Foreground function for Container Widget and WidgetExt

giannissc:


druid/src/widget/container.rs
```diff
@@ -77,6 +79,36 @@ impl<T: Data> Container<T> {
         self.background = None;
     }
 
+    /// Builder-style method for setting the foreground for this widget.
+    ///
+    /// This can be passed anything which can be represented by a [`BackgroundBrush`];
```
xStrom, line 84: I guess `BackgroundBrush` is no longer a great name. Not a change we need to worry about in this PR though.

druid/src/widget/widget_ext.rs
```diff
@@ -303,16 +310,20 @@ mod tests {
...
 
         // this should be Container<Slider>
         let widget = Slider::new()
             .background(Color::BLACK)
-            .border(Color::BLACK, 1.0);
+            .border(Color::BLACK, 1.0)
+            .foreground(Color::WHITE);
         assert!(widget.background_is_some());
         assert!(widget.border_is_some());
+        assert!(!widget.foreground_is_some());
```
jaredoconnell, line 326: This test is failing because you set it to ! (not), when it is some because you set foreground above. You probably just copied it from above. Removing the ! in the lower one should help.

druid/src/widget/container.rs
```diff
@@ -170,6 +207,11 @@ impl<T: Data> Widget<T> for Container<T> {
             ctx.request_paint();
         }
         self.child.update(ctx, data, env);
+            if let Some(brush) = self.foreground.as_mut() {
+            trace_span!("update foreground").in_scope(|| {
+                brush.update(ctx, old_data, data, env);
+            });
+        }
```
xStrom, line 214: I don't think this needs to be after `self.child.update`. Let's move it immediately after the background update block.

jaredoconnell:
I left a comment about a test failure.

There are also some formatting problems that need to be fixed. There is some valuable information here: https://github.com/linebender/druid/blob/master/CONTRIBUTING.md

Lastly, don't forget to add your PR to the changelog. You can also add yourself to the authors file.

druid/src/widget/widget_ext.rs
```diff
@@ -115,6 +115,13 @@ pub trait WidgetExt<T: Data>: Widget<T> + Sized + 'static {
         Container::new(self).background(brush)
     }
 
+    /// Wrap this widget in a [`Container`] with the provided `foreground`.
+    ///
+    /// See [`Container::background`] for more information.
```
xStrom, line 120:
```suggestion
    /// Wrap this widget in a [`Container`] with the provided foreground `brush`.
    ///
    /// See [`Container::foreground`] for more information.
```

Also, please fix the above `background` method's docs to also say ```background `brush` ``` as there is no variable `background`.

xStrom (commented):
Generally looks fine, but I've got some inline comments.

druid/src/widget/widget_ext.rs
```diff
@@ -303,16 +310,20 @@ mod tests {
         let widget = Slider::new()
             .background(Color::BLACK)
             .align_left()
-            .border(Color::BLACK, 1.0);
+            .border(Color::BLACK, 1.0)
+            .foreground(Color::WHITE);
         assert!(widget.border_is_some());
         assert!(!widget.background_is_some());
+        assert!(!widget.foreground_is_some());
```
jaredoconnell, line 317: For this test case, you should remove the foreground call, because you're testing for there to be no foreground.

xStrom: What you're saying makes sense, but what is going on with the background check here? Not really a problem with this PR, but looks weird.

druid/src/widget/widget_ext.rs
giannissc: Just changed that one

druid/src/widget/widget_ext.rs
jaredoconnell: The one group of tests makes sure that fields that weren't set stay unset, and the other ones ensures that ones that were set stay set.

xStrom: Aha, it's a double use of `Container`. The background is applied to the inner `Container` but the assert is for the outer `Container`. So the `foreground` method call should not be removed, but instead be moved earlier in the callhain, before `align_left`.

giannissc: Just moved that. I think it is ready for another round of comments, if any! I will do another pass but I think all the tests should pass now

druid/src/widget/container.rs
giannissc: Yeah I was thinking the same since I was preparing the PR. Since this didn't make it in time should I prepare another PR for the `BackgroundBrush` change?

xStrom: If you wish, sure. After we merge this one here.

giannissc: Should I just rename it to `Brush`? I am also thinking `PrimitivesBrush`

xStrom: `Brush` sort-of clashes with Piet's [`Brush`](https://docs.rs/piet/latest/piet/trait.RenderContext.html#associatedtype.Brush) so perhaps not. Given that it is defined in `painter.rs` I'm also thinking maybe `PaintBrush`. Although we could maybe go with something non-brush even, given that it supports a full `Painter` widget as a variant. 🤔

druid/src/widget/widget_ext.rs
```diff
@@ -115,6 +115,13 @@ pub trait WidgetExt<T: Data>: Widget<T> + Sized + 'static {
         Container::new(self).background(brush)
     }
 
+    /// Wrap this widget in a [`Container`] with the provided foreground `brush`.
+    ///
+    /// See [`Container::foreground`] for more information.
+    fn foreground(self, brush: impl Into<BackgroundBrush<T>>) -> Container<T> {
+        Container::new(self).foreground(brush)
+    }
```
xStrom, line 123: The comment on this `foreground` method is now fine. Please do me a favor and also update the comment on the old `background` method that is just before `foreground`. Specifically it still has the incorect text ```with the provided `background`.``` -- please replace it with ```with the provided background `brush`.``` for consistency.

xStrom (changes requested):
Please add a `CHANGELOG.md` entry for these changes. You probably need to rebase on `master` first to avoid merge conflicts.

The main entry in the unreleased version's *added* section at the top of the file:
```md
### Added

- `foreground`, `set_foreground`, and `clear_foreground` methods to `Container` and `WidgetExt::foreground` method for convenience. ([#2346] by [@giannissc])
```

Then also define your profile link and the PR link at the very bottom of the file.
```md
[@giannissc]: https://github.com/giannissc

[...]

[#2346]: https://github.com/linebender/druid/pull/2346
```

giannissc:
Ready to be merged! @xStrom

xStrom (approved):
Thanks!

jaredoconnell (approved):

PR #2348, opened 2023-01-27
Fix docs.rs building and docs.rs `README.md` displaying.

xStrom:
The hotfixes continue. 😮

Turns out that we are using a stale nightly compiler flag in our docs.rs config, which means that no docs are being built. I'm commenting out that flag for now, because a simple flag change didn't solve it.

```
> cargo +nightly doc -Zunstable-options -Zrustdoc-scrape-examples=examples
error: flag -Zrustdoc-scrape-examples does not take a value, found: `examples`
> cargo +nightly doc -Zunstable-options -Zrustdoc-scrape-examples
warning: Rustdoc did not scrape the following examples because they require dev-dependencies: cursor, image, invalidation, list, svg, tabs, widget_gallery, anim, async_event, blocking_function, calc, custom_widget, disabled, either, event_viewer, flex, game_of_life, hello, identity, layout, lens, markdown_preview, multiwin, open_save, panels, scroll, scroll_colors, slider, split_demo, styled_text, sub_window, switches, text, textbox, timer, transparency, view_switcher, z_stack, edit_text, invalidate, perftest, quit, shello
    If you want Rustdoc to scrape these examples, then add `doc-scrape-examples = true`
    to the [[example]] target configuration of at least one example.
    Finished dev [unoptimized + debuginfo] target(s) in 0.16s
```

The path forward here is to address this properly in a future release along with proper CI testing with the nightly `rustdoc`.

Also the crates.io `README.md` fix I did in #2347 doesn't seem to address the issue for docs.rs, even though they claim to respect `Cargo.toml`. For now I'm just going to copy-paste our `README.md` to the `druid` crate directory. I'll figure out what the proper way to do this is later in a throwaway testing crate.

The goal right now is to get Druid 0.8 into a representable state.


jneem (approved):
:fire:

jneem:
I don't quite understand why CI thought README was a symlink, but I guess you figured it out already...

xStrom:
Yeah apparently just deleting the file and creating a new one isn't enough. I had to use `git rm` and `git add` too, otherwise some inner git db still claims it's a symlink.

PR #2352, opened 2023-01-30
Synchronize `kurbo` and `image` imports with `piet-common`.

xStrom:
Druid v0.8 ended up shipping with a different major version of `image` than Piet, which causes severe issues due to their incompatibility, as pointed out by @jpochyla. The ultimate solution to this is to import the `image` and `kurbo` crates from `piet-common` so that they remain synchronized.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ You can find its changes [documented below](#082---2023-01-27).
 
 ### Maintenance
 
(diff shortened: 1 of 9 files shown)


jaredoconnell (approved):
This looks good to me. Most importantly it looks less error prone.

longmathemagician (approved):
Looks good!

PR #2353, opened 2023-01-31
Fix `rustdoc` example scraping configuration.

xStrom:
[Docs.rs failed to build Druid v0.8.0 docs](https://docs.rs/crate/druid/0.8.0) because we were specifying a stale unstable nightly rustdoc option in our `Cargo.toml`.

I temporarily disabled this option in #2348.

This PR here brings back example scraping. Unfortunately the [rustdoc book is also stale](https://doc.rust-lang.org/rustdoc/scraped-examples.html#how-to-use-this-feature) so I had to go [deep into the tracking issue](https://github.com/rust-lang/rust/issues/88791#issuecomment-1028240911) to figure out how it changed. The feature can now be disabled on a per example basis, but I enabled it for all of them like was intended before.

I also enabled all the features for docs.rs builds. We used to disable `svg` and `image` due to duplicate artifacts. The duplicate artifacts tracking issue is still open, however `rustdoc` no longer seems to complain for our builds. Perhaps the code changed or the issue is more limited now? Not sure, but I think we can keep the features enabled for now.

I tested this on my own machine with the nightly toolchain and everything seems to work fine.

Critically, I also updated our CI tests. The docs are now built with the nightly toolchain, as that matches docs.rs. I also updated the flags to be a closer match to what docs.rs is configured to use. This way we can hopefully prevent another event where a published build fails on docs.rs. On top of that I made sure we enable our various features in all sorts of CI tests and removed `libx11-dev` as a seemingly stale Linux dependency.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,10 +34,10 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: install libx11-dev
+      - name: install linux dependencies
         run: |
           sudo apt update
-          sudo apt install libx11-dev libpango1.0-dev libxkbcommon-dev libxkbcommon-x11-dev
+          sudo apt install libpango1.0-dev libxkbcommon-dev libxkbcommon-x11-dev
         if: contains(matrix.os, 'ubuntu')
 
       - name: install stable toolchain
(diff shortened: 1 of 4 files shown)


jaredoconnell (approved):
Looks good to me.

jneem (approved):

PR #2356, opened 2023-02-08
Fix doc visibility issues.

### LINKED ISSUE #2355: Documentation for LensExt and Widget is missing from druid docs
The pages for LensExt and Widget both cannot be found in the 8.2 druid docs.

xStrom:
@cbondurant reported in #2355 that the docs for `Widget` and `LensExt` are missing. Indeed that is the case, as is for `Lens`, `WidgetId`, and `WidgetExt`. I think I checked these just a month ago, and regardless I'm confident that these used to work way earlier, with no code/attribute changes since then.

My best guess is that `rustdoc` changed how `#[doc(hidden)]` is inherited. The way we use this attribute is to hide the docs in the `widget` module but then show them in the re-exported root module. This no longer works and the docs are hidden in the root too. Interestingly the reverse is still possible, we could hide the docs in the root but have them show up in the `widget` module.

In this PR I decided to just have the docs visible in both the `widget` module and the root module. We're only talking a few traits here and they're quite fundamental.

Additionally, I made two adjacent changes.
1. I hid the docs for all deprecated items. We already did this for a bunch, now it is consistent.
2. I made sure that all re-exported Druid items (e.g. `Env`) show up as inline docs, not as re-exports.

Fixes #2355

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ You can find its changes [documented below](#082---2023-01-27).
 
 ### Docs
 
+- `Widget`, `WidgetExt`, `WidgetId`, `Lens` and `LensExt` docs are visible again. ([#2356] by [@xStrom])
+- Deprecated items are now hidden. ([#2356] by [@xStrom])
 - Fixed `rustdoc` example scraping configuration. ([#2353] by [@xStrom])
 - Added info about git symlinks to `CONTRIBUTING.md`. ([#2349] by [@xStrom])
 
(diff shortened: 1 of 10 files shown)


jneem (approved):
Odd, I wonder if this was an intentional change in rustdoc.

I also wonder
- whether adding `doc(inline)` changes things at all? (i.e. if something is `doc(inline)` at the top-level and `doc(hidden)` in a submodule, what happens?)
- whether making it `pub(crate)` in the submodule and then `doc(inline)` at top-level whould have the desired effect?

But it's too late for me right now to experiment with this. Also I guess the `pub(crate)` variant would be a breaking change, so it isn't worth the trouble even if it works...

PR #2357, opened 2023-02-11
Update xilem_tokio branch from master

azymohliad:
Besides just merging `master` into `xilem_tokio`, I only updated the usage of `druid_shell::menu::Menu::add_item` ([commit](https://github.com/linebender/druid/pull/2357/commits/76bffbc2f0f18685ba33c948056f483bd051ee31)), as its signature has slightly changed. Xilem examples seem to work well.


xilem/src/app_main.rs
```diff
@@ -58,8 +58,8 @@ impl<T: Send + 'static, V: View<T> + 'static> AppLauncher<T, V> {
             QUIT_MENU_ID,
             "E&xit",
             Some(&HotKey::new(SysMods::Cmd, "q")),
+            Some(false),
             true,
-            false,
```
azymohliad, line 62: Btw, the order of arguments `enabled` and `selected` has changed in druid-shell ([commit](https://github.com/linebender/druid/commit/7c08b326578aa2b25057f5f1af09f9a6b1f51da4#diff-0f63e54d7096d613a02825d47679656d3cd119abea8bc85c59cba648269b6ae0)), so I swapped true & false here too. But the upstream [xilem](https://github.com/linebender/xilem/blob/51ae4d2bfc9d9bcf9e16f7f90a2f85860022cb99/src/app_main.rs#L71) has it in the old order, while [glazier](https://github.com/linebender/glazier/blob/bb4bc453a949b74fc4de16ce6be66b901dcf4ab4/src/menu.rs#L69) matches the latest druid-shell. So I guess it's a regression in Xilem and the quit menu should be unchecked but enabled, is that right? Happy to make a patch to Xilem.

xStrom: Yes this should be fixed in Xilem proper too. A PR would be wonderful.

raphlinus (approved):
I verify that this works as intended. I'm happy for you to commit as soon as `cargo fmt` is taken care of. I'm sending an invite for commit access. Thanks!

xilem/src/app_main.rs
azymohliad: on it

azymohliad: [Done](https://github.com/linebender/xilem/pull/42)

azymohliad:
Thank you @raphlinus! So should I just merge it myself?

xStrom:
Yeah the way we do it is that once a PR is approved, the author gets to choose whether they want to do additional changes in the PR or if it's time to merge.

Generally I would advise the *Squash and merge* button for merging into `master`, but this here is a different kind of PR so I guess the more standard merge that retains the individual commits is in order.

raphlinus:
Yes, our usual pattern is for the author to merge. (This gives an opportunity for a final check, in case the author feels like there might be another change needed before actually merging)

PR #2359, opened 2023-02-15
Emphasized discontinuation with warning label

mattrob33:
It's easy to overlook that this project has been discontinued. A Github "Warning" flag helps call attention to this fact.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,8 @@ For an overview of some key concepts, see the (work in progress) [Druid book].
 
 ## Project status
 
-**The Druid project is being discontinued by the core developer team.**
+> **Warning**
+> **The Druid project is being discontinued by the core developer team.**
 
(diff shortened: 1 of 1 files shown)


xStrom:
Did you yourself or anyone you know invest significant time into using Druid that this would have prevented?

It feels a bit like it's targeting someone who only casually looks at Druid. In that case it's not super important that they notice this message.

It is a minor change and this discussion does seem like bikeshedding. However I'm also thinking about how this would look like on crates.io. Do you perhaps know of any published Rust crate that uses this warning sign or even just quoting (`>`) in their readme? I would like to know how it looks on that site. This readme needs to deliver its impact not just on [GitHub](https://github.com/linebender/druid/blob/master/README.md), but also on [crates.io](https://crates.io/crates/druid), and it shouldn't look too horrible on [docs.rs](https://docs.rs/crate/druid/latest).

PR #2363, opened 2023-02-28
Satisfy clippy 1.68 beta.

xStrom:
Clippy lints addressed:
* [`derivable_impls`](https://rust-lang.github.io/rust-clippy/master/index.html#derivable_impls)


jaredoconnell (approved):
Looks good to me. That is a much more concise way to do it, so it makes sense that Clippy is requesting it.

Zarenor (approved):

PR #2378, opened 2023-05-04
Fix Window light theme detection being backwards

Insprill:
In the Windows backend, when dark mode was enabled, `should_use_light_theme` would return true, and vice-versa. This caused the title bar's color to be the opposite of the system theme. This fixes that behavior so the title bar matches the system theme.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ You can find its changes [documented below](#083---2023-02-28).
 ### Fixed
 
 - `syn` feature `extra-traits` is now always enabled. ([#2375] by [@AtomicGamer9523])
+- Title bar color was opposite of the system theme on Windows. ([#2378] by [@Insprill])
 
 ### Visual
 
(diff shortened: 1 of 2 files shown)


xStrom:
This doesn't seem right to me. I'm using the dark theme and that registry value is `0` for me.

Are you claiming that if the `AppsUseLightTheme` property in registry is `0` then apps should use the light theme, and when it is `1` they should not use the light theme?

If so, does this match the registry value you see on your setup?

Insprill:
Apologies, after looking at it again, it's where the theme is set that's backward. `DWMWA_USE_IMMERSIVE_DARK_MODE` sets whether *dark* mode should be used, however, the value of the registry key was being passed in which states whether *light* mode should be used. I've renamed the function to `should_use_dark_theme` to match the change in return value.

xStrom (approved):
Aha, that does seem like a bug. I confirmed that this fixes the issue. 🎉 

Thanks!

jaredoconnell (approved):

PR #2388, opened 2023-08-05
```self.value.push('.')``` calculator . fix __display

### PR DESCRIPTION
**Pull request by YoungHaKim7**
ex )
36. 3

If you press 36.3, 36 is followed by 3. I think this is better.

```rust
89            '.' => {
90                if !self.in_num {
91                    self.value.push('.');
92                    self.in_num = true;
93                }
```

### DIFF
--- a/druid/examples/calc.rs
+++ b/druid/examples/calc.rs
@@ -88,7 +88,7 @@ impl CalcState {
             }
             '.' => {
                 if !self.in_num {
-                    self.value = "0".to_string();
+                    self.value.push('.');
                     self.in_num = true;
                 }
                 if self.value.find('.').is_none() {


xStrom:
I can not reproduce the bug you are describing.

Also, this patch breaks the following sequence: `2` `x` `.` `5` `=` by giving the answer 5 instead of 1.

Do you still get 1 as the answer to this test?

YoungHaKim7:
```rust 

66                 if self.in_num {
                    self.value = "0".to_string();
                    self.in_num = true;
                }
 ```
 It's my mistake. I'm sorry.
 I skipped an exclamation point on the 66th line. I'm sorry. I'll close it.

PR #2398, opened 2024-02-23
Update the CI and pin the Rust toolchain version.

xStrom:
This PR updates the CI to bring it more in sync with the current Linebender standard.

* Various CI action versions were bumped
* The Rust toolchain version is now pinned
* Beta toolchain jobs were removed
* Preparation for enabling merge queues

### DIFF
--- a/druid-shell/src/backend/wayland/keyboard.rs
+++ b/druid-shell/src/backend/wayland/keyboard.rs
@@ -384,7 +384,6 @@ impl Manager {
                         if let Some(windata) = winhandle.data() {
                             windata.with_handler({
                                 let windata = windata.clone();
-                                let evt = evt;
                                 move |handler| match evt.state {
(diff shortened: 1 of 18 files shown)


xStrom:
Unfortunately Rust 1.70 would no longer compile the workspace so I had to update to Rust 1.76. That meant also dealing with new Clippy lints. I addressed some and added a "temporary" ignore attribute to the difficult ones.

druid-shell/src/backend/wayland/keyboard.rs
```diff
@@ -384,7 +384,6 @@ impl Manager {
                         if let Some(windata) = winhandle.data() {
                             windata.with_handler({
                                 let windata = windata.clone();
-                                let evt = evt;
```
jneem, line 387: :thinking:

PR #2400, opened 2024-05-03
RadioGroup size and padding

josefc12:
With this you can change the size of the radio group's radios by passing 'size' and 'padding'.
The 'size' adjusts the scaling generally for the button and the label. Padding sets the distance from the button to the label.


josefc12:
I'm very sorry about the commits

PR #2402, opened 2024-05-21
Add set_mouse_pass_through and is_foreground_window

AlexKnauth:
- `set_mouse_pass_through` sets whether the mouse passes through the window to whatever is behind.
- `is_foreground_window` returns true if the window is the foreground window or this is unknown, and returns false if a different window is known to be the foreground window.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ You can find its changes [documented below](#083---2023-02-28).
 ### Added
 
 - Type name is now included in panic error messages in `WidgetPod`. ([#2380] by [@matthewgapp])
+- `set_mouse_pass_through` sets whether the mouse passes through the window to whatever is behind. ([#2402] by [@AlexKnauth])
+- `is_foreground_window` returns true if the window is the foreground window or this is unknown, and returns false if a different window is known to be the foreground window. ([#2402] by [@AlexKnauth])
 
 ### Changed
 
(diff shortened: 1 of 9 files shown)


jaredoconnell:
What is the intended use of these features?
These types of features are not the easiest to test.

AlexKnauth:
My intended use is for LiveSplit One, which is a speedrun timer meant to be used while playing a game. If a user is playing the game using their mouse, they will want their mouse events to pass through the timer window (which will be semi-transparent). They will *not* want to be able to accidentally click on the timer and have that pull focus away from their game.

AlexKnauth:
For testing these features, would it help more to add a new example to `druid/examples`, or to modify an existing example like `input_region.rs`?

jacksongoode:
@linebender Would it make sense to set someone as a maintainer of this project so new fixes and PR's can come in?

jaredoconnell:
> @linebender @cmyr Would it make sense to set someone as a maintainer of this project so new fixes and PR's can come in?

I believe the current intention is to archive the project once Xilem is a reasonable replacement of Druid. At that point, if someone would like to fork the project, that's the only path forward.

xStrom (approved):
The `input_region` example modification seems somewhat flakey, as once the passthrough gets switched on I can't switch it back off even if the window is in the foreground again.

Even so, the code in this PR doesn't really break any existing features and has some value, so I'll go ahead and merge.

Thanks!

PR #2409, opened 2024-10-22
Update Rust stable to 1.82 and satisfy Clippy.

xStrom:
Contains one API change:

```
--- pub struct CustomCursor(Arc<HCursor>);
+++ pub struct CustomCursor(Rc<HCursor>);
```


jaredoconnell (approved):
The code looks good, and I can validate that there are no problems running basic examples on MacOS.

